PR #273, opened 2016-04-28
Fix broken webpack config

### LINKED ISSUE #272: Webpack configuration is broken
With f8edd24a4cdfe93abcf02d032767e47bbde249a0, including externals in webpack.config.js breaks the client in the browser's javascript development console (require url not resolving)

But npm test is broken with those removed.  We should probably restore webpack.test.config.js and reference it for npm test again

jeffstieler:
Fixes #272.


jkudish:
Argh sorry for breaking this. I'd still like us to consolidate back to a single file or less duplication at least though.

PR #277, opened 2016-04-30
Modals: Clean up layout

kellychoffman:
Cleaning up the Add a Package modal.

Before:

Small screen:
before - small

Larger screen:
before- large

After:
after - small

screen shot 2016-04-30 at 11 35 02 am

@jkudish or @jeffstieler for review.


jkudish:
Code looks good and tests well visually on various resolutions/browsers (include ie!).

:shipit:

PR #279, opened 2016-04-30
Add release npm task

jkudish:
This PR seeks to add a new `npm run release` task that can be used to package up the plugin for future releases. It performs the following steps:

* Clear out the `release` folder from previous task completions
* Runs `npm run dist` to compile the plugin's Javascript files for production 
* Create `release/woocommerce-connect` with a copy of the plugin's markdown files, php files and the compiled js
* Zip that folder up

This produces a final `release` folder (which is git ignored) with the plugin files as a folder and as a zip ready for distribution.

To test:

* Checkout this branch
* Run `npm run release` -- verify that you get a warning about needing to `npm install`
* Run `npm install`
* Run `npm run release` -- at the prompt, type `no` or something else that's not `yes` -- verify that script abandons ship
* Run `npm run release` -- at the prompt, type `yes` -- verify that script runs as expected and produces a `release` folder that is complete as per the description above
* Upload the produced zip to a test site and make sure that the plugin works correctly

cc @allendav

### DIFF
--- a/tasks/release.js
+++ b/tasks/release.js
@@ -1,0 +1,59 @@
+/*eslint no-process-exit: 0, no-undef: 0 */
+require( 'shelljs/global' );
+const colors = require( 'colors' );
+const confirm = require( 'confirm-simple' );
+const archiver = require( 'archiver' );
+const fs = require( 'fs' );
+
+// some config
+const releaseFolder = 'release';
+const targetFolder = 'release/wocommerce-connect';
+const dirsToCopy = [
+	'assets',
+	'classes',
+	'dist',
+	'vendor',
+];
+
+confirm( colors.cyan( 'Howdy! This script is going to create a release folder with a compiled ' +
+	'zipped up plugin ready for release. This script assumes you\'ve already checked out the correct branch, ' +
+	'are running the latest code, and have run tests as needed. You good with that?' ), ( ok ) => {
+	if ( ! ok ) {
+		console.log( 'OK. Abandoning ship'.magenta );
+		process.exit( 0 );
+	}
+
+	// run npm dist
+	exec( 'npm run dist' );
+
+	// start with a clean release folder
+	rm( '-rf', releaseFolder );
+	mkdir( releaseFolder );
+	mkdir( targetFolder );
+
+	// copy all markdown information files
+	cp( '*.md', targetFolder );
+
+	// copy the main php file
+	cp( 'woocommerce-connect-client.php', targetFolder );
+
+	// copy the directories to the release folder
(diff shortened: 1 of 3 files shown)


jkudish:
BTW given my Afk for the next few days, if this gets a :shipit: please merge it on my behalf.

tasks/release.js
```diff
@@ -0,0 +1,64 @@
+try {
+    require( 'shelljs/global' );
+    const colors = require( 'colors' );
+    const confirm = require( 'confirm-simple' );
+    const archiver = require( 'archiver' );
+    const fs = require( 'fs' );
+} catch( e ) {
+    console.error( 'Please run `npm install` first.'.red );
```
nabsul: Seems like it's not a good idea to use `.red` when the failure could be due to not npm installing the colors library.

tasks/release.js
```diff
@@ -0,0 +1,64 @@
+try {
+    require( 'shelljs/global' );
+    const colors = require( 'colors' );
+    const confirm = require( 'confirm-simple' );
+    const archiver = require( 'archiver' );
+    const fs = require( 'fs' );
+} catch( e ) {
```
nabsul: Simply not having a this try-catch statement will cause node to throw something like:`Cannot find module 'confirm-simple'`, which seems like a more useful message than this.

tasks/release.js
```diff
@@ -0,0 +1,64 @@
...
+const releaseFolder = 'release';
+const targetFolder = 'release/wocommerce-connect';
+const dirsToCopy = [
+    'assets',
+    'classes',
+    'dist',
+    'vendor',
+];
+
+confirm( colors.cyan( 'Howdy! This script is going to create a release folder with a compiled zipped up plugin ready for release. This script assumes you\'ve already checked out the correct branch, are running the latest code, and have run tests as needed. You good with that?' ), (ok) => {
```
nabsul: Can we split this into multiple lines so that it can be read more easily?

nabsul:
looks like the files need to eslinted. Spaces for tabs, spacing around parentheses and all that.

tasks/release.js
jkudish: True enough. I had started with something more complicated and the try/catch stayed... I'll simplify this.

tasks/release.js
jkudish: Sure.

jkudish:
Thanks for the feedback @nabsul! I've cleaned everything up, rebased, squashed ans forced pushed.

nabsul:
:shipit:

PR #287, opened 2016-05-03
City isn’t required to get a shipping quote.

### LINKED ISSUE #222: Not originally calculating rates on cart page
Video: http://cld.wthms.co/1idPg/3xDfwnbR

In an incognito session, I added a product to the cart and tried to calculate shipping from the cart. It doesn't show any shipping methods after entering my zip and state. It only shows shipping rates after going to checkout.

jeffstieler:
Fixes #222.

Only Country, State, and Postcode are sent from the Cart page.

To test:
* Check out https://github.com/Automattic/woocommerce-connect-server/tree/fix/247-fetch-rates-without-city-or-address and start local server
* Add item to cart
* View cart
* Click "calculate shipping"
* Enter valid Country/State/Postcode combo
* Verify rates are displayed

### DIFF
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -191,13 +191,12 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 		 */
 		public function is_valid_package_destination( $package ) {
 
-			$city     = isset( $package['destination']['city'] ) ? $package['destination']['city'] : '';
 			$country  = isset( $package['destination']['country'] ) ? $package['destination']['country'] : '';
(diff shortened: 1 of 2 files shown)


nabsul:
This change looks good :shipit: 
But I think we need to think more carefully about this `is_valid_package_destination()` method. Do we want the client to check this? I think it's going to be different depending on the specific shipping method we use.

jeffstieler:
> Do we want the client to check this? I think it's going to be different depending on the specific shipping method we use.

I agree this will likely differ between shipping vendors, but IMO, Country, State, and Postcode (as validated by WC) is a good minimum check to avoid rates requests we know will fail.

jkudish:
I agree with @jeffstieler. Let's ship it :shipit:

Issue #131, opened 2016-04-06
Generate PR for WooCommerce


allendav:
And retire this repo

jkudish:
A reminder that when we do this we'll need to adjust the behaviour of tracking opt-in and opt-out events (see #289)

jkudish:
We'll also need to keep an eye on #337

allendav:
No longer a task, as Connect will NOT be delivered via a PR to WooCommerce

PR #289, opened 2016-05-05
Add tracks integration

jkudish:
This PR seeks to introduce integrations with the Automattic Tracks metrics system. Specifically it:

* Adds a new Tracks class, and integrates it into our plugin
* Uses Jetpack's existing tracks integration to record events
* Adds the first two events which track opting in and opting out of WooCommerce Connect (activating/deactivating the plugin -- we'll need to adjust that when we merge into WC Core, noted for #131)
* Adds phpunit tests accordinglt

To test:

* You'll need a site that is actually connected via Jetpack, no fakes (but localhost works as long as it was once connected)
* Checkout this branch
* Run phpunit tests -- make sure they pass
* Deactivate the WooCommerce Connect plugin, verify that the following is added to your debug file:
`Tracked the following event: woocommerceconnect_opted_out`
* Re-activate the WooCommerce Connect plugin, verify that the following is added to your debug file:
`Tracked the following event: woocommerceconnect_opted_in`
* Wait approximately 5 minutes (enjoy a good cup of coffee!) -- there is an expected lag before the events are added to Tracks
* Visit the Tracks live event page (ping me if you need help finding it)
* Type in `woocommerceconnect_opted_out` and/or `woocommerceconnect_opted_in` into the `Event Name` box
* Ensure events with your username were recorded appropriately

### DIFF
--- a/woocommerce-connect-client.php
+++ b/woocommerce-connect-client.php
@@ -65,6 +65,25 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 
 		protected $service_object_cache = array();
 
+		static function load_tracks_for_activation_hooks() {
+			require_once( plugin_basename( 'classes/class-wc-connect-logger.php' ) );
+			require_once( plugin_basename( 'classes/class-wc-connect-tracks.php' ) );
+			$logger = new WC_Connect_Logger( new WC_Logger() );
+			return new WC_Connect_Tracks( $logger );
+
+		}
+
+		static function plugin_activation() {
+			$tracks = self::load_tracks_for_activation_hooks();
+			$tracks->opted_in();
+		}
+
+		static function plugin_deactivation() {
+			$tracks = self::load_tracks_for_activation_hooks();
+			$tracks->opted_out();
+
+		}
+
 		public function __construct() {
 			add_action( 'woocommerce_init', array( $this, 'init' ) );
 		}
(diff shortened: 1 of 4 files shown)


woocommerce-connect-client.php
```diff
@@ -101,6 +120,14 @@ public function set_service_settings_store( WC_Connect_Service_Settings_Store $s
 			$this->service_settings_store = $settings_store;
 		}
 
+		public function get_tracks() {
```
nabsul: So, where is/will this method used?

nabsul:
So, I've functionally tested this and can confirm the events getting added to tracks.
The code also looks good.
I have not however run the php unit tests. If you can get someone to do that, then :shipit:  😄

woocommerce-connect-client.php
jkudish: it wasn't strictly needed but all the other classes in the plugin have this and I wanted to remain consistent

jkudish:
> I have not however run the php unit tests. If you can get someone to do that, then 

I ran them many times, and Travis has ran them several times, so I'm just going to :ship:

woocommerce-connect-client.php
nabsul: What I find odd about all of this is that you're only using tracks via the static methods above. What's the point of initializing an instance of this class if it's not being used anywhere?

jkudish: Future compatibility. I have a PR incoming that will use it elsewhere

PR #290, opened 2016-05-05
Don't request rates if settings are empty.

### LINKED ISSUE #193: Do not run rate requests for instances of USPS with no saved settings
i.e. if you add WooCommerce Connect USPS to a Zone but never Save Changes on its settings page, then the settings for that instance are probably empty in options.

We should 1) not generate shipping rate requests from instances in that state and 2) log when a shipping rate request was so requested.

Otherwise, we will probably log confusing messages as rate requests fail (since things like origin zip code which is required will be empty.)

jeffstieler:
Fixes #193.

To test:
* Add a new USPS shipping instance to a zone and do not enter any settings for it
* Add an item to cart and attempt a checkout (that would fall under the above zone)
* Verify that a "Service settings empty." warning is logged

### DIFF
--- a/tests/php/test_class-wc-connect-shipping-method.php
+++ b/tests/php/test_class-wc-connect-shipping-method.php
@@ -127,4 +127,36 @@ class WP_Test_WC_Connect_Shipping_Method extends WP_UnitTestCase {
 
 	}
 
+	/**
+	 * @covers WC_Connect_Shipping_Method::calculate_shipping
+	 */
+	public function test_invalid_service_settings_calculate_shipping() {
+
+		$loader = new WC_Connect_Loader();
+		$loader->load_dependencies();
+
+		$shipping_method = $this->getMockBuilder( 'WC_Connect_Shipping_Method' )
+			->setConstructorArgs( array( 1 ) )
+			->setMethods( array( 'is_valid_package_destination', 'get_service_settings', 'log' ) )
+			->getMock();
+
+		$shipping_method->expects( $this->any() )
+			->method( 'is_valid_package_destination' )
+			->will( $this->returnValue( true ) );
+
+		$shipping_method->expects( $this->any() )
+			->method( 'get_service_settings' )
+			->will( $this->returnValue( new stdClass() ) );
+
+		$shipping_method->expects( $this->once() )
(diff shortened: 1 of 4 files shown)


tests/php/test_class-wc-connect-shipping-method.php
```diff
@@ -99,4 +99,64 @@ public function test_is_valid_package_destination( $package, $expected ) {
...
+			->will( $this->returnValue( false ) );
+
+		$shipping_method->expects( $this->once() )
+			->method( 'log' )
+			->with(
+				$this->stringContains( 'Package destination failed validation.' ),
+				$this->anything()
+			);
+
+		$shipping_method->calculate_shipping();
```
jkudish: should this test that `$shipping_method->calculate_shipping();` returns early?

tests/php/test_class-wc-connect-shipping-method.php
```diff
@@ -99,4 +99,64 @@ public function test_is_valid_package_destination( $package, $expected ) {
...
+			->with(
+				$this->stringContains( 'Service settings empty.' ),
+				$this->anything()
+			);
+
+		$shipping_method->calculate_shipping();
+
+	}
+
 }
```
jkudish: should we add a test for valid settings too?

jkudish:
Added some feedback about tests, but otherwise this works well, tests well and code looks good.

tests/php/test_class-wc-connect-shipping-method.php
jeffstieler: I was trying to test the code I added without going overboard and getting full coverage for unrelated changes. Though yes, eventually this should be 100% covered.

tests/php/test_class-wc-connect-shipping-method.php
jeffstieler: In order to do so, we'd need to mock the API client as well, and so on. It's rather involved because we can't simply check the return value.. we have to make sure other methods weren't called.

jkudish: For a future PR then perhaps

jkudish:
:shipit:

PR #298, opened 2016-05-06
Update/make modal stateless

nabsul:
This makes the add-package sub component completely stateless so that it follows the pattern of most other components.

Also found some props that need to be added to the PropTypes.

To test: Nothing to test. The Add & Edit Modals should behave exactly as they did before.

@jeffstieler @jkudish @allendav

PR #300, opened 2016-05-06
Update/preset packaging readonly

### LINKED ISSUE #260: Add package modal: No way to cancel
Right now, there's no obvious way to back out of adding a package. You can click outside the modal, but there should also be a Cancel button. Which means we should move the position of the 'Save package to use in other shipping methods' checkbox. 

Design:

screen shot 2016-04-27 at 1 05 24 pm

### LINKED ISSUE #303: Cleanup AddPackages Modal Issues
The following issues were noted:

- Not all strings are localized
- `checked` and `isVisible` don't need to be explicitly set to `true`
- Not all locales use the dot for decimal separators (see placeholders for dimensions)
- `is_letter` in `useDefaultField()` doesn't need to be ternary 
- In FormSettingsExplanation: Maybe say "Defines" instead of "Define". Also there is a space ahead of Define that probably needs to go

### LINKED ISSUE #307: Flat Rate Box Dimensions should probably not be editable
otherwise the merchant could make the box dimensions wrong (e.g. bigger) and then box packing could put a too-large item in the box and then the merchant would potentially have to pony up more $ at the post office to ship the item

should-flat-rate-boxes-be-editable

cc @kellychoffman for thoughts

nabsul:
This change causes the add-package modal to disable dimension and weight fields when a preset package is used. A `read_only` field is needed in the box fields to support this. The server needs to support this (PR pending).

Notes: You'll need the corresponding server branch to be able to save. Everything else works.

To test:
- Check this branch out and try switching around the presets in the drop-down when adding a package
- Check that after saving the box, editing results in the correct fields being read only.
- See that everything saves correctly, etc.

Closes #260 
Closes #303 
Closes #307

@allendav @jeffstieler @jkudish

### DIFF
--- a/client/components/shipping/packages/add-package.js
+++ b/client/components/shipping/packages/add-package.js
@@ -75,6 +75,11 @@ const useDefaultField = ( value, updatePackagesField, setModalReadOnly ) => {
 	updatePackagesField( {
 		index: null,
 		is_letter: 'envelope' === value ? true : false,
+		name: '',
+		outer_dimensions: '',
+		inner_dimensions: '',
+		box_weight: '',
+		max_weight: '',
 	} );
 	setModalReadOnly( false );
 };
--- a/client/components/shipping/packages/add-package.js
+++ b/client/components/shipping/packages/add-package.js
@@ -54,7 +54,7 @@ const renderOuterDimensions = ( showOuterDimensions, packageData, value, updateP
 				name="outer_dimensions"
 				placeholder={ exampleDimensions( 100.25, 25.25, 5.75 ) }
 				value={ value }
-				className="flat-rate-package__outer-dimensions__read-only"
+				className={ is_user_defined ? '' : 'flat-rate-package__outer-dimensions__read-only' }
 				onChange={ ( event ) => updateFormTextField( event, updatePackagesField ) }
 				disabled={ ! is_user_defined }
 			/>
@@ -140,7 +140,7 @@ const AddPackageDialog = ( props ) => {
 					name="inner_dimensions"
 					placeholder={ exampleDimensions( 100, 25, 5.5 ) }
 					value={ inner_dimensions }
-					className="flat-rate-package__inner-dimensions__read-only"
+					className={ is_user_defined ? '' : 'flat-rate-package__inner-dimensions__read-only' }
 					onChange={ ( event ) => updateFormTextField( event, updatePackagesField ) }
 					disabled={ ! is_user_defined }
 				/>
@@ -155,7 +155,7 @@ const AddPackageDialog = ( props ) => {
 						name="box_weight"
 						placeholder={ __( 'Weight of box' ) }
 						value={ box_weight }
-						className="flat-rate-package__package-weight__read-only"
+						className={ is_user_defined ? '' : 'flat-rate-package__package-weight__read-only' }
 						onChange={ ( event ) => updateFormTextField( event, updatePackagesField ) }
 						disabled={ ! is_user_defined }
 					/>
@@ -167,7 +167,7 @@ const AddPackageDialog = ( props ) => {
 						name="max_weight"
 						placeholder={ __( 'Max weight' ) }
 						value={ max_weight }
-						className="flat-rate-package__max-weight__read-only"
+						className={ is_user_defined ? '' : 'flat-rate-package__max-weight__read-only' }
 						onChange={ ( event ) => updateFormTextField( event, updatePackagesField ) }
 						disabled={ ! is_user_defined }
 					/>
--- a/client/components/shipping/packages/add-package.js
+++ b/client/components/shipping/packages/add-package.js
@@ -5,18 +5,22 @@ import FormLabel from 'components/forms/form-label';
 import FormTextInput from 'components/forms/form-text-input';
 import FormSettingExplanation from 'components/forms/form-setting-explanation';
 import FormButton from 'components/forms/form-button';
-import FormCheckbox from 'components/forms/form-checkbox';
+// import FormCheckbox from 'components/forms/form-checkbox';
 import Dialog from 'components/dialog';
 import AddPackagePresets from './add-package-presets';
 import { translate as __ } from 'lib/mixins/i18n';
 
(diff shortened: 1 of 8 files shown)


jkudish:
#298 is parked, do you still want this (#300) reviewed?

allendav:
@jkudish - @nabsul and I chatted a bit - we're going to set 298 aside and review this instead, since it has all of 298's changes in it anyways

client/components/shipping/packages/add-package.js
```diff
@@ -5,175 +5,177 @@ import FormLabel from 'components/forms/form-label';
...
-import FormCheckbox from 'components/forms/form-checkbox';
+// import FormCheckbox from 'components/forms/form-checkbox';
 import Dialog from 'components/dialog';
 import AddPackagePresets from './add-package-presets';
 import { translate as __ } from 'lib/mixins/i18n';
 
-class AddPackageDialog extends React.Component {
+const getDialogButtons = ( mode, savePackage, packageData ) => {
+	return [
+		/*
```
allendav: Since box sharing is out of MVP, lets just go ahead and remove this now

client/components/shipping/packages/add-package.js
```diff
@@ -5,175 +5,177 @@ import FormLabel from 'components/forms/form-label';
...
+	if ( ! showOuterDimensions && ! packageData.outer_dimensions ) {
+		return (
+			<a
+				href="#"
+				className="form-setting-explanation"
+				onClick={ ( evt ) => {
+					evt.preventDefault();
+					toggleOuterDimensions();
+				} }>
+				Define exterior dimensions
```
allendav: Localization here and elsewhere please

client/components/shipping/packages/add-package.js
```diff
@@ -5,175 +5,177 @@ import FormLabel from 'components/forms/form-label';
...
-			packageData,
-			toggleOuterDimensions,
-		} = this.props;
+const renderOuterDimensions = ( showOuterDimensions, packageData, value, updatePackagesField, modalReadOnly ) => {
+	return ( showOuterDimensions || packageData.outer_dimensions ) ? (
+		<FormFieldset>
+			<FormLabel>Outer Dimensions (L x W x H)</FormLabel>
+			<FormTextInput
+				name="outer_dimensions"
+				placeholder="100.25 x 25.25 x 5.75"
```
allendav: Quebec will use commas instead of decimal points.  Noting here, but OK to fix under #303

client/components/shipping/packages/add-package.js
```diff
@@ -5,175 +5,177 @@ import FormLabel from 'components/forms/form-label';
...
-				</a>
-			);
-		}
-	}
+const usePresetValues = ( preset, updatePackagesField, setModalReadOnly ) => {
+	updatePackagesField( {
+		index: null,
+		...preset,
+	} );
+	setModalReadOnly( true );
```
allendav: nice

client/components/shipping/packages/add-package.js
```diff
@@ -5,175 +5,177 @@ import FormLabel from 'components/forms/form-label';
...
+};
+
+const useDefaultField = ( value, updatePackagesField, setModalReadOnly ) => {
+	updatePackagesField( {
+		index: null,
+		is_letter: 'envelope' === value ? true : false,
```
allendav: is `? true : false` actually needed?

client/components/shipping/packages/add-package.js
```diff
@@ -5,175 +5,177 @@ import FormLabel from 'components/forms/form-label';
...
-				<FormLabel>Outer Dimensions (L x W x H)</FormLabel>
+				<FormLabel htmlFor="name">Package name</FormLabel>
 				<FormTextInput
-					name="outer_dimensions"
-					placeholder="100.25 x 25.25 x 5.75"
-					value={ value }
-					onChange={ this.updateFormTextField }
+					id="name"
+					name="name"
+					placeholder="The customer will see this during checkout"
```
allendav: Please localize placeholders too

client/components/shipping/packages/add-package.js
```diff
@@ -5,175 +5,177 @@ import FormLabel from 'components/forms/form-label';
...
-					</div>
-					<FormSettingExplanation> Define both the weight of the empty box and the max weight it can hold</FormSettingExplanation>
-				</FormFieldset>
-			</Dialog>
-		);
-	}
-}
+					<span className="wcc-shipping-add-package-weight-unit">{ weightUnit }</span>
+				</div>
+				<FormSettingExplanation> Define both the weight of the empty box and the max weight it can hold</FormSettingExplanation>
```
allendav: Leading unwanted space before Define. Also, should Define be "Defines"

allendav:
> After a preset package is added to the list of packages, it can be edited like any normal package

I don't think we want to allow this - see #307 cc @kellychoffman

kellychoffman:
> I don't think we want to allow this

Agree.

Could both the dimension and weight`fieldset`s for flat rate packages have a unique class so I could style the dimensions to appear more like a definition and less like an input? Open to other ways of doing this, if you have any suggestions. This is what I'm thinking:

screen shot 2016-05-06 at 8 53 54 pm

Also, the 'Define exterior dimensions' link should be changed to 'View exterior dimensions'.

nabsul:
@kellychoffman: Please add the new classes. I can plug them in separately.

kellychoffman:
Thanks @nabsul. 

065b516: I added the class names to the read-only inputs. As long as they can be only applied when a flat rate is selected, I'll be able to work my style magic.

allendav:
@nabsul - i'm sorry for not thinking of this sooner, but is read_only the best name? 'cause you can still edit the package name, right? Maybe it should be 'read_only_size_and_weight'

allendav:
Tests nicely. After changing `read_only` to `is_user_defined`, I think we can 🚢

kellychoffman:
Updated styles in 1a946db. 

Just one more thing we should probably change:  'Define exterior dimensions' link should be changed to 'View exterior dimensions'.

PR #313, opened 2016-05-07
Package name: add href

kellychoffman:
Fixes: https://github.com/Automattic/woocommerce-connect-client/issues/308 and partially: https://github.com/Automattic/woocommerce-connect-client/issues/310

Adding an href to the package name link so it gets all the special link properties that it deserves.

Before:

After:
href


nabsul:
:shipit:

PR #316, opened 2016-05-09
Add/basic client side validation

jkudish:
This the beginning of supporting client-side validation. More to come, but I thought I'd start small to make reviewing and testing easier :)

This adds client side validation for text fields, specifically `Title` and `Origin` (since Account doesn't have any validation rules). It also adds a re-usable, and testable `isError` utility that we'll be able to re-use for other field types.

To test:

* You'll need to communicate with a server that is patched with PR 262 from the server
* Make sure to set `WOOCOMMERCE_CONNECT_SERVER_URL` accordingly
* Checkout this branch
* You may need to set `WOOCOMMERCE_CONNECT_FREQUENT_FETCH` or otherwise force a fetch
* Visit the settings page for a USPS instance
* Modify the title. Verify that the validation hint is displayed if you leave the title blank, but not otherwise
* Modify the origin. Verify that the validation hint is displayed if you leave the zip code blank or if you enter an invalid (not 5 digits or 5+4) zip code, but not otherwise.

image

Note: Further improvements to come so that we don't display the validation error right away (e.g. it's annoying to see it pop up while you type) but I wanted to get something basic working before I got too far down a rabbit hole.

cc @jeffstieler @allendav @nabsul

### DIFF
--- a/client/lib/utils/test/error.js
+++ b/client/lib/utils/test/error.js
@@ -1,0 +1,114 @@
+import { expect } from 'chai';
+import {
+	isFieldError,
+} from '../error';
+
+describe( 'Error utils', () => {
+	it( '#isFieldError(): required, no pattern, empty value', () => {
+		const required = true;
+		const schema = {
+			pattern: '',
+		};
+		const value = '';
+
+		const isError = isFieldError( required, schema, value );
+
+		expect( isError ).to.be.true;
+	} );
+
+	it( '#isFieldError(): required, no pattern, not empty value', () => {
+		const required = true;
+		const schema = {
+			pattern: '',
+		};
+		const value = '1234';
+
+		const isError = isFieldError( required, schema, value );
+
+		expect( isError ).to.be.false;
+	} );
+
+	it( '#isFieldError(): required, pattern, empty value', () => {
+		const required = true;
+		const schema = {
+			pattern: /^1234/,
+		};
+		const value = '';
+
+		const isError = isFieldError( required, schema, value );
+
+		expect( isError ).to.be.true;
+	} );
+
+	it( '#isFieldError(): not required, no pattern, empty value', () => {
+		const required = false;
+		const schema = {
+			pattern: '',
+		};
+		const value = '';
+
+		const isError = isFieldError( required, schema, value );
+
+		expect( isError ).to.be.false;
+	} );
+
+	it( '#isFieldError(): not required, pattern, empty value', () => {
+		const required = false;
+		const schema = {
+			pattern: /^1234/,
+		};
+		const value = '';
+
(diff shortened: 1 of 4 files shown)


jeffstieler:
Are you planning on implementing a full JSON schema validator, or will an existing library replace this?

jkudish:
> Are you planning on implementing a full JSON schema validator, or will an existing library replace this?

Addressing one thing at a time, wanted to get something basic working first. Will look into what's needed for other fields next, and if that involves an existing library then I'll pick one. But I hadn't gone that far yet.

allendav:
If an invalid field lacks a validation_hint, you get an odd display.

screen shot 2016-05-09 at 4 07 53 pm


We should probably _not_ render `FormInputValidation` if validation_hint is absent. (I had to hack the server to remove the validation_hint for title to get this, fwiw.)

client/lib/utils/test/error.js
```diff
@@ -0,0 +1,102 @@
+import { expect } from 'chai';
+import {
+	isFieldError,
```
allendav: I'm curious - why multiline here?

jkudish: Cuz copy paste from another test and cuz it allows to add more functions later one.

jkudish:
Good point @jeffstieler will adjust

client/lib/utils/test/error.js
```diff
@@ -0,0 +1,102 @@
...
+		const schema = {
+			pattern: /^1234/,
+		};
+		const value = '1235';
+
+		const isError = isFieldError( required, schema, value );
+
+		expect( isError ).to.be.true;
+	} );
+} );
```
allendav: You have no test covering the following common/important combination: required: true, pattern: /^1234/, value: 1234

allendav: If one of these tests fail, its going to take a developer a little bit of time to figure out which of them to look at, since the names of the tests are so similar and there are so many of them.  May I suggest including a unique identifier in the description, e.g. "FE_TEST_1", "FE_TEST_2", etc

jkudish: I don't like that... The test titles are pretty descriptive and it should be easy enough to search for it

jkudish: > You have no test covering the following common/important combination: required: true, pattern: /^1234/, value: 1234

Added in a merged rebase

jkudish:
> We should probably not render FormInputValidation if validation_hint is absent. (I had to hack the server to remove the validation_hint for title to get this, fwiw.)

Fixed in the latest rebase

allendav:
Tests nicely. :shipit:

PR #318, opened 2016-05-10
PreventDefault so that the click doesn't cause a scroll up on the page

nabsul:
To test: Try editing an existing package. Before the fix it should scroll up, after it should not.

@kellychoffman @jkudish @allendav @jeffstieler


jkudish:
Works as expected and LGTM. :shipit:

PR #323, opened 2016-05-10
Packages: Clean up ui

kellychoffman:
Clean up the packages UI. Fixes the rest of https://github.com/Automattic/woocommerce-connect-client/issues/310 ( I don't want to wait for Calypso and I think we should add an outline here)

Before:

screen shot 2016-05-10 at 11 10 35 am

After:

screen shot 2016-05-10 at 11 52 43 am
screen shot 2016-05-10 at 12 00 12 pm

- Lots of alignment tweaks, including small screen
- Allows for up to 3 decimal point dimensions without breaking on to another line (large screen view)
- Changes to package type icon: alignment, color, stylesheet cleanup
- Correct cursor type
- Focus outline for removing a package

### DIFF
--- a/client/components/shipping/packages/style.scss
+++ b/client/components/shipping/packages/style.scss
@@ -10,13 +10,14 @@
         width: 10%;
         @include breakpoint( '<660px' ) {
             width: 12%;
-    }
+        }
 
-    .gridicon {
-        border-radius: 50%;
-        background-color: $gray;
-        fill: white;
-        padding: 4px;
+        .gridicon {
+            border-radius: 50%;
+            background-color: lighten( $gray, 10% );
+            fill: white;
+            padding: 4px;
+            margin-bottom: -4px;
         }
     }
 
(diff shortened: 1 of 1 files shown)


client/components/shipping/packages/style.scss
```diff
@@ -9,31 +13,38 @@
...
+            background-color: lighten( $gray, 10% );
+            fill: white;
+            padding: 4px;
+            margin-bottom: -4px;
         }
     }
 
     .package-name {
-        width: 55%;
+        width: 51%;
```
jkudish: awfully specific :laughing:

client/components/shipping/packages/style.scss
```diff
@@ -97,6 +112,10 @@
     &:hover {
         color: $alert-red;
     }
+
+    &:focus {
+        box-shadow: 0 0 0 1px $blue-medium,0 0 2px 1px rgba(30,140,190,.8);
```
jkudish: space after `$blue-medium` and in the `rgba` function?

jkudish:
Looks good in a variety of browsers and resolutions. I left a couple minor code spacing comments, but otherwise :shipit:

client/components/shipping/packages/style.scss
kellychoffman: haha. there's math involved here! i had to take out my calculator and everything.

client/components/shipping/packages/style.scss
kellychoffman: space added. oops, rgba function uses color variable now.

PR #335, opened 2016-05-11
Fix/334 render div by default

### LINKED ISSUE #334: Generate a div by default for unrecognized form layout element (group) types, don't require items
Right now, if the form layout array element has a type other than 'fieldset' or 'actions' (e.g. 'submit') then we get the following error because SettingsGroup returns nothing from its render:

```
warning.js:45 Warning: SettingsGroup(...): No `render` method found on the returned component instance: you may have forgotten to define `render`, returned null/false from a stateless component, or tried to render an element whose type is a function that isn't a React component.
```

We should at least return a div in such cases, instead of blowing up entirely.  We should also not require items, since an layout array element of type actions could potentially have a single item.  This will avoid the following error for such layouts:

```
Warning: Failed propType: Required prop `group.items` was not specified in `SettingsGroup`. Check the render method of `Connect(SettingsGroup)`.
```

allendav:
Fixes #334 

This PR seeks to make it less likely that a unrecognized type in form layout will cause all rendering of the form to break.

To test:
Hack a local server to remove the type from a fieldset, e.g. /lib/shipping/usps/service-settings.js form_layout[0].type
Refresh the client, receiving the new schema
Ensure the remainder of the form renders

### DIFF
--- a/client/components/wcc-settings-form/settings-group.js
+++ b/client/components/wcc-settings-form/settings-group.js
@@ -14,14 +14,14 @@ const SettingsGroup = ( { group, schema, storeOptions, settings, form, formActio
 				<CompactCard className="settings-group-card">
 					<FormSectionHeading className="settings-group-header">{ group.title }</FormSectionHeading>
 					<div className="settings-group-content">
-						{ group.items.map( item => (
+						{ group.items ? group.items.map( item => (
 							<SettingsItem
 								key={ item.key ? item.key : item }
 								layout={ item }
 								schema={ schema }
 								storeOptions={ storeOptions }
 							/>
-						) ) }
+						) ) : null }
 					</div>
 				</CompactCard>
 			);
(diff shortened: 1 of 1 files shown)


client/components/wcc-settings-form/settings-group.js
```diff
@@ -43,13 +49,20 @@ const SettingsGroup = ( { group, schema, storeOptions, settings, form, formActio
 					/>
 				</CompactCard>
 			);
+
+		default:
+			return (
+				<div className="settings-group-default">
+					{ group.items ? renderGroupItems( group.items, schema, storeOptions ) : null }
```
jkudish: the repetition between this and line 30 is giving me a bit of a twitch

jkudish:
Tests nicely. Just left one comment about DRY. Otherwise :shipit:

client/components/wcc-settings-form/settings-group.js
allendav: Yeah, the lack of recursion between this and settings-item is giving me a twitch. I might need to do something about that.

allendav:
ty!

PR #343, opened 2016-05-13
Timestamp setting updates

jkudish:
* Add a new last_rate_request timestamp bumped whenever there's a shipping rate request (at most once an hour though to prevent too many writes)
* Send `0` as the fallback timestamp instead of `false`

Can only be merged after https://github.com/Automattic/woocommerce-connect-server/pull/266 is merged. See that PR for testing instructions

Fixes https://github.com/Automattic/woocommerce-connect-client/issues/216


nabsul:
Works with the corresponding server branch. Once that's reviewed, :shipit:

Issue #306, opened 2016-05-06
We should validate fields in the modal before leaving the modal


allendav:
![we-really-should-validate-while-in-the-modal](https://cloud.githubusercontent.com/assets/1595739/15086650/124e60e8-1397-11e6-81cb-90d9f353a715.gif)

cc @kellychoffman - do you concur?

jkudish:
@jeffstieler and I just discussed this. We agreed that:

* We'll have client side validation for the package to be able to show inline errors
* We'll also have server side validation for packages. Most users won't encounter this (since they'll hit client side validation first), but we'll need to be safe. Any returned errors from the server can highlight the package in question with a message to go and fix that package. We can then show the specific error in the modal. We'll need some design for this @kellychoffman

kellychoffman:
> do you concur?

I do indeed.

kellychoffman:
> We'll also have server side validation for packages. Most users won't encounter this (since they'll hit client side validation first), but we'll need to be safe. Any returned errors from the server can highlight the package in question with a message to go and fix that package.

Can you better explain this? Maybe with an example? I'm not sure I fully understand.

jkudish:
> Can you better explain this? Maybe with an example? I'm not sure I fully understand.

Yeah sorry, re-reading myself I see that it's not 100% clear. 

In other words: when you add/edit a package we'll have instant client-side validation so you can immediately see any errors. We can disable the "add package"/"save changes" button until validation passes. However, we'll also need/have server-side validation which occurs after one clicks the main "Save changes" button of the form. Most users will have failed client validation so won't trigger an error with packages at this stage, but if they do, we'll need something to indicate which of the already saved boxes has an issue they need to address. 

Does that make sense?

kellychoffman:
Here's what I'm thinking:

We disable the 'Add packages' button until all the required fields are filled out. This way no packages will get added unless they are valid. In theory, anyway. When I was talking to @jeffstieler earlier, he mentioned it might be a good idea to plan for this anyways. 

**Add packages modal** 

On open: 
(disabled primary button)

<img width="511" alt="screen shot 2016-05-16 at 7 46 39 pm" src="https://cloud.githubusercontent.com/assets/5835847/15308879/f34c5974-1b9e-11e6-86ec-8745e1f02def.png">

With error: 
(contextual error message, primary button still disabled)

[...]

Issue #188, opened 2016-04-15
Allow decimal service adjustments


allendav:
Right now, the shipping adjustment must be an integer.  If I wanted to add $9.95 (a common shipping and handling amount) to the rate, I could not.

![screen shot 2016-04-15 at 12 17 20 pm](https://cloud.githubusercontent.com/assets/1595739/14572507/0c41f280-0304-11e6-8103-92682fde9b95.png)

PR #348, opened 2016-05-18
Show errors inline and prevent form submit when errors present.

### LINKED ISSUE #206: Contextual validation: Client-side validation of form values before allowing submit.
Form data should be validated against the JSON Schema as values change.

Validation errors should display immediately alongside the field that triggered them.

Related: https://github.com/Automattic/woocommerce-connect-client/issues/205

jeffstieler:
Requires https://github.com/Automattic/woocommerce-connect-server/pull/273 to test.

Fixes #206 and incorporates some error UI design from #306.

To test:
* Erase a required field value (like Title or Origin)
* Verify the error message is displayed
* Verify the "Save changes" button is disabled
* Enter an invalid zip code for Origin and
* Verify the error message is displayed
* Verify the "Save changes" button is disabled
* Add a package with only a name
* Verify that the package list item shows an error state
* Verify the "Save changes" button is disabled
* Force a bad value for a shipping service adjustment in the database by changing the serialized `i:0` value to `s:1:"a"`
* Reload form
* Verify that the group containing that service is expanded, and the service adjustment is highlighted red with an icon


jeffstieler:
I'm opening this up for review even though it's missing one piece that I originally considered as part of this effort, which is showing errors from the server inline on the form.

client/components/shipping/packages/packages-list-item.js
```diff
@@ -1,36 +1,56 @@
...
-				<Gridicon icon="cross-small" size={ 18 } />
-			</Button>
+	errors,
+} ) => {
+	const hasError = !! errors.length;
+	const classes = classNames( 'wcc-shipping-packages-list-item', {
+		'wcc-error': hasError,
+	} );
+	return (
+		<div className={ classes }>
```
allendav: Optional. No need to assign to classes, just call classNames here

client/components/shipping/packages/packages-list.js
```diff
@@ -8,24 +8,34 @@ const PackagesList = ( {
 	dimensionUnit,
 	removePackage,
 	editPackage,
+	errors,
 } ) => {
+	const renderPackageListItem = ( pckg, idx ) => {
+		const itemErrors = errors.length ? errors.filter( ( error ) => {
```
allendav: Is this logic basically finding the errors related to this particular package?

client/components/shipping/services/entry.js
```diff
@@ -10,35 +12,44 @@ const ShippingServiceEntry = ( {
...
-				checked={ enabled }
-				onChange={ ( event ) => updateValue( 'enabled', event.target.checked ) }
+	hasError,
+} ) => {
+	const classes = classNames( 'wcc-shipping-service-entry', {
+		'wcc-error': hasError,
+	} );
+
+	return (
+		<div className={ classes }>
```
allendav: Could simplify the code a bit by not defining classes, but just using classNames here. Optional

client/components/shipping/packages/packages-list.js
```diff
@@ -8,24 +8,34 @@ const PackagesList = ( {
...
+			return ( 0 === error.indexOf( idx + '.' ) );
+		} ).map( ( error ) => error.substr( 2 ) ) : false;
```
allendav: What happens if I have more than 9 packages. Wouldn't idx = "10" then? Does this substr still pull the expected part of the string?

package.json
```diff
@@ -61,6 +61,7 @@
     "babel-preset-stage-1": "^6.5.0",
     "debug": "2.2.0",
     "dompurify": "^0.7.4",
+    "is-my-json-valid": "^2.13.1",
```
allendav: I think you need to add classnames too, technically?

client/components/shipping/packages/packages-list-item.js
```diff
@@ -1,36 +1,56 @@
...
-		<div className="package-dimensions">
-			<span>{ data.inner_dimensions } { dimensionUnit }</span>
-		</div>
-		<div className="package-actions">
-			<Button compact borderless className="remove-package" onClick={ onRemove }>
-				<Gridicon icon="cross-small" size={ 18 } />
-			</Button>
+	errors,
+} ) => {
+	const hasError = !! errors.length;
```
jkudish: `!!` is kinda sketchy as far as good practices go

also, why not just pass in `hasError` here directly since we never need the errors themselves in this component. I think the least info the components require, the better

client/components/shipping/packages/packages-list.js
jkudish: yes this is kinda hard to grok

client/components/shipping/packages/packages-list.js
```diff
@@ -8,24 +8,34 @@ const PackagesList = ( {
...
+
+		return (
+			<PackagesListItem
+				key={ idx }
+				index={ idx }
+				data={ pckg }
+				dimensionUnit={ dimensionUnit }
+				onRemove={ () => removePackage( idx ) }
+				editPackage={ editPackage }
+				errors={ itemErrors }
```
jkudish: see note above about passing `hasErrors` directly

client/components/shipping/services/group.js
```diff
@@ -39,6 +40,7 @@ const ShippingServiceGroup = ( {
 			compact
 			actionButton={ actionButton }
 			actionButtonExpanded={ actionButton }
+			expanded={ !! errors.length }
```
jkudish: `!!` is not great best practice

client/components/wcc-settings-form/index.js
```diff
@@ -24,4 +47,37 @@ WCCSettingsForm.propTypes = {
 	saveFormData: PropTypes.func.isRequired,
 };
 
-export default WCCSettingsForm;
+const getFormErrors = ( schema, data ) => {
+	const validate = validator( schema, { greedy: true } );
+	const success = validate( data );
+
+	if ( ! success && validate.errors && validate.errors.length ) {
```
jkudish: kinda hard to grok what this does exactly

client/components/wcc-settings-form/settings-group.js
```diff
@@ -2,42 +2,54 @@ import React, { PropTypes } from 'react';
...
+	schema,
+	storeOptions,
+	form,
+	saveForm,
+	dismissSuccess,
+	errors,
+} ) => {
+	const renderSettingsItem = ( item ) => {
+		const key = item.key ? item.key : item;
+		const itemErrors = errors.length ? errors.filter( ( error ) => {
```
jkudish: that's one long ternary. I think it'd be clearer to split this up. Also some comments wouldn't hurt.

client/components/wcc-settings-form/settings-group.js
```diff
@@ -2,42 +2,54 @@ import React, { PropTypes } from 'react';
...
+		} ).map( ( path ) => {
+			if ( 0 === path.indexOf( key + '.' ) ) {
+				return path.substr( key.length + 1 );
+			}
+			return path;
+		} ) : false;
+
+		return (
 			<SettingsItem
-				key={ item.key ? item.key : item }
```
jkudish: are you sure this is safe to remove?

client/components/wcc-settings-form/settings-group.js
```diff
@@ -2,42 +2,54 @@ import React, { PropTypes } from 'react';
...
 				layout={ item }
-				schema={ schema }
-				storeOptions={ storeOptions }
+				{ ...{
+					key,
+					schema,
+					storeOptions,
+				} }
+				errors={ itemErrors }
```
jkudish: i think this one should go above the spread operator to help readability a bit

jkudish:
Do you think Redux has a part to play in error handling? Would it simplify passing `error` objects all over the different components?

client/components/shipping/packages/packages-list.js
jeffstieler: Yeah, it's matching the errors to the the package using the array index.

The error returned from `is-my-json-valid` looks like `packages.1.inner_dimensions`.

client/components/shipping/packages/packages-list.js
jeffstieler: Nope - that's a bug. I'll fix.

client/components/wcc-settings-form/index.js
jeffstieler: `is-my-json-valid` prepends all errors (paths to fields) with `data`, so this map removes that.

client/components/wcc-settings-form/settings-group.js
jeffstieler: It's not being removed, it's part of the spread operator below.

jkudish: if I understand the spread operator correctly, then what you're doing below becomes:

```javascript
<SettingsItem
     key={ item.key }
```

So the ternary from the existing line is removed

jeffstieler: `const key = item.key ? item.key : item;`

This ☝️  was added above the usage of the spread. As I understand it, this should function identically to the old code.

jeffstieler:
> Do you think Redux has a part to play in error handling? Would it simplify passing `error` objects all over the different components?

I initially started with that, but then landed on [this documentation](http://redux.js.org/docs/recipes/ComputingDerivedData.html) (by way of https://github.com/reactjs/redux/issues/749) that made me rethink it.

allendav:
I deleted all my instances and added a fresh instance of Canada Post.  I see this in dev console when editing it for the first time:

screen shot 2016-05-23 at 3 30 48 pm

allendav:
Immediately after the above, if I click on Add Package, I get the following:

screen shot 2016-05-23 at 3 33 13 pm

jeffstieler:
@allendav I can't reproduce your console errors. Can you try again now that I've rebased?

allendav:
@jeffstieler i don't get those errors anymore since the rebase. i do get this now:

validator-notice-keys

jeffstieler:
@allendav as we discussed, the `key` warning in `NoticesList` occurs on master as well. Is this good to go?

nabsul:
Since we don't allow the user to submit invalid form data, maybe we should relax the adjustment value field so that it allows the user to input invalid characters.

The way it works now is inconsistent with other text field above (zip, title, etc.). This would also hopefully resolve this issue: #188

allendav:
> Since we don't allow the user to submit invalid form data, maybe we should relax the adjustment value field so that it allows the user to input invalid characters.

I'd rather we didn't

allendav:
> @allendav as we discussed, the key warning in NoticesList occurs on master as well. Is this good to go?

@jeffstieler - I think so - captured key issue as a new issue here : #360

jeffstieler:
> But if we can't figure out a way to allow only certain characters in a field (like decimals) I guess we may have no choice but to relax this.

I think this is the key problem with our current implementation. I agree with @nabsul on this one, we should relax the `onChange` handler, and let the schema do the validation.

PR #350, opened 2016-05-20
Add global notices

### LINKED ISSUE #322: Global save validation: Update UI
When 'Save changes' is pressed, the success notice is within the button bar at the bottom of the page:

screen shot 2016-05-10 at 10 58 46 am

We should take advantage of [calypso's global notice](https://wpcalypso.wordpress.com/devdocs/design)  component instead. Its meant for a "page" wide notice, is fixed so it will always show, and can be configured to disappear after a set amount of time.  An example:

global notice

jkudish:
This PR seeks to replace the existing inline save/error notices and add global notices instead. It re-uses the Calypso `GlobalNotices` component and the associated state tree (reducer/actions). In order to get that to work properly, this PR also introduces some changes to the i18n boot/mixin system so that it works with `this.translate` in Calypso components properly.

Success:

image

Error:

image

In addition, this PR upgrades Calypso and all its sub-dependencies to their latest versions, by upgrading our `npm-shrinkwrap.json`

To test:

* Option one is to just run `npm install`, in theory npm should do a good job of just upgrading what it needs to. Option two is to delete the `node_modules` folder and `npm cache clean && npm install`.
* Checkout this PR
* Ensure everything compiles as expected with `npm start` and `npm run dist`
* Ensure tests pass for `npm test`
* Save the USPS form with different conditions generating error and success states
* Ensure that notices are generated as expected
* Success notices should disappear after 2250ms
* Error notices persist for 7s. I chose this semi-arbitrarily, so definitely welcome to change this up.

cc @kellychoffman for a design review (there's definitely a few small css tweaks to be made here)

cc @nabsul @allendav @jeffstieler 

 Fixes #322

### DIFF
--- a/assets/stylesheets/_components.scss
+++ b/assets/stylesheets/_components.scss
@@ -53,7 +53,7 @@
 //@import 'components/forms/range/style';
 @import 'components/forms/sortable-list/index';
 //@import 'components/gauge/style';
-//@import 'components/global-notices/style';
+@import 'components/global-notices/style';
 //@import 'components/gravatar/style';
 @import 'components/gridicon/style';
 //@import 'components/happiness-support/style';
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -139,6 +139,17 @@
         font-size: 13px;
     }
 
+    // Global notices
+    // adjust until system fonts get in wp-admin
+    .global-notices .notice__text,
+    .wcc-modal .global-notices .notice__text {
+        padding: 10px 8px 8px;
+
+        @include breakpoint( '<660px' ) {
+            padding: 16px;
+        }
+    }
+
     // Custom styles
     max-width: 700px;
 
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -143,9 +143,23 @@
     }
 
     // Global notices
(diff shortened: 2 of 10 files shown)


kellychoffman:
Nice. I tweaked the padding a bit in the main stylesheet, as I believe this to be because we are still using Open Sans and Calypso is using system fonts. I hear [system fonts are coming wp-admin](https://core.trac.wordpress.org/ticket/36753), which will better align with Calypso-esque styles.

nabsul:
Doesn't look right on my screen:

image

jkudish:
Yeah we need some better margins/max weight. @kellychoffman wanna adjust it?

jkudish:
FYI, the warning that was being generated has been fixed upstream in Calypso. If you run it with the latest, you shouldn't get any warnings.

kellychoffman:
Thanks for the heads up. None of my messages were that long when I tested :/ 

Should be all good now.

jkudish:
FYI, I updated Calypso and its sub dependencies in the latest commit, so that we get the warning fixed in master. This means this PR should be tested for any possible regressions through the client code. Please take a look :eyes:

nabsul:
Overall looks good and tests fine. :shipit:

PR #355, opened 2016-05-23
Update webpack config

### LINKED ISSUE #278: Simplify webpack config
See https://github.com/Automattic/woocommerce-connect-client/pull/273

jkudish:
This PR seeks to simplify the test webpack config so that it inherits from the main one instead and adds the only required change (node externals) by extending that base config. This helps us reduce duplication and make future maintenance easier.

In addition, this PR:

* Remove `npm run lint` from `npm test` command since webpack runs the linter already
* Remove nodeExternals require from base webpack config. It wasn't used.

To test:

* Make sure tests pass `npm run test`
* Make sure things compile correctly `npm start` and `npm run dist`

Fixes #278

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,7 +1,6 @@
 const webpack = require( 'webpack' ),
 	path = require( 'path' ),
-	ExtractTextPlugin = require( 'extract-text-webpack-plugin' ),
-	nodeExternals = require( 'webpack-node-externals' );
+	ExtractTextPlugin = require( 'extract-text-webpack-plugin' );
 
 const babelSettings = {
 	cacheDirectory: true,
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,6 +1,6 @@
-const webpack = require( 'webpack' ),
-	path = require( 'path' ),
-	ExtractTextPlugin = require( 'extract-text-webpack-plugin' );
+const webpack = require( 'webpack' );
+const path = require( 'path' );
+const ExtractTextPlugin = require( 'extract-text-webpack-plugin' );
 
 const babelSettings = {
 	cacheDirectory: true,
--- a/webpack.test.config.js
+++ b/webpack.test.config.js
@@ -1,65 +1,12 @@
-var webpack = require( 'webpack' ),
-	path = require( 'path' ),
-	nodeExternals = require( 'webpack-node-externals' );
+const webpack = require( 'webpack' );
+const path = require( 'path' );
+const nodeExternals = require( 'webpack-node-externals' );
+const webpackBaseConfig = require( './webpack.config' );
 
-module.exports = {
-	cache: true,
-	module: {
-		loaders: [
-			{
-				test: /\.json$/,
-				loader: 'json-loader'
-			},
-			{
-				test: /\.html$/,
-				loader: 'html-loader'
-			},
-			{
-				test: /\.scss$/,
-				loader: 'noop-loader',
-			},
-			{
-				test: /\.jsx?$/,
-				loader: 'babel-loader',
-				query: {
-					cacheDirectory: true,
-					presets: [
-						'es2015',
-						'stage-1',
-						'react'
-					],
-					plugins: [
-						"add-module-exports",
-					],
-					babelrc: false,
-				}
-			}
-		]
-	},
-	resolve: {
-		alias: {
-			'react': path.join( __dirname, 'node_modules', 'react' ),
-			'react-dom': path.join( __dirname, 'node_modules', 'react-dom' ),
-			'redux': path.join( __dirname, 'node_modules', 'redux' ),
-			'lib/mixins/i18n': path.join( __dirname, 'client', 'lib', 'mixins', 'i18n' )
-		},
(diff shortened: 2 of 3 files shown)


webpack.config.js
```diff
@@ -1,7 +1,6 @@
 const webpack = require( 'webpack' ),
 	path = require( 'path' ),
-	ExtractTextPlugin = require( 'extract-text-webpack-plugin' ),
```
nabsul: In most of our code, we we never use this comma-separated `const` syntax. Let's stick with the same style everywhere.

webpack.test.config.js
```diff
@@ -1,65 +1,12 @@
-var webpack = require( 'webpack' ),
+const webpack = require( 'webpack' ),
 	path = require( 'path' ),
-	nodeExternals = require( 'webpack-node-externals' );
+	nodeExternals = require( 'webpack-node-externals' )
```
nabsul: Same comment about `const`

jkudish: sure. I only used it because it was already like that in the webpack files. I'll change it up.

jkudish: Fixed in c51d178

webpack.config.js
jkudish: Fixed in c51d178

webpack.test.config.js
```diff
@@ -1,65 +1,12 @@
...
-	},
-	target: 'node',
-	externals: [
-		nodeExternals( {
-			whitelist: [ 'wp-calypso' ]
-		} )
-	]
-};
\ No newline at end of file
+webpackConfig.externals = [
```
nabsul: If I'm not mistaken, this code will actually add the `externals` member directly into the module `./webpack.config`, which will cause problems. You probably want to do an `object.assign()` here.

webpack.test.config.js
```diff
@@ -1,65 +1,12 @@
...
+	nodeExternals( {
+		whitelist: [ 'wp-calypso' ]
+	} )
+];
+
+module.exports = webpackConfig;
```
nabsul: Missing EOL

webpack.test.config.js
jkudish: I am not 100% sure that's true, but even if it was, what kind of problems do you think will produce? This `webpack.test.config.js` is only loaded for tests, so it won't interfere with the normal course of operations. Can you produce an issue with doing this?

webpack.test.config.js
jkudish: Argh, `editorconfig` was not running in phpStorm... 8c579df

webpack.test.config.js
nabsul: Are you sure `webpack.test.config.js` is only loaded in a test? Perhaps it's loaded but not used.

Either way, editing a module like this is a terrible idea:

```
// mod.js
module.exports = { test: 'hello' };

//mod1.js
const mod = require( './mod' );
mod.test2 = 'hi';
module.exports = mod;

//index.js
const mod = require( './mod' );
const mod1 = require( './mod1' );
console.log( 'mod' );
console.dir( mod );
console.log( 'mod1' );
console.dir( mod1 );

//output
PS> node .\index.js
mod
{ test: 'hello', test2: 'hi' }
mod1
{ test: 'hello', test2: 'hi' }

```

jkudish: > Are you sure webpack.test.config.js is only loaded in a test? Perhaps it's loaded but not used.

Loaded by what?

jkudish: Alright, alright. Let's play it safe :wink: 

077a355

nabsul: In this particular situation, I guess there's no risk. But 90% of the time, editing a module's contents is a bad idea. Let's avoid it.

webpack.test.config.js
```diff
@@ -1,65 +1,12 @@
-var webpack = require( 'webpack' ),
-	path = require( 'path' ),
-	nodeExternals = require( 'webpack-node-externals' );
+const webpack = require( 'webpack' );
+const path = require( 'path' );
+const nodeExternals = require( 'webpack-node-externals' );
+const webpackBaseConfig = require( './webpack.config' );
 
-module.exports = {
```
nabsul: I'm looking at this code side by side with `webpack.config`. There are some significant different. `sassLoader`, `entry`, and several other parameters were not being used. The loader for .jsx files is also different (at least missing eslint). 

Is all this intentional? Have you gone over all of these differences?

nabsul:
Looks good. Besides my questions about the differences between the base and the test config, :shipit:

webpack.test.config.js
jkudish: Those were the kinds of differences that pushed towards this "merger" in the first place. It was difficult to keep the two in sync and the differences were minor and annoying to manage.

Technically we don't need the sass to compile for tests but the time spent on it (anecdotally at least) is not significant enough to worry about. 

The main thing is that the tests continue to run as expected and aren't slow, which they aren't.

jkudish:
Rebased+squashed. Once the automated tests pass here, I'll merge.

PR #356, opened 2016-05-24
Remove outer_dimensions when it is empty

nabsul:
This field must be completely removed if it's an empty string. 

The server requires that this field either conform to the `L x W x H` format, or not be included at all. An empty string is not valid.

To test:

First confirm the error with `outer_dimensions`:
1) Try adding a new packages, but be sure to first enter a character in the `outer_dimensions` field, and then delete it. This will result in an empty string getting stored in that field.
2) See that the new package is added to the package list
3) Try saving your package and you should get a server error

Next, repeat those steps on this branch and see that everything works fine.

Also: Try adding a package without a title to see the "Untitled" text that is displayed in that situation.
Also: Try out the clickable icon for the package. It does the same thing as clicking the name.

@allendav @jkudish @jeffstieler @kellychoffman


client/components/shipping/packages/add-package.js
```diff
@@ -79,7 +87,6 @@ const useDefaultField = ( value, updatePackagesField ) => {
 		is_letter: 'envelope' === value,
 		name: null,
 		is_user_defined: true,
-		outer_dimensions: null,
```
jeffstieler: Why did you remove this line?

nabsul: In the case of no outer_dimensions, we want to delete that member altogether. If we save `packageData` with `outer_dimensions` equal to null or empty string, the form validation will fail.

nabsul: Side note: I did look into making the server more flexible, but since we're using Enjoi to generate the Joi validation, there's no good way to allow it in JSON schema.

jeffstieler: `null` values here are pruned in the reducer - I put it there as a mechanism to delete object properties by way of an action. See: https://github.com/Automattic/woocommerce-connect-client/blob/master/client/state/form/packages/reducer.js#L52

nabsul:
@jeffstieler I tried simply setting `outer_dimensions` to null instead of deleting it, but saving still fails.

However, the way it is now works.

client/components/shipping/packages/add-package.js
```diff
@@ -10,6 +10,11 @@ import AddPackagePresets from './add-package-presets';
 import { translate as __ } from 'lib/mixins/i18n';
 
 const getDialogButtons = ( mode, dismissModal, savePackage, packageData ) => {
+	// replace empty string with null so that reducer removes the field altogether
+	if ( '' === packageData.outer_dimensions ) {
+		delete packageData.outer_dimensions;
+	}
+
```
jeffstieler: I don't think `getDialogButtons()` is the right place for this logic - how about moving it to the `AddPackageDialog` render method?

nabsul:
Now that you mention it, I think the best place for this is directly in the reducer. I basically extended `isNull()` to be `isNullOrEmpty()`

nabsul:
While I was in here, I found that you can create/edit a package and remove the box name. It gets added to the list as invalid but then you can no longer re-edit it to get it fixed.

To solve this, I made the icon clickable as well.

jkudish:
I know it's not directly related to your changes but the `data` proptypes shouldn't be required for `PackagesListItem`. Right now they cause warnings if left empty (it should be a user validation but not a code-warning, if that makes sense)

client/state/form/packages/reducer.js
```diff
@@ -10,6 +10,10 @@ import {
 import omitBy from 'lodash/omitBy';
 import isNull from 'lodash/isNull';
 
+const isNullOrEmpty = ( value ) => {
```
jkudish: I think lodash's [`isEmpty`](https://lodash.com/docs#isEmpty`) would be better fit here

jkudish:
Minor comment in the code. And some more complete testing instructions would be great :+1:

client/state/form/packages/reducer.js
nabsul: Good point. I was initially thrown off by it returning true for values like `12`, but it does work fine for strings.

nabsul:
@jkudish:

> the data proptypes shouldn't be required for PackagesListItem. 

I kind of see what you mean, but actually, this is an error that happens as a result of a bug in our code. `data.name` and `data.inner_dimensions` should never be missing. Once I get #364 done, it should become impossible for this to happen.

jkudish:
The lodash seems to be causing a test to fail. Not sure why without digging in more.

> data.name and data.inner_dimensions should never be missing. Once I get #364 done, it should become impossible for this to happen.

Fair enough I suppose

nabsul:
Wow. Good thing that unit test was there. `isEmpty()` results in dropping values of `false`.

Going back to original approach.

allendav:
@nabsul @kellychoffman - what would you think about displaying "(no title)" when a box has had its title edited away but they haven't attempted saving the settings yet?

(no title) is a WordPress convention for things like untitled posts, fwiw

kellychoffman:
Sure, either that or `Untitled`, which is what we are using in Calypso.

nabsul:
@kellychoffman Please review the way "untitled" is rendered and feel free to make improvements to the formatting.

kellychoffman:
I made it italics so its different than the others, but still looks like a clickable/tappable link:

screen shot 2016-05-27 at 6 10 30 pm

kellychoffman:
The only strange thing is that after I added all the boxes in the above screenshot, I am unable to Save changes:

screen shot 2016-05-27 at 6 15 17 pm

nabsul:
Great catch! The untitled text should actually be displayed in red as a validation error, which means you can't save.






On Fri, May 27, 2016 at 5:19 PM -0700, "Kelly Hoffman" <notifications@github.com> wrote:












The only strange thing is that after I added all the boxes in the above screenshot, I am unable to Save changes:

client/state/form/packages/reducer.js
```diff
@@ -8,7 +8,8 @@ import {
 	TOGGLE_OUTER_DIMENSIONS,
 } from './actions';
 import omitBy from 'lodash/omitBy';
-import isNull from 'lodash/isNull';
+
+const isNullOrEmpty = ( value ) => null === value || '' === value;
```
jeffstieler: I think we should trim whitespace before the empty part of this check.

Putting a single space character in the outer dimensions field still results in a value being set.

jeffstieler:
Small piece of feedback, but otherwise this looks good!

jeffstieler:
Code looks good, and tests well! :ship: it.

Issue #274, opened 2016-04-28
Packages: UI for zero packages


kellychoffman:
Right now, when there aren't any packages, the UI looks a bit broken:

<img width="547" alt="screen shot 2016-04-28 at 2 57 37 pm" src="https://cloud.githubusercontent.com/assets/5835847/14901335/0447e6c2-0d52-11e6-86e8-6664c422bac3.png">

Let add some copy there when there's no packages: 

<img width="701" alt="screen shot 2016-04-28 at 2 57 29 pm" src="https://cloud.githubusercontent.com/assets/5835847/14901445/9f5ad8c2-0d52-11e6-87ba-bb2fe93f1dde.png">

PR #357, opened 2016-05-24
Add special display for when there are zero packages

nabsul:
#274

@kellychoffman

### DIFF
--- a/client/components/shipping/packages/packages-list-item.js
+++ b/client/components/shipping/packages/packages-list-item.js
@@ -11,7 +11,7 @@ const renderIcon = ( isLetter, isError ) => {
 		icon = isLetter ? 'mail' : 'flip-horizontal';
 	}
 	return (
-		<Gridicon icon={ icon } size={ isError ? 29 : 18 } />
+		<Gridicon icon={ icon } className="package-type-icon" size={ isError ? 29 : 18 } />
 	);
 }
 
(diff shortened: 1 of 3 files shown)


kellychoffman:
@nabsul: I've added the icon, messaging and the styles to go along with it. 

screen shot 2016-05-24 at 4 32 43 pm

jkudish:
What happens when a user doesn't have packages defined? I think that's what the copy should explain, and then offer to add custom/predefined ones.

nabsul:
Oh, if we have no packages, then we should force the user to select the "ship individually" radio button as well.

kellychoffman:
Mentioning the predefined packages is a good idea. 

> Oh, if we have no packages, then we should force the user to select the "ship individually" radio button as well.

What do you mean? My thought that this is the messaging they'd see when they first hit the settings page and thus haven't added any info, packages included.

nabsul:
@kellychoffman Regarding the "ship individually" comment, I'll create a separate issue around this topic. There's a couple of routes we can take and should discuss before deciding.

nabsul:
Rebased the code.

jeffstieler:
This looks good and tests well. I'd recommend splitting the package list rendering logic into a few functions.

:shipit: or refactor. Please set back to "needs review" if you change code.

PR #362, opened 2016-05-25
Allow non-float values to be entered into service adjustment inputs.

### LINKED ISSUE #188: Allow decimal service adjustments

jeffstieler:
This provides a more natural experience to a user that might be typing a decimal number, or who accidentally enters a non-digit character.

Fixes #188.

**Note: test with a USPS instance, there is currently a bug with Canada Post showing errors on service fields.**

To test:
* Enter the value `a` into a service adjustment input
* Verify that the input is marked with an error icon, and the "save changes" button is disabled
* Clear the input
* Verify that the input is marked with an error icon, and the "save changes" button is disabled
* Enter `1`
* Verify that the input is **not** marked with an error icon, and the "save changes" button is **enabled**
* Enter `1.`
* Verify that the input is marked with an error icon, and the "save changes" button is disabled
* Enter `1.1`
* Verify that the input is **not** marked with an error icon, and the "save changes" button is **enabled**

### DIFF
--- a/client/components/shipping/services/entry.js
+++ b/client/components/shipping/services/entry.js
@@ -64,7 +64,7 @@ ShippingServiceEntry.propTypes = {
 	adjustment: PropTypes.oneOfType( [
 		PropTypes.string,
 		PropTypes.number,
-	] ).isRequired,
+	] ),
 	adjustment_type: PropTypes.string.isRequired,
 	currencySymbol: PropTypes.string.isRequired,
 	updateValue: PropTypes.func.isRequired,
--- a/client/components/shipping/services/entry.js
+++ b/client/components/shipping/services/entry.js
@@ -28,9 +28,20 @@ const ShippingServiceEntry = ( {
 				disabled={ ! enabled }
 				value={ adjustment }
 				onChange={ ( event ) => {
-					const value = event.target.value ? event.target.value : 0;
-					if ( ! isNaN( value ) ) {
-						updateValue( 'adjustment', Number.parseFloat( value ) );
+					const value = event.target.value || null;
+					const floatValue = Number.parseFloat( value );
+
+					/*
+					 * If the adjustment value isn't a valid float, or ends in a non-integer, pass
+					 * the value through unmodified and let the schema validation catch it.
+					 *
+					 * If the adjustment *is* a valid float, update the settings value with the
+					 * parsed version so it doesn't fail schema validation.
+					 */
+					if ( isNaN( floatValue ) || value.match( /.*[^\d]$/ ) ) {
+						updateValue( 'adjustment', value );
+					} else {
+						updateValue( 'adjustment', floatValue );
 					}
 				} }
 				isError={ hasError }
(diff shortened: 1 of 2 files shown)


nabsul:
Let's make the following changes:

1) `1.` is valid and gets saved as `1`
2) '1.' doesn't trigger the error response
3) `ShippingServiceGroup` is going to have to accept string valued prop types so that we don't get propType error messages in the console. This is OK. We don't need the component validation to be perfect, we have validation on the server and this real-time highlighting validation.

jkudish:
Yeah I agree with all of @nabsul's points. Currently typing `1.1` results in a temporary red field which is confusing and janky.

Additionally:

* Should erasing the field result in zero?
* It's pretty common to type `.5` instead of `0.5`, currently typing `.` into the field results in a validation error
* Don't think it's a huge concern, but could there be performance implications with running the update on each change. Should this be debounced or throttled?

jeffstieler:
There is quite a bit of complexity to support `1.` and empty field == `0`, unless we're OK with changing the schema for the field to be a string. Since it's a number, even a numeric string value causes the validation to fail.

In addition, the string value `1.` gets parsed to a number with the value `1`, so we can't send the parsed value through to the redux state tree if it ends in a `.`, because then we're back to the odd "insert a period in between two numbers" flow. (in order to enter `1.1` you first have to type `11`, then insert a `.` between)

@nabsul @jkudish

jkudish:
Yeah I realize it's complex. I would suggest maybe shipping this PR more or less as-is and then keep iterating on this later on. The current experience is fine for now just would be best to have something more polished eventually.

jeffstieler:
> Yeah I realize it's complex. I would suggest maybe shipping this PR more or less as-is and then keep iterating on this later on. The current experience is fine for now just would be best to have something more polished eventually.

I agree - so are we good to :ship: this?

jkudish:
We should fix the prop warnings now, I think.

jeffstieler:
> We should fix the prop warnings now, I think.

Which ones? There are several in `master` that I think an issue exists for already.

jkudish:
> 3) ShippingServiceGroup is going to have to accept string valued prop types so that we don't get propType error messages in the console.

jkudish:
Removing the value still results in a warning:

image

jeffstieler:
> Removing the value still results in a warning:

Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/362/commits/8ed0665f4f8f30c0e321c0672f0a72868cadf8dd.

jkudish:
Looks and works better.

A non-blocker question: should we be allowing negative values? (we currently do)

jeffstieler:
> should we be allowing negative values? (we currently do)

I'd think not, but you never know?

PR #364, opened 2016-05-26
Add validation inside the add/edit packages Modal

### LINKED ISSUE #297: Packages form element allows duplicate package names
image

### LINKED ISSUE #306: We should validate fields in the modal before leaving the modal

### LINKED ISSUE #324: Packages: Make dimension spacing consistent
Right now, when you add your package dimensions, it takes whatever spacing you add. Lets make it consistent like the original design: 

screen shot 2016-05-10 at 12 20 33 pm

This is how its currently behaving:

dimension spacing

cc: @nabsul

nabsul:
I've added validation to the add/edit packages modal. It does the following:
1) Check if name field is empty 
2) Check if box name already exists in the list of boxes
3) requires inner dimensions and checks the format
4) if outer dimensions are provided, checks the format
5) makes sure box and max weights are numbers and not empty
6) Auto-correct spacing around "x" in the dimensions definitions

To Test:

1) Open USPS or Canada Post settings
2) Make sure you have a few boxes entered
3) Try to add a new box
4) Make sure it follows all the rules above
5) Try editing an existing box
6) Make sure it follows the rules above
7) Make sure you're able to save data after making changes
8) Experiment with variations of spacing around the "x" in the dimensions. It should be auto-corrected.

@kellychoffman @jeffstieler @allendav @jkudish

Closes #297 
Closes #306 
Closes #324

### DIFF
--- a/client/components/shipping/packages/add-package.js
+++ b/client/components/shipping/packages/add-package.js
@@ -3,6 +3,7 @@ import FormSectionHeading from 'components/forms/form-section-heading';
 import FormFieldset from 'components/forms/form-fieldset';
 import FormLabel from 'components/forms/form-label';
 import FormTextInput from 'components/forms/form-text-input';
+import FormInputValidation from 'components/forms/form-input-validation';
 import FormSettingExplanation from 'components/forms/form-setting-explanation';
 import FormButton from 'components/forms/form-button';
 import Dialog from 'components/dialog';
@@ -10,9 +11,9 @@ import AddPackagePresets from './add-package-presets';
 import { translate as __ } from 'lib/mixins/i18n';
 import { sprintf } from 'sprintf-js';
 
-const getDialogButtons = ( mode, dismissModal, savePackage, packageData ) => {
+const getDialogButtons = ( mode, dismissModal, savePackage, packageData, error ) => {
 	return [
-		<FormButton onClick={ () => savePackage( packageData ) }>
+		<FormButton onClick={ () => savePackage( packageData ) } disabled={ error }>
 			{ ( 'add' === mode ) ? __( 'Add package' ) : __( 'Apply changes' ) }
 		</FormButton>,
 		<FormButton onClick={ () => dismissModal() } isPrimary={ false }>
@@ -113,12 +114,14 @@ const AddPackageDialog = ( props ) => {
 		is_user_defined,
 	} = packageData;
 
+	const nameError = '' === name ? 'Name cannot be empty' : null;
+
 	return (
 		<Dialog
 			isVisible={ showModal }
 			additionalClassNames="wcc-modal wcc-shipping-add-edit-package-dialog"
 			onClose={ dismissModal }
-			buttons={ getDialogButtons( mode, dismissModal, savePackage, packageData ) }>
+			buttons={ getDialogButtons( mode, dismissModal, savePackage, packageData, nameError ) }>
 			<FormSectionHeading>{ ( 'edit' === mode ) ? __( 'Edit package' ) : __( 'Add a package' ) }</FormSectionHeading>
 			{ ( 'add' === mode ) ? (
 				<AddPackagePresets
@@ -137,7 +140,9 @@ const AddPackageDialog = ( props ) => {
 					placeholder={ __( 'The customer will see this during checkout' ) }
 					value={ name }
 					onChange={ ( event ) => updateFormTextField( event, updatePackagesField ) }
+					isError={ nameError }
 				/>
+				{ nameError ? <FormInputValidation isError text={ nameError } /> : '' }
 			</FormFieldset>
 			<FormFieldset>
 				<FormLabel>{ sprintf( __( 'Inner Dimensions (L x W x H) %s' ), dimensionUnit ) }</FormLabel>
--- a/client/components/shipping/packages/add-package.js
+++ b/client/components/shipping/packages/add-package.js
@@ -12,6 +12,7 @@ import { translate as __ } from 'lib/mixins/i18n';
 import { sprintf } from 'sprintf-js';
 import modalErrors from './modal-errors';
 import difference from 'lodash/difference';
+import trim from 'lodash/trim';
 
 const getDialogButtons = ( mode, dismissModal, savePackage, packageData, error ) => {
 	return [
@@ -53,6 +54,12 @@ const updateFormTextField = ( event, updatePackagesField ) => {
 	updatePackagesField( { [name]: value } );
 };
 
+const FieldInfo = ( { isError, text } ) => {
+	return isError
+		? <FormInputValidation isError text={ text } />
+		: <FormInputValidation icon="info" text={ text } />;
+};
+
 const renderOuterDimensions = ( showOuterDimensions, dimensionUnit, packageData, value, updatePackagesField, is_user_defined, error ) => {
 	return ( showOuterDimensions || packageData.outer_dimensions ) ? (
 		<FormFieldset>
@@ -66,6 +73,7 @@ const renderOuterDimensions = ( showOuterDimensions, dimensionUnit, packageData,
 				disabled={ ! is_user_defined }
 				isError={ error }
 			/>
+			<FieldInfo isError={ error } text="Outer dimensions of the box are required" />
 		</FormFieldset>
 	) : null;
 };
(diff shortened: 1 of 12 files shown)


kellychoffman:
I went to test this as a new person who hadn't added any packages yet. Upon clicking the 'Add packages' button, I was presented with all the errors: 

screen shot 2016-05-26 at 4 00 01 pm

Can this only happen after I try and fail to enter in values?

kellychoffman:
> 1) Right now there is no text warnings for any of the fields besides "name"
> 2) I think the red highlighting around the fields is clear enough, do we really want text?
> 3) The text for the name field seems necessary (red in the case of duplicate names isn't quite obvious)...

Yeah, its a bit strange to have some and not others. Thinking about it from a user's perspective, I might get frustrated if I didn't know 'why' I had an error. My vote would be text for everything.

> ... but I also feel that it makes the modal box "jerk" quite a bit when it comes and goes. Suggestions?

Maybe it would it be worth it to add helper text underneath. We could move the current placeholder text as helper text and the placeholder could repeat the title or just 'Name'?

jeffstieler:
Is there a reason you didn't use the `is-my-json-valid` module to validate?

nabsul:
@jeffstieler 
> Is there a reason you didn't use the is-my-json-valid module to validate?

Yes! I went down that path but ran into several issues:
1) No easy way to also check for duplicate path names
2) Weight fields are strings, which always cause the validator to fail
3) When you add a character to `outer_dimensions` and then delete that character, the validator continues to fail because it doesn't consider the empty string to be the same as null.

The whole thing got so convoluted I gave up and decided to do it this way instead. Overall, this validation is very specific to this component's special behavior needs, so I think this approach of building it into the component is ok.

jeffstieler:
> 1) No easy way to also check for duplicate path names

JSON Schema has `uniqueItems`, which is supported by `is-my-json-valid`.

> 2) Weight fields are strings, which always cause the validator to fail

Just have to parse them to `Number`s.

nabsul:
@jeffstieler: The uniqueItems would require us to add the new item to the array, validate, and then revert if validation fails. I expect getting it all to work would take me at least another 2 days and would include updates to the server-side settings schema. I don't think it's worth the effort.

jeffstieler:
> The uniqueItems would require us to add the new item to the array, validate, and then revert if validation fails. I expect getting it all to work would take me at least another 2 days and would include updates to the server-side settings schema. I don't think it's worth the effort.

Hmm ok - I forgot that we weren't actually adding items to the `boxes` array until the "save" button was clicked.

jeffstieler:
@nabsul I feel like you could still use the `is-my-json-valid` module to validate a yet-to-be-added package. You're passing the `boxes.item` schema into the `Packages` component, so you have it available.

This would enable you to remove a lot of new code - you'd only need to parse the weight fields as numbers, and do the unique name check yourself.

What do you think?

nabsul:
Yep, I saved a full 10 lines of code.

nabsul:
@kellychoffman: What do you think about this approach? 

image

kellychoffman:
Green usually means you successfully performed an action. I would keep them red with the `!` but only after you have tried and failed to enter something, or skipped the field entirely.

My concern wasn't with all the errors, it was when the errors where appearing: on open and before I did anything.

nabsul:
Thanks for the feedback! I think I'm ready for another round of review (both user testing and code review)

nabsul:
@kellychoffman When "box weight" is invalid and "max weight" is valid, the gray text underneath gets squished to the side. I wasn't sure how to fix this. Could you take a look?

kellychoffman:
>  When "box weight" is invalid and "max weight" is valid, the gray text underneath gets squished to the side. I wasn't sure how to fix this. Could you take a look?

Fixed this in 4875d47. Also cleaned up the error styles so they are better aligned.

kellychoffman:
I noticed one weird thing: I get an error on the dimensions if I don't add a space in between the `x`s. 

x issue

nabsul:
I have some ideas to resolve this, but it's beyond the scope of this PR.

kellychoffman:
@nabsul Sounds good. Added a new issue: https://github.com/Automattic/woocommerce-connect-client/issues/377

nabsul:
I'm going to add the autocorrect here. It'll be easier to test it all together.

jkudish:
I played around with this a bunch just now and tried to break it :stuck_out_tongue: 

Overall this works pretty well and is pretty sleek. I did manage to get into some odd edge-case situations though.

After adding a package, when I went to add a second package I got this:

image

If I canceled and tried again, then it got reset properly.

---

When I edit a value incorrectly, I get a message about the field being required, which is not exactly accurate:

image

Also, shouldn't `.3` be allowed? It should just convert to 0.3 (like we do for service adjustments)

---

Same thing here, it's not the right message:

image

---

I expanded to edit the exterior dimensions and as I started to type, it gave me this:

image

which feels unfair given I was still typing.

---

After playing around with the modal for a few minutes, there started to be a lag when typing. Too many change events or something not getting garbage collected?

--- 

Why is max weight required? Couldn't a box accept infinitely heavy items?

nabsul:
Great catches @jkudish. Not sure if I'll fix all of them, but I'll see what I can do.

nabsul:
@jkudish I've resolved everything except for the comment about max_weight being required. That change should be discussed separately and would anyways need to be changed at the server.

allendav:
One oddity that we could probably punt to a new issue is that the dialog isn't cleared out between Cancel and starting a new Add flow:

reset

client/components/shipping/packages/add-package.js
```diff
@@ -113,45 +67,95 @@ const AddPackageDialog = ( props ) => {
...
+							placeholder={ exampleDimensions }
+							value={ outer_dimensions }
+							className={ fieldClassName }
+							onChange={ updateTextField }
+							disabled={ ! is_user_defined }
+							isError={ modalErrors.outer_dimensions }
+						/>
+						{ fieldInfo( 'outer_dimensions' ) }
+					</FormFieldset> )
+				: ''
```
allendav: Why empty string and not `null` like we typically do?

allendav:
Leading spaces confuse the logic. Please trim whitespace to avoid this. I've seen this flummox clients and support engineers in the past.

screen shot 2016-06-06 at 12 43 22 pm

client/components/shipping/packages/input-filters.js
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	return value;
+};
+
+const dimensionRegex = /^(\S+)\s*x\s*(\S+)\s*x\s*(\S+)$/;
+const dimensions = ( value ) => {
+	const result = dimensionRegex.exec( value );
+	if ( result ) {
+		return result.splice( 1, 4 ).map( number ).join( ' x ' );
```
allendav: I'm really trying to understand this. So, the regex exec will return something like 

`["5 x 6 x 66", "5", "6", "66", index: 0, input: "5 x 6 x 66"]`

and then splice( 1,  4 ) yields

`["5", "6", "66"]`

but that doesn't make sense to me - I thought splice( 1,  4 ) would remove 4 elements from the array starting with the 1st element?

I mean this works, but I don't understand why. Can you explain? :)

allendav:
@nabsul - a handful of comments and questions for you

client/components/shipping/packages/input-filters.js
nabsul: I don't know what documentation I was reading, but it said that the first parameter is the start index you want to keep, and the last parameter was the index where you want to stop.

I'm just as baffled as you are about how/why this is working. I'm going to simplify it into something more understandable.

client/components/shipping/packages/add-package.js
nabsul: I didn't know we had a rule about this.

nabsul:
@allendav The behavior of not clearing the contents of the modal is intentional. The thinking was that the user might exit the modal accidentally, and it would be nice if they didn't lose all their typing in the process.

client/components/shipping/packages/add-package.js
allendav: Not a rule - this is just different

allendav:
Almost there. One last nit: attempting to create a package with a leading space for the dimension results in a validation warning after saving the package:

super-box

allendav:
Nevermind the space problem above - my copy of the branch was 1 commit behind :shipit:

PR #366, opened 2016-05-26
Fix error field path parsing.

### LINKED ISSUE #365: Canada Post service errors not displayed
Entering bad values for Canada Post service adjustments results in validation failure, but no UI response.

The current parsing strategy of string splitting on `.` doesn't work for validation results such like: `services["DOM.EP"]`.

jeffstieler:
Fixes #365.

To test:
* Cause any validation error for any field, make sure error behavior is unchanged
* Edit a servive adjustment value in a Canada Post instance to be a non-number, like `a` (edit in DB)
* Reload settings page
* Verify that the service group with the errant value is expanded, and the adjustment input decorated with an error


client/components/wcc-settings-form/index.js
```diff
@@ -79,10 +80,11 @@ const getFormErrors = ( schema, data ) => {
 		 * This removes the `data.` prepending all errors, to facilitate easier matching to form fields.
 		 */
 		return validate.errors.map( ( error ) => {
-			if ( 0 === error.field.indexOf( 'data.' ) ) {
-				return error.field.substr( 5 );
+			const errorPath = ObjectPath.parse( error.field );
```
nabsul: Can't this all be done in one line?  `error.field.replace( 'data.', '' )`

nabsul: Does `ObjectPath.parse()` do anything more than `error.field.split('.')` ?

npm-shrinkwrap.json
```diff
@@ -2429,6 +2429,11 @@
       "version": "2.0.0",
       "from": "object.omit@>=2.0.0 <3.0.0"
     },
+    "objectpath": {
```
nabsul: If you decide to remove ObjectPath, don't forget to undo this.

client/components/wcc-settings-form/index.js
jeffstieler: > Does `ObjectPath.parse()` do anything more than `error.field.split('.')` ?

Yes. It also normalizes the path string. See: https://www.npmjs.com/package/objectpath

jeffstieler: This PR solves a problem with the usage of `.split( '.' )` - the service IDs for Canada Post contain `.` characters.

nabsul:
Works great :shipit:

PR #368, opened 2016-05-27
hide packing method when there are no packages

### LINKED ISSUE #361: Packing method radio button when there are no packages defined.
image

When there are no packages defined, the radio buttons above the package list make no sense. We could:

1) Just hide it altogether.
2) Gray it out and set the value to "pack individually"
3) Just leave it as is for simplicity.

Suggestions?

@kellychoffman @allendav @jeffstieler @jkudish @DanReyLop

nabsul:
closes #361 

This change hides the packing method radio buttons when there are no boxes defined. The reason is that this parameter has no effect on settings unless there is at least one package defined in packages.

I also went ahead and did some cleanup to the React component:
- redux `connect()` is now only done once in index.js
- Update proptypes to properly reflect properties used by the subcomponents`

To Test:
1) Go to settings page (Canada Post or USPS)
2) Observer that the radio buttons are hidden if there are no packages defined
3) Try saving, refreshing the page, and cancelling a few actions
4) Make sure unit tests pass and no errors are thrown in the console.

Here are sample images of what you should see:
image
image

### DIFF
--- a/client/components/wcc-settings-form/index.js
+++ b/client/components/wcc-settings-form/index.js
@@ -10,18 +10,18 @@ import isString from 'lodash/isString';
 import { translate as __ } from 'lib/mixins/i18n';
 import validator from 'is-my-json-valid';
 import ObjectPath from 'objectpath';
+import * as SettingsActions from 'state/settings/actions';
+import * as PackagesActions from 'state/form/packages/actions';
 
-const WCCSettingsForm = ( {
-	storeOptions,
-	schema,
-	layout,
-	settings,
-	form,
-	saveFormData,
-	formActions,
-	noticeActions,
-	errors,
-} ) => {
+const WCCSettingsForm = ( props ) => {
+	const {
+		layout,
+		settings,
+		saveFormData,
+		formActions,
+		noticeActions,
+	} = props;
+
 	const setIsSaving = ( value ) => formActions.setField( 'isSaving', value );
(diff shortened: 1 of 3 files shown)


client/components/wcc-settings-form/index.js
```diff
@@ -10,18 +10,18 @@ import isString from 'lodash/isString';
 import { translate as __ } from 'lib/mixins/i18n';
 import validator from 'is-my-json-valid';
 import ObjectPath from 'objectpath';
+import * as SettingsActions from 'state/settings/actions';
```
allendav: This is a neat syntax that I didn't know about before this. Cool :)

client/components/wcc-settings-form/settings-item.js
```diff
@@ -3,10 +3,6 @@ import Indicators from 'components/indicators';
 import TextField from 'components/text-field';
 import RadioButtons from 'components/radio-buttons';
 import ShippingServiceGroups from 'components/shipping/services';
-import { connect } from 'react-redux';
-import * as SettingsActions from 'state/settings/actions';
-import * as PackagesActions from 'state/form/packages/actions';
-import { bindActionCreators } from 'redux';
```
allendav: Nice clean up - removing redux from this component is a win

allendav:
Tests nicely. :shipit:

PR #369, opened 2016-05-27
Add hook to update schemas when woocommerce settings are updated

### LINKED ISSUE #261: Update service settings when woocommerce settings get changed
The scenario we're thinking about specifically is if the user changes their weight or dimension units. The client will need to re-fetch the service settings (for example, the preset box sizes) so that the server returns new values with the correct units.

nabsul:
closes #261 

If settings any woo commerce settings are changed, we should re-fetch settings from the server. This is because the default values and settings will change if for example the units of measurements are changed.

To Test:

- Load this branch
- Enable logging
- Make sure your frequent fetch flag is NOT enabled
- Save your woocommerce settings (you don't need to change the settings)
- Observer that a `fetch_service_schemas_from_connect_server` event is logged in the logs

@allendav @jkudish @jeffstieler @DanReyLop


jeffstieler:
LGTM and tests well! :shipit:

PR #370, opened 2016-05-27
Add select all in services selection

### LINKED ISSUE #198: UX Improvement: add a check all checkbox for services
From @kellychoffman:

There's a bulk select component in calypso that we could take advantage of

I think it would look best if instead of the count, we had the column title there. Design:

Originally posted here by accident: https://github.com/Automattic/woocommerce-connect-server/issues/197

nabsul:
closes #198 

This adds a "select-all" checkbox in the header of the service foldable card, like so:
image

To Test: Check out his branch and try some of the following scenarios:
- Check-all when none/some/all services are already checked
- Un-check-all
- Saving the form and making sure the check-all renders correctly upon initialization

@DanReyLop @jeffstieler @jkudish @allendav @kellychoffman

### DIFF
--- a/client/components/shipping/services/group.js
+++ b/client/components/shipping/services/group.js
@@ -1,6 +1,7 @@
 import React, { PropTypes } from 'react';
 import ShippingServiceEntry from './entry';
 import FoldableCard from 'components/foldable-card';
+import CheckBox from 'components/forms/form-checkbox';
 import Gridicon from 'components/gridicon';
 import { translate as __ } from 'lib/mixins/i18n';
 import { sprintf } from 'sprintf-js';
@@ -14,6 +15,10 @@ const summaryLabel = ( services ) => {
 	}
 	const format = ( 1 === numSelected ) ? __( '%d service selected' ) : __( '%d services selected' );
 	return sprintf( format, numSelected );
+};
+
+const getCheckbox = ( title ) => {
+	return <div><CheckBox />{ title }</div>
 };
 
 const ShippingServiceGroup = ( {
@@ -33,7 +38,7 @@ const ShippingServiceGroup = ( {
 	);
 	return (
 		<FoldableCard
-			header={ title }
+			header={ getCheckbox( title ) }
 			summary={ summary }
 			expandedSummary={ summary }
 			clickableHeader={ true }
--- a/client/components/shipping/services/group.js
+++ b/client/components/shipping/services/group.js
@@ -17,8 +17,19 @@ const summaryLabel = ( services ) => {
 	return sprintf( format, numSelected );
 };
 
-const getCheckbox = ( title ) => {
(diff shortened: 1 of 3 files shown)


nabsul:
@jkudish @jeffstieler @kellychoffman @allendav @DanReyLop 

I'm a little stuck here: The select all checkbox logic is working, but it's also triggering expand/contract of the foldable card. What's the best way to prevent this from happening?

DanReyLop:
Adding a ```return false;``` to the onClick event handler should do the trick. If it doesn't work, try ```event.stopPropagation();``` , because I'm quite sure that React does ```stopPropagation``` for you when you return ```false```, but not 100% sure.

DanReyLop:
Looks like my knowledge is outdated. You should call ```event.stopPropagation()``` manually.

Here's what the React docs say:

Note:
As of v0.12, returning false from an event handler will no longer stop event propagation. Instead, e.stopPropagation() or e.preventDefault() should be triggered manually, as appropriate.

nabsul:
Thanks @DanReyLop

kellychoffman:
Sorry if the design in #198 was unclear. The multi select checkbox was supposed to be within the content inside a header to also make room for labeling the [price adjustment so that was more clear](https://github.com/Automattic/woocommerce-connect-client/issues/219). 

screen shot 2016-06-01 at 10 28 41 am

While I think the current way you have it implemented will save you a click, we were informed that most people don't select all the options in a grouping. We also talked about auto expanding the first box. There's also a confusing thing that happens when you only have a few selected: you see an _unselected_, but then also read '5 services _selected_' to the right of it:

screen shot 2016-06-01 at 10 19 15 am

nabsul:
@kellychoffman: Can you please take a look at the code now? The select-all check box is in the right place, but it'll still need some styling.

kellychoffman:
Styled what was there. Should the Price Adjustment header be added in here or a different PR?

nabsul:
Let's definitely take care of Price adjustment as well. I've added a placehold span for the text.

kellychoffman:
👍  Price adjustment styled.

client/components/shipping/services/group.js
```diff
@@ -16,21 +18,28 @@ const summaryLabel = ( services ) => {
...
+		errors,
+	} = props;
 	const summary = summaryLabel( services );
 	const actionButton = (
 		<button className="foldable-card__action foldable-card__expand" type="button">
 			<span className="screen-reader-text">{ __( 'Expand Services' ) }</span>
 			<Gridicon icon="chevron-down" size={ 24 } />
 		</button>
 	);
+	const allChecked = every( services, ( service ) => service.enabled );
```
allendav: Nice touch

allendav:
Tests nicely. :shipit:

PR #374, opened 2016-05-31
Show errors from the server inline.

jeffstieler:
This overrides the calculated client errors with values from state (which are from the server).

This PR seeks to display errors returned from the server inline on the client form.

It changes the form state `error` property to `errors` and sets the value to the array returned by the server during validation failure.

The form component's error property now defaults to this `form.errors` state, falling back to any calculated client-side errors if none are found in state.

Upon interacting with the form and causing a `state.settings` update, the `form.errors` state is cleared, allowing client-side validation to take over.

To test:
- Use this testing branch of the server (forces "title" to error) https://github.com/Automattic/woocommerce-connect-server/tree/test/205-server-errors-on-client
- Submit an error-free settings form
- Verify that an error notice is displayed and the title field shows a validation message
- Add a character to the title field
- Verify that the title error disappears

### DIFF
--- a/client/components/wcc-settings-form/index.js
+++ b/client/components/wcc-settings-form/index.js
@@ -7,6 +7,7 @@ import { bindActionCreators } from 'redux';
 import * as FormActions from 'state/form/actions';
 import { successNotice, errorNotice } from 'state/notices/actions';
 import isString from 'lodash/isString';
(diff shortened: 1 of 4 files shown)


jkudish:
Works and tests as expected. Code looks good too. :shipit:

PR #378, opened 2016-06-01
auto-correct dimensions string

### LINKED ISSUE #324: Packages: Make dimension spacing consistent

### LINKED ISSUE #377: Packages: Error on dimensions when no space is present
Originally brought up here: https://github.com/Automattic/woocommerce-connect-client/pull/364#issuecomment-222883503

 I get an error on the dimensions if I don't add a space in between the `x`s.

nabsul:
closes #377 
closes #324
This change causes the dimension strings to be auto-correct for number of spaces in the " x " string that's supposed to separate the dimensions.

To test: Open up an add/edit modal and observe the auto correction on the strings such as the following:
- "1x1x2"
- "1x 1x2"
- "1x                 1                    x                         2"

Notes:
- The code seems to be a little laggy. Any suggestions to improve performance?
- It really only corrects the separating X's, nothing else. So it'll also autocorrect: "AxBxc"
- This will make a lot more sense once #364 is merged

@kellychoffman @DanReyLop @jkudish @jeffstieler @allendav

### DIFF
--- a/client/components/shipping/packages/add-package.js
+++ b/client/components/shipping/packages/add-package.js
@@ -50,6 +50,24 @@ const updateFormTextField = ( event, updatePackagesField ) => {
 	updatePackagesField( { [name]: value } );
 };
 
+const dimensionRegex = /^(\S+)\s*x\s*(\S+)\s*x\s*(\S+)$/;
+const dimensionStringFilter = ( dims ) => {
+	const result = dimensionRegex.exec( dims );
+	if ( result ) {
+		return result[1] + ' x ' + result[2] + ' x ' + result[3];
+	}
+
+	return dims;
+};
+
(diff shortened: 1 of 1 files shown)


kellychoffman:
Cool. I like how it auto adds the spaces. 

Only weird thing is when I move the cursor to the right of one of the `x`s and backspace, my cursor jumps to the end of the string.

nabsul:
I think I have a solution that resolves both the performance and the cursor position issue. I've made it so that the filter is run when the save button is clicked.

nabsul:
Moved to the modal validation PR

PR #380, opened 2016-06-01
Translate measurement units

### LINKED ISSUE #359: Figure out how to translate measurement units
Right now, we are outputting "in" and "lbs", etc. without any translation.

Should this be done via the calypso translator? Or from the server?

nabsul:
Closes #359 

This will provide the WooCommerce translation of dimension and weight units.

To Test:
- Switch your site language to French (not all languages have translations of the units, French does)
- See that the new unit names are displayed in the box packer settings of USPS and CanadaPost

@jeffstieler @allendav @jkudish @DanReyLop

### DIFF
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -37,8 +37,8 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 
 			return array(
 				'currency_symbol' => $currency_symbol,
-				'dimension_unit' => $dimension_unit,
-				'weight_unit' => $weight_unit
+				'dimension_unit' => $this->translate_unit( $dimension_unit ),
+				'weight_unit' => $this->translate_unit( $weight_unit ),
 			);
 		}
 
@@ -159,6 +159,32 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 			}
 
 			return 'woocommerce_' . $service_id . '_' . $service_instance . '_form_settings';
+		}
+
+		private function translate_unit( $value ) {
+			switch ( $value ) {
+				case 'kg':
+					return __('kg', 'woocommerce');
+				case 'g':
+					return __('g', 'woocommerce');
+				case 'lbs':
+					return __('lbs', 'woocommerce');
+				case 'oz':
+					return __('oz', 'woocommerce');
+				case 'm':
+					return __('m', 'woocommerce');
+				case 'cm':
+					return __('cm', 'woocommerce');
+				case 'mm':
+					return __('mm', 'woocommerce');
+				case 'in':
+					return __('in', 'woocommerce');
+				case 'yd':
+					return __('yd', 'woocommerce');
+				default:
+					error_log( 'Unexpected measurement unit: ' . $value );
+					return $value;
+			}
 		}
 	}
 }
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -182,7 +182,7 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 				case 'yd':
 					return __('yd', 'woocommerce');
 				default:
-					error_log( 'Unexpected measurement unit: ' . $value );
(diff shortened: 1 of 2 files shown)


nabsul:
image

@kellychoffman: It seems that longer weight unit names cause some issues for the box and max weight fields. Can you take a look?

kellychoffman:
@nabsul: I added room for ~6 characters. Are there units with more?

nabsul:
@kellychoffman 
Let me see if I can find that out.

nabsul:
@kellychoffman  And the winner is: `lbs (ポンド)`

kellychoffman:
Looking good at all the screen sizes now. 

screen shot 2016-06-02 at 11 03 42 pm
screen shot 2016-06-02 at 11 04 05 pm

nabsul:
Sorry @kellychoffman. Still fails for Japanese :-(

image

nabsul:
The only thing left to tweak is the display of longer weight units in different languages. So I think this is fine to review already.

kellychoffman:
Hmm, its showing up fine for me even with the parenthesis added, which I failed to account for before. I added a bit more wiggle room. Let me know if you are still seeing breakage. 

screen shot 2016-06-03 at 11 24 14 am

classes/class-wc-connect-service-settings-store.php
```diff
@@ -160,5 +160,31 @@ protected function get_service_settings_key( $service_id, $service_instance = fa
 
 			return 'woocommerce_' . $service_id . '_' . $service_instance . '_form_settings';
 		}
+
+		private function translate_unit( $value ) {
+			switch ( $value ) {
+				case 'kg':
+					return __('kg', 'woocommerce');
+				case 'g':
+					return __('g', 'woocommerce');
```
allendav: Let's use _x here ( https://codex.wordpress.org/Function_Reference/_x )

classes/class-wc-connect-service-settings-store.php
```diff
@@ -160,5 +160,31 @@ protected function get_service_settings_key( $service_id, $service_instance = fa
...
+				case 'lbs':
+					return __('lbs', 'woocommerce');
+				case 'oz':
+					return __('oz', 'woocommerce');
+				case 'm':
```
allendav: Let's use _x here ( https://codex.wordpress.org/Function_Reference/_x )

classes/class-wc-connect-service-settings-store.php
```diff
@@ -160,5 +160,31 @@ protected function get_service_settings_key( $service_id, $service_instance = fa
...
+				case 'cm':
+					return __('cm', 'woocommerce');
+				case 'mm':
+					return __('mm', 'woocommerce');
+				case 'in':
+					return __('in', 'woocommerce');
+				case 'yd':
+					return __('yd', 'woocommerce');
+				default:
+					error_log( 'Unexpected measurement unit: ' . $value );
```
allendav: We should use the WCC logger here

allendav:
Looks like this still needs _x and rebasing too

classes/class-wc-connect-service-settings-store.php
nabsul: @allendav: If we do this, we'll break away from the standard used in WooCommerce itself. I think we should stick with the way it's implemented now, unless this changes in the woocommerce plugin.

```
						<select id="dimension_unit" name="dimension_unit" class="wc-enhanced-select">
							<option value="m" <?php selected( $dimension_unit, 'm' ); ?>><?php echo __( 'm', 'woocommerce' ); ?></option>
							<option value="cm" <?php selected( $dimension_unit, 'cm' ); ?>><?php echo __( 'cm', 'woocommerce' ); ?></option>
							<option value="mm" <?php selected( $dimension_unit, 'mm' ); ?>><?php echo __( 'mm', 'woocommerce' ); ?></option>
							<option value="in" <?php selected( $dimension_unit, 'in' ); ?>><?php echo __( 'in', 'woocommerce' ); ?></option>
							<option value="yd" <?php selected( $dimension_unit, 'yd' ); ?>><?php echo __( 'yd', 'woocommerce' ); ?></option>
						</select>
```

allendav:
:shipit:

Issue #363, opened 2016-05-26
Add a package modal cut off on short browsers


jkudish:
Minor browser quirk:

![image](https://cloud.githubusercontent.com/assets/260253/15560128/53069a3c-229d-11e6-8999-4092d4da87f8.png)

The modal is cut off when I have the developer console, and it would be the same if the user had a short browser window for some reason. We should make sure it let's you scroll to the buttons

allendav:
Yeah, this is annoying. Please also try/fix mobile.

nabsul:
Does anyone know off the top of their head the css needed to fix this? Or should I research it?

@kellychoffman @allendav @jeffstieler @jkudish

kellychoffman:
I think we should do what we're doing in Calypso when you add media:

I think the only major change we'd need to make is: absolutely positioning the modal and the bottom save container. And doing a bunch of percentage based height and width adjustments. I did it quick in the browser and it works pretty good. I can take this one over @nabsul.

![quick mock](https://cloud.githubusercontent.com/assets/5835847/15622968/d0f05c1e-242b-11e6-8768-4b4685d9abcb.gif)

PR #382, opened 2016-06-03
Dialog card: Always display action buttons even on short screens

kellychoffman:
Fixes: #363 

Before:
screen shot 2016-06-03 at 4 55 41 pm

After:
scroll baby scroll

To test, open the Add a package modal and resize your browser so the height is small. The action buttons should still display. 

The caveat here is that the height will be long if you have a tall browser. I think thats fine. I didn't want to restrict the height because some modals will be long, and this modal in particular has two different heights depending if the exterior dimensions are present.

### DIFF
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -120,15 +120,39 @@
 
     // Dialog
     &.dialog.card {
-        padding: 0;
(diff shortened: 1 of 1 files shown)


nabsul:
Works great! :shipit:

Issue #336, opened 2016-05-11
If Jetpack is_staging_site, add notice to health page and don't bump stats


jkudish:
Jetpack has a `Jetpack::is_staging_site()` function which, if true, should prevent "syncing" of the site to WP.com. I don't think we should consider it as a full "killswitch" flag for WCC, but we may want to scale some things back (like not bumping any stats) when it returns true, and perhaps showing that on the health page (e.g. "This is a staging site, some features may not work as expected") (cc @allendav) as well.

h/t @kraftbj

allendav:
I agree. Let's kill bumping stats and add this to the health page. Changing this question to bug.

PR #384, opened 2016-06-06
Disable bumping stats when the site is a Jetpack staging site

DanReyLop:
Solves #336

When the site is a Jetpack staging site (```Jetpack::is_staging_site()``` returns ```true```), the requests sent to the WCC server will have a flag to prevent stats to be bumped for the site.

Also, a notice in the health page will be displayed.

I'm not sure how to test this, since I don't know how to put a site in "staging" mode, I just put a ```return true;``` at the start of ```Jetpack::is_staging_site()``` and check that the notice on the Health page is shown, and that all the requests that the client sends to the WCC server have ```settings.enable_stats = false```.

@jeffstieler @allendav @nabsul

### DIFF
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -193,6 +193,7 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 				'last_heartbeat' => get_option( 'wc_connect_last_heartbeat', 0 ),
 				'last_rate_request' => get_option( 'wc_connect_last_rate_request', 0 ),
 				'active_services' => $this->wc_connect_loader->get_active_services(),
+				'disable_stats' => Jetpack::is_staging_site(),
 			) );
 
 			$body = wp_json_encode( apply_filters( 'wc_connect_api_client_body', $body ) );
--- a/classes/class-wc-connect-help-provider.php
+++ b/classes/class-wc-connect-help-provider.php
@@ -145,6 +145,14 @@ if ( ! class_exists( 'WC_Connect_Help_Provider' ) ) {
 					__( 'Please connect Jetpack to WordPress.com', 'woocommerce' ),
 					''
 				);
+			} else if ( Jetpack::is_staging_site() ) {
+				$health_item = $this->build_indicator(
+					'jetpack_indicator',
+					'notice',
+					'indicator-warning',
+					__( 'This is a Jetpack staging site', 'woocommerce' ),
+					''
+				);
 			} else {
 				$health_item = $this->build_indicator(
 					'jetpack_indicator',


classes/class-wc-connect-api-client.php
```diff
@@ -193,6 +193,7 @@ protected function request( $method, $path, $body = array() ) {
 				'last_heartbeat' => get_option( 'wc_connect_last_heartbeat', 0 ),
 				'last_rate_request' => get_option( 'wc_connect_last_rate_request', 0 ),
 				'active_services' => $this->wc_connect_loader->get_active_services(),
+				'enable_stats' => ! Jetpack::is_staging_site(),
```
jkudish: I'd like to suggest that this be `disable_stats` instead. It makes more sense semantically to have this be a flag to turn it off and have it be on by default.

This will simplify the server PR a little bit too, checking for a truthy value is easier to read then checking for `false` explicitly.

Also I think this value should get a filter on it; though it's optional

classes/class-wc-connect-help-provider.php
```diff
@@ -145,6 +145,14 @@ protected function get_health_items() {
 					__( 'Please connect Jetpack to WordPress.com', 'woocommerce' ),
 					''
 				);
+			} else if ( Jetpack::is_staging_site() ) {
+				$health_item = $this->build_indicator(
+					'jetpack_indicator',
+					'notice',
+					'indicator-error',
+					__( 'This is a Jetpack staging site, some features may not work as expected', 'woocommerce' ),
```
jkudish: Would be nice to be more specific here, for users' sake

jkudish:
Don't forget to label/milestone and assign your PRs :)

classes/class-wc-connect-api-client.php
DanReyLop: Hum... I agree on the server logic part, but I've never been a fan of "inverse booleans" such as ```var doesntHaveSomething``` or ```somethingIsDisabled```, it makes the code harder to read.

What do you mean by filter?

classes/class-wc-connect-help-provider.php
DanReyLop: Well, for now everything works, except for our internal stats, so I don't know what else to put here. Is there anything that isn't supposed to work in Jetpack Staging?

classes/class-wc-connect-api-client.php
jkudish: I guess it's a bit of a personal opinion, but I think it almost always makes sense to use the one that is not the "most common scenario". In this case, disabling stats is the exception not the rule, thus checking for `disable_stats` makes more sense. If we only wanted stats for certain requests, then I'd agree with you.

---

Filter: I mean a WordPress filter (`apply_filters( ... )`) so that someone (mostly us) can toggle this as needed.

classes/class-wc-connect-help-provider.php
jkudish: I'll defer to @allendav on this one.

classes/class-wc-connect-api-client.php
DanReyLop: Agree, in this case it makes more sense. I've changed the flag to ```disable_stats```.

Adding a filter for this seems an overkill at this point, though.

allendav: @jkudish - can you give an example use case for the filter? There is already a `jetpack_is_staging_site` filter in Jetpack - would that suffice?

classes/class-wc-connect-help-provider.php
allendav: Instead of error, let's make this a warning and say simply "This is a Jetpack staging site"

There is no user facing impact at this time, right? The value presently is just for an agency/dev to have some context for their work. Down the road we may want to tie other features to staging/not-staging and then we can consider adding additional text, e.g. "This is a Jetpack staging site, only test shipping labels can be generated"

jkudish: Agreed

allendav:
@DanReyLop wrote:

> I'm not sure how to test this, since I don't know how to put a site in "staging" mode

I suggest temporarily hooking a function into the `jetpack_is_staging_site` filter for testing. I use a "catch all" dev.php to do little hacks like this.

ref: https://github.com/Automattic/jetpack/blob/c424be4df0284fb30f85088b8a0f921747da0b82/class.jetpack.php#L5652

jkudish:
Fine to skip the filter for now.

classes/class-wc-connect-api-client.php
```diff
@@ -193,6 +193,7 @@ protected function request( $method, $path, $body = array() ) {
 				'last_heartbeat' => get_option( 'wc_connect_last_heartbeat', 0 ),
 				'last_rate_request' => get_option( 'wc_connect_last_rate_request', 0 ),
 				'active_services' => $this->wc_connect_loader->get_active_services(),
+				'disable_stats' => Jetpack::is_staging_site(),
```
allendav: If Jetpack is not active, will this code explode? We may need to test for the class + method first.

classes/class-wc-connect-help-provider.php
```diff
@@ -145,6 +145,14 @@ protected function get_health_items() {
 					__( 'Please connect Jetpack to WordPress.com', 'woocommerce' ),
 					''
 				);
+			} else if ( Jetpack::is_staging_site() ) {
```
allendav: Same comment as above. We should check for the class + method before attempting to use it since Jetpack may not be activated (or possibly even installed.)

classes/class-wc-connect-api-client.php
DanReyLop: This method checks for Jetpack way before reaching this code. This is the moment of sending a request to the WCC server, so at this point we are sure that Jetpack is alive and running.

classes/class-wc-connect-help-provider.php
DanReyLop: If Jetpack is not connected the ```else if``` branches above this one will execute instead, so this won't ever get checked if Jetpack is not connected / installed.

classes/class-wc-connect-api-client.php
allendav: Cool. Fair enough, then :)

classes/class-wc-connect-help-provider.php
allendav: Cool

PR #385, opened 2016-06-07
Add fallback message on shipping settings page

### LINKED ISSUE #338: Send users to debug page if form fails to render
Here's an idea: if the form fails to render because of a recent form schema update, we shouldn't wait 24 hours to re-fetch the schema. Perhaps we can re-queue a one time event to fetch 30m later or something?

We'll need to be careful because we also don't want to hammer the server with requests in the event where we introduce an error.

I don't have an exact technical implementation in mind, but something to potentially explore.

jkudish:
This PR seeks to add a fallback message that appears (via a css animation) on the shipping settings form if the form hasn't rendered within 3 seconds. The message prompts the user to visit
the debug page to find some troubleshooting steps.

image

The easiest way to test it is to apply the following diff to `client/main.js`:

```
diff --git a/client/main.js b/client/main.js
index daa853c..d2bc95f 100644
--- a/client/main.js
+++ b/client/main.js
@@ -27,7 +27,7 @@ let render = () => {
        const WCCShippingSettings = require( 'views/shipping' );
        ReactDOM.render(
                <Provider store={ store }>
-                       <WCCShippingSettings
+                       <WCCShippingSetting
                                storeOptions={ storeOptions }
                                schema={ formSchema }
                                layout={ formLayout }
@@ -54,7 +54,7 @@ if ( module.hot ) {
                try {
                        renderApp();
                } catch ( error ) {
-                       renderError( error );
+                       // renderError( error );
                }
        };
```

Though anything that breaks the form from rendering (including changing the schema) is a good way to test this.

Fixes #338 

cc @kellychoffman for a design review (anything we should prettify here?)
cc @nabsul @allendav @jeffstieler @DanReyLop

### DIFF
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -219,3 +219,23 @@
     @import '../../client/views/shipping/style';
 
 }
+
+@-webkit-keyframes fadeIn { from { opacity: 0; } to { opacity: 1; } }
+@-moz-keyframes fadeIn { from { opacity: 0; } to { opacity: 1; } }
+@keyframes fadeIn { from { opacity: 0; } to { opacity: 1; } }
+
+#wc-connect-admin-container .form-troubles {
+	opacity: 0;
+	-webkit-animation: fadeIn ease-in 1;
+	-moz-animation: fadeIn ease-in 1;
+	animation: fadeIn ease-in 1;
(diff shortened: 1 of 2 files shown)


allendav:
:shipit:

PR #388, opened 2016-06-07
Fix rate adjustment input fields so they let the user write floats like "2.05"

### LINKED ISSUE #381: Service rate adjustment has some bugs
I've noticed the following two strange behaviors:

- "2." is considered invalid
- Trying to type "0" after "2." reverts the value to simply "2"

@jeffstieler

DanReyLop:
Fixes #381 

This is not an ideal solution. The input will only be validated when ```onBlur``` is fired, and the input value won't be reformatted, so the user can submit for example "04.10", the server will receive "4.1", but that change won't be reflected on the form until the user refreshes the page.

To test it, simply go to the USPS form, select one rate and type a value like "2.05" or "invalid123" into the "Adjustment" field.

To make this work flawlesly, the ```ShippingServiceEntry``` component (or a sub-component) would need to have state (and since our tree is fully stateless right now, that would be a bummer). Tell me if you want me to go that way, or if this solution is good enough.

@nabsul @jkudish @allendav @jeffstieler

### DIFF
--- a/client/components/shipping/services/entry.js
+++ b/client/components/shipping/services/entry.js
@@ -1,9 +1,9 @@
 import React, { PropTypes } from 'react';
 import FormCheckbox from 'components/forms/form-checkbox';
 import FormSelect from 'components/forms/form-select';
-import FormTextInput from 'components/forms/form-text-input';
 import Gridicon from 'components/gridicon';
 import classNames from 'classnames';
+import NumberInput from './number-input';
 
 const numberRegex = /^\d+(\.\d+)?$/;
 const parseNumber = ( value ) => {
@@ -38,7 +38,7 @@ const ShippingServiceEntry = ( props ) => {
 				{ name }
 			</label>
 			{ hasError ? <Gridicon icon="notice" /> : null }
-			<FormTextInput
+			<NumberInput
 				disabled={ ! enabled }
 				value={ adjustment }
 				onChange={ ( event ) => updateField( 'adjustment', parseNumber( event.target.value ) ) }
--- a/client/components/shipping/services/entry.js
+++ b/client/components/shipping/services/entry.js
@@ -5,9 +5,12 @@ import Gridicon from 'components/gridicon';
 import classNames from 'classnames';
 import NumberInput from './number-input';
 
-const numberRegex = /^\d+(\.\d+)?$/;
 const parseNumber = ( value ) => {
-	return numberRegex.test( value ) ? Number.parseFloat( value ) : value;
+	if ( '' === value ) {
+		return 0;
+	}
+	const float = Number.parseFloat( value );
+	return isNaN( float ) ? value : float;
 };
 
 const ShippingServiceEntry = ( props ) => {
--- a/client/components/shipping/services/number-input.js
+++ b/client/components/shipping/services/number-input.js
@@ -1,0 +1,54 @@
+import React, { PropTypes } from 'react';
+import FormTextInput from 'components/forms/form-text-input';
+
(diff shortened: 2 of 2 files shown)


jkudish:
I don't think we should push this as it is.

nabsul:
What if we validate the input when the save button is clicked? That's sort of how I did things in the add-package modal.

DanReyLop:
Second try: I created a component that wraps the existing FormTextInput and gives it a little of logic (and state) to handle this. I think the behaviour is ideal now:
- Real-time validation while typing, so when the user types "2." the field goes red.
- The input doesn't change while the user has focus, so he can type "2", "2.", "2.0", "2.05".
- After the field loses focus, the input is changed if needed, so "2.10" becomes "2.1".
- The redux part has the same behaviour that it has in ```master```: it gets a float when the input is valid, and a string when it's not.
- When the field doesn't have focus, its behaviour is the same as it would be if it was stateless, so if shouldn't complicate testing.

I'm willing to iterate over this, just tell me if I'm in the right way (the NPM modules to fix this use case seemed totally overkill).

nabsul:
Let's park this for a little while. In the end, we may want to consider refactoring our redux state to solve this issue in a more general way. 

Since we're using redux for state, I'm not a huge fan of also having state inside the React component. Let's also talk about this in tomorrow's meeting.

allendav:
> Let's park this for a little while. In the end, we may want to consider refactoring our redux state to solve this issue in a more general way.

> Since we're using redux for state, I'm not a huge fan of also having state inside the React component. Let's also talk about this in tomorrow's meeting.

In the interest of supporting our users now until we craft a more comprehensive approach later, let's move forward with reviewing and merging this small PR as discussed during today's meeting.

allendav:
Each of these inputs should NOT result in a red box + exclamation mark

Input, Treated as
(empty), 0
1., 1
-1, -1
1.1, 1.1
.5, 0.5

@DanReyLop - how difficult would it be to support these relatively common user inputs in this PR?

DanReyLop:
Pretty easy, I've just commited it.

As an added bonus, we get rid of a RegExp, that's always good news :)

Side effect: everything that Javascript can parse to a Number, we will accept it, so `1potato` will be interpreted as `1`.

allendav:
Much better, but we should redbox on `1potato` - maybe we could test for any char not in 0123456789.-

Also seems a ci test is failing

DanReyLop:
I say if JavaScript swallows it, we should too. If the user types `1potato`, then the field loses focus it will be converted to just `1`, so I don't think adding a RegExp for this particular super-edge case would do us any good.

I've tried adding the property `type="number"` to the `<input>` field, but with the Calypso style it looks quite bad (too much padding):
number_input

So the options are:
* Using `type="number"` and correcting the styles (fun fact: it disallows every letter except `e`, for the merchants who like to input their rates in scientific notation)
* RegExp disallowing all characters except numbers
* Letting it go :snowman: :snowflake: 

Your call :)

About CI, yeah, something wonky happened with SVN, but only in some machines, in others it worked fine. Since this is a JS-only change, I would say we're good to go.

nabsul:
There's a way to tell CI to try again. I've never done it though...

allendav:
> If the user types 1potato, then the field loses focus it will be converted to just 1, so I don't think adding a RegExp for this particular super-edge case would do us any good.

Yeah, ya know let's go with this - I restarted the test (you click on the Details button and then there is a recycle button that will start it again.  Should be done shortly.

allendav:
:shipit:

PR #387, opened 2016-06-07
Improved error handling

jkudish:
This PR seeks to improve the catching of errors on the client. Specifically, it adds support for errors returned from the WP-API endpoint as strings, such as problems with communicating with
the server and other errors that are returned directly from that endpoint. This makes sure these errors do not produce a javascript warning and output the error as a notice.

To test:

* Checkout this branch
* `npm start`
* Define `define( 'WOOCOMMERCE_CONNECT_SERVER_URL', ' https://fake.url' );` in your `wp-config.php`
* Attempt to save settings. Verify that you get the following (and no javascript errors):

image

Fixes #386

cc @allendav @nabsul @DanReyLop @jeffstieler

PR #389, opened 2016-06-07
Enables Javascript minification in production builds

DanReyLop:
Production builds weren't minifying the Javascript properly. It was because in the development webpack config it was specified that modules were contatenated using `eval` calls, and that couldn't be minified, since the entire modules were strings.

I've also cleaned-up the webpack production config file, it was way more complex than it needs to be, I'm not sure if it was an historical reason for that.

To test this, just run `npm run dist`, check that the JS is now less than 500KB, and try to use the client in "production" mode (without the Webpack dev server).

cc @allendav @nabsul @jkudish @jeffstieler

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "scripts": {
     "start": "cross-env NODE_ENV=development webpack-dev-server --hot --inline --watch --content-base dist --port 8085",
-    "dist": "cross-env NODE_ENV=production webpack -p --config webpack.production.config.js",
+    "dist": "cross-env NODE_ENV=production webpack --config webpack.production.config.js",
     "eslint": "eslint tasks/ client/",
     "test": "karma start --single-run --no-auto-watch",
     "test-watch": "karma start --auto-watch --no-single-run",
--- a/webpack.production.config.js
+++ b/webpack.production.config.js
@@ -1,16 +1,22 @@
 var webpack = require( 'webpack' ),
-	webpackBaseConfig = require( './webpack.config.js' ),
-	plugins = webpackBaseConfig.plugins || [],
-	output = webpackBaseConfig.output || {};
+	config = require( './webpack.config.js' );
 
-delete output.publicPath;
+delete config.output.publicPath;
+delete config.devtool;
 
-var productionConfig = Object.assign( {}, webpackBaseConfig, {
-		output,
-		plugins
-	} );
+config.plugins.push( new webpack.DefinePlugin( {
+	'process.env.NODE_ENV': '"production"'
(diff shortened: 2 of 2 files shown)


allendav:
Not sure this is a showstopper, but this does increase the npm run dist time by 5 seconds for me (from 25 before the change to about 30 afterwards)

jkudish:
IMO that's a pretty negligible difference given that: 1) this isn't a task that's run often, 2) the performance/bandwith gains make it wortwhile

allendav:
There is something odd about the script created by this. If I test using `npm run dev` on master, I get this expected error when attempting to save settings using a bogus token against the live server:

screen shot 2016-06-08 at 5 17 45 pm

But if I run this branch under the same circumstances, I get no error notice but instead this error in my console:

screen shot 2016-06-08 at 5 35 18 pm

jkudish:
@allendav i think because #387 hasn't been rebased onto this branch

DanReyLop:
The increased time on ```npm run dist``` is perfectly expected. Webpack is doing something now that it wasn't doing before, so it needs to take more time (and 5 seconds seems acceptable to make the generated JS 3 times smaller).

I've just rebased with ```master```. But really, this change should be quite straight-forward, either everything works or nothing works :)

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "scripts": {
     "start": "cross-env NODE_ENV=development webpack-dev-server --hot --inline --watch --content-base dist --port 8085",
-    "dist": "cross-env NODE_ENV=production webpack -p --config webpack.production.config.js",
+    "dist": "cross-env NODE_ENV=production webpack --config webpack.production.config.js",
```
allendav: Removing `-p` removes `--optimize-minimize` and `--optimize-occurence-order` per https://webpack.github.io/docs/cli.html#production-shortcut-p

I see you add uglify back in below, but not occurrence order (dedupe isn't the same) - and I was wondering why we don't want occurrence order anymore? (Curious)

DanReyLop: I've just tried OccurenceOrderPlugin, it shaves 1KB more off the bundle size... not much, but doesn't hurt to add it :)

I've removed the ```-p``` option because I find it better to have the equivalent options detailed in the config file, it's more explicit this way and it allows us to fine-tune the minification process.

allendav:
Tests nicely. Much smaller too:

```
woocommerce-connect-client.js  478 kB       0  [emitted]  woocommerce-connect-client
woocommerce-connect-client.css  177 kB       0  [emitted]  woocommerce-connect-client
```

:shipit:

PR #395, opened 2016-06-09
Added a tool to generate all the JS strings needed for translation

DanReyLop:
Added a tool to generate all the JS strings needed for translation, and a way to inject them to the client from PHP.

To generate the strings file, run: ```npm run i18n```
This will generate the JS bundle, parse the AST looking for ```translate``` calls and generate (or overwrite) the file ```i18n/strings.php```. If you already have the JS bundle compiled, running ```node tasks/i18n``` will be faster.

The ```strings.php``` file will have an array like this:
```$i18nStrings = array(
"String in English" => __( "String in English" ),
"Another" => __( "Another" ),
...
);```

Since all the ```__()``` calls that appear in PHP code will be translated by the normal WordPress plug-in translation process, at runtime if the user is (for example) Spanish that array will look like:
```$i18nStrings = array(
"String in English" => "Cadena en Español",
"Another" => "Otra",
...
);```

The array will be inserted as-is to JS, since our i18n helper expects an object with that format.

**NOTE:** In the current state of our i18n system on JS side, we **don't** support plurals. We only support mappings of the form ```"string literal" => "translated string"```. Right now we are translating plurals manually, like in [this case](https://github.com/Automattic/woocommerce-connect-client/blob/master/client/components/shipping/services/group.js#L17), but that won't work for languages that have different rules for prurals, such as some Eastern-European languages. But for now (English, French, Spanish) we are fine.

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -40,6 +41,7 @@
     "json-loader": "^0.5.4",
     "karma": "^0.13.22",
     "karma-chai": "^0.1.0",
+    "karma-cli": "^1.0.0",
```
DanReyLop: I needed to add this to make ```npm test``` work (at least in Windows), I hope you don't mind :)

tasks/i18n.js
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env node
```
DanReyLop: This file is heavily based on this one: https://github.com/Automattic/wp-calypso/blob/master/server/i18n/index.js

tasks/i18n.js
```diff
@@ -0,0 +1,105 @@
...
+
+	// prepend the matches array with this content to open the php file
+	matches.unshift( [
+		'<?php',
+		'\n/* THIS IS A GENERATED FILE. DO NOT EDIT DIRECTLY */',
+		'\n$' + arrayName + ' = array('
+	].join( '' ) );
+
+	// append the generated file comment to the bottom of the array as well
+	matches.push( '\n);\n/* THIS IS THE END OF THE GENERATED FILE */' );
```
nabsul: Let's add one more `\n` to the generated file at the end.

i18n/strings.php
```diff
@@ -0,0 +1,38 @@
+<?php
+/* THIS IS A GENERATED FILE. DO NOT EDIT DIRECTLY */
+$i18nStrings = array(
+"Saving…" => __( "Saving…" ),
```
nabsul: A tab would be nice for everything inside `array()`
Also, any reason for using double quotes? We generally try to use single quotes by default, and double quotes when there is an advantage.

DanReyLop: I used the same code as Calypso, so the better solution would be to modify ```preprocess-xgettextjs-match.js``` inside ```wp-calypso``` to use single quotes.

Also, a tab is easy to add, I'll do it, but let's not forget that this is an auto-generated file and shouldn't be edited or even read by humans.

i18n/strings.php
```diff
@@ -0,0 +1,38 @@
...
+	"Add a package" => __( "Add a package" ),
+	"Package name" => __( "Package name" ),
+	"The customer will see this during checkout" => __( "The customer will see this during checkout" ),
+	"This field must be unique" => __( "This field must be unique" ),
+	"Inner Dimensions (L x W x H) %s" => __( "Inner Dimensions (L x W x H) %s" ),
+	"Outer Dimensions (L x W x H) %s" => __( "Outer Dimensions (L x W x H) %s" ),
+	"Package weight" => __( "Package weight" ),
+	"Weight of box" => __( "Weight of box" ),
+	"Max weight" => __( "Max weight" ),
+	"Max weight" => __( "Max weight" ),
```
nabsul: Are my eyes deceiving me, or are there two "Max weight" string here...?

i18n/strings.php
```diff
@@ -0,0 +1,38 @@
...
+	"Defines both the weight of the empty box and the max weight it can hold" => __( "Defines both the weight of the empty box and the max weight it can hold" ),
+	"Add a package" => __( "Add a package" ),
```
nabsul: "Add a package also appears twice"

i18n/strings.php
DanReyLop: Oops! Looks like this tool doesn't remove duplicates (I wonder if they are aware of this in Calypso). Great catch!

:construction_worker: _Implementing de-duplication_

DanReyLop: Done

nabsul:
I'm also curious how/where it gets the following strings: 	

`"Saving…", "Save Settings", "More", "Dismiss"`

But this isn't a deal breaker for me.

nabsul:
I'll trust that you'll get de-duping done before merging :wink: :shipit:

DanReyLop:
These strings are from Calypso code that is included in the bundle. For example, "Dismiss" is from the Notice component.

And don't worry, I've implemented de-duping :)

PR #396, opened 2016-06-09
Add Number support to text fields (in preparation for fallback rate support)

jkudish:
This PR seeks to add `Number` support to the `textField` component, in preparation for supporting the fallback rate setting.

This PR should be tested in conjunction with https://github.com/Automattic/woocommerce-connect-server/pull/300 -- where full testing instructions are included.

See https://github.com/Automattic/woocommerce-connect-client/issues/265

cc @nabsul @allendav @jeffstieler @DanReyLop

### DIFF
--- a/client/components/shipping/services/entry.js
+++ b/client/components/shipping/services/entry.js
@@ -4,11 +4,7 @@ import FormSelect from 'components/forms/form-select';
 import FormTextInput from 'components/forms/form-text-input';
 import Gridicon from 'components/gridicon';
 import classNames from 'classnames';
-
-const numberRegex = /^\d+(\.\d+)?$/;
-const parseNumber = ( value ) => {
-	return numberRegex.test( value ) ? Number.parseFloat( value ) : value;
-};
+import parseNumber from 'lib/utils/parse-number';
 
 const ShippingServiceEntry = ( props ) => {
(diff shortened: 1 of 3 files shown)


allendav:
@jkudish @DanReyLop - I'm inclined to not mix number and non-number handling in one component - I think it might make maintenance more straightforward.  What do you think about shelving this PR in favor of the approach in https://github.com/Automattic/woocommerce-connect-client/pull/388

DanReyLop:
#388 is an ad-hoc solution for the rate adjustment fields (which don't have schema), this is a bit more general solution for having fields with `number` schema type. This solution has the same problems that #388 seeks to solve (you can't type `2.05` in the Fallback rate field).

I think the quickest solution for shipping this with Alpha 2 is discarding this PR and re-implementing it using the `<NumberInput>` component (it'll be easier too when we decide to refactor if all the numeric inputs use the same component).

Let me know if you agree and I'll start working on it.

nabsul:
Is this still being worked on?

DanReyLop:
I've implemented a solution using the general idea of this PR and the `NumberField` component, here it is: https://github.com/Automattic/woocommerce-connect-client/pull/410

I'm closing this one.

PR #400, opened 2016-06-11
Remove the Package validation error log, it only added noise to the logs

### LINKED ISSUE #251: Let's not log Package destination failed validation any more
This gets into the logs a bunch and will only get worse on active sites.  All it tells us is that the customer is in the process of filling out the checkout form.  Let's not log this any more and just `return`

https://github.com/Automattic/woocommerce-connect-client/blob/b62e9c7f39dbb902f6b2e97fc4e17dcc958e8e26/classes/class-wc-connect-shipping-method.php#L224

DanReyLop:
Fixes #251 

### DIFF
--- a/tests/php/test_class-wc-connect-shipping-method.php
+++ b/tests/php/test_class-wc-connect-shipping-method.php
@@ -103,31 +103,6 @@ class WP_Test_WC_Connect_Shipping_Method extends WP_UnitTestCase {
 	/**
 	 * @covers WC_Connect_Shipping_Method::calculate_shipping
 	 */
-	public function test_invalid_destination_calculate_shipping() {
-
-		$shipping_method = $this->getMockBuilder( 'WC_Connect_Shipping_Method' )
-			->disableOriginalConstructor()
-			->setMethods( array( 'is_valid_package_destination', 'log' ) )
-			->getMock();
-
-		$shipping_method->expects( $this->any() )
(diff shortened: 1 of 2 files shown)


tests/php/test_class-wc-connect-shipping-method.php
```diff
@@ -103,31 +103,6 @@ public function test_is_valid_package_destination( $package, $expected ) {
 	/**
 	 * @covers WC_Connect_Shipping_Method::calculate_shipping
 	 */
-	public function test_invalid_destination_calculate_shipping() {
```
nabsul: It seems to me that you didn't have to completely remove this test because of the removal of the log message. Can you explain?

DanReyLop: The test just mocks everything and asserts that the log is called, so without log, this test doesn't actually do anything.

allendav: I think @DanReyLop is correct - we could however use a test for a not-yet-setup service schema - to make sure we handle that correctly. @DanReyLop - would you write up an issue to add a test_not_yet_setup_shipping_method test that ensures log is called if a service settings mock get_service_settings call returns an empty array

With that, I think we can move forward with this pr

allendav:
After writing up the issue to add a service not-yet-setup test mentioned above, you can :shipit:

PR #401, opened 2016-06-12
Prompt user before leaving the page when there are unsaved changes

### LINKED ISSUE #383: Prompt user if unsaved changes exist
when they attempt to navigate away from a shipping method instance settings form with unsaved changes.

Also make sure we don't cause this issue again - https://github.com/Automattic/woocommerce-connect-client/issues/221

DanReyLop:
Fixes #383 

I've added a boolean property to the form state, named ```pristine```. I'm not sure if this is an Angular-only concept, I'm willing to change the name if it's confusing.

@allendav @jeffstieler @nabsul


client/components/wcc-settings-form/index.js
```diff
@@ -15,64 +15,84 @@ import ObjectPath from 'objectpath';
...
-		layout,
-		saveFormData,
-		formActions,
-		noticeActions,
-	} = props;
-
-	const setIsSaving = ( value ) => formActions.setField( 'isSaving', value );
-	const setSuccess = ( value ) => {
-		formActions.setField( 'success', value );
+const WCCSettingsForm = React.createClass( {
```
allendav: Is there something ES6 syntax cannot support? I'd rather not lose ES6 syntax if possible.

client/state/form/reducer.js
```diff
@@ -34,9 +27,13 @@ export default function form( state = {}, action ) {
 		} );
 	}
 
-	// Allow client-side form validation to take over error state when inputs change
-	if ( settingsActions.hasOwnProperty( action.type ) && newState.hasOwnProperty( 'errors' ) ) {
-		delete newState.errors;
+	if ( settingsActions.hasOwnProperty( action.type ) ) {
+		newState.pristine = false;
```
allendav: It seems like it would be simpler for this to be part of the SET_FIELD reducer instead.

The SET_FIELD reducer could replace this

```
if ( 'success' === action.field && action.value ) {
   newObj.pristine = true;
}```

with something like this then
```
newObj.pristine = ( 'success' === action.field && action.value )
```

File: client/state/form/test/reducer.js [hunk 3]
```diff
@@ -38,23 +35,6 @@ describe( 'Settings reducer', () => {
 		} );
 	} );
 
-	it( 'UPDATE_FORM_ELEMENT_FIELD', () => {
```
allendav: Thank you for cleaning this old code up

client/components/wcc-settings-form/index.js
DanReyLop: Since now the component needs lifecycle methods (`componentDidMount`, `componentWillUnmount`), it can't be a functional component anymore. So it's either this, or ES6 classes (which I've read somewhere that are fundamentally broken).

client/state/form/reducer.js
DanReyLop: The form becomes "dirty" when a setting changes, and only becomes `pristine` when it's finished saving the changes. Inputting some text in a text field doesn't call the `SET_FIELD` reducer, so what you proposed wouldn't work here.

Side note: the `SET_FIELD` name is the most confusing name ever. It changes a property of the form (such as `isSaving`), not form fields.

client/components/wcc-settings-form/index.js
allendav: Good point. So in that case, what do you think about:

1) not implementing this in `WCCSettingsForm` so we can keep it as a functional component
and
2) moving the addEventListener up to client/main.js where we have access to the store and can look at store.getState() to see if pristine is false?

client/state/form/reducer.js
allendav: Ah - I see. Yeah, that action name is poor.  Wanna change it to SET_FORM_PROPERTY or something instead?

client/components/wcc-settings-form/index.js
DanReyLop: Done

allendav:
Works nicely. :shipit:

PR #402, opened 2016-06-12
Added the possibility to display a validation error on the Services fieldset

DanReyLop:
Allows the Services entire fieldset to have an error state as a whole (not only for sub-fields), and a ```validation_hint```.

Server-side PR: https://github.com/Automattic/woocommerce-connect-server/pull/301

How to test: You will to update your server to https://github.com/Automattic/woocommerce-connect-server/pull/301 , after that just go to configure a USPS provider and leave all the Services unchecked. After you submit the changes you should see an error in the Service fieldset.

cc/ @allendav @nabsul @jeffstieler 

cc/ @kellychoffman if you want to tweak the design for the error field

### DIFF
--- a/client/components/wcc-settings-form/settings-item.js
+++ b/client/components/wcc-settings-form/settings-item.js
@@ -1,4 +1,5 @@
 import React, { PropTypes } from 'react';
+import some from 'lodash/some';
 import Indicators from 'components/indicators';
 import TextArea from 'components/text-area';
 import TextField from 'components/text-field';
@@ -26,7 +27,10 @@ const SettingsItem = ( {
 	const fieldRequired = ( -1 !== schema.required.indexOf( id ) );
 	const fieldValue = settings[ id ];
 	const fieldSchema = schema.properties[ id ];
-	const fieldError = ( errors && errors.length ) ? ( layout.validation_hint || '' ) : false;
+
+	// Check if the response has an error for this concrete field (not any subfields)
+	const hasDirectError = errors && some( errors, error => ! error.length );
+	const fieldError = hasDirectError ? ( layout.validation_hint || '' ) : false;
 
 	switch ( layout.type ) {
 		case 'radios':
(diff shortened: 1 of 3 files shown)


kellychoffman:
> If you want to tweak the design for the error field

Can you add a screencap here? (And any PRs that have front end changes)

DanReyLop:
Sure, here it goes:

services_error

kellychoffman:
Thanks. Updated styles so that each container is highlighted, it looked a bit off when there was one service expanded. 

Now:

screen shot 2016-06-13 at 12 41 02 pm

kellychoffman:
This reminds me, we should have a title and description above the services containers. Maybe:

**Services**
_Choose at least one service to return rates for your customers._

Something along those lines. Hopefully will reduce the amount of blank entries we get?

DanReyLop:
Thanks, that definitely looks better :)

This PR (and specially the corresponding server-side PR) are already quite big, so I've moved your suggestion to a new issue: https://github.com/Automattic/woocommerce-connect-client/issues/404

I'll work on it right away.

kellychoffman:
Good call. Thanks!

client/components/wcc-settings-form/settings-item.js
```diff
@@ -26,7 +27,10 @@ const SettingsItem = ( {
 	const fieldRequired = ( -1 !== schema.required.indexOf( id ) );
 	const fieldValue = settings[ id ];
 	const fieldSchema = schema.properties[ id ];
-	const fieldError = ( errors && errors.length ) ? ( layout.validation_hint || '' ) : false;
+
+	// Check if the response has an error for this concrete field (not any subfields)
+	const hasDirectError = errors && some( errors, error => ! error.length );
```
nabsul: How about hasFieldError`, or `hasOuterError`. "Direct" doesn't seem clear enough to me.

nabsul:
Works great. Besides the suggestion to rename the boolean value, :shipit: when the server side is done.

PR #410, opened 2016-06-17
Add number field type support

### LINKED ISSUE #265: Add Optional Fall Back Rate Field to USPS and Canada Post Settings
While comparing the existing wooThemes USPS extension and the WCC USPS method, one of things that came up was that the old extension supports an optional fall back rate – one that is returned to the customer during checkout in the event that no rates are returned by USPS for whatever reason.

I’m starting to think we should have that too in our method – here’s my rationale – if for whatever reason we don’t return a rate (and we’ve seen lately how easy it is to get into that situation, although hopefully we’ve now closed some of those holes) and the merchant only has USPS shipping for the customer’s zone, that customer will simply NOT be able to complete checkout. Period.

As a merchant, this would give me much angst and reflect badly on the WCC USPS method, even if it later turns out it was due to something I did wrong (e.g. selecting a bad set of USPS services)

At least seeing the fall back rate 1) will let a customer proceed with checkout and 2) reassure me that the extension is probably working and that I need to fix a setting.

As with any other USPS service, the Edit Order page should reflect that the Fall Back Rate was the “selected” service

I realize it adds more visual clutter to the settings UI, but I think the upside greatly outweighs that downside.

cc @kellychoffman

DanReyLop:
Fixes #265 

See #396 for the previous approach.

The `NumberField` component is pretty much similar to the `TextField` component, but once we DRY `renderFieldDescription` and `renderFieldError` it will look a little better (there's already a PR for that).

To test this, go to the Settings page of a Canada Post service and input whatever you want in the "Fallback Rate" field, it should let you type whatever you want but convert it to a number (if possible) on blur.


allendav:
Tests very nicely. No changes needed to code. :shipit:

PR #414, opened 2016-06-20
Make the JS root container be selected dinamically depending on the wcConnectData contents

DanReyLop:
The Root component of the React bundle will be determined by the contents of the `wcConnectData` payload, instead of it being always the `<WCCSettingsForm>`.

This can be merged as-is, the existing code will keep working.

In the near-ish future, the `rootView` for the "Label printing" meta-box will be a button that opens a dialog, instead of a full `SettingsForm`.


client/views/root/canada_post.js
```diff
@@ -0,0 +1,3 @@
+import RootView from '../generic-settings-form';
```
allendav: Is there a way to do this without having a specific root view for each service? This means if we want to add a new shipping service to WooCommerce Connect we'd have to roll the client.  We want to avoid that.

client/main.js
```diff
@@ -34,10 +35,10 @@ window.addEventListener( 'beforeunload', ( event ) => {
 const rootEl = document.getElementById( 'wc-connect-admin-container' );
 
 let render = () => {
-	const WCCShippingSettings = require( 'views/shipping' );
+	const RootView = require( 'views/root/' + rootView );
```
nabsul: I don't think having a variable in your require is a good idea. It makes the code very unpredictable; I'm not even sure how WebPack works in this case...

client/views/root/canada_post.js
nabsul: So far (and we intend on keeping it that way), there is no difference between USPS and CanadaPost in the eye of the plugin.

client/main.js
DanReyLop: WebPack loads in the bundle all the modules that match the pattern `views/root/*` , so it's basically as expected.

nabsul:
I think we have a couple of better options to solve this issue:

1) Create a second React App for the purpose of Label printing.
2) Add some logic in mounting the root element. Right now we simply try to mount the div with ID `wc-connect-admin-container`. We could extend this part to first check for that ID and mount based on that. Then we'd just have a different ID for the Label printing stuff.

Option 2 is very easy to do. I have some code samples in my hobby project if you want to compare: https://github.com/Fermiac/woocommerce-siftscience/blob/master/app/index.js#L11

 1 is more complx, but would hopefully result in two smaller webpacked files, instead of one large one.

client/main.js
nabsul: Interesting. So WebPack doesn't do anything smart like ignoring unused code and modules?

Still, I would really prefer to keep it so that `requires()` take constant parameters.

client/views/root/canada_post.js
DanReyLop: @allendav Good point. I don't think basing the RootView in the <div> we find would be a easy thing to implement... I will make that all the shipping services use a `shipping` RootView, so right now we would have `shipping`, `self-help`, and `shipping-label`. Is that ok?

client/main.js
DanReyLop: It can't be **that** smart. If you require `"something/" + arbitraryString`, the bundle must have all the possible bundles ready.

DanReyLop:
The vast majority of our code footprint are the dependencies. So instead of, for example, having a single 500KB bundle, we would have 2 bundles of 450KB... that doesn't sound like a win to me :)

Option 2 is mostly what I have done, just with a different implementation detail. Somehow, from the PHP code we need to tell the bundle which view to load. Either with an ID in the <div> or with a string in the `wcConnectData`, that's the smallest of our problems.

client/views/root/canada_post.js
allendav: > so right now we would have shipping, self-help, and shipping-label. Is that ok?

Yes, I think something like that would work. Is shipping-label more specific than needed? I.e. if we wanted to add another Button -> Dialog thing down the road to Edit Order.

DanReyLop: We'll see. If I can make the Shipping Label dialog fully data-driven, I'll change this name to something less specific. If not, we'll need an entry point just for the labels.

allendav:
> The vast majority of our code footprint are the dependencies. So instead of, for example, having a single 500KB bundle, we would have 2 bundles of 450KB... that doesn't sound like a win to me :)

I agree - for this reason let's try not to have multiple apps, at least not at this time.

> WebPack loads in the bundle all the modules that match the pattern views/root/* , so it's basically as expected.

The technique you've proposed in this PR ( the `+ rootView` approach ) gave me pause as well. My gut would prefer what @nabsul mentioned in Option 2 above (i.e. conditionals for the each root views), but I do not have a technical reason to not proceed with the `+ rootView` approach

allendav:
@DanReyLop - are you following a data-driven approach for the shipping-label ui?

If so, why do we need distinct root views at all? Self-help and Shipping instances both use the `WCCSettingsForm` today. Maybe we just need to rename `WCCSettingsForm` to something like `WCCDataDrivenContainer` or something :)

cc @nabsul

DanReyLop:
I'm trying to make the flow data-driven, but data-drive the button and opening the dialog just seemed too much of a stretch. Specially if we are going to show different information in place of the button if the label was already printed.

allendav:
Fair enough. How about this then - we create a second root view for the new container, require it and the old container ( WCCSettingsForm - which we could rename to WCCDataDrivenContainer or something ) explicitly instead of by variable and use a simple conditional like @nabsul mentioned to render one or the other. That might make it easier to re-integrate down the road if we want and avoids the wildcarding that feels odd.

DanReyLop:
Ok for me. I'm closing this PR and will implement it directly in the labels branch.

PR #415, opened 2016-06-20
Refactored and cleaned-up some PHP code

DanReyLop:
Sorry for the ambiguous PR title, these were just a few things that I encountered and I couldn't leave as they were. They are separated in independent commits for easier reviewing.


jeffstieler:
@DanReyLop it looks like this broke the tests. Can you look into it?

DanReyLop:
No way of running PHPUnit on Windows :cry: 
This can take a while.

DanReyLop:
I've just removed the tests that were failing. Before, the scripts and styles were being included conditionally, but that's not really needed because "registering" a script/style doesn't actually include it in the page. So I've removed that logic, and the test that covered that logic.

woocommerce-connect-client.php
```diff
@@ -419,48 +419,9 @@ public function woocommerce_payment_gateways( $payment_gateways ) {
...
-				// If we are on a system status page, make sure it is the connect tab
-				if ( 'connect' !== $tab ) {
-					return;
-				}
-			} else {
-				// Don't recognize the hook? Go no further
-				return;
-			}
-
+		public function admin_enqueue_scripts() {
```
allendav: Won't this cause our big ol' bundles to be loaded on every single wp-admin page?

client/views/shipping/style.scss
```diff
@@ -1,12 +0,0 @@
-.wcc-form-success {
-    animation: disappear .3s ease-in-out 2s;
```
allendav: Why this change?

DanReyLop: This file was not used, dead code removal :)

woocommerce-connect-client.php
DanReyLop: `wp_register_script( 'handler_name', 'path/to/script' )` just *registers* that for that name (`'handler_name'`) those files need to be included. The bundle (and the CSS, and the translations) won't actually be included unles `wp_enqueue_script( 'handler_name' )` and `wp_enqueue_style( 'handler_name' )` are called.

So this code is pretty harmless on its own, the extra logic to not register the script and styles when not needed was not worth it, in my opinion.

You can try it if you don't believe me 100%, the result is obvious with the Chrome DevTools open :)

jeffstieler:
Looks good and tests well (doesn't appear to break the self help page). :shipit:

PR #416, opened 2016-06-20
Moved the form errors calculation to a selector

DanReyLop:
In my ongoing quest to move as much logic as possible from the components to the Redux tree, I've moved the calculation of the form errors to a selector.

I'm working on moving the `formSchema` and `formLayout` to the Redux tree, to have most (if not all) of the complex logic needed for a multi-step form encapsulated in the reducers. If you have any objection about that better say it as soon as possible :)


jeffstieler:
Nice work @DanReyLop - I like that this provides an avenue for reusing the form error selector. :shipit:

PR #417, opened 2016-06-20
Make the CI build fail on ESLint errors

DanReyLop:
CI is not failing even when ESLint is givin errors. This is a fix for that.


DanReyLop:
I've fixed the commands that the CI uses and then the build failed (as expected), you can see it in the Travis history. After that I commited the actual fixes for the Lint errors, and now the build should pass :)

@nabsul @allendav @jeffstieler

jeffstieler:
:+1:

PR #418, opened 2016-06-20
Moved the save form logic to a Redux action, using redux-thunk

DanReyLop:
In my series of "move everything possible away from the components" PRs, this is the last one :)

Instead of having the Settings component configure the `saveForm` logic and attaching all the callbacks, it's now done in the "Redux backend", using redux-thunk.

In last week's Calypso meeting `redux-thunk` was discussed (and quickly dismissed) as an approoach to async actions such as this one. Let's see if it can work for our case :)

### DIFF
--- a/client/components/wcc-settings-form/index.js
+++ b/client/components/wcc-settings-form/index.js
@@ -6,58 +6,19 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import * as FormActions from 'state/form/actions';
 import { successNotice, errorNotice } from 'state/notices/actions';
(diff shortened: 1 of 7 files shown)


jeffstieler:
@DanReyLop this code looks good, but I can no longer get the error notice to appear.

jeffstieler:
:shipit:.

DanReyLop:
Ty! :+1: 

@allendav and @nabsul , feel free to comment in this changeset even if it's already merged, how the `thunk` works can be a mindf**k at first, it's better you understand it well before moving on :)

Issue #419, opened 2016-06-20
Integrate with Create a Ticket on WooThemes.com


allendav:
In self-help, ideally passing the siteURL and the log snippet to the page

laurendavissmith:
Closing, the links provided in the extension are enough for now

Issue #153, opened 2016-04-12
Add stateful self-help dashboard


allendav:
~~Needs design.~~ [Design below](https://github.com/Automattic/woocommerce-connect-client/issues/153#issuecomment-213117660)

Adding self-help will be essential to avoiding an avalanche of support requests.  Our client has only the logs right now.

I think we need to add something, before launching the Beta, for WCC to give merchants a single page to assess the health of their WCC setup before they reach out to support (and to give them guidance or even automation around opening a productive ticket if need be.)

I am thinking a new tab in wp-admin > WooCommerce > System Status called “WooCommerce Connect” with:

- an overall red, yellow or green indicator for the health of the connection
- red because jetpack is not installed, active, or connected or if the merchant’s site has never successfully hit the /services endpoint (and a prompt to take care of that)
- yellow if the merchant’s site has successfully hit the /services endpoint but the last time it tried it failed (along with a try again button)
- green otherwise
- a shortcuts to each of the enabled service’s settings pages (e.g. each shipping zone method instance for wcc powered usps) – with a red indicator if the service instance setting is missing a required field or green otherwise
- a prompt if they haven’t set up any WCC services to do so:)
- NEW - warn the user if they are using a CUSTOM PERMALINK STRUCTURE ( @nabsul and people on the Internet have gotten 404s when attempting to access WC API endpoints when CUSTOM is selected for permalinks.  Plain may also be a problem.)
- a display with how long ago the merchant’s site last was able to successfully hit the /services endpoint to get available services from the WCC server (we can call this the “last check for updates” or something)
- a display with how long ago the last /shipping-rates request was and whether it was successful or not (including the complete response hidden under an expander)
- a control to enable/disable detailed debug logging for WCC
- a button to force a re-fetch of /services
- move the WCC connection test button @jkudish added to the tools page to this page instead
- display a reverse chronological snippet of the most recent dozen or so entries in the WCC debug log, if any (with a link to the full log)
- display the WordPress.com shadow blog ID for aid in helping support
- and lastly, a button to open a ticket ( / send an email?), ideally being able to pre-fill some info like the shadow blog id, url, WCC debug log and other site details / system status 

The other neat thing about this page is it could serve as a destination for “something is wrong with WCC, click here to fix it” notices wherever we display them

cc @kellychoffman @mikeyarce

allendav:
Setting myself as assignee to start coding this

allendav:
[...]

PR #420, opened 2016-06-20
Add basic support links

allendav:
This is the last piece of #153 - it adds a pair of support links to the bottom of the self-help page.

I've gone ahead and opened a new issue ( #419 ) for when we can do something more magical with https://woothemes.com/my-account/create-a-ticket/

To test, open self-help (wp-admin > WooCommerce > System Status > WooCommerce Connect), scroll to the bottom and voila - support links :P

### DIFF
--- a/client/components/text/index.js
+++ b/client/components/text/index.js
@@ -1,11 +1,34 @@
 import React, { PropTypes } from 'react';
 import FormFieldset from 'components/forms/form-fieldset';
+import FormLegend from 'components/forms/form-legend';
+import { sanitize } from 'dompurify';
+
+const renderTitle = ( title ) => {
+	if ( ! title ) {
+		return null;
+	}
+
+	return (
+		<FormLegend>{ title }</FormLegend>
+	);
+};
+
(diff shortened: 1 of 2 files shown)


client/components/text/index.js
```diff
@@ -1,11 +1,34 @@
...
+	return (
+		<FormLegend>{ title }</FormLegend>
+	);
+};
+
+const renderText = ( text ) => {
+	/* for dangerouslySetInnerHTML we need to pause linting for a narrow scope */
+	/* eslint-disable */
+	return (
+		<span dangerouslySetInnerHTML={ { __html: sanitize( text, { ADD_ATTR: ['target'] } ) } }>
```
jeffstieler: Why do you need to allow `target` attributes for "text"?

jeffstieler:
This largely looks good, I just had one question @allendav.

client/components/text/index.js
allendav: We don't presently - this is in case we ever want to be able to use this component to pass up an anchor with a target attribute.

PR #425, opened 2016-06-22
Removed form schema from the Redux state tree

DanReyLop:
Follow-up from #416 

As discussed, the Redux tree shouldn't have the form schema, so I rewrote the `getFormErrors` selector so it gets the schema from the component's `props` instead of from the state.

@jeffstieler


jeffstieler:
Thanks for fixing this @DanReyLop! Code looks good, and validation still works. 🚢  it.

PR #428, opened 2016-06-23
Label printing: super early draft

DanReyLop:
First approach to the label printing flow.

I see this PR more as a "sanity check" to see if we all are in the same page, and if the general architecture of the Label Printing dialog is ok.

How to test: Create an order, and go to the "Edit Order" admin page. Click the "Print Label" button in the sidebar, and enjoy the super-duper dialog.

What works:
* Opening and closing the dialog.
* Tabbed navigation (you can go back to a previous step in the form by clicking its title in the "tab" list).
* Data-driveness.
* Getting the destination address from the order.
* The dialog will crash if you reach the "Preview" step. Just a heads up :)


classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,310 @@
...
+
+			foreach( $order->get_address( 'shipping' ) as $key => $value ) {
+				$formData[ 'orig_' . $key ] = '';
+				$formData[ 'dest_' . $key ] = $value;
+			}
+
+			return $formData;
+		}
+
+		protected function get_form_layout() {
```
DanReyLop: The form layout / schema should come from the WCC server I think, to give us more flexibility without needing to update the client.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,310 @@
...
+					'key' => 'state',
+				),
+				array(
+					'key' => 'postcode',
+				),
+				array(
+					'key' => 'country',
+				),
+			);
+			$address_summary = '{first_name} {last_name}\\n{address_1} {address_2}\\n{city}, {postcode} {state}, {country}';
```
DanReyLop: This sort-of template is what will be displayed when an step has a "suggestion" received from the server. So it will appear side-by-side the "Summary" of the adress, and the "Summary" of the suggested address. None of that is implemented yet.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,310 @@
...
+
+			$layout[] = array(
+				'type' => 'step',
+				'tab_title' => 'Origin',
+				'title' => 'Generate shipping label: Address verification (origin)',
+				'description' => 'asdfghjkjhgfdsaASDFGHJJHGFDSA',
+				'items' => array( array( 'items' => $orig_address_fields ) ),
+				'bypass_suggestion_flag' => 'orig_bypass_suggestion',
+				'summary' => str_replace( '{', '{orig_', $address_summary ),
+				'sugggestion_hint' => 'vnkdjfitysncgkgse,jhxccvnaluerghsldjfvbzdj',
```
DanReyLop: This `suggestion_hint` will be shown in the yellow notice, according to the mock-ups.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,310 @@
...
+			);
+
+			$layout[] = array(
+				'type' => 'step',
+				'tab_title' => 'Packages',
+				'title' => 'Shopping cart contents',
+				'items' => array(
+					array(
+						'items' => array(
+							array(
```
DanReyLop: So. Much. Nesting.

An "step" needs to have an array of "items", and each of those "items" must be a "group" (like a `fieldset` for example), and each of those groups will have an array of "items", each one of them will be an actual field.

Another approach could be to treat an "step" as a special type of group... I don't know, I have mixed feelings either way.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,310 @@
...
+								),
+							),
+						),
+					),
+				),
+				'summary' => '{rate}',
+			);
+
+			$layout[] = array(
+				'type' => 'summary',
```
DanReyLop: This is a special type of step, which will need to have access to the whole schema / layout / data. It will show the summary of every past step. There are no mock-ups of this step yet, but I assume something like this will be needed.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,310 @@
...
+					'length' => $length,
+					'quantity' => $item['qty'],
+					'weight' => $weight,
+					'width' => $width
+				);
+			}
+			$formData[ 'cart' ] = $contents;
+
+			foreach( $order->get_address( 'shipping' ) as $key => $value ) {
+				$formData[ 'orig_' . $key ] = '';
```
DanReyLop: The origin address will need to be stored somewhere, for sure. I just don't know where yet. After it's introduced for the first time, the flow will just automatically skip past the "Origin address" step.

Because we need the origin address, right?

client/components/wcc-settings-form/index.js
```diff
@@ -9,19 +9,76 @@ import { successNotice, errorNotice } from 'state/notices/actions';
...
+		return (
+			<FormButtonsBar>
+				<FormButton type="button" onClick={ props.formActions.nextStep }>
+					{ label }
+				</FormButton>
+			</FormButtonsBar>
+		)
+	};
+
+	const renderTab = ( label, index ) => {
```
DanReyLop: Probably all of this `renderWhatever` methods should be its own component. I still don't know how fine-grained to go with stuff like this. But I haven't given this much thought, 95% of this code will need to be rewritten for sure :)

client/components/wcc-settings-form/index.js
```diff
@@ -9,19 +9,76 @@ import { successNotice, errorNotice } from 'state/notices/actions';
...
+	};
+
 	return (
 		<div>
 			<GlobalNotices id="notices" notices={ notices.list } />
-			{ props.layout.map( ( group, idx ) => (
-				<WCCSettingsGroup
+			{ 'step' === props.layout[ 0 ].type
+				? renderMultiStepForm()
+				: <WCCSettingsStep
```
DanReyLop: So, I generalized our current forms as just "a multi-step form of just one step". A bit hacky but it makes every use case we have fit into the same data model.

client/main.js
```diff
@@ -32,10 +33,10 @@ window.addEventListener( 'beforeunload', ( event ) => {
 const rootEl = document.getElementById( 'wc-connect-admin-container' );
 
 let render = () => {
-	const WCCShippingSettings = require( 'views/shipping' );
+	const RootView = 'shipping-label' === rootView ? require( 'views/shipping-label' ) : require( 'views/shipping' );
```
DanReyLop: `require` without wildcards so you can sleep peacefully at night ;)

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,310 @@
...
+				'tab_title' => 'Preview',
+				'title' => 'dfhdfghdfgh',
+				'action_label' => 'Print',
+				'confirmation_flag' => 'confirm',
+			);
+
+			return $layout;
+		}
+
+		protected function get_form_schema() {
```
DanReyLop: The idea with the schema is that the Hydra server will receive a plain array (not nested). So, for example, after the first step, the server will receive all the `orig_address`, `orig_city`, `orig_whatever` fields filled, and the rest of the fields blank. In the server-side validation, it will return something for the origin address' individual fields (`OK`, `ERROR`, `suggestion`), and simply return errors for the rest of the fields.

So, the idea is that the client will send the entire form, and the server will short-circuit validation when appropiate.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,310 @@
...
+				'sugggestion_hint' => 'vnkdjfitysncgkgse,jhxccvnaluerghsldjfvbzdj',
+			);
+
+			$layout[] = array(
+				'type' => 'step',
+				'tab_title' => 'Destination',
+				'title' => 'Generate shipping label: Address verification',
+				'description' => 'asdfghjkjhgfdsaASDFGHJJHGFDSA',
+				'items' => array( array( 'items' => $dest_address_fields ) ),
+				'bypass_suggestion_flag' => 'dest_bypass_suggestion',
```
DanReyLop: Field to set to a `truthy` value when sending the form to the server, to indicate that the client chose this address even though the server suggested another.

client/components/wcc-settings-form/index.js
```diff
@@ -9,19 +9,76 @@ import { successNotice, errorNotice } from 'state/notices/actions';
...
+import FormButton from 'components/forms/form-button';
+import FormButtonsBar from 'components/forms/form-buttons-bar';
 
 const WCCSettingsForm = ( props ) => {
+	const currentStep = props.form.currentStep;
+	const currentStepLayout = props.layout[ currentStep ];
+
+	const renderCurrentStep = () => {
+		if ( ! currentStepLayout ) {
+			return <span>... spinner or something ...</span>;
```
DanReyLop: So here's the thing: The form starts in the step `-1`. During this time, the form should display an spinner or something.
Right when the user lands in a step, `something` must be triggered so that the form will auto-advance to the next field, and so on, if the current step's form is in a correct state (according to the client **and** the server). Since this auto-advance needs server-side validation, some kind of spinner is needed here.

Of course, none of this is implemented yet, so you must manually click the "Next" button to get to the first form.

DanReyLop:

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,310 @@
...
+					'title' => 'Postal code',
+					'minLength' => 1,
+				),
+				'country' => array(
+					'type' => 'string',
+					'title' => 'Country',
+					'minLength' => 1,
+				),
+			);
+			$required_address_fields = ['first_name', 'address_1', 'city', 'state', 'postcode', 'country'];
```
jeffstieler: This needs to be PHP 5.2.x compatible. Please switch to `array()` syntax.

jeffstieler:
@DanReyLop why all the changes to `false !== error` ?

DanReyLop:
@jeffstieler Errors were not showing when the `validation_hint` for a field is empty. When that happened, the `error` property was set to the empty string, which was being detected as `false`-ish.

jeffstieler:
@DanReyLop that was a pseudo-enforced requirement to have a `validation_hint` for all fields.

Why not just add validation hints?

DanReyLop:
Ok, I didn't know about that requirement. I'll revert that change and add validation_hint for every field.

classes/class-wc-connect-shipping-label.php
allendav: Yep, we will need the complete origin address. Let's handle a single origin address at this time (yes, there could be different origin addresses for different shipping zones for multi-warehouse merchants, but they are a corner case and I would rather we not confuse the UI at this time.)

I think we should add fields to wp-admin > WooCommerce > Settings > General, up near the top with Base Location for the origin address. Please write a separate WooCommerce Connect issue to filter whatever fields you need into WooCommerce core.

classes/class-wc-connect-shipping-label.php
allendav: I think so perhaps someday, but I think trying to do that now will endanger Beta.  Feel free to write an issue on the server to add an endpoint for the form content for a given carrier.

client/main.js
allendav: Better, but what about conditional rendering instead of conditional instantiation? Its more familiar a pattern.

client/state/form/reducer.js
```diff
@@ -14,6 +20,14 @@ reducers[ SET_FORM_PROPERTY ] = ( state, action ) => {
 	return Object.assign( {}, state, newObj );
 };
 
+reducers[ NEXT_STEP ] = ( state ) => {
+	return Object.assign( {}, state, { currentStep: state.currentStep + 1 } );
```
allendav: To avoid explosions, should the reducer only advance when we aren't on the last step?

client/state/form/reducer.js
```diff
@@ -14,6 +20,14 @@ reducers[ SET_FORM_PROPERTY ] = ( state, action ) => {
...
+};
+
+reducers[ GO_TO_STEP ] = ( state, action ) => {
+	return Object.assign( {}, state, { currentStep: action.step } );
```
allendav: Should we do bounds checking here as well?

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,316 @@
...
+			);
+			$required_fields[] = 'rate';
+
+			return array(
+				'required' => $required_fields,
+				'properties' => $properties,
+			);
+		}
+
+		public function meta_box( $post ) {
```
allendav: An order may have multiple packages, and may not have been box packed yet at all (e.g. they merchant uses free shipping). Please add a TODO and an issue to wit. We are going to have to handle 1 (typical) to N (possible) labels for each order.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,316 @@
...
+
+		public function __construct( WC_Connect_Service_Settings_Store $settings_store ) {
+			$this->settings_store = $settings_store;
+		}
+
+		protected function get_form_data( WC_Order $order ) {
+			$formData = array();
+			$contents = array();
+
+			foreach( $order->get_items() as $item ) {
```
allendav: We may (or may not) already have rates and box packing solutions (e.g. if the customer selected a USPS rate during checkout). Please add a TODO and an issue

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,316 @@
...
+			return $formData;
+		}
+
+		protected function get_form_layout() {
+			$layout = array();
+
+			$address_fields = array(
+				array(
+					'key' => 'first_name',
+					'validation_hint' => 'This field is required.',
```
allendav: Please __( ) this and all merchant (or customer) facing text

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,316 @@
...
+				array(
+					'key' => 'postcode',
+					'validation_hint' => 'This field is required.',
+				),
+				array(
+					'key' => 'country',
+					'validation_hint' => 'This field is required.',
+				),
+			);
+			$address_summary = '{first_name} {last_name}\\n{address_1} {address_2}\\n{city}, {postcode} {state}, {country}';
```
allendav: WooCommerce core has a formatted address function https://docs.woothemes.com/wc-apidocs/source-class-WC_Abstract_Order.html#1096-1124

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,316 @@
...
+			$dest_address_fields = $address_fields;
+			foreach( $address_fields as $index => $_ ) {
+				$orig_address_fields[ $index ][ 'key' ] = 'orig_' . $address_fields[ $index ][ 'key' ];
+				$dest_address_fields[ $index ][ 'key' ] = 'dest_' . $address_fields[ $index ][ 'key' ];
+			}
+
+			$layout[] = array(
+				'type' => 'step',
+				'tab_title' => 'Origin',
+				'title' => 'Generate shipping label: Address verification (origin)',
```
allendav: Same comment re `__()` here and below

classes/class-wc-connect-shipping-label.php
DanReyLop: We don't even really need a dedicated settings form for that. We could use this flow, and after a successful label purchasing, store the origin address *somewhere* in the database (I'm not familiar enough with WooCommerce, but all that we need is a place to store it). This way, the merchant would fill this step naturally the first time he prints a label, but the next times the flow will auto-advance past it.

client/state/form/reducer.js
DanReyLop: Yep, this needs to be checked somewhere, but maybe it will be better in the action creator (conditional dispatch, one of the primary use-cases of `redux-thunk`). I still need to figure this out.

client/state/form/reducer.js
DanReyLop: Same answer as above :)

classes/class-wc-connect-shipping-label.php
DanReyLop: Yep, I didn't really want to dedicate much care to the texts, since they were mostly put there as a placeholder until we have the mock-ups and we decide that this general solution is what we want. For more examples of placeholder text, see `sfdgsdkjfhasdlkfhfadd` a few lines below this :)

classes/class-wc-connect-shipping-label.php
DanReyLop: This "template string" will be sent to the React client as-is and transformed into a meaningful summary in JS code, so the WooCommerce function doesn't work for this case.

client/main.js
DanReyLop: In this case, I think having a conditional `require` is a reasonable compromise. If we `require` the 2 modules and then decide which one to render, then we are pulling the entire dependency trees of both the components in every case. This may count as micro-optimization, but if it makes start-up 50ms quicker I think is worth having one "weird" line.

allendav:
Let's wrap this up.  Here's what I suggest:

- [ ] Focus this pr to the box-packing-solution-exists part of the problem - write a separate issue to handle not-yet-box-packed orders
- [ ] Instead of passing the cart, pass the box-packed-solution
- [ ] Resolve bounds checking design for back/next

And then, once this is code reviewed, we can go from there in subsequent PRs. Not-yet-box-packed-orders could wait until after we get labels kinda working for box-packing-solution-exists orders.

DanReyLop:
I have a PR (I will send it today) that implements the auto-advancing of the steps (and even with server-side validation). It takes care of boundary checking when going back / forward. It's a separate PR to not clutter this one even more.

About the cart / packages, it's way beyond the scope of this PR. This PR's only focus should be to agree on the high-level architecture of the flow, not focus on implementing the individual steps. Since there wasn't any major disagreement on this PR, I'll proceed to create several issues to implement each individual step, starting with address and packages.

Should the next PRs be made against this branch, at least until it's in a more reasonable shape? Right now it doesn't look good enough to be merged to `master`...

kellychoffman:
Very cool to see this. Adding all my comments, even tho some may be addressed in a later PR.

- Should 'Address (cont.)' be 'Address line 2'?
- Are some fields going to be converted to dropdowns, ie `State` and `Country` in a later PR? 
- Can the 'Next' button be added to the `dialog__footer`? And have the next step appended to it, ie: 
screen shot 2016-06-28 at 11 32 08 am
- We'll also need a `Cancel` button in the footer.
- Similarly, the tab/navigation list should be in `dialog__header`. Were you going to take a stab at the layout there or were you waiting on me?

DanReyLop:
> Should 'Address (cont.)' be 'Address line 2'?
Yes. I implemented the address step ad a "proof of concept", I'm re-doing it for real right now.

> Are some fields going to be converted to dropdowns, ie State and Country in a later PR?
Same as above. The `State` field is going to be tricky to data-drive, because it depends on another field (country), but at least the Country dropdown is gonna be there.

This was a "technical" PR so I didn't pay attention to the layout, but at least I'll make the changes you pointed out, seems easy enouugh :)

kellychoffman:
> This was a "technical" PR so I didn't pay attention to the layout, but at least I'll make the changes you pointed out, seems easy enouugh :)

Thanks! Yeah, I figured but just wanted to make a note either way.

DanReyLop:
Actually, I was implementing adding the next step's name to the button label, and I realized that it may be confusing if there are skipped steps. For example, consider this flow: `Origin address` -> `Destination address` -> `Packaging`. If the Origin Address is empty but the Destination Address is pre-filled and valid, the user will fill the Origin address, click the button named `Next: Destination Address`, and be presented with the packaging form. Do you think that's better than a generic "Next"?

kellychoffman:
Good point. Lets keep the generic 'Next' for now until more of this is in place. We can always change it later if we need to.

kellychoffman:
> Added cancel button to the dialog...

Could they be added to the footer div - that empty space below the scrolling content?

screen shot 2016-06-28 at 1 48 03 pm

DanReyLop:
Short answer: no :(

Long answer: we need to change the style of the dialog anyway, if we want the tabs to stay on top when the user scrolls. Right now the whole form (including tabs and buttons) lives *inside* the `dialog__content` container, so it could be put as a normal form instead of a dialog (remember this is intended as a generic data-driven solution and as much as possible not tied with the current Printing Label requirements). Putting the buttons _outside_ of the `dialog__content` container would require a significant amount of work.

I've left the HTML code ready to be styled (put dummy `<div>` elements wrapping the buttons and the tabs), so it should be easy to style with the HTML structure as-is.

kellychoffman:
> we need to change the style of the dialog anyway, if we want the tabs to stay on top when the user scrolls. 

Aren't we using the same `dialog` component as the media modal in the calypso editor? Header and footer are fixed, content scrolls: 

header footer fixed

DanReyLop:
The `Dialog` component only allows 2 children: `dialog__content` and `dialog__action-buttons`. The media modal has its tabs _inside_ the `dialog__content` element. The only element we could re-use from the `<Dialog>` component would be the buttons bar, but right now the structure of the printing label dialog is like this:

```
<Dialog>
  <div class="dialog__content">
    // Here starts the data-driven logic-dense multi-step form component
    <ul>
      TABS
    </ul>
    { Current step form }
    <div>
      { Cancel button }
      { Next button }
    </div>
    // Here ends the data-driven logic-dense multi-step form component
  </div>
</Dialog>
```
Just thinking about moving the buttons bar outside the Form component makes my brain hurt :(

kellychoffman:
> The Dialog component only allows 2 children: dialog__content and dialog__action-buttons. The media modal has its tabs inside the dialog__content element. 

My bad. Carry on.

allendav:
> Should the next PRs be made against this branch, at least until it's in a more reasonable shape? Right now it doesn't look good enough to be merged to master...

I'm ok with this

classes/class-wc-connect-shipping-label.php
jeffstieler: Could an alternative be to exclude this summary from the layout schema, and have it rendered as part of a custom component?

This would require a new top-level component, instead of using `WCCSettingsForm`.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,316 @@
...
+		public function __construct( WC_Connect_Service_Settings_Store $settings_store ) {
+			$this->settings_store = $settings_store;
+		}
+
+		protected function get_form_data( WC_Order $order ) {
+			$formData = array();
+			$contents = array();
+
+			foreach( $order->get_items() as $item ) {
+				$product = $order->get_product_from_item( $item );
```
jeffstieler: I had this line fail locally when testing. Please handle the non-`WC_Product` return cases.

See: https://docs.woothemes.com/wc-apidocs/source-class-WC_Abstract_Order.html#1879-1896

classes/class-wc-connect-shipping-label.php
jeffstieler: Additionally, this should be coming from the server before this feature is launched.

classes/class-wc-connect-shipping-label.php
jeffstieler: You have a type in the key name.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,316 @@
...
+
+			$layout[] = array(
+				'type' => 'step',
+				'tab_title' => 'Destination',
+				'title' => 'Generate shipping label: Address verification',
+				'description' => 'asdfghjkjhgfdsaASDFGHJJHGFDSA',
+				'items' => array( array( 'items' => $dest_address_fields ) ),
+				'bypass_suggestion_flag' => 'dest_bypass_suggestion',
+				'summary' => str_replace( '{', '{dest_', $address_summary ),
+				'sugggestion_hint' => 'qwertyuiopqwertyuiopqwerty uiopqwertyuiopqwertyuiop',
```
jeffstieler: Same typo as https://github.com/Automattic/woocommerce-connect-client/pull/428/files#diff-cc0cefd4ac82fa521d8d0ed20fc65bbbR104.

jeffstieler:
Instead of making a "step" the parent component to a "group", what do you think about making it an equivalent or alternative component at the same nesting level?

I'm thinking we should be able to have a section in a form that contains a multi-step workflow. Does that make sense?

classes/class-wc-connect-shipping-label.php
DanReyLop: Should it? I've added an issue for that (#436) but I don't know how important is it for Beta.

classes/class-wc-connect-shipping-label.php
DanReyLop: That doesn't feel right... I'll just park this decision until we have mock-ups for the last step. For now the flow isn't even considering the last step, so there's no work to be undone.

DanReyLop:
> I'm thinking we should be able to have a section in a form that contains a multi-step workflow. Does that make sense?

With that requirement, the whole thing would fall apart. Also, I don't think UX-wise it makes sense, because then you would have the multi-step button ("Next, next, next") and somewhere below the real form submit button.

> Instead of making a "step" the parent component to a "group", what do you think about making it an equivalent or alternative component at the same nesting level?

Right now the steps are the top components. The real question is if we would want several "groups" inside an "step". If that's the case, we would need the 2 levels of nesting. If we don't, then we can treat a "step" as a special case of "group".

classes/class-wc-connect-shipping-label.php
allendav: > Additionally, this should be coming from the server before this feature is launched.

Agreed, but let's wait for after Beta for getting shipping label form schema and layout from the server

allendav:
@jeffstieler wrote

> Instead of making a "step" the parent component to a "group", what do you think about making it an equivalent or alternative component at the same nesting level?

@DanReyLop wrote:

> Right now the steps are the top components. The real question is if we would want several "groups" inside an "step". If that's the case, we would need the 2 levels of nesting. If we don't, then we can treat a "step" as a special case of "group".

I think Jeff has a great idea, and I think that we don't need to support 2 levels of nesting for the foreseeable future.  Basically if boils down to what if WCCSettingsForm could render `WCCSettingsGroup`s as a vertical list (like it does today) OR as tabs.  We would also need to have WCCSettingsForm filter out the `actions` group if present (only one should be allowed going forward) and render that separately (which would allow us to hook into calypso's Dialog better.)

We could probably use a new container WCCFormContent that contains the `WCCSettingsGroup`s and takes tab-style or vertical list style as a prop, something like this maybe?

img_0534

I realize this would require this PR to be dramatically rewritten, but I think this would simplify the design by eliminating the need for methods like `renderTab` and `renderMultiStepForm` and give us a firmer foundation.

What do you think @DanReyLop @jeffstieler ?

DanReyLop:
The printing label dialog is not just a form with several tabs. The logic of advancing from one step to the next validating with the server means that a "flow" if very different than just a collection of tabs.

I will make the "steps" an special kind of "groups", and I will try to abstract the "actions" group into something that applies to both cases.

client/components/wcc-settings-form/settings-group.js
```diff
@@ -56,13 +57,26 @@ const SettingsGroup = ( props ) => {
...
+			const buttons = group.items.map( ( button ) => {
+				let onClick = noop;
+				let isDisabled = false;
+				let label = button.title;
+				let isPrimary = false;
+				if ( 'submit' === button.type ) {
+					isPrimary = true;
+					if ( form.isSaving ) {
+						isDisabled = true;
+						label = button.progressTitle || label;
```
DanReyLop: Before we were actually ignoring the `layout` of the `actions` group. We can now add the `Saving...` string to the server-generated existing layouts if we want.

DanReyLop:
I've made the "steps" to be a special case of grup, removing 1 level of nesting here: https://github.com/Automattic/woocommerce-connect-client/pull/428/commits/a479bb786fd7a41f531145604cab526bc7aa0fb8

And I've extracted the Action Buttons into a separate component here: https://github.com/Automattic/woocommerce-connect-client/pull/428/commits/acdb63b73e23cb9e9d15bdb2cde63089d8148ff7

allendav:
@DanReyLop - ok to resume review? please toggle label if so - thx!

classes/class-wc-connect-shipping-label.php
DanReyLop: I've added an issue to handle that part of the form: https://github.com/Automattic/woocommerce-connect-client/issues/445

How did you get it to fail? With a product variation instead of a normal product?

classes/class-wc-connect-shipping-label.php
DanReyLop: Added an issue for this: https://github.com/Automattic/woocommerce-connect-client/issues/434

DanReyLop:
@allendav Yes, ready to resume reviewing.

As part of the work in the address steps in #442, I localized all the strings in the form schema / layout, and put sensible placeholders. So that change is not in this review but it's done :)

I think after #442 and the address suggestion branch, we should merge ths big one into `master`, or at least mark it somehow that it doesn't need more reviewing.

allendav:
This PR is plenty big and gets us off to a good start. I think we should :shipit: - do you concur, @jeffstieler  ?

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,309 @@
...
+			$required_fields[] = 'rate';
+
+			return array(
+				'required' => $required_fields,
+				'properties' => $properties,
+			);
+		}
+
+		public function meta_box( $post ) {
+			global $theorder;
```
jeffstieler: I think we should just use the `$post` object that's passed in to the metabox callback instead of relying on a global variable.

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -0,0 +1,78 @@
...
+	 */
+	protected $rest_base = 'connect/shipping-label';
+
+	/**
+	 * Register the routes for shipping labels printing.
+	 */
+	public function register_routes() {
+		register_rest_route( $this->namespace, '/' . $this->rest_base . '', array(
+			array(
+				'methods'             => WP_REST_Server::EDITABLE,
```
jeffstieler: Let's use `CREATABLE` to make this respond to `POST`.

classes/class-wc-connect-shipping-label.php
DanReyLop: The `$theorder` variable is used this way across all the other meta-boxes. I guess it's expensive to retrieve and there are no more cache layers than a global variable (which is a bit sad). I'm OK either way, but when there's a pattern there's usually a reason :)

classes/class-wc-rest-connect-shipping-label-controller.php
DanReyLop: @jeffstieler The `PUT` method is hardcoded in the client: https://github.com/Automattic/woocommerce-connect-client/blob/master/client/lib/save-form/index.js#L4

How much do you care about HTTP verb semantics? :D

client/components/action-buttons/index.js
```diff
@@ -0,0 +1,33 @@
...
+const ActionButtons = ( { buttons } ) => {
+	return (
+		<FormButtonsBar>
+			{ buttons.map( ( button, idx ) => (
+				<FormButton
+					type="button"
+					key={ idx }
+					disabled={ button.isDisabled }
+					onClick={ button.onClick }
+					isPrimary={ !! button.isPrimary }>
```
jeffstieler: We've foregone the use of `!!` for boolean casting in the past. Let's remove it.

client/components/wcc-settings-form/index.js
jeffstieler: For now, can we replace `... spinner or something ...` with an actual spinner, or a gridicon.. something?

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,309 @@
...
+
+			if ( ! is_object( $theorder ) ) {
+				$theorder = wc_get_order( $post->ID );
+			}
+
+			$order = $theorder;
```
jeffstieler: Let's remove this and rename the variable below to `$theorder`.

classes/class-wc-rest-connect-shipping-label-controller.php
jeffstieler: `saveForm` should be updated to take a `method` argument, since it's grown beyond it's original single usage of saving the shipping method instance settings.

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -0,0 +1,78 @@
...
+				'callback'            => array( $this, 'update_items' ),
+				'permission_callback' => array( $this, 'update_items_permissions_check' ),
+			),
+		) );
+	}
+
+	// TODO: remove this when the real server-side validation is implemented
+	private function validation_error( $fields ) {
+		return new WP_Error('validation_failed',
+			'Error!!!',
```
jeffstieler: Let's make this less excited and more meaningful. :)

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -0,0 +1,78 @@
...
+				),
+			)
+		);
+	}
+
+	public function update_items( $request ) {
+		$request_body = $request->get_body();
+		$settings = json_decode( $request_body, false, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
+
+		// TODO: use the real WCC server endpoint to validate
```
jeffstieler: For posterity, quoting @allendav:

> maybe we should go ahead and add a address validation method to our api class and have update_items call that
> the TODO (to replace Hawk St with actual validation) could live in the api class then

client/state/form/actions.js
```diff
@@ -7,3 +14,68 @@ export const setFormProperty = ( field, value ) => {
...
 };
+
+export const goToStep = ( stepIndex ) => {
+	return {
+		type: GO_TO_STEP,
+		step: stepIndex,
+	};
+};
+
+export const nextStep = () => ( dispatch, getState, { callbackURL, nonce, formSchema, formLayout } ) => {
```
jeffstieler: I've never seen the syntax `() => ( args ) =>`. What does it do?

jeffstieler: Wait, does this just yield another function? If so.. why have `nextStep()` generate a function if it's not dependent on an initial set of arguments?

Can you explain this block a bit more?

DanReyLop: See `react-thunk` documentation: https://github.com/gaearon/redux-thunk

With `react-thunk`, actions can either be plain objects (such as the value returned from `gotoStep(stepIndex)` ), or functions that receive a way to dispatch more actions and access state directly. It's a bit hard to grasp at first.

So the `nextStep` action creator yields a function that, when invoked with `dispatch`, `getState` and a bunch of arguments (configured in `main.js` line 22), it executes all that logic. That logic can fully read the state (using the `getState` arg) and can dispatch other actions (using the `dispatch` arg).

This allows to have complex logic inside an action creator, even async logic such as a network request.

DanReyLop: The "extra" argument, in this case an object with the properties `{ callbackURL, nonce, formSchema, formLayout }` is configured when initializing the `thunk` middleware (`main.js`, line 22) and so they are global and immutable.

client/state/form/reducer.js
```diff
@@ -28,6 +42,12 @@ export default function form( state = {}, action ) {
 		} );
 	}
 
+	if ( state.hasOwnProperty( 'shippingLabel' ) || shippingLabelActions.hasOwnProperty( action.type ) ) {
```
jeffstieler: We agreed not to use `hasOwnProperty()`

client/state/form/shipping-label/actions.js
```diff
@@ -0,0 +1,14 @@
+import * as FormActions from 'state/form/actions';
+
+export const OPEN_PRINTING_FLOW = 'OPEN_PRINTING_FLOW';
+export const EXIT_PRINTING_FLOW = 'EXIT_PRINTING_FLOW';
+
+export const openPrintingFlow = () => ( dispatch ) => {
```
jeffstieler: There's that syntax again.. are you meaning to use this?

jeffstieler:
In addition to the comments left inline, please double check all of your code formatting to make sure it meets standards.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,309 @@
...
+				),
+				admin_url( 'admin.php' )
+			) );
+
+			$admin_array = array(
+				'storeOptions' => $this->settings_store->get_shared_settings(),
+				'formSchema'   => $this->get_form_schema(),
+				'formLayout'   => $this->get_form_layout(),
+				'formData'     => $this->get_form_data( $order ),
+				'callbackURL'  => get_rest_url( null, "/wc/v1/connect/shipping-label" ),
```
jeffstieler: Please use single quotes here.

classes/class-wc-connect-shipping-label.php
```diff
@@ -0,0 +1,309 @@
...
+					'title' => 'Postal code',
+					'minLength' => 1,
+				),
+				'country' => array(
+					'type' => 'string',
+					'title' => 'Country',
+					'minLength' => 1,
+				),
+			);
+			$required_address_fields = array('first_name', 'address_1', 'city', 'state', 'postcode', 'country');
```
jeffstieler: Please add spaces before/after closing/opening parens.

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -0,0 +1,78 @@
...
+				'methods'             => WP_REST_Server::EDITABLE,
+				'callback'            => array( $this, 'update_items' ),
+				'permission_callback' => array( $this, 'update_items_permissions_check' ),
+			),
+		) );
+	}
+
+	// TODO: remove this when the real server-side validation is implemented
+	private function validation_error( $fields ) {
+		return new WP_Error('validation_failed',
```
jeffstieler: Please add a space after the opening paren.

jeffstieler:
Let's not delay this any further. :shipit:.

DanReyLop:
partying

Issue #397, opened 2016-06-09
Number fields should validate when empty


jkudish:
As a follow-up to #396, the client validation should allow number fields to be empty, rather than forcing them to be zero.

allendav:
@DanReyLop - with https://github.com/Automattic/woocommerce-connect-client/pull/410 we can close this one, right?

DanReyLop:
It would be better that the field could be empty instead of forcing a `0` there, but that's not easy to do until we decouple the form UI values from the Redux state (as we discussed).

I'll leave this open as a reminder, but descope it from Alpha 2.

jeffstieler:
Fixed in https://github.com/Automattic/woocommerce-connect-client/issues/430.

PR #430, opened 2016-06-27
Separate form values from schema'd settings.

### LINKED ISSUE #426: Separate form value state from resultant settings schema'd object.
Having the form input values and the resultant settings object be one in the same causes issues in the UX for the settings form. A good example is number-type input (see: https://github.com/Automattic/woocommerce-connect-client/issues/188)

This issue should include the following:
- [x] Change `settings` state to be "form values"
- [x] Add method to convert "form values" into a settings schema object (coerce values, etc)
- [x] https://github.com/Automattic/woocommerce-connect-client/issues/397

jeffstieler:
Fixes #426 and #397.

To test: (basically kick the tires and make sure all functionality works as before, with the exception of)
* Make sure that service adjustment will accept an empty input
* Make sure that service adjustment will accept the value `1.`


client/components/wcc-settings-form/settings-item.js
```diff
@@ -23,18 +22,19 @@ const SettingsItem = ( {
...
-	const updateSubSubValue = ( key, subKey, val ) => settingsActions.updateSettingsObjectSubField( id, key, subKey, val );
-	const removeArrayItem = ( idx ) => settingsActions.removeSettingsArrayFieldItem( id, idx );
+	const updateSubValue = ( key, val ) => settingsActions.updateSettingsField( [ id, key ], val );
+	const removeArrayItem = ( idx ) => settingsActions.removeSettingsField( [ id, idx ] );
 	const savePackage = ( packageData ) => packagesActions.savePackage( id, packageData );
 	const fieldRequired = ( -1 !== schema.required.indexOf( id ) );
-	const fieldValue = settings[ id ];
+	const fieldValue = form.settings[ id ];
 	const fieldSchema = schema.properties[ id ];
+	const fieldType = layout.type || fieldSchema.type || '';
```
DanReyLop: The type should always come from the `layout`, the `type` field in the layout is mandatory (or it should be).

jeffstieler: I'm not sure that's enforced yet - so I was going for being "liberal" with what we accept.

client/lib/utils/coerce-values.js
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Retrieve a field's schema, handling referenced schema definitions if need be.
+ *
+ * @param {Object} fieldSchema - Schema object for field.
+ * @param {Object} definitions - Common definitions.
+ * @returns {Object} - Schema object for field, potentially pulled from $ref's definition.
+ */
+const getFieldSchema = ( fieldSchema, definitions ) => {
+	if ( fieldSchema.hasOwnProperty( '$ref' ) ) {
+		const definitionKey = fieldSchema.$ref.match( /^#\/definitions\/(.+)/ );
```
DanReyLop: I think `fieldSchema.$ref.split('#/definitions/')[1]` would be cleaner (but yet again, I hate RegExps too much, so feel free to ignore me)

client/lib/utils/coerce-values.js
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Retrieve a field's schema, handling referenced schema definitions if need be.
+ *
+ * @param {Object} fieldSchema - Schema object for field.
+ * @param {Object} definitions - Common definitions.
+ * @returns {Object} - Schema object for field, potentially pulled from $ref's definition.
+ */
+const getFieldSchema = ( fieldSchema, definitions ) => {
+	if ( fieldSchema.hasOwnProperty( '$ref' ) ) {
```
DanReyLop: This `hasOwnProperty` thing always seemed overly-verbose to me. Wouldn't it be easier to just check if `fieldSchema.$ref` is `truthy`? Same thing applies to your other uses of `hasOwnProperty`.

client/lib/utils/coerce-values.js
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return value;
+};
+
+/**
+ * Coerce a form values object using the provided schema.
+ *
+ * @param {Object} schema - Schema containing type declarations.
+ * @param {Object} values - Form values.
+ * @param {Object} [definitions=null] - Optional. Schema definitions, parsed from schema if omitted.
```
DanReyLop: Is there any use case for this 3rd argument? If not, it's better to just remove it and add it later if needed.

client/state/form/settings/test/actions.js
```diff
@@ -0,0 +1,50 @@
...
+		const action = updateSettingsField( path, value );
+
+		expect( action ).to.eql( {
+			type: UPDATE_SETTINGS_FIELD,
+			path: 'testField',
+			value: 'testValue',
+		} );
+	} );
+
+	it( '#removeeSettingsField()', () => {
```
DanReyLop: Typo in `removee`

package.json
```diff
@@ -84,6 +85,7 @@
     "redux-thunk": "^2.1.0",
     "reselect": "^2.5.1",
     "sprintf-js": "1.0.3",
+    "type-coerce": "^0.1.0",
```
DanReyLop: Not sure if I trust this library. Looking at the repo: https://github.com/wacky6/type-coerce/blob/master/lib/type-coerce.js it seems that it doesn't have any stars, there was only one commit, with no issues or PRs... the code is simple enough, but this is rougly the equivalent of copying and pasting from StackOverflow. I'd suggest looking for another library, or implementing it ourselves, or reviewing the `type-coerce` library as if it was part of this PR.

DanReyLop:
The adjustment input works as expected, but the Fallback Rate input still has the previous behaviour (empty value is converted to 0 on blur).

Also, everything :boom: in the Self-Help page.

client/lib/utils/coerce-values.js
jeffstieler: It's used on lines 74 and 76 below. It's essentially a mechanism to pass down the `definitions` in recursive calls to `coerceFormValues()`.

client/lib/utils/coerce-values.js
jeffstieler: `fieldSchema.$ref.split('#/definitions/')[1]` can fail with a malformed `$ref` value. I'm going to pass on this for now.

jeffstieler:
> Also, everything 💥 in the Self-Help page.

Fixed!

package.json
jeffstieler: I agree, see https://github.com/Automattic/woocommerce-connect-client/pull/430/commits/1b18857558bafd04e2f59f461cf560a8e30faa58

DanReyLop:
:shipit: 

If you are not going to remove the `NumberInput` component in this PR, better write an issue so we don't forget. This behaviour is better (less "hacky") that the `NumberInput` one.

PR #431, opened 2016-06-28
Multi-step form auto-advancing logic and validation

DanReyLop:
This PR builds on #428 , I've separated it into a new PR to ease the code review.

This PR includes logic to:
* Disable the "Next" button when the current form step has errors.
* Automatically advance the flow until a step with errors is reached.
* For auto-advance purposes, validate client-side, and then server-side only if the client-side validation succeded. Note that the server can say that there are errors, but if they are in future steps the form must auto-advance.

I've created a silly REST endpoint mock, it has these rules:
* It will fail an address if the address line 1 isn't `Hawk St`.
* It will always fail the `Packaging` step (because from that step onward, it's uncharted lands).

So you can just play with it. If you create an order with destination to `Hawk St`, you will be able to auto-advance from the `Origin address` step to the `Packaging` step.
Note that you won't be able to advance to the next step until the current step is valid, and even if it's valid for the client, a server-side validation request will be triggered and the form won't advance to the next step if the server-side validation fails (try, for example, to put `Falcon St` in the address).

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -157,6 +157,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 				'type' => 'summary',
 				'tab_title' => 'Preview',
 				'title' => 'dfhdfghdfgh',
+				'items' => array(),
 				'action_label' => 'Print',
 				'confirmation_flag' => 'confirm',
 			);
--- a/classes/class-wc-rest-connect-shipping-label-controller.php
+++ b/classes/class-wc-rest-connect-shipping-label-controller.php
@@ -37,14 +37,33 @@ class WC_REST_Connect_Shipping_Label_Controller extends WP_REST_Controller {
 		) );
 	}
 
-	public function update_items( $request ) {
+	private function validation_error( $fields ) {
+		return new WP_Error('validation_failed',
+			'Error!!!',
+			array(
+				'status' => 400,
+				'error' => 'validation_failure',
+				'data' => array(
+					'fields' => $fields,
+				),
+			)
+		);
+	}
 
+	public function update_items( $request ) {
 		$request_body = $request->get_body();
 		$settings = json_decode( $request_body, false, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
 
-		var_dump( $settings );
+		if ( $settings->orig_address_1 !== 'Hawk St' ) {
+			return $this->validation_error( array( 'orig_address_1' ) );
 
-		return new WP_REST_Response( array( 'success' => true ), 200 );
+		}
+
+		if ( $settings->dest_address_1 !== 'Hawk St' ) {
+			return $this->validation_error( array( 'dest_address_1' ) );
+		}
+
+		return $this->validation_error( array( 'cart', 'rate' ) );
 	}
 
 	/**
--- a/classes/class-wc-rest-connect-shipping-label-controller.php
+++ b/classes/class-wc-rest-connect-shipping-label-controller.php
@@ -1,0 +1,57 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( class_exists( 'WC_REST_Connect_Shipping_Label_Controller' ) ) {
+	return;
+}
+
+class WC_REST_Connect_Shipping_Label_Controller extends WP_REST_Controller {
+
+	/**
+	 * Endpoint namespace.
+	 *
+	 * @var string
+	 */
+	protected $namespace = 'wc/v1';
+
+	/**
+	 * Route base.
+	 *
+	 * @var string
+	 */
+	protected $rest_base = 'connect/shipping-label';
+
+	/**
+	 * Register the routes for order notes.
+	 */
+	public function register_routes() {
+		register_rest_route( $this->namespace, '/' . $this->rest_base . '', array(
+			array(
+				'methods'             => WP_REST_Server::EDITABLE,
+				'callback'            => array( $this, 'update_items' ),
+				'permission_callback' => array( $this, 'update_items_permissions_check' ),
+			),
+		) );
+	}
+
+	public function update_items( $request ) {
+
+		$request_body = $request->get_body();
+		$settings = json_decode( $request_body, false, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
+
+		var_dump( $settings );
+
+		return new WP_REST_Response( array( 'success' => true ), 200 );
+	}
+
+	/**
+	 * Validate the requester's permissions
+	 */
+	public function update_items_permissions_check( $request ) {
+		return current_user_can( 'manage_woocommerce' );
+	}
+
+}
--- a/classes/class-wc-rest-connect-shipping-label-controller.php
+++ b/classes/class-wc-rest-connect-shipping-label-controller.php
@@ -25,7 +25,7 @@ class WC_REST_Connect_Shipping_Label_Controller extends WP_REST_Controller {
 	protected $rest_base = 'connect/shipping-label';
 
 	/**
-	 * Register the routes for order notes.
+	 * Register the routes for shipping labels printing.
 	 */
 	public function register_routes() {
 		register_rest_route( $this->namespace, '/' . $this->rest_base . '', array(
@@ -37,6 +37,7 @@ class WC_REST_Connect_Shipping_Label_Controller extends WP_REST_Controller {
 		) );
 	}
 
+	// TODO: remove this when the real server-side validation is implemented
 	private function validation_error( $fields ) {
 		return new WP_Error('validation_failed',
 			'Error!!!',
@@ -54,6 +55,7 @@ class WC_REST_Connect_Shipping_Label_Controller extends WP_REST_Controller {
 		$request_body = $request->get_body();
 		$settings = json_decode( $request_body, false, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
 
+		// TODO: use the real WCC server endpoint to validate
 		if ( $settings->orig_address_1 !== 'Hawk St' ) {
 			return $this->validation_error( array( 'orig_address_1' ) );
 
@@ -70,7 +72,7 @@ class WC_REST_Connect_Shipping_Label_Controller extends WP_REST_Controller {
 	 * Validate the requester's permissions
 	 */
 	public function update_items_permissions_check( $request ) {
-		return current_user_can( 'manage_woocommerce' );
+		return current_user_can( 'edit_shop_order' );
 	}
 
 }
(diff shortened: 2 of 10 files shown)


client/state/form/actions.js
```diff
@@ -10,15 +16,67 @@ export const setFormProperty = ( field, value ) => {
...
-};
-
 export const goToStep = ( stepIndex ) => {
 	return {
 		type: GO_TO_STEP,
 		step: stepIndex,
 	};
 };
+
+export const nextStep = () => ( dispatch, getState, { callbackURL, nonce, formSchema, formLayout } ) => {
```
DanReyLop: Maybe this action is _too_ complex, but that's because the actions responsibility (auto-advance the form) has a lot of rules (client-side validation, async server-side validation, etc). If you feel this is hard to read I could try separating this logic into smaller chunks.

client/state/form/shipping-label/actions.js
```diff
@@ -1,9 +1,13 @@
...
 export const EXIT_PRINTING_FLOW = 'EXIT_PRINTING_FLOW';
 
-export const openPrintingFlow = () => ( {
-	type: OPEN_PRINTING_FLOW,
-} );
+export const openPrintingFlow = () => ( dispatch ) => {
+	dispatch( { type: OPEN_PRINTING_FLOW } );
+	dispatch( FormActions.setFormProperty( 'pristine', false ) );
+	dispatch( FormActions.nextStep() );
+};
```
DanReyLop: I don't know if this kind of code is what we want to achieve. `redux-thunk` allows this kind of things to be done inside actions, so actions stop being plain objects and start becoming "recipes", or even places to put complex logic. What do you think about this vs having several reducers interpret the `OPEN_PRINTING_FLOW` action?

allendav: You want those to happen in order, right? It appears that the "redux way" to do this might be this way: https://github.com/tshelburne/redux-batched-actions

classes/class-wc-connect-shipping-label.php
```diff
@@ -157,6 +157,7 @@ protected function get_form_layout() {
 				'type' => 'summary',
 				'tab_title' => 'Preview',
 				'title' => 'dfhdfghdfgh',
+				'items' => array(),
 				'action_label' => 'Print',
```
allendav: Please go ahead and __( 'Print', 'woocommerce' ) now

classes/class-wc-connect-shipping-label.php
```diff
@@ -157,6 +157,7 @@ protected function get_form_layout() {
 				'type' => 'summary',
 				'tab_title' => 'Preview',
```
allendav: Please go ahead and `__( 'Preview', 'woocommerce' )` now

classes/class-wc-connect-shipping-label.php
```diff
@@ -157,6 +157,7 @@ protected function get_form_layout() {
...
 				'title' => 'dfhdfghdfgh',
```
allendav: How will this be used? Go ahead and put something like `__( 'TODO Title', 'woocommerce' )` now

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -0,0 +1,76 @@
...
+
+	/**
+	 * Route base.
+	 *
+	 * @var string
+	 */
+	protected $rest_base = 'connect/shipping-label';
+
+	/**
+	 * Register the routes for order notes.
```
allendav: Copy pasta - I probably messed this up in the other endpoint too

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -0,0 +1,76 @@
...
+				'methods'             => WP_REST_Server::EDITABLE,
+				'callback'            => array( $this, 'update_items' ),
+				'permission_callback' => array( $this, 'update_items_permissions_check' ),
+			),
+		) );
+	}
+
+	private function validation_error( $fields ) {
+		return new WP_Error('validation_failed',
+			'Error!!!',
```
allendav: Let's go ahead and at least `__( 'TODO Error', 'woocommerce' )` this now. If we get too many of these we're going to miss something eventually :)

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -0,0 +1,76 @@
...
+				),
+			)
+		);
+	}
+
+	public function update_items( $request ) {
+		$request_body = $request->get_body();
+		$settings = json_decode( $request_body, false, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
+
+		if ( $settings->orig_address_1 !== 'Hawk St' ) {
```
allendav: Please add a `\\ TODO call server to validate address` here for good form

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		return $this->validation_error( array( 'cart', 'rate' ) );
+	}
+
+	/**
+	 * Validate the requester's permissions
+	 */
+	public function update_items_permissions_check( $request ) {
+		return current_user_can( 'manage_woocommerce' );
```
allendav: Is this capability higher than needed? Is there a capability that precludes changing settings but allows Edit Order work? IDK  (I used manage_woocommerce for the other endpoints because they change settings.)

client/components/wcc-settings-form/index.js
```diff
@@ -35,7 +35,10 @@ const WCCSettingsForm = ( props ) => {
 		const label = ( currentStepLayout || {} ).action_label || __( 'Next' );
 		return (
 			<FormButtonsBar>
-				<FormButton type="button" onClick={ props.formActions.nextStep }>
+				<FormButton
+					type="button"
+					disabled={ ! currentStepLayout || 0 < props.stepErrors.length || props.form.isSaving }
```
allendav: Let's make this logic a const like label (I don't like _complicated_ logic in the renders if it can be avoided

client/components/wcc-settings-form/index.js
```diff
@@ -35,7 +35,10 @@ const WCCSettingsForm = ( props ) => {
...
+					onClick={ props.formActions.nextStep }>
```
allendav: I wonder if we should have a separate component for these multi-step forms to reduce the risk of collateral damage to the settings and self-help views. Keeping multi-step and single-step logic clean in one component could be challenging.  We could DRY things with utility requires.

@jeffstieler - what do you think?

client/lib/save-form/index.js
```diff
@@ -34,7 +33,7 @@ const saveForm = ( setIsSaving, setSuccess, setError, url, nonce, formData ) =>
 			}
 
 			return setError( JSON.stringify( json ) );
-		} );
+		} ).then( () => setIsSaving( false ) );
```
allendav: I'm curious - why this change?

client/state/form/actions.js
allendav: Since this action submits the form, maybe a better name would be `submitStep` since `nextStep` could be casually misinterpreted to mean simply mutating the step in state

client/state/form/shipping-label/actions.js
allendav: Like we talked about in chat this morning, this might be a good place to use https://github.com/tshelburne/redux-batched-actions - especially as you point out for rendering performance

client/state/selectors.js
```diff
@@ -38,7 +38,22 @@ const getRawFormErrors = ( schema, data ) => {
...
 	( state ) => state.form.errors,
-	( state, props ) => props.schema,
+	( state, schema ) => schema,
 	( state ) => state.settings,
 	( errors, schema, data ) => removeErrorDataPathRoot( errors || getRawFormErrors( schema, data ) )
 );
+
+export const getStepFormErrors = createSelector(
+	( state ) => state,
+	( state, schema ) => schema,
```
allendav: Let's DRY at least this (should improve readability too)

classes/class-wc-rest-connect-shipping-label-controller.php
DanReyLop: I'm gonna put `edit_shop_order` then.

client/components/wcc-settings-form/index.js
DanReyLop: The single-step form doesn't have any logic here. If you look at the `render` method, if the form is single-step, then it just render a `<WCCSettingsStep>` component. So it's very unlikely that all this logic would break the single-step forms (famous last words)

client/lib/save-form/index.js
DanReyLop: To make sure that `setIsSaving(false)` is the last thing this function does. I use it in the `nextStep` action to detect when the form finished validating.

client/state/selectors.js
DanReyLop: What do you mean? This is how `reselect` works. Really, it passes the 3 parameters (`state`, `schema` and `layout`) to all the functions, but since they are not needed I just omitted them.

client/state/form/shipping-label/actions.js
DanReyLop: I've tried, but... https://github.com/tshelburne/redux-batched-actions/issues/3
We can't use (for now) batched actions with thunks inside. It would be a good way to contribute back if we feel that our life is empty without this feature :)

DanReyLop:
@allendav I promise I will clean-up the layout strings and add i18n in the main branch, but please don't hold this PR hostage until I do :D

I thought that the schema & layout was going to be generated in the WCC server, so I didn't care much for how I implemented in the merchant host. But since it's possible we release Beta with the schema generated in the host, I'm gonna fix it :)

client/state/selectors.js
allendav: I meant adding something like this befpre `getStepFormErrors` and `getFormErrors`

```
const getSchema = ( state, schema ) => schema;
```

and then, in `getStepFormErrors` and `getFormErrors` do something like this instead:

```
export const getStepFormErrors = createSelector(
	( state ) => state,
	getSchema,
```

IDK, its a small DRY - maybe too small

DanReyLop: I personally think that would even hurt readability :(

allendav:
It is OK to merge this into #428 as is and then we'll wrap up comments there.

classes/class-wc-rest-connect-shipping-label-controller.php
DanReyLop: The change looked innocent enough so I didn't test it, but actually changing this to `edit_shop_order` makes my admin account throw a 403 error.

According to this document: https://docs.woothemes.com/document/roles-capabilities/
`manage_woocommerce` should be pretty much the only permission we should care about.

I was confused because I looked at this code and it seemed that there were more fine-grained permissions: https://github.com/woothemes/woocommerce/blob/2267f961347f928749b59e8003e628d1deb99203/includes/class-wc-install.php#L673

PR #442, opened 2016-07-01
[Labels] Address form

### LINKED ISSUE #435: [Labels] Address step
Mock-up:
screen-shot-2016-06-21-at-4-54-42-pm1

Notes:
* The list of states should change when the Country is changed
* Not all countries have something equivalent to `State`, so that field shouldn't be required
* Name and Company fields of the Destination Address shouldn't be editable (@allendav Why? They are editable in `wp-admin` anyways)

DanReyLop:
Fixes #435 

In this PR:
* Put the real, localized strings of the Address steps
* Implemented the country and state selectors
* Got the address form(s) ready for design

To test this, just go to a pending WooCommerce order with a physical product and click the "Create Label" button.

We have 3 different cases for the State selector:
* There are countries with States, and WooCommerce has a list of them. Example: US, Spain
* Some countries have states (or an equivalent concept) but WoCommerce doesn't have a list of them. In that case, a text field is displayed instead of a dropdown. Example: UK
* Some countries doesn't have states. In that case, we hide the entire field. Example: Sri Lanka

Screenshot:
address

@allendav @jeffstieler @nabsul 

@kellychoffman Do you want to start with the styling of this step? Or is it still too early in the design process?


classes/class-wc-connect-shipping-label.php
```diff
@@ -36,89 +36,118 @@ protected function get_form_data( WC_Order $order ) {
...
 				array(
 					'key' => 'city',
-					'validation_hint' => 'This field is required.',
+					'validation_hint' => __( 'Required.', 'woocommerce' ),
 				),
 				array(
 					'key' => 'state',
-					'validation_hint' => 'This field is required.',
+					'type' => 'state',
+					'dataset' => $this->get_states_map(),
```
DanReyLop: This massive array will end up being duplicated (once for the origin address form and another time for the destination address form). Do you have any suggestion to send only 1 array in the payload? This doesn't fit our `#definition` model...

nabsul:
Right now, fields are highlighted in red before I even start typing things into the form. It shouldn't be this way.

The fields should only get highlighted in red if I try to submit invalid values...

nabsul:
Also, if I type something in the form and then click cancel, I get that annoying "Are you sure you want to leave this page" question when I try to leave the page.

DanReyLop:
> Right now, fields are highlighted in red before I even start typing things into the form. It shouldn't be this way.

That's how it works with the rest of the forms. If you create a new USPS service, the `Postal Code` field will highlight red before you even type anything. If we provide client-side real-time validation, I can't see other way to do it with the architecture we have right now :(
I don't think this would be very important, because 99.9% of the time the addresses will be pre-filled, with at least the required fields.

> Also, if I type something in the form and then click cancel, I get that annoying "Are you sure you want to leave this page" question when I try to leave the page.

Yep, I noticed that, I fixed it in a later commit (it belongs to another PR).

nabsul:
> That's how it works with the rest of the forms. 

Definitely not true! https://github.com/Automattic/woocommerce-connect-client/pull/364

We spent a lot of time discussing exactly how forms and validations should behave in that PR. I think it's important to stay consistent, otherwise the user is going to get (subconsciously) confused by these subtle differences in behavior.

> If you create a new USPS service, the Postal Code field will highlight red before you even type anything. 

It might mean we need to fix those forms. 

> I can't see other way to do it with the architecture we have right now :(

It's definitely being done in the "Add Packages" form.

DanReyLop:
Well, you got me there :)

The "Add Packages" form seems to live outside of our data-driven bubble, so it has its own separate logic for validation.

> It might mean we need to fix those forms.

I would say that's out of the scope of this PR. Before this PR it wasn't very noticeable because the only required field in a data-driven form that wasn't prefilled was the Postal Code. This form just brought a pre-existing issue to light.

I'll create a separate issue to fix this in a way that affects all our data-driven forms.

classes/class-wc-connect-shipping-label.php
jeffstieler: We'll likely want to keep populating these with data from the WC instance.. so perhaps a custom `type` that renders with a specific country/state component makes the most sense.

jeffstieler:
Do you have a use in mind for the `id + '_container'` fieldset IDs?

classes/class-wc-connect-shipping-label.php
DanReyLop: I'm not sure I follow... do you mean having both the state and country fields as one single field in our layout?

DanReyLop:
@jeffstieler Yes, that's needed to style the fields. For example, in the mock-ups, City/State/ZIP are displayed `inline`. It's a bit hacky to do it this way, specially if the formLayout is going to be in the WCC server, but it's the only way I can think to do it without making ad-hoc field types for city, state and postal code.

classes/class-wc-connect-shipping-label.php
jeffstieler: No, I don't think the state and country fields should be the same component - that was a definite oversight on my part.

We should bootstrap the country-to-state data from the WC store, and pass the dataset as a prop which gets derived from state + the bootstrapped data. (i.e. the "state" dataset prop is calculated like the "errors" prop is)

DanReyLop: So, just send the dataset as a global JS variable? Or as part of the `wcConnectData` object? That's an easy change, but I want to double-check we are in the same page :)

jeffstieler: As part of the `wcConnectData` object - then pass it through to the component via `props`.

DanReyLop: Done. It seemed reasonable to use the same data for the country dropdown and the state dropdown, so I had to make a custom component for the country too.

classes/class-wc-connect-shipping-label.php
```diff
@@ -36,89 +36,111 @@ protected function get_form_data( WC_Order $order ) {
...
 			}
 
 			$layout[] = array(
 				'type' => 'step',
-				'tab_title' => 'Origin',
-				'title' => 'Generate shipping label: Address verification (origin)',
-				'description' => 'asdfghjkjhgfdsaASDFGHJJHGFDSA',
+				'tab_title' => __( 'Origin', 'woocommerce' ),
+				'title' => __( 'Generate shipping label: Origin address', 'woocommerce' ),
+				'description' => __( 'Confirm the address of the recipient is correct to ensure the package will get to the correct destination', 'woocommerce' ),
```
allendav: I think this should say something like "Confirm the address you are shipping from is correct..." right?

classes/class-wc-connect-shipping-label.php
```diff
@@ -224,42 +242,42 @@ protected function get_form_schema() {
...
-						'title' => 'Width',
+						'title' => __( 'Width', 'woocommerce' ),
 					),
 				),
 			);
 
 			$properties[ 'cart' ] = array(
 				'type' => 'array',
-				'title' => 'Items to send',
+				'title' => __( 'Items to send', 'woocommerce' ),
```
allendav: Don't forget, packages !== items, and the label dialog doesn't really need to know about the concept of the "cart" - it really needs to know about the packages array (or singular package for this intiial revision.)

classes/class-wc-connect-shipping-label.php
```diff
@@ -287,13 +305,14 @@ public function meta_box( $post ) {
...
-				'formLayout'   => $this->get_form_layout(),
-				'formData'     => $this->get_form_data( $order ),
-				'callbackURL'  => get_rest_url( null, "/wc/v1/connect/shipping-label" ),
-				'nonce'        => wp_create_nonce( 'wp_rest' ),
-				'rootView'     => 'shipping-label',
+				'storeOptions'  => $this->settings_store->get_shared_settings(),
+				'formSchema'    => $this->get_form_schema(),
+				'formLayout'    => $this->get_form_layout(),
+				'formData'      => $this->get_form_data( $order ),
+				'countriesData' => $this->get_states_map(),
```
allendav: Instead of calling this countriesData, could we be more specific and call it countriesStatesMap

client/components/state-dropdown/index.js
```diff
@@ -0,0 +1,44 @@
+import React, { PropTypes } from 'react';
+import Dropdown from 'components/dropdown';
+import TextField from 'components/text-field';
+import { translate as __ } from 'lib/mixins/i18n';
+
+const StateDropdown = ( props ) => {
+	const statesMap = ( props.countriesData[ props.countryCode ] || {} ).states;
+
+	if ( ! statesMap ) { // We don't have a list of states for this country
```
allendav: If we don't have a list of states for a country, would it be safe to assume the country has no states and therefore we wouldn't even need this TextField? Or is that not a safe assumption.

client/main.js
```diff
@@ -40,6 +41,7 @@ let render = () => {
 				storeOptions={ storeOptions }
 				schema={ formSchema }
 				layout={ formLayout }
+				countriesData={ countriesData }
```
allendav: Instead of having countriesData at all, why not add it to storeOptions?  I would hate to see us get into a habit of adding too many objects that we end up needing to "wire through" to their consumers this way.

allendav:
What's the secret sauce to get my origin address to validate?

screen shot 2016-07-08 at 11 18 59 am

allendav:
We should be able to prefill the state for origin from the base location too

screen shot 2016-07-08 at 11 24 40 am

client/components/state-dropdown/index.js
DanReyLop: I mirrored the way WC core builds the states map. For each country:
* If the states map is an associative array, then that's the list of the states for that country.
* If the states map is an **empty** array, then the country has no states.
* If the states map doesn't have an entry for that country, then the country **may** have states, but we don't have a list of them, so show a TextField just in case.

client/main.js
DanReyLop: Agreed

DanReyLop:
@allendav: Secret sauce to validate: Street address must be `Hawk St` .

classes/class-wc-connect-shipping-label.php
DanReyLop: Since it needs to know also about the contents of the package, I believed that the "Shopping Cart" concept was a good fit here. The way I see it, the merchant will be phisically putting the items into the boxes while he looks at that step of the form (and weighting them), so the list of items that must be inside each box is important too.

DanReyLop:
@allendav: I'll add pre-filling of those fields to https://github.com/Automattic/woocommerce-connect-client/pull/450

PR #444, opened 2016-07-04
Improved the performance of form validation by caching the validator.

### LINKED ISSUE #443: Input lags when writting on a data-driven form field
For example, going to the USPS service configuration page and writing random garbage on the "Postal Code" field.

I've done a quick profile, and every React update takes about 100ms in my computer. React is smart enough to do several changes in the same update, but it's still a poor user experience. While I'm typing (if I type fast enough), 1 CPU core is being used constantly at 100%.

I don't know for sure what is responsible of this, but I would guess that JS is doing too much for each keystroke (type coercion, form validation, component update). The easiest solution would be to throttle or debounce those tasks, if it's doable to do so in the React world.

DanReyLop:
This makes each form validation to take about 1ms. Before, in my machine,
it took between 10 and 15ms.

Fixes #443 

Turns out, the **mayor** lag I was noticing was caused by React validating PropTypes, and so it doesn't happen in release builds. Still, trying to fix it I run across this optimisation (caching the `is-my-json-valid` validator).

@jeffstieler @allendav @nabsul

### DIFF
--- a/client/components/shipping/packages/modal-errors.js
+++ b/client/components/shipping/packages/modal-errors.js
@@ -3,6 +3,9 @@ import trim from 'lodash/trim';
 import omitBy from 'lodash/omitBy';
 import reduce from 'lodash/reduce';
 import validator from 'is-my-json-valid';
+import memoize from 'lodash/memoize';
+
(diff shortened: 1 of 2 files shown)


nabsul:
very cool. I'd never heard of `memoize` before.

:shipit:

PR #450, opened 2016-07-06
Store the merchant address so it's prefilled the next times the label printing dialog is opened

### LINKED ISSUE #434: [Labels] Remember the merchant's warehouse address
The first time the merchant prints a label, he has to input the origin address from which to send the package (there's no way around that).

When a label is successfully printed, we should extract the origin address from the request made to the WCC server and store it somewhere in the merchant host's database. Any subsequent times the merchant wants to print a label, that information should be pre-filled in the form.

DanReyLop:
Fixes #434 

To test it:
* Open the label printing dialog.
* Input a complete origin address.
* Click the "Next" button.
* Reload the page and open the label printing dialog again.

The second time you open the label printing dialog, the Origin step should be pre-filled, and if it's "valid" for the server (remember, only `Hawk St` addresses are acceptable for now) it will auto-skip the step entirely.

Ideally, the merchant address should only be "remembered" if it was used to actually purchase a label, but since the integration between the merchant host and the WCC server isn't implemented yet, that's not possible.


classes/class-wc-connect-shipping-label.php
```diff
@@ -38,8 +38,10 @@ protected function get_form_data( WC_Order $order ) {
 			}
 			$formData[ 'cart' ] = $contents;
 
+			foreach( $this->settings_store->get_origin_address() as $key => $value ) {
+				$formData[ 'orig_' . $key ] = $value;
+			}
```
allendav: This would be a place where you could pre-fill a bunch of stuff ( https://github.com/Automattic/woocommerce-connect-client/pull/442#issuecomment-231435906 )

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -51,10 +60,29 @@ private function validation_error( $fields ) {
 		);
 	}
 
+	private function starts_with( $haystack, $needle ) {
+		 return $needle === substr( $haystack, 0, strlen( $needle ) );
+	}
+
+	private function store_origin_address( $settings ) {
```
allendav: This makes me a little nervous. A client could add arbitrary orig_* fields this way. Maybe we should have a whitelist instead?

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -51,10 +60,29 @@ private function validation_error( $fields ) {
...
+		}
+		$this->settings_store->update_origin_address( $address );
+	}
+
 	public function update_items( $request ) {
 		$request_body = $request->get_body();
 		$settings = json_decode( $request_body, false, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
 
+		if ( true /* TODO: Trigger this only when the label was correctly purchased */ ) {
+			$this->store_origin_address( $settings );
```
allendav: I think it is OK to store the origin address once it validates.

classes/class-wc-rest-connect-shipping-label-controller.php
DanReyLop: What would the client gain by sending unexpected garbage to our server?

classes/class-wc-connect-service-settings-store.php
```diff
@@ -42,6 +42,14 @@ public function get_shared_settings() {
 			);
 		}
 
+		public function get_origin_address() {
+			return get_option( 'woocommerce_connect_origin_address', array() );
```
allendav: We are using the pattern `wc_connect_` for our other options. Let's do that here too

classes/class-wc-connect-service-settings-store.php
```diff
@@ -42,6 +42,14 @@ public function get_shared_settings() {
...
+		}
+
+		public function update_origin_address( $address ) {
+			return update_option( 'woocommerce_connect_origin_address', $address );
```
allendav: Same comment

classes/class-wc-rest-connect-shipping-label-controller.php
allendav: Who knows :) But I'd prevent it nonetheless

PR #455, opened 2016-07-13
[Labels] Address form

### LINKED ISSUE #435: [Labels] Address step

DanReyLop:
Fixes #435

See #442 
This PR has the exact same code that #442, but the destination branch is `master`. Apparently you can't change the destination branch of an existing PR, so there you go, new one.

In this PR:
* Put the real, localized strings of the Address steps
* Implemented the country and state selectors
* Got the address form(s) ready for design

To test this, just go to a pending WooCommerce order with a physical product and click the "Create Label" button.

We have 3 different cases for the State selector:
* There are countries with States, and WooCommerce has a list of them. Example: US, Spain
* Some countries have states (or an equivalent concept) but WoCommerce doesn't have a list of them. In that case, a text field is displayed instead of a dropdown. Example: UK
* Some countries doesn't have states. In that case, we hide the entire field. Example: Sri Lanka

Screenshot:
address

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -82,6 +82,8 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 				array(
 					'key' => 'country',
 					'validation_hint' => 'This field is required.',
+					'type' => 'dropdown',
+					'titleMap' => WC()->countries->get_countries(),
 				),
 			);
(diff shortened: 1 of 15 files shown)


nabsul:
:shipit:

PR #456, opened 2016-07-13
Store the merchant address so it's prefilled the next times the label printing dialog is opened

### LINKED ISSUE #434: [Labels] Remember the merchant's warehouse address

DanReyLop:
Fixes #434 

See #450 
This PR has the exact same code that #450, but the destination branch is `master`. Apparently you can't change the destination branch of an existing PR, so there you go, new one.

To test it:
* Open the label printing dialog.
* Input a complete origin address.
* Click the "Next" button.
* Reload the page and open the label printing dialog again.

The second time you open the label printing dialog, the Origin step should be pre-filled, and if it's "valid" for the server (remember, only `Hawk St` addresses are acceptable for now) it will auto-skip the step entirely.

Ideally, the merchant address should only be "remembered" if it was used to actually purchase a label, but since the integration between the merchant host and the WCC server isn't implemented yet, that's not possible.

### DIFF
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -43,11 +43,19 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 		}
 
 		public function get_origin_address() {
-			return get_option( 'woocommerce_connect_origin_address', array() );
+			$wc_address_fields = array();
+			$wc_address_fields[ 'company' ] = get_bloginfo( 'name' );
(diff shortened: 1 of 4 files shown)


nabsul:
:shipit: after conflict resolution.

PR #458, opened 2016-07-13
[Labels] Address suggestions

### LINKED ISSUE #439: [Labels] Address suggestion
Mock-up:
screen-shot-2016-06-22-at-7-27-05-am

Requirements:
* After validating the address step with the server, the address validation can have 3 possible results: `valid`, `invalid` and `suggestion`.
* If the server offers a suggestion, the form shouldn't advance to the next step.
* The suggested address will have all the differences with the original address highlighted.
* The suggested address will be pre-selected.
* The user can do one of these 3 actions:
  * Select the suggested address and advance. This will put the suggested address data in the form values and continue to the next step without further validation.
  * Select the original address and advance. In that case, the form will have an special flag set to indicate the server that it should accept that address instead of trying to suggest an alternative.
  * Edit the original address. That case is essentially to "go back" to the address form. When the user clicks "Next" again, the same server-side validation will be performed and maybe the address suggestion will pop up again.

DanReyLop:
Fixes #439 

This PR is big so if you want you can review individual commits.

* https://github.com/Automattic/woocommerce-connect-client/commit/8eff5da3fc7a15c843197627ccac00a29dd654df has an important change: instead of the server returning just a list of errant fields, for each field it can return an arbitrary object. So, for example, for each field it can return a custom error message, or even indicate the level of severity of that error.
* https://github.com/Automattic/woocommerce-connect-client/commit/afe7fa2f0261f33fc857bf5cdbab2bbf5942a3fe and https://github.com/Automattic/woocommerce-connect-client/commit/6d5ff532c8e3a921df6c862909b338e6cf2e6889 implement the actual address suggestion feature, implemented without caring much for code structure (so you can actually focus on the feature implementation).
* The rest of commits are refactors and clean-ups.

Current state of the UI:
address_suggestion

See #439 for the expected behaviour.

Instructions to test (magic sauce):
* The server will accept an address if the address name is `Hawk St`
* If the address name is exactly `Awk St`, the server will _suggest_ to be changed to `Hawk St`
* Any other street name will be a plain validation error. To show the power of the new error fields structure, the server will return a custom error message (`The server doesn't like that address!`)


client/state/form/reducer.js
```diff
@@ -21,8 +22,12 @@ reducers[ SET_FORM_PROPERTY ] = ( state, action ) => {
...
 };
 
-reducers[ NEXT_STEP ] = ( state ) => {
-	return Object.assign( {}, state, { currentStep: state.currentStep + 1 } );
+reducers[ BACK_FROM_SUGGESTION ] = ( state ) => {
+	return [
+		setFormProperty( 'errors', {} ),
+		setFormProperty( 'acceptSuggestion', undefined ),
+		setFormProperty( 'pristine', false ),
+	].reduce( form, state );
```
DanReyLop: I implemented it this way on purpose to trigger a discussion. What do you think about this way of implementing an action which must perform a sequence of tasks? The other way to do it would be to make an action creator with several `dispatch()` calls. I can see pros and cons either way. We want to keep reducers simple, but if we go too far then the Redux log will just be plagued with `SET_PROPERTY`, `UPDATE_FIELD` actions that are _not_ semantic at all.

classes/class-wc-connect-api-client.php
```diff
@@ -95,16 +95,36 @@ private function validation_error( $fields ) {
...
+			if ( $label_settings->orig_address_1 === 'Awk St' ) {
+				if ( ! $label_settings->orig_bypass_suggestion ) {
+					$errors[ 'orig_address_1' ] = array(
+						'value' => 'Hawk St',
+						'level' => 'suggestion',
+					);
+				}
+			} else if ( $label_settings->orig_address_1 !== 'Hawk St' ) {
+				$errors[ 'orig_address_1' ] = array(
+					'value' => 'The server doesn\'t like that street!',
```
allendav: Please add a TODO to ensure we clean this up

client/components/suggestion/index.js
```diff
@@ -0,0 +1,60 @@
+import React, { PropTypes } from 'react';
```
allendav: Suggestion is kinda broad. Let's be more specific and rename this component to something like suggested-address instead

client/components/suggestion/index.js
```diff
@@ -0,0 +1,60 @@
...
+		</FormLabel>
+	);
+};
+
+const Suggestion = ( { acceptSuggestion, formValues, formActions, layout, suggestions } ) => {
+	return (
+		<div>
+			<RadioButton
+				checked={ ! acceptSuggestion }
+				onChange={ () => formActions.setFormProperty( 'acceptSuggestion', false ) } >
```
allendav: This (suggestion) is a low level component. Should the onChange callback be passed in as a prop instead of defined here to avoid tight coupling to our form implementation?

client/components/suggestion/index.js
```diff
@@ -0,0 +1,60 @@
...
+				{ layout.suggestion_original_title }
+				<Summary
+					formValues={ formValues }
+					layoutItems={ layout.items }
+					summaryTemplate={ layout.template } />
+				<Button onClick={ formActions.backFromSuggestion } >
```
allendav: What does this button do? The action name isn't clear. Also, should this callback be passed in as a prop (same comment basically as above)

client/components/suggestion/index.js
```diff
@@ -0,0 +1,60 @@
...
+					{ __( 'Edit' ) }
+				</Button>
+			</RadioButton>
+			<RadioButton
+				checked={ acceptSuggestion }
+				onChange={ () => formActions.setFormProperty( 'acceptSuggestion', true ) } >
```
allendav: Same comment as above about passing in props

client/components/suggestion/index.js
```diff
@@ -0,0 +1,60 @@
...
+		</div>
+	);
+};
+
+Suggestion.propTypes = {
+	layout: PropTypes.shape( {
+		template: PropTypes.string,
+		items: PropTypes.array,
+		suggestion_original_title: PropTypes.string,
+		suggestion_corrected_title: PropTypes.string,
```
allendav: Instead of having these two props, what if we had an array of suggestions+titles for our address picker. At some point in MVP we will probably need to support choosing from among > 2 addresses (think about amazon's checkout where it shows multiple addresses and asks you to pick one.)

client/components/summary/index.js
```diff
@@ -0,0 +1,96 @@
+import React, { PropTypes } from 'react';
+import find from 'lodash/find';
+
+const Summary = ( { overrideFields, formValues, layoutItems, summaryTemplate } ) => {
```
allendav: This component name is also too broad. I have no idea what it does from the name nor the code. Please choose a more descriptive name.

client/components/tabs/index.js
```diff
@@ -0,0 +1,30 @@
+import React, { PropTypes } from 'react';
```
allendav: Instead of Tabs, how About TabBar since this only renders a Tab Bar and not the Tab content.

client/components/wcc-settings-form/index.js
```diff
@@ -8,43 +8,41 @@ import * as FormActions from 'state/form/actions';
...
-		}
-		return <span>{ label }</span>;
-	};
-
-	const renderMultiStepForm = () => {
+	const getActionButtons = () => {
 		const buttons = [];
+		const currentStepLayout = props.layout[ currentStep ];
+		let submitEnabled = currentStepLayout && isEmpty( props.stepErrors ) && ! props.form.isSaving;
+		if ( ! isEmpty( props.stepSuggestions ) ) {
```
allendav: Please add a comment here explaining why we want to enable submit in this case to aid the next person editing this logic.

client/components/wcc-settings-form/settings-group.js
```diff
@@ -57,6 +47,26 @@ const SettingsGroup = ( props ) => {
 			);
 
 		case 'step':
+			if ( undefined !== form.acceptSuggestion ) {
```
allendav: are you missing a `typeof` here?

client/components/suggestion/index.js
```diff
@@ -0,0 +1,60 @@
...
+	} ).isRequired,
+	acceptSuggestion: PropTypes.bool.isRequired,
```
allendav: Instead of acceptSuggestion, what about `selected` and treat this more like a radio button array (see also comment above about > 2 addresses)

client/state/form/actions.js
```diff
@@ -15,67 +16,47 @@ export const setFormProperty = ( field, value ) => {
...
-		const setError = ( value ) => dispatch( setFormProperty( 'errors', value ) );
-
-		dispatch( setFormProperty( 'pristine', true ) );
-		saveForm( setIsSaving, setSuccess, setError, callbackURL, nonce, submitMethod, getState().form.values );
-	};
-
-	const tryAutoAdvance = ( isManualAction ) => {
-		const currentStep = getState().form.currentStep;
-		const currentStepLayout = formLayout[ currentStep ] || {};
+export const backFromSuggestion = () => ( {
```
allendav: What does `backFromSuggestion` mean? I think we need a more descriptive name here.

client/state/form/actions.js
```diff
@@ -15,67 +16,47 @@ export const setFormProperty = ( field, value ) => {
...
-			return;
-		}
+export const goToStep = ( stepIndex ) => ( dispatch, getState, { formLayout } ) => {
+	const stepLayout = formLayout[ stepIndex ] || {};
 
-		if ( getStepFormErrors( getState(), formSchema, formLayout ).length ) {
-			return;
-		}
+	// Clear the "bypass suggestion" flag if the user is coming *back* to this step
+	if ( getState().form.currentStep > stepIndex && stepLayout.bypass_suggestion_flag ) {
```
allendav: Instead of `bypass_suggestion_flag` would `skip_showing_suggestions` be easier to comprehend?

client/state/form/actions.js
```diff
@@ -15,67 +16,47 @@ export const setFormProperty = ( field, value ) => {
...
+	} );
+};
 
-			tryAutoAdvance( false );
-		} );
-	};
+export const resetFlow = () => ( dispatch ) => {
+	dispatch( setFormProperty( 'pristine', true ) );
+	dispatch( goToStep( -1 ) );
+	dispatch( setFormProperty( 'acceptSuggestion', undefined ) );
```
allendav: Instead of acceptSuggestion, if we support N addresses this should probably be something like 'selectDestinationAddress' and 0

client/components/suggestion/index.js
DanReyLop: When clicking the button, the form flow goes _back_ _from_ the _suggestion_ screen. That means the address step will be displayed again so the merchant can manually edit it. I'm not very good at naming, do you have a better alternative? :)

client/components/suggestion/index.js
DanReyLop: Nothing in this component is specific to addresses. The server could potantially return a suggestion for any step. For example, if we detect that the merchant chose Media Mail for something that's _suspiciously_ big to be a book (but not over the official limit), we could warn him.

client/components/suggestion/index.js
DanReyLop: See my answer above. This is not just a component that cares about displaying a couple of addresses. It's a component that implement the entire logic of suggestions (present the 2 options, modify the state of the flow if the user chooses any of them or goes back), so it makes sense that the component can actually call Redux actions.

client/components/suggestion/index.js
DanReyLop: :dizzy_face: :boom: <- Mind-blown

As we discussed, we'll talk about this in our chat :)

client/components/summary/index.js
DanReyLop: It renders the contents of a step (for example, the address) in a text-only format (or at least, in a non-editable format). I can't think on any better name. Maybe `StepSummary`?

client/components/wcc-settings-form/settings-group.js
DanReyLop: Nope, this is a valid way of checking if something is `undefined`. Of course, it's not the only one, because JavaScript.

client/state/form/reducer.js
DanReyLop: @allendav ^^

client/state/form/actions.js
DanReyLop: `bypass_suggestion_flag` contains the *name* of the field that's used to indicate that the suggestion will be ignored. For example, for the Origin Address step:
`bypass_suggestion_flag: 'orig_bypass_suggestion'`

So if the field `orig_bypass_suggestion` is True (or "true-ish"), then the suggestion will be ignored. That's why the `bypass_suggestion_flag` has `flag` in its name, because it's *not* storing a boolean value, but giving the name of the actual *flag* that contains the boolean value.

I don't know how to name it to make it less confusing. `skip_showing_suggestions` makes it look like a boolean value  :(

DanReyLop:
For the record, we decided that we **don't** want to support more than 2 addresses to choose from.

client/components/wcc-settings-form/settings-group.js
allendav: True, we just generally have used the typeof way -- TIL that this (comparing directly to undefined) is safe since ECMAScript 5

client/state/form/reducer.js
allendav: Sorry, meant to comment on this - I had to do a bit of digging - this approach seems sound and is pretty much what abramov recommends here - https://twitter.com/dan_abramov/status/656074974533459968

In the comments on that tweet, he also advises against several dispatch calls as non-performant.

DanReyLop: He uses an actions that encapsulates several actions. Which is more or less what we are doing with `redux-thunk`. But what I did here was to call several reducers **inside** a reducer, and that seems a little more weird.

About performance... I wouldn't worry about that, not at this level at least.

allendav:
Noteworthy - ran into this during test - not in scope for Sprint 3 of course :)
https://github.com/Automattic/woocommerce-connect-client/issues/465

allendav:
It seems that once you've entered an origin address and hit next, that even if the label flow is not completed, that origin address should be retained on the merchant's host. Do you agree? Don't need to fix it here, but if you agree, please write a new issue.

cc @kellychoffman

allendav:
I have an order containing a single 3 in x 4 in x 0.2 in x 24 oz item that costs $2. I got a quote for Priority Mail 2-Day of $12.40 for it during checkout, which I selected.

I fill in the Origin step with everything including Awk St and hit Next.
I am advanced all the way to an empty Buy & Print step and see the following malformed response in Chrome's developer tools:

```
<br />
<b>Notice</b>:  Undefined property: stdClass::$orig_bypass_suggestion in <b>/Applications/MAMP/htdocs/wp-content/plugins/woocommerce-connect-client/classes/class-wc-connect-api-client.php</b> on line <b>99</b><br />
<br />
<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /Applications/MAMP/htdocs/wp-content/plugins/woocommerce-connect-client/classes/class-wc-connect-api-client.php:99) in <b>/Applications/MAMP/htdocs/wp-includes/rest-api/class-wp-rest-server.php</b> on line <b>1196</b><br />
<br />
<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /Applications/MAMP/htdocs/wp-content/plugins/woocommerce-connect-client/classes/class-wc-connect-api-client.php:99) in <b>/Applications/MAMP/htdocs/wp-includes/rest-api.php</b> on line <b>395</b><br />
<br />
<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /Applications/MAMP/htdocs/wp-content/plugins/woocommerce-connect-client/classes/class-wc-connect-api-client.php:99) in <b>/Applications/MAMP/htdocs/wp-includes/rest-api.php</b> on line <b>396</b><br />
<br />
<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /Applications/MAMP/htdocs/wp-content/plugins/woocommerce-connect-client/classes/class-wc-connect-api-client.php:99) in <b>/Applications/MAMP/htdocs/wp-includes/rest-api.php</b> on line <b>397</b><br />
{"code":"validation_failed","message":"One or more fields of your request are invalid.","data":{"status":400,"error":"validation_failure","data":{"fields":{"orig_address_1":{"value":"Hawk St","level":"suggestion"},"dest_address_1":{"value":"The server doesn't like that street!"},"cart":{"value":""},"rate":{"value":""}}}}}
```

client/components/summary/index.js
```diff
@@ -0,0 +1,132 @@
...
+		}
+		if ( line.length ) {
+			children.push( { value: line } );
+		}
+		return (
+			<p key={ idx }>
+				{ children.map( ( field, index ) => (
+					<span
+						key={ index }
+						style={ field.override ? { backgroundColor: 'cyan' } : {} } >
```
kellychoffman: Can you add a class to this instead and define in a stylesheet?

kellychoffman:
I have a styling question. Where are the divs that wrap these elements created?

```
'suggestion_original_title' => __( 'Address entered', 'woocommerce' ),
'suggestion_corrected_title' => __( 'Suggested address', 'woocommerce' ),
```

Can they have a class applied to them?

Is there a stylesheet already that I'm not seeing or should I create one? Where do you think would be the best place for that?

kellychoffman:
Feel free to disregard my style comments, we can move it to: https://github.com/Automattic/woocommerce-connect-client/issues/461

DanReyLop:
> It seems that once you've entered an origin address and hit next, that even if the label flow is not completed, that origin address should be retained on the merchant's host. Do you agree? Don't need to fix it here, but if you agree, please write a new issue.

@allendav Already implemented :) See #456

DanReyLop:
> With WP_DEBUG false, this tests out, so please fix the Undefined property: stdClass::$orig_bypass_suggestion

This is one of the reasons I hate PHP. Accessing an undefined property happily returns `NULL` (which is what I expect, a "falsy" value), but in certain error reporting configurations it also spits a `NOTICE`. I fixed it.

Issue #461, opened 2016-07-20
[Labels] Style all the things


allendav:
The tab bar and the content for each tab need some styling :)

@DanReyLop please get this roughly styled by the end of Sprint 3 and then @kellychoffman can give it a UX review and deep styling while we're on support week

kellychoffman:
Related comment: https://github.com/Automattic/woocommerce-connect-client/pull/458#issuecomment-235147298

TL;DR: If you could just add some classes to the surrounding div, ul, and li's I can take it from there. Or show me how to do it :)

DanReyLop:
I'm trying to style it myself, I'll push what we have to this issue in a day or 2, and then all the classes will be in place, you will only need to clean-up my mess :D

kellychoffman:
Thanks @DanReyLop :) Let me know I can help.

kellychoffman:
Updating this issue definition to be: 

At the end of Sprint 5, shipping labels will have a production-ready level of design.

allendav:
@kellychoffman - let's open new issues for style probs and close this, ok?

Issue #432, opened 2016-06-29
[Labels] Only show the Create Label button when the order qualifies


DanReyLop:
We should only show the `Create Label` button that opens the label printing flow when the shipping order is adequate. For the moment, that means that:
* The order can be shipped (it's not a downloadable product)
* The origin and destination addresses are in the US

If there was already a label printed for that order, show some information about the label (when it was printed, how many labels, price?) instead of the button to open the dialog.

PR #467, opened 2016-07-26
[Labels] Packaging step

### LINKED ISSUE #445: [Labels] Packaging step
Mock-up: TBD

In this step, the UI must display the user's "Shopping Cart". For each box the merchant will have to send, we must display at least:
* Dimensions of the box (so he knows what box to pick)
* List of products that should go into that box
* Total weight of the box and its contents
* The weight should be editable

There are 2 use cases:
* The user selected a shipping method powered by WCC, so the order metadata has all the packaging information. In that case, that information should be used.
* The user selected free shipping. In that case, we don't have the packaging information in the order metadata. In that case, **for the first iteration**, all purchased items will be "packaged" individually, in a box with the same size and weight as the item itself.

DanReyLop:
Fixes #445 

Things to consider:
* The UI is very crude. I just made a few loops with plain `<ul>` and `<li>` tags.
* Since everything you see in the Packaging step is already valid, that step is skipped by default. So after the address steps you will appear in a (not implemented) Rates step. You can go back clicking the "Packages" link at the top. If we decide that the last step will have a summary of everything, I think this behaviour is fine. Anyway, we can change our minds later.
* If there is no information in the order `meta` about packaging (for example, if the user selected Free Shipping, or the order was placed using the code in the current `master` branch), then all the items will be packed individually (not a perfect fallback, but a fallback).
* All the relevant information per-package is visible (dimensions, weight), but only the weight is editable.
* In each package, it's displayed a list of the products that will go in that package.
* It works flawlessly with product variations! (Of course, y'all should try to challenge this fact).

@jeffstieler @allendav

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -91,7 +91,6 @@ module.exports = {
 		'react/jsx-no-undef': 2,
 		'react/jsx-no-duplicate-props': 1,
 		'react/react-in-jsx-scope': 2,
-		'react/no-danger': 2,
 		'react/no-did-mount-set-state': 1,
 		'react/no-did-update-set-state': 1,
 		'react/jsx-no-bind': [ 1, { 'allowArrowFunctions': true } ],
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -38,7 +38,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 						'items' => array(
 							array(
 								'height' => $height,
-								'product_id' => $item['product_id'],
+								'product_id' => $item[ 'product_id' ],
 								'length' => $length,
 								'quantity' => 1,
 								'weight' => $weight,
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -10,12 +10,13 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 			$this->settings_store = $settings_store;
 		}
 
-		protected function get_form_data( WC_Order $order ) {
-			$form_data = array();
-			$contents = array();
-
+		protected function get_individual_packages( $order ) {
+			$packages = array();
 			foreach( $order->get_items() as $item ) {
 				$product = $order->get_product_from_item( $item );
+				if ( ! $product ) {
+					continue;
+				}
 				$height = 0;
 				$length = 0;
 				$weight = $product->get_weight();
@@ -26,17 +27,57 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 					$length = $product->get_length();
 					$width  = $product->get_width();
 				}
+				$name = $product->get_formatted_name();
 
-				$contents[] = array(
-					'height' => $height,
-					'product_id' => $item['product_id'],
-					'length' => $length,
-					'quantity' => $item['qty'],
-					'weight' => $weight,
-					'width' => $width
-				);
+				for ( $i = 0; $i < $item[ 'qty' ]; $i++ ) {
+					$packages[] = array(
+						'height' => $height,
+						'length' => $length,
+						'weight' => $weight,
+						'width' => $width,
+						'items' => array(
+							array(
+								'height' => $height,
+								'product_id' => $item['product_id'], // TODO: Make this work with product variations
+								'length' => $length,
+								'quantity' => 1,
+								'weight' => $weight,
+								'width' => $width,
+								'name' => $name,
+							),
+						),
+					);
+				}
 			}
-			$form_data[ 'cart' ] = $contents;
+			return $packages;
+		}
+
+		protected function get_packaging_data( WC_Order $order ) {
+			$shipping_methods = $order->get_shipping_methods();
+			if ( empty( $shipping_methods ) || ! isset( $shipping_methods[ 0 ][ 'wc_connect_packages' ] ) ) {
+				return $this->get_individual_packages( $order );
+			}
+
+			$packages = json_decode( $shipping_methods[ 0 ][ 'wc_connect_packages' ], true );
+
+			foreach( $packages as $package_index => $package ) {
+				foreach( $package[ 'items' ] as $item_index => $item ) {
+					// TODO: Make this work with product variations (make the WCC server handle and return 'variation_id' as well as 'product_id')
+					$product = $order->get_product_from_item( $item );
+					if ( ! $product ) {
+						continue;
+					}
+					$packages[ $package_index ][ 'items' ][ $item_index ][ 'name' ] = $product->get_formatted_name();
+				}
+			}
+
+			return $packages;
+		}
+
+		protected function get_form_data( WC_Order $order ) {
+			$form_data = array();
+
+			$form_data[ 'cart' ] = $this->get_packaging_data( $order );
 
 			foreach( $this->settings_store->get_origin_address() as $key => $value ) {
 				$form_data[ 'orig_' . $key ] = $value;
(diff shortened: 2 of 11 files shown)


.eslintrc.js
```diff
@@ -91,7 +91,6 @@ module.exports = {
 		'react/jsx-no-undef': 2,
 		'react/jsx-no-duplicate-props': 1,
 		'react/react-in-jsx-scope': 2,
-		'react/no-danger': 2,
```
DanReyLop: `dangerouslySetInnerHTML` is so cumbersome to use (on purpose) that it's highly unlikely to be used accidentally. I think this rule only adds noise.

classes/class-wc-connect-shipping-label.php
```diff
@@ -10,12 +10,13 @@ public function __construct( WC_Connect_Service_Settings_Store $settings_store )
 			$this->settings_store = $settings_store;
```
DanReyLop: Git messed up this file diff a bit. It's better if you look at the final code of `get_individual_packages()` and `get_packaging_data()`

.eslintrc.js
allendav: I'm ok with this :)

client/components/shopping-cart/index.js
```diff
@@ -0,0 +1,54 @@
+import React, { PropTypes } from 'react';
+import NumberInput from 'components/number-field/number-input';
+import Gridicon from 'components/gridicon';
+import { sanitize } from 'dompurify';
+
+const ShoppingCart = ( { packages, updateValue, dimensionUnit, weightUnit, errors } ) => {
```
allendav: Instead of `ShoppingCart`, which sounds like a front side component, could we use something like `OrderContents` for the name for this component instead?

classes/class-wc-connect-shipping-label.php
```diff
@@ -10,12 +10,13 @@ public function __construct( WC_Connect_Service_Settings_Store $settings_store )
...
 		}
 
-		protected function get_form_data( WC_Order $order ) {
-			$form_data = array();
-			$contents = array();
-
+		protected function get_individual_packages( $order ) {
```
allendav: Please rename this function as `get_items_as_individual_packages` to make it clearer what it does

classes/class-wc-connect-shipping-method.php
```diff
@@ -290,7 +290,8 @@ public function calculate_shipping( $package = array() ) {
 							)
 						),
 						'cost'     => $rate->rate,
-						'calc_tax' => 'per_item'
+						'calc_tax' => 'per_item',
+						'meta_data' => array( 'wc_connect_packages' => json_encode( $rate->packages ) ),
```
allendav: What happens if we don't `json_encode $rate->packages`? I take it something bad happens if the key points directly at `$rate->packages`?

allendav:
Its OK that the UI is crude - we'll style the hell out of this as part of #461

kellychoffman:
> Its OK that the UI is crude - we'll style the hell out of this as part of #461

Heck ya we will. 🎨

kellychoffman:
> So after the address steps you will appear in a (not implemented) Rates step. You can go back clicking the "Packages" link at the top.

What if, for some odd reason, a store owner didn't fill out the shipping section on the product page? We let them publish products without any shipping info.

I didn't enter any dimensions or weight on one of my test products, and when I got to the packages point, its just blank. Should we add some error messaging here instead?

client/components/shopping-cart/index.js
DanReyLop: `OrderPackages` sounds clearer I think. Will do :)

classes/class-wc-connect-shipping-method.php
DanReyLop: According to the method signature of `WC_Shipping_Rate->add_meta_data()` , both the key and the value must be `string`. I don't know where it's being all serialized, so I just wanted to play it safe here :)

DanReyLop:
> What if, for some odd reason, a store owner didn't fill out the shipping section on the product page? We let them publish products without any shipping info.

Related: https://github.com/Automattic/woocommerce-connect-client/issues/223

I'll investigate how the other parts of Connect handle this scenario and I'll get back to you.

DanReyLop:
@kellychoffman: Getting shipping rates (with Connect) doesn't even work if one or more items don't have weight (not even fallback rate will be returned). How did you do to create an order with items with no weight?

I managed to create an order manually. Once I got to the Packages step, I got this, which is not *so* bad ("Air" is my weightless product):
weightless

allendav:
@kellychoffman wrote

> I didn't enter any dimensions or weight on one of my test products, and when I got to the packages point, its just blank. Should we add some error messaging here instead?

If a product is marked shippable and is missing a dimension, I think we should display notices in both the front-side product view (when an admin is logged in) and in the labels UI.  Do you concur, @kellychoffman  ?  If so we should write up a core issue and a connect issue, respectively.

kellychoffman:
> How did you do to create an order with items with no weight?

I manually created an order with a product that had no weight or dimensions. 

> If a product is marked shippable and is missing a dimension, I think we should display notices in both the front-side product view (when an admin is logged in) and in the labels UI.

Yes, that sounds like a good idea. 

Related, the 'Create label' button still shows when I manually create an order with a only a virtual product type. I'm thinking we should only show the 'create label' button when its a shippable product. If ya'll agree, I'll make an issue for that as well.

DanReyLop:
> Related, the 'Create label' button still shows when I manually create an order with a only a virtual product type. I'm thinking we should only show the 'create label' button when its a shippable product. If ya'll agree, I'll make an issue for that as well.

There's an ~~app~~ issue for that: https://github.com/Automattic/woocommerce-connect-client/issues/432 :)

Ok, I managed to reproduce what you got. You manually created an empty order, added a product, but you **didn't** save the changes in the order, so when you entered the Print Label flow, there were no items in the order. Does that make sense to you? The "Create Label" button should not have been there in that situation, #432 will fix that.

kellychoffman:
> You manually created an empty order, added a product, but you didn't save the changes in the order

I still see the button even after I save changes.

DanReyLop:
> I still see the button even after I save changes.

Yes, the button is always showing because #432 is not implemented yet. But, if you open the Create Label dialog **after** you saved your changes, you *should* see the item you added in the Packages step, even if it doesn't have weight or dimensions. It's a bit weird because it will display `0in x 0in x 0in` dimensions, but at least it will be there and the weight will be editable.

kellychoffman:
> Yes, the button is always showing because #432 is not implemented ye

Ah, got it. Thanks for the clarification.

client/components/order-packages/index.js
```diff
@@ -0,0 +1,54 @@
...
+const OrderPackages = ( { packages, updateValue, dimensionUnit, weightUnit, errors } ) => {
+	const renderItemInfo = ( item, itemIndex ) => (
+		<li key={ itemIndex }>
+			<span dangerouslySetInnerHTML={ { __html: sanitize( item.name ) } } />
+			{ 1 < item.quantity ? ' (x' + item.quantity + ')' : null }
+		</li>
+	);
+
+	const renderPackageDimensions = ( pckg ) => {
+		return [ pckg.length, pckg.width, pckg.height ].map( dim => dim + dimensionUnit ).join( ' x ' );
```
allendav: We should have a space between each dimension and its unit.  E.g. this shows the problem:

screen shot 2016-07-27 at 12 40 51 pm

classes/class-wc-connect-shipping-label.php
```diff
@@ -26,17 +27,56 @@ protected function get_form_data( WC_Order $order ) {
...
+				for ( $i = 0; $i < $item[ 'qty' ]; $i++ ) {
+					$packages[] = array(
+						'height' => $height,
+						'length' => $length,
+						'weight' => $weight,
+						'width' => $width,
+						'items' => array(
+							array(
+								'height' => $height,
+								'product_id' => $item['product_id'],
```
allendav: Spaces within brackets please

allendav:
Please fix these and then :shipit: 

https://github.com/Automattic/woocommerce-connect-client/pull/467/files#r72508968
https://github.com/Automattic/woocommerce-connect-client/pull/467/files#r72508805

Issue #393, opened 2016-06-08
Let's DRY renderFieldDescription, renderError


allendav:
These two functions appear in a few of our components, especially after #376 lands

PR #469, opened 2016-07-27
Added style for the Create Label dialog

DanReyLop:
Fixes (partially) #461 

I've added styles for the dialog, the tabs and the steps that are fully implemented (Origin Address and Destination Address), including the Address Suggestion case.

@kellychoffman I hope that's close enough for you to start tweaking :) If you can't find something or I implemented anything in a terribly wrong way, please tell me.

@allendav @jeffstieler


kellychoffman:
This is looking great!

Marking as in progress while I clean some things up.

kellychoffman:
Styled the address sections up, including the top tabs. 

@DanReyLop: One (hopefully) small request. 

All the section titles start with "Create shipping label:". I want to deemphasize that first part. 

Is it possible to break up "Create shipping label:" and "Destination address" into two parts? I want to style them differently. Can we keep the `h3` on the second half, but give the first part another class?

DanReyLop:
> All the section titles start with "Create shipping label:". I want to deemphasize that first part.

@allendav @jeffstieler Are you guys ok allowing limited HTML in the data-driven titles (and maybe other data-driven strings, such as descriptions)? A simple `<strong>` would solve that *small* request, and it can give us more flexibility, but if we go down that path it can get messy.

@kellychoffman if the answer is "no way in hell", then you're out of luck, but I've heard they are reasonable people :)

kellychoffman:
Haha, no worries if not. I have other ideas if the answer is no :)

allendav:
@DanReyLop - i am not a fan of the old non semantic tags from the early days of HTML. Although we could do a wp_kses style of filtering, to allow a whitelist of tags through, I'd prefer we decorated the form elements with classnames, kinda like we can presently with indicators:

https://github.com/Automattic/woocommerce-connect-client/blob/1cd846293d0f82341dd5d30b37b8f3871d9e9301/classes/class-wc-connect-help-provider.php#L51

What would you think about that approach instead?

DanReyLop:
@allendav That's not gonna work in this case. The thing is, we need to be able to apply 2 different styles to *the same* string, so "Create Label: Origin Address" needs to become "Create Label: **Origin Address**", so we could set the title value to `"Create Label: <strong>Origin Address</strong>"` and apply some styles to `.step-title strong`.

Looks like we are already allowing HTML in some data-driven elements, such as `text-area` (and more `descriptions`): https://github.com/Automattic/woocommerce-connect-client/blob/master/client/components/text-area/index.js#L11
Maybe we should extend it to all data-driven strings? I would be fine restricting the HTML, but I can see how adding links, bold/italics and such could be useful for things like titles or descriptions.

allendav:
> Looks like we are already allowing HTML in some data-driven elements, such as text-area (and more descriptions): https://github.com/Automattic/woocommerce-connect-client/blob/master/client/components/text-area/index.js#L11
Maybe we should extend it to all data-driven strings? I would be fine restricting the HTML, but I can see how adding links, bold/italics and such could be useful for things like titles or descriptions.

OK, let's do it and copy the sanitize approach

DanReyLop:
Well, this turned out to be quite bigger than I expected.

Since we use stateless components and sometimes we return `null` in their `render` function, we **need** React v15 (see [React v15 changelog](https://facebook.github.io/react/blog/2016/04/07/react-v15.html#functional-components-can-now-return-null-too)) I sincerely don't know why this didn't bite us before. I've upgraded the ReactJS version to `15.1.0` (same as the one `wp-calypso` is using) in https://github.com/Automattic/woocommerce-connect-client/pull/469/commits/9a5cd7a2d85c6c8c802ad93c8ab68a5f0667db4b

In https://github.com/Automattic/woocommerce-connect-client/pull/469/commits/f7df1e4c6785e05f4d64446e1885800d9d94ae16 I added HTML support to all the data-driven strings I could find. That includes titles, descriptions, errors, etc. You can see the allowed HTML tags and attributes here: https://github.com/Automattic/woocommerce-connect-client/pull/469/commits/f7df1e4c6785e05f4d64446e1885800d9d94ae16#diff-7fe2d3cc6e0977e2515f9f048166556bR1 If you find that a potentially useful tag is missing (`<blink>` or `<marquee>`, for example), just add it to that file.

While I was at it, I refactored the rendering of descriptions & errors, so once this PR lands we can close #393 too.

@kellychoffman does this work for you? I didn't add the possibility to add custom CSS classes to parts of the data-driven fields because that could get really messy.

kellychoffman:
>  I didn't add the possibility to add custom CSS classes to parts of the data-driven fields because that could get really messy.

Didn't realize that would be a caveat. I was hoping to do something more custom here:

screen shot 2016-08-02 at 2 39 53 pm

 Is there a reason the prepending `Create shipping label:` has to be data driven? 

Regardless, this doesn't have to hold up this PR. I'd rather get what we currently have in. We can punt this to Sprint 4.

allendav:
@DanReyLop wrote:

> PS: There are merge conflicts with master. You can just ignore it, once we are done with this PR I'll rebase.

Please go ahead and rebase now - part of the benefit of the review is to have multiple people watching out for collateral merge damage.

assets/stylesheets/style.scss
```diff
@@ -224,8 +255,11 @@
 
     // WCC Components
     @import '../../client/components/shipping/packages/style';
-    @import '../../client/components/indicators/style';
-    @import '../../client/components/text/style';
+	@import '../../client/components/indicators/style';
```
allendav: @DanReyLop I see you switched to tabs for these lines

@kellychoffman - unlike our JS with tabs, we seem to use spaces in our SCSS, including this file - any idea why?

kellychoffman: @allendav no idea why :/ 

changed this file to use tabs.

DanReyLop:
> Is there a reason the prepending `Create shipping label:` has to be data driven?

Data-driveness is all or nothing, so everything you see from the tabs down to the action buttons needs to be data-driven (unless we axe data-driveness altogether). One of the downsides of data-driveness is the lack of fine-grained design tweaking :(

If you are ok parking this until the next sprint, then that settles it for now :)

kellychoffman:
Got it. Thanks for clarifying. 

> If you are ok parking this until the next sprint, then that settles it for now :)

👍  Yeah, lets do that.

DanReyLop:
@kellychoffman I've just rebased with `master` and force-pushed, so if you plan on keep working on this branch, please pull the latest version of it.

kellychoffman:
I'll make a new PR for further style and UI changes. I say we get this one in if @allendav is done reviewing.

allendav:
:shipit:

PR #470, opened 2016-07-27
Add autoprefixer to the Webpack build

### LINKED ISSUE #468: Use autoprefixer to support non-standard CSS vendor prefixes
We are currently not using `autoprefixer`, so CSS rules like `display: flex` will be ignored by older browsers that only understand vendor-specific syntax (`-webkit`, `-moz`, etc).

We should use `autoprefixer` and while we are at it, remove vendor-specific CSS rules from our code, since they will be auto-generated.

DanReyLop:
Fixes #468 

We had already installed `autoprefixer` as a dependency, but for some reason it wasn't integrated into the Webpack build :)

The easiest way to test is:
* Don't forget to do `npm install` and re-start the build script.
* Go to a Canada Post settings configuration page, and remove the amount in the "Fallback Rate" field.
* If the placeholder (`Rate amount`) appears in light gray, then autoprefixer is working.
* If the placeholder appears in darker gray, then it's not working.

This is because the `::placeholder` pseudo-class is not standarized even in modern browsers, so if the vendor-prefixed versions aren't generated, the styling for that pseudo-class (make it lighter gray) is ignored.

Good:
placeholder_good

Bad:
placeholder_bad

@allendav @jeffstieler

### DIFF
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -64,15 +64,9 @@
         @extend %form-field;
     }
 
-    @mixin placeholder {
-        &::-moz-placeholder { @content }
-        &::-webkit-input-placeholder { @content }
-        &:-ms-input-placeholder { @content }
-    }
-
     %form-field {
         box-shadow: none;
-        @include placeholder {
+        &::placeholder {
             color: lighten( $gray, 10% );
         }
     }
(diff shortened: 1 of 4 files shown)


allendav:
@DanReyLop - under what circumstances were the vendor prefixed versions not generated? I tested master on chrome and it seemed fine there. Is this only an IE8 and 9 problem?

DanReyLop:
The `::placeholder` thing in particular worked fine because we were using a SASS `mixin` to manually do what `autoprefixer` gives us for free. If you look at my commit, I removed that `mixin` because it's no longer needed.

I don't know of any particular rule that can fail in modern versions of Firefox/Chrome/Safari that you can test. But we were including `autoprefixer` in our `package.json` dependencies, and it wasn't used anywhere, so I assume that the plan was to integrate it in the build but it somehow got lost in the way. It really is a must-use tool to code CSS without worrying too much about browser compatibility (I assume @kellychoffman can vouch for this).

allendav:
:shipit:

PR #473, opened 2016-08-01
[Labels] Rates selection step

### LINKED ISSUE #452: [Labels] Rates step
Design: TBD

Based on the origin address, destination address and packages, this step must show a list of possible shipping options to use. Those options will be provided by the WCC server. Each option should display a readable title ( e.g: `1-Day Priority Mail` ) and the price.

If the buyer selected a shipping service during checkout, that service should be pre-selected, and thus the entire step can be automatically skipped.

There can be a different rate for each package in the order, and so the merchant will need to see a list of rates for each package.

DanReyLop:
Fixes #452 

As usual, crude UI and everything mocked in the server, but only 1 more step to go!

For the UI of that step, I did a simple loop over all the packages in the order, and for each one, show the package dimensions (as a way for the merchant to remember which package is the rate referring to) and a `<select>` with all available rates for each package (with the price alongside).

Since this step required a more "dynamic" communication between client and server, I've added 2 small things:
* Apart from returning `errors`, the server can now return an array of `field_options`. This assoc array will have the structure `"field.path" => "arbitrary data"`. That can be used to feed the client with dynamic options for a field (because the `formLayout` is immutable).
* There is a new kind of `error`. You have the plain error, the `suggestion` (used by returning an error with `"level" => "suggestion"`), and now, `overwrite` (used by returning an error with the structure `"level" => "overwrite", "value" => "newFieldValue"`). The form value for that field will be overwritten with the value provided by the server, and then the error will be discarded. This allows the server to not only return the possible rates, but to also `pre-select` a rate for the user (normally the cheapest one), so the client can even skip that step. This "feeding the client with form values" will come useful in the Packaging step for the Free Shipping case.

If you don't understand something, please point it out so I can add documentation.

@jeffstieler @allendav

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -73,10 +73,31 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 			return $packages;
 		}
 
+		protected function get_selected_rates( WC_Order $order ) {
+			$shipping_method = reset( $order->get_shipping_methods() );
+			if ( ! $shipping_method || ! isset( $shipping_method[ 'wc_connect_packages' ] ) ) {
+				return array();
+			}
+
+			$packages = json_decode( $shipping_method[ 'wc_connect_packages' ], true );
+			$rates = array();
+
+			foreach( $packages as $package ) {
+				if ( ! $package[ 'service_id' ] ) {
+					return array();
+				}
+				$rates[] = $package[ 'service_id' ];
+			}
+
+			return $rates;
+		}
+
 		protected function get_form_data( WC_Order $order ) {
 			$form_data = array();
 
 			$form_data[ 'cart' ] = $this->get_packaging_data( $order );
+
+			$form_data[ 'rates' ] = $this->get_selected_rates( $order );
 
 			foreach( $this->settings_store->get_origin_address() as $key => $value ) {
 				$form_data[ 'orig_' . $key ] = $value;
@@ -207,16 +228,12 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 				'title' => 'TODO: Not implemented yet',
 				'items' => array(
 					array(
-						'key' => 'rate',
-						'type' => 'radios',
-						'titleMap' => array(
-							'media' => 'Media Mail ($1.00)',
-							'pri_1' => 'Priority 1-day ($9.99)',
-							'todo' => 'TODO: These are NOT real rates',
-						),
+						'key' => 'rates',
+						'type' => 'rates',
+						'packages_field' => 'cart',
 					),
 				),
-				'summary' => '{rate}',
+				'summary' => '{rates}',
 			);
 
 			$layout[] = array(
@@ -357,12 +374,15 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 			);
 			$required_fields[] = 'cart';
 
-			$properties[ 'rate' ] = array(
-				'type' => 'string',
-				'title' => __( 'Shipping rate', 'woocommerce' ),
-				'enum' => array( 'pri_1', 'media', 'todo' ),
+			$properties[ 'rates' ] = array(
+				'type' => 'array',
+				'title' => __( 'Shipping rates', 'woocommerce' ),
+				'items' => array(
+					'type' => 'string',
+					'title' => __( 'Shipping rate', 'woocommerce' ),
+				),
 			);
-			$required_fields[] = 'rate';
+			$required_fields[] = 'rates';
 
 			return array(
 				'required' => $required_fields,
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -53,7 +53,8 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		}
 
 		protected function get_packaging_data( WC_Order $order ) {
-			$shipping_method = reset( $order->get_shipping_methods() );
+			$shipping_methods = $order->get_shipping_methods();
+			$shipping_method = reset( $shipping_methods );
 			if ( ! $shipping_method || ! isset( $shipping_method[ 'wc_connect_packages' ] ) ) {
 				return $this->get_items_as_individual_packages( $order );
 			}
@@ -74,7 +75,8 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		}
 
 		protected function get_selected_rates( WC_Order $order ) {
-			$shipping_method = reset( $order->get_shipping_methods() );
+			$shipping_methods = $order->get_shipping_methods();
+			$shipping_method = reset( $shipping_methods );
 			if ( ! $shipping_method || ! isset( $shipping_method[ 'wc_connect_packages' ] ) ) {
 				return array();
 			}
--- a/client/components/shipping-rates/index.js
+++ b/client/components/shipping-rates/index.js
@@ -3,9 +3,7 @@ import Dropdown from 'components/dropdown';
 
 const ShippingRates = ( { id, selectedRates, availableRates, packages, updateValue, dimensionUnit, weightUnit, currencySymbol, errors, layout } ) => {
 	const renderTitle = ( pckg ) => {
-		const dimensions = [ pckg.length, pckg.width, pckg.height ].map( dim => dim + ' ' + dimensionUnit ).join( ' x ' );
-		const weight = pckg.weight + ' ' + weightUnit;
-		return dimensions + ' ' + weight;
+		return `${pckg.length} ${dimensionUnit} x ${pckg.width} ${dimensionUnit} x ${pckg.height} ${dimensionUnit} (${pckg.weight} ${weightUnit})`;
 	};
 
 	const renderSinglePackage = ( pckg, index ) => {
--- a/client/components/shipping-rates/index.js
+++ b/client/components/shipping-rates/index.js
@@ -1,0 +1,54 @@
+import React, { PropTypes } from 'react';
+import Dropdown from 'components/dropdown';
+
+const ShippingRates = ( { selectedRates, availableRates, packages, updateValue, dimensionUnit, weightUnit, currencySymbol, errors, layout } ) => {
+	const renderTitle = ( pckg ) => {
+		const dimensions = [ pckg.length, pckg.width, pckg.height ].map( dim => dim + ' ' + dimensionUnit ).join( ' x ' );
+		const weight = pckg.weight + ' ' + weightUnit;
+		return <span>{ dimensions + ' ' + weight }</span>;
+	};
+
+	const renderSinglePackage = ( pckg, index ) => {
+		const selectedRate = selectedRates[ index ] || '';
+		const errorObject = ( errors[ index ] || {} )[ '' ];
+		const error = errorObject ? ( errorObject.value || layout.validation_hint ) : null;
+
+		const titleMap = {};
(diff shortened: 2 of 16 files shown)


jeffstieler:
@DanReyLop can you describe an example user flow where the "new error type" is utilized? I'm not really getting the idea..

classes/class-wc-connect-shipping-label.php
```diff
@@ -73,11 +73,32 @@ protected function get_packaging_data( WC_Order $order ) {
 			return $packages;
 		}
 
+		protected function get_selected_rates( WC_Order $order ) {
+			$shipping_method = reset( $order->get_shipping_methods() );
```
jeffstieler: I get a `PHP Strict Standards:  Only variables should be passed by reference in php shell code` from this line.

Can you accomplish this same logic without using `reset()` or by storing the shipping methods in a variable first?

DanReyLop:
@jeffstieler Basically this: https://github.com/Automattic/woocommerce-connect-client/pull/473/files#diff-57dabbbbd8e057f1082af23aa58d3f5aR146

The server calculates the rates and it returns them to the client using the `field_options` array. Then, it uses an error of type `overwrite` to make the client auto-select one rate (the cheapest one, unless we define more complex rules in the server).

Another case: An order with free shipping. When entering the Print Label flow, the form is submitted to check the origin&destination addresses. At that point, the server detects that the items in the order are not packaged, so it runs the box-packer algorithm and it returns the new `packages` array inside a `overwrite` error field. That way, when the merchant arrives to the Packaging step, it will show the correct packages.

Another: When normalizing an address, it can be that all the changes made to that address are "trivial changes". For example, changing `Main Street` to `MAIN ST`, adding the +4 digits to the ZIP code, etc. If that's the case, instead of showing the merchant the "Choose between the original address and the suggested address" UI, the server can simply send the changes in `overwrite` errors and directly jump to the next steps.

classes/class-wc-connect-shipping-label.php
DanReyLop: Sure, I'll store it in a temp variable. Which PHP version are you on? I'm in `5.6.21` and I had no warning even with `define('WP_DEBUG', true);`.

jeffstieler:
@DanReyLop OK - that makes sense to me, but not as some type of "error". None of the scenarios you described are actual errors.

Can we rename this to make more sense for how it's used?

client/lib/utils/coerce-values.js
```diff
@@ -1,3 +1,5 @@
+let coerceFormValues;
```
jeffstieler: This doesn't appear to be necessary.

client/lib/utils/coerce-values.js
```diff
@@ -66,7 +81,7 @@ const coerceValue = ( value, type ) => {
  * @param {Object} [definitions=null] - Optional. Schema definitions, parsed from schema if omitted.
  * @returns {Object} - Coerced values based on schema.
  */
-const coerceFormValues = ( schema, values, definitions = null ) => {
+coerceFormValues = ( schema, values, definitions = null ) => {
```
jeffstieler: Can't this go back to a `const` ?

client/lib/utils/coerce-values.js
DanReyLop: Sadly, it is, because of [this ESLint rule](http://eslint.org/docs/rules/no-use-before-define).

`coerceFormValues` is called from `coerceValue`, and `coerceValue` is called from `coerceFormValues`. Either we disable the ESLint rule, or `coerceFormValues` needs to be declared on the top of the file.

jeffstieler: Well it's also because of the refactor in this PR - why does it make more sense to have `coerceValue` call `coerceFormValues`?

DanReyLop: Before, when validating a field of type `array`, `coerceFormValues` was called recursively for each of its elements. The problem is, `coerceFormValues` only operates on fields of type `object` (like the entire `schema`, or the `box` definition). Since the Rates field definition is an `array` of plain `string`s, this code was blowing up.

I refactored it using my best judgement without completely flipping everything upside-down (I didn't expect the ESLint failure here). So, `coerceFormValues` works on `object`s, and `coerceValue` works on everything else (including `array`s). Do you have a better alternative?

jeffstieler: I think either refactoring `coerceFormValues` to handle non-objects or refactoring the array items map callback to choose between `coerceFormValues` and `coerceValue` based on the `type` in `fieldSchema.items` would work.

DanReyLop:
@jeffstieler I took another stab at refactoring `coerce-values.js`. I've also renamed `form.errors` to `form.fieldsStatus` in every place I could find, since now that structure can contain not only errors, but also other kinds of `stuff`™

classes/class-wc-connect-shipping-label.php
jeffstieler: The result of `$order->get_shipping_methods()` is what needs to be stored in a variable. `reset()` resets the internal array pointer, and expects to receive the variable by reference.

client/components/shipping-rates/index.js
```diff
@@ -0,0 +1,54 @@
+import React, { PropTypes } from 'react';
+import Dropdown from 'components/dropdown';
+
+const ShippingRates = ( { selectedRates, availableRates, packages, updateValue, dimensionUnit, weightUnit, currencySymbol, errors, layout } ) => {
+	const renderTitle = ( pckg ) => {
+		const dimensions = [ pckg.length, pckg.width, pckg.height ].map( dim => dim + ' ' + dimensionUnit ).join( ' x ' );
+		const weight = pckg.weight + ' ' + weightUnit;
+		return <span>{ dimensions + ' ' + weight }</span>;
```
jeffstieler: What do you think about using `sprintf-js` to format the title? It's already a dependency, and would remove all the array mapping/joining and concatenation.

client/components/shipping-rates/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+	const renderSinglePackage = ( pckg, index ) => {
+		const selectedRate = selectedRates[ index ] || '';
+		const errorObject = ( errors[ index ] || {} )[ '' ];
+		const error = errorObject ? ( errorObject.value || layout.validation_hint ) : null;
+
+		const titleMap = {};
+		Object.keys( availableRates[ index ] ).forEach( ( serviceId ) => {
+			const rateObject = availableRates[ index ][ serviceId ];
+			titleMap[ serviceId ] = rateObject.name + ' (' + currencySymbol + rateObject.rate.toFixed( 2 ) + ')';
```
jeffstieler: Same note about `sprintf-js`.

jeffstieler:
I'm encountering PHP errors/warnings when trying to test this.

`PHP Notice:  Undefined offset: 0 in /srv/www/wordpress-default/wp-content/plugins/woocommerce-connect-client/classes/class-wc-connect-api-client.php on line 146`

`PHP Notice:  Only variables should be passed by reference in /srv/www/wordpress-default/wp-content/plugins/woocommerce-connect-client/classes/class-wc-connect-shipping-label.php on line 56`

`PHP Notice:  Only variables should be passed by reference in /srv/www/wordpress-default/wp-content/plugins/woocommerce-connect-client/classes/class-wc-connect-shipping-label.php on line 77`

client/components/shipping-rates/index.js
DanReyLop: I've always found the `sprintf` syntax quite hard to read. What do you think about ES6 template strings? The `dimensions` string could become like this:
```
const dimensions = `${pckg.length} ${dimensionUnit} x ${pckg.width} ${dimensionUnit} x ${pckg.height} ${dimensionUnit}`;
```

DanReyLop:
Weird. I'm supposedly running PHP in "cry-baby" mode too and I haven't seen the warnings. Anyway, they should be fixed now.

client/components/shipping-rates/index.js
jeffstieler: The template string is much better.

Just for kicks, here's the sprintf version: ` sprintf( '%1$s %4$s x %2$s %4$s x %3$s %4$s %5$s %6$s ', pckg.length, pckg.width, pckg.height, dimensionUnit, pckg.weight, weightUnit )`

I think we'll need a tie breaker - you up for it @allendav? :P

allendav: sprintf :)

allendav: Actually, now that I think of it, we might need to do sprintf ultimately, since this isn't localization friendly. OK to open a separate issue for that against the next sprint.

allendav:
> also pre-select a rate for the user (normally the cheapest one), so the client can even skip that step. This "feeding the client with form values" will come useful in the Packaging step for the Free Shipping case

I set up an instance of USPS with origin 98290.
I added a 2" x 2" x 2" thing weighing 8 ounces to my cart.
I gave it a 98272 address in checkout and selected "First-Class Mail - Parcel: $3.40" and completed checkout.
In Edit Order, on the rate step, that rate did not appear (looks like those rates are hardcoded), so please write a separate issue for rate pre-select, which should have been in this sprint, but I think is going to have to happen in Sprint 4.

allendav:
@DanReyLop - please add an issue to sanitize or at least whitelist the content of label_settings before we use it (e.g. [here](https://github.com/Automattic/woocommerce-connect-client/blob/5ce8b15286fab4722f0ab8a5ab69ad1829c6589d/classes/class-wc-rest-connect-shipping-label-controller.php#L74))

allendav:
@DanReyLop - please add an issue to fix this error which gets thrown on opening the labels ui:

```
index.js?d4d5:16 POST http://localhost:8888/wp-json/wc/v1/connect/shipping-label 400 (Bad Request)
```

allendav:
@DanReyLop - thank you for persevering with this labels UI - we've almost got all the basics!

@jeffstieler - I'm ok with deferring the sprintf/template/localization issue to Sprint 4 if you are, so that we can get this PR in and so that @DanReyLop can focus on getting the components in place for the last step so that @kellychoffman can proceed with styling and so that we can begin replacing dummy behavior with behavior backed by wired-through endpoints.

If you're ok with this Jeff, let's mark this as ready to merge.

Then, in Sprint 5, I'd like us to carefully consider having a non data driven container for the labels UI so we can iterate faster, and take the gamble that we be ok with requiring the merchant to update the plugin to gain any new functionality we add for labels down the road.

DanReyLop:
> In Edit Order, on the rate step, that rate did not appear (looks like those rates are hardcoded)

Yep, those rates are hard-coded in the PHP code. When I wire PHP with the real endpoints this should work as advertised though :)

> index.js?d4d5:16 POST http://localhost:8888/wp-json/wc/v1/connect/shipping-label 400 (Bad Request)

This just means that the server returned a validation error.If we prefer to not throw errors to the console, we could change our "validation error" responses to return a `200 OK` with an error payload (we are ignoring the HTTP status code anyway).

allendav:
> This just means that the server returned a validation error.If we prefer to not throw errors to the console, we could change our "validation error" responses to return a 200 OK with an error payload (we are ignoring the HTTP status code anyway).

I think we should do that. Please open a separate issue for Sprint 4 for that.

DanReyLop:
No need to create a separate issue, I only needed to change one line :)

client/components/shipping-rates/index.js
DanReyLop: I went ahead and changed the strings to use ES6 templates. At this moment this string isn't going to change for any language ("W x L x H" is universal), so there's no need to localize it.

jeffstieler:
Thanks for your perseverance on this PR @DanReyLop! :ship: it.

PR #479, opened 2016-08-07
Added Labels Purchase&Print step

### LINKED ISSUE #460: [Labels] Purchase & Print label step
Design: TBD

This is the last step of the label printing flow.

It should display a summary of the previous steps, so if the form was auto-advanced until this point, the merchant can take a look at the packages, the address, the chosen rates, etc.

When the user clicks the "Purchase" button, a request should be made to the server to proceed with the actual purchase of the label. The server will return a PDF with the label, and it should be automatically opened in a pop-up window (if possible, the native Print dialog should be automatically opened).

Question: Should we allow the merchant to select the paper size? I say we always return the 4x6" PDF. If the merchant has a regular A4 printer, he will just have to uncheck the "stretch contents to fill page" option in the Print dialog.

DanReyLop:
Fixes #460 

In the last step, this is implemented:
* Show a summary of the previous steps, with a button to take the user back to any of those steps (if it's not obvious that the tabs are clickable).
* Select paper size (I just put `A4` and `4x6` for now).
* When clicking the "Buy" button, the Print Labels dialog is automatically closed and the native "Print" dialog is shown instead, with the label file returned by the server ready to print.

This PR allows to do a complete printing flow, but all the server communication is still mocked.


DanReyLop:
PS: I've tested the native printing dialog in IE11, Edge, Firefox and Chrome (everything in Windows 10). I would appreciate someone tests it in Safari and in some other OS X browser.

client/components/order-packages/index.js
```diff
@@ -52,3 +51,28 @@ OrderPackages.propTypes = {
 };
 
 export default OrderPackages;
+
+export const Summary = ( { packages, dimensionUnit, weightUnit } ) => {
+	const renderPackageInfo = ( pckg, pckgIndex ) => {
+		return (
+			<li key={ pckgIndex }>
+				{ `${renderPackageDimensions( pckg, dimensionUnit )} (${pckg.weight} ${weightUnit})` }
```
jeffstieler: Calling `renderPackageDimensions` from a template string seems unnecessary here. Can you store the result in a variable to use instead? I think that would be more easily read.

client/components/wcc-settings-form/settings-group.js
```diff
@@ -98,6 +104,34 @@ const SettingsGroup = ( props ) => {
...
+					<Button compact
+							onClick={ () => formActions.goToStep( stepIndex ) } >
+						{ __( 'Edit' ) }
+					</Button>
+				</div>
+			);
+
+			return (
+				<div className="settings-group-default">
+					<FormSectionHeading dangerouslySetInnerHTML={ sanitizeHTML( group.title ) } />
```
jeffstieler: Do we actually need to use `dangerouslySetInnerHTML` here?

Does `<FormSectionHeading>{ sanitizeHTML( group.title ) }</FormSectionHeading>` suffice?

classes/class-wc-connect-api-client.php
```diff
@@ -152,6 +152,15 @@ public function send_shipping_label_request( $label_settings ) {
 				}
 			}
 
+			if ( empty( $errors ) && $label_settings->confirm ) {
```
jeffstieler: This line threw a PHP warning for me.

client/components/order-packages/index.js
```diff
@@ -3,24 +3,23 @@ import NumberInput from 'components/number-field/number-input';
...
-const OrderPackages = ( { packages, updateValue, dimensionUnit, weightUnit, errors } ) => {
-	const renderItemInfo = ( item, itemIndex ) => (
-		<li key={ itemIndex }>
-			<span dangerouslySetInnerHTML={ { __html: sanitize( item.name ) } } />
-			{ 1 < item.quantity ? ' (x' + item.quantity + ')' : null }
-		</li>
-	);
+const renderItemInfo = ( item, itemIndex ) => (
+	<li key={ itemIndex }>
+		<span dangerouslySetInnerHTML={ { __html: sanitize( item.name ) } } />
```
jeffstieler: Looks like this was missed in a previous review, but same comment here as https://github.com/Automattic/woocommerce-connect-client/pull/479/files#r73906121

client/components/wcc-settings-form/settings-group.js
DanReyLop: If you use what you propose, the HTML characters would be escaped. The only way to bypass that is to use `dangerouslySetInnerHTML`.

jeffstieler: I just tested my snippet, and it worked as I expected it to:

screen shot 2016-08-08 at 5 05 04 pm

allendav:
The only problem I saw I wrote up here : #484 and that is unrelated to this PR

@jeffstieler - good to go?  If so, then let's :shipit:

jeffstieler:
Good to go. :ship: it.

PR #485, opened 2016-08-12
Remove data-driveness that was added for the Labels UI code

DanReyLop:
This (big) PR does the following things:

Remove the over-complications in data-driven logic that were added just for supporting the Labels UI flow. This includes:
* Multi-step forms
* Suggestions
* Auto-advance logic
* Component summary logic
* Steps & Summary subtypes of layout groups
* Tab bar

These are the parts of the data-driveness that were kept, because I feel they will be useful (feel free to disagree violently):
* Custom field error messages pushed from the server at validation time
* Being able to "overwrite" a field's value from the server
* `Dropdown`, `CountryDropdown` and `StateDropdown` components
* Allow a limited set of HTML tags for all the data-driven strings (titles, descriptions, etc)

Also, a *big* directory reestructuring. All the basic components are in `/client/components`, and those components are not dependant on receiving a `formSchema` or a `formLayout`, instead they receive the actual values they need.

The main entry point is still `main.js`, but now that file acts as a kind of "Router". The existing data-driven code (`WCCSettingsForm`) is under the `client/settings` directory. All the React components that are specific to that "sub-application" are in `client/settings/views`, and its state tree is in `client/settings/state`.

See specially the `client/main.js` file and `client/settings/index.js` file.

For every new "sub-application" (the `Payment selection` UI will be one, for example, and so will be the `Labels UI`), create a new directory tree in `client/applicationName`, add an `index.js` file that respects the interface required by `main.js`, and after that whatever you do under that directory tree will be completely independent of the rest of the bundle.

PS: Right now the Labels UI doesn't work. At all. But since @allendav is soon starting with the "Payment selection" UI work, I feel we need to review this as soon as possible.

@jeffstieler @nabsul @robobot3000 @allendav


assets/stylesheets/style.scss
```diff
@@ -6,7 +6,7 @@
 //@import 'shared/colors'; // import all of our wpcom colors
 @import 'components/dialog/style';
 
-#wc-connect-admin-container,
+.wc-connect-container,
```
allendav: Could we keep the `admin` part please? When we add front-side containers (e.g. in my-account or elsewhere) it will probably be useful to be able to distinguish Calypso like styles from theme inspired styles.

assets/stylesheets/style.scss
```diff
@@ -267,20 +267,17 @@
...
 	@import '../../client/components/text-area/style';
 	@import '../../client/components/toggle/style';
+	@import '../../client/settings/views/packages/style';
 
 }
 
 @keyframes fadeIn { from { opacity: 0; } to { opacity: 1; } }
 
-#wc-connect-admin-container .form-troubles {
+.wc-connect-container .form-troubles {
```
allendav: Same suggestion as above

classes/class-wc-connect-help-provider.php
```diff
@@ -670,13 +670,15 @@ public function page() {
 				$this->get_support_items()
 			);
 
+			$root_view = 'wc-connect-help';
```
allendav: I like this technique, but how about `wc-connect-admin-help`

classes/class-wc-connect-help-provider.php
```diff
@@ -687,7 +689,7 @@ public function page() {
 				<h2>
 					<?php _e( 'WooCommerce Connect Status', 'woocommerce' ); ?>
 				</h2>
-				<div id="wc-connect-admin-container"></div>
+				<div class="wc-connect-container" id="<?= $root_view ?>"></div>
```
allendav: We can't use `<?=` since that requires PHP 5.4 and WordPress can run as low as 5.2.4.

Also, please esc_attr or esc_attr_e $root_view

classes/class-wc-connect-shipping-label.php
```diff
@@ -426,23 +149,22 @@ public function meta_box( $post ) {
...
+				'rootView'     => $root_view,
 			);
 
 			wp_localize_script( 'wc_connect_admin', 'wcConnectData', $admin_array );
 			wp_enqueue_script( 'wc_connect_admin' );
 			wp_enqueue_style( 'wc_connect_admin' );
 
 			?>
-			<div id="wc-connect-admin-container"><span class="form-troubles" style="opacity: 0"><?php printf( __( 'Settings not loading? Visit the WooCommerce Connect <a href="%s">debug page</a> to get some troubleshooting steps.', 'woocommerce' ), $debug_page_uri ); ?></span></div>
+			<div class="wc-connect-container" id="<?= $root_view ?>"><span class="form-troubles" style="opacity: 0"><?php printf( __( 'Settings not loading? Visit the WooCommerce Connect <a href="%s">debug page</a> to get some troubleshooting steps.', 'woocommerce' ), $debug_page_uri ); ?></span></div>
```
allendav: Same comment as above concerning `<?=` and esc_attr

client/main.js
```diff
@@ -1,48 +1,49 @@
...
-	nonce,
-	submitMethod,
-	rootView,
-} = wcConnectData;
+const Route = ( ( rootView ) => {
+	switch ( rootView ) {
+		case 'wc-connect-create-shipping-label':
+			return ShippingLabel;
+		case 'wc-connect-settings':
+		case 'wc-connect-help':
```
allendav: Same comment as above re adding `admin` for this one

allendav:
@DanReyLop thank you for pulling this together so quickly - I tested each of our non label views and they all work. I have a few suggestions (classnames) and a few must dos (PHP 5.4 isms and escaping) and then this is good to go.

DanReyLop:
I really need to read PHP changelogs, I've always worked with 5.4 and I take too many things for granted. Nice catch!

I've fixed the rest of your comments too.

allendav:
:shipit:

PR #488, opened 2016-08-17
Add Settings View for Payment Method Selection

### LINKED ISSUE #471: Add a payment method selection UI
The payment method selection needs to be sent in label requests.

This relies on a to-be-built payment methods enumeration endpoint on the server.

The UI should also prompt the merchant to save a credit card on WordPress.com if they haven't already done so.  **NOTE: If the logged in user is NOT the primary user (i.e. the user that has the actual Jetpack connection) - the UI should instead prompt the current user to have the primary user add a saved card**

This isn't self-help and it isn't shipping zone method settings, so I'm thinking we'll need a new view - a new tab for Connect in wp-admin > WooCommerce > Settings

Something like this - what do you think, @kellychoffman (sorry, gave up on Sketch for the moment - ran out of time)

img_1972

cc @peterbutler @janm6k

allendav:
Fixes #471 

To test:
First, add credentials to your local server, in local.json, e.g.:

```
        "dummy_auth_credentials": {
                "authenticated": true,
                "blog_id": 123456,
                "blog_owner_id": 123456
        },
```

you'll have to use real blog_id and blog_owner_ids, of course.

Then, open your local client and go to wp-admin > WooCommerce > Settings > WooCommerce Connect
Verify you see all your stored payment methods
Verify the save button enables if you change the selection
Verify the save button disables while saving

Caveats:
I'm still trying to get webpack + svgs to work (there are supposed to be card logos. Will fix before I merge this to master.

Note:
I created a new lib, form-base, that contains actions and reducers for forms like this. I did not want to deep link from the shared-settings container into the settings container for its state management.  We can use this new lib for other forms.

I also make the shared settings root view where the notices actions are connected, since that way the root view can control the text that appears (e.g. "Payment methods updated.") more appropriately.


classes/class-wc-connect-api-client.php
```diff
@@ -135,6 +135,15 @@ public function get_shipping_rates( $services, $package ) {
 
 
 		/**
+		 * Asks the WooCommerce connect server for an array of payment methods
+		 *
+		 * @return true|WP_Error
```
DanReyLop: This doesn't return `true` :)

classes/class-wc-connect-payment-methods-store.php
```diff
@@ -0,0 +1,128 @@
...
+				return;
+			}
+
+			$this->logger->log( 'Successfully loaded payment methods from server response.', __FUNCTION__ );
+			$this->update_last_fetch_timestamp();
+
+			// If we made it this far, it is safe to store the object
+			$this->update_payment_methods( $payment_methods );
+
+			// Potentially auto-update the selected method
```
DanReyLop: If you're going to put an entire sentence for a method's name, this kind of comment is redundant

classes/class-wc-connect-payment-methods-store.php
```diff
@@ -0,0 +1,128 @@
...
+				}
+			}
+
+			// Just one? Go ahead and use it
+			if ( 1 === $payment_method_count ) {
+				$this->service_settings_store->set_selected_payment_method_id( $payment_method_ids[ 0 ] );
+				return;
+			}
+
+			// Is the stored method id not in the list? Select the first one
```
DanReyLop: This seems like too big of an assumption. I don't think we should pre-select a payment method if there's more than one.

classes/class-wc-connect-payment-methods-store.php
```diff
@@ -0,0 +1,128 @@
...
+
+		public function get_payment_methods() {
+			return get_option( 'wc_connect_payment_methods', array() );
+		}
+
+		protected function update_payment_methods( $payment_methods ) {
+			update_option( 'wc_connect_payment_methods', $payment_methods );
+		}
+
+		public function get_last_fetch_timestamp() {
```
DanReyLop: This function is never called. Do you plan on using it in the future? I get what it can be used for, but I also don't like dead code :fire:

classes/class-wc-connect-payment-methods-store.php
```diff
@@ -0,0 +1,128 @@
...
+			return get_option( 'wc_connect_payment_methods_last_update', null );
+		}
+
+		protected function update_last_fetch_timestamp() {
+			update_option( 'wc_connect_payment_methods_last_update', time() );
+		}
+
+		protected function get_payment_methods_from_response_body( $response_body ) {
+			if ( ! is_object( $response_body ) ) {
+				return new WP_Error( 'payment_method_response_body_type', 'Expected but did not receive object for response body.' );
```
DanReyLop: Are these errors going to be presented to the user (in a notice)? Because they look like errors from a test case, not very user-friendly.

classes/class-wc-connect-payment-methods-store.php
```diff
@@ -0,0 +1,128 @@
...
+				return new WP_Error( 'payment_method_response_body_missing_payment_methods', 'Expected but did not receive payment_methods in response body.' );
+			}
+
+			$payment_methods = $response_body->payment_methods;
+			if ( ! is_array( $payment_methods ) ) {
+				return new WP_Error( 'payment_methods_type', 'Expected but did not receive array for payment_methods.' );
+			}
+
+			foreach ( (array) $payment_methods as $payment_method ) {
+				$required_keys = array( 'payment_method_id', 'name', 'card_type', 'card_digits', 'expiry' );
```
DanReyLop: Do we need this kind of defensive programming? Shouldn't we trust the server on this?

classes/class-wc-connect-settings-view.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+if ( ! class_exists( 'WC_Connect_Settings_View' ) ) {
+
+	class WC_Connect_Settings_View {
```
DanReyLop: The `Packages Manager` is going to be included in this page? If this is only for payment method selection, then the `WC_Connect_Settings_View` name is too generic.

client/lib/form-base/actions.js
```diff
@@ -0,0 +1,82 @@
...
+	//
+	// 				return setError( JSON.stringify( json ) );
+	// 			} ).then( () => setIsSaving( false ) );
+	// 		} ).catch( ( e ) => {
+	// 			setError( e );
+	// 			setIsSaving( false );
+	// 		} );
+	// 	};
+	// };
+//};
```
DanReyLop: ^^ This ^^

client/components/payment-method-selector/index.js
```diff
@@ -0,0 +1,55 @@
...
+				/>
+			);
+		} )
+	);
+
+	const hasMethods = 0 < paymentMethods.length;
+
+	return (
+		<FormFieldset className="payment-method-selector">
+			<FormLegend dangerouslySetInnerHTML={ sanitizeHTML( title ) } />
```
DanReyLop: Since this component is not data-driven, this `dangerouslySetInnerHTML` is not needed. If for some reason in the future we need to add mark-up to the title, we can just pass a React Component as the title instead of a string.

client/components/payment-method-selector/index.js
```diff
@@ -0,0 +1,55 @@
+import React, { PropTypes } from 'react';
```
DanReyLop: Since this is a component that will only be used in the Payment Selection screen, it's better to put it in `client/shared-settings/views` and leave `client/components` only for generic, reusable components (to reduce the clutter). Same for the `PaymentMethod` component.

client/components/payment-method/index.js
```diff
@@ -0,0 +1,47 @@
...
+		}
+	);
+
+	return (
+		<FormLabel className="payment-method__label">
+			<FormRadio
+				className="payment-method__radio"
+				value={ value }
+				checked={ value === currentValue }
+				onChange={ () => onChange( value ) }
```
DanReyLop: The `PaymentMethodSelection` `onChange` property is ignoring its argument, so you can just do here:
`onChange={ onChange }`

client/components/payment-method/index.js
```diff
@@ -0,0 +1,47 @@
+import React, { PropTypes } from 'react';
+import FormLabel from 'components/forms/form-label';
+import FormRadio from 'components/forms/form-radio';
+import { sprintf } from 'sprintf-js';
+import { translate as __ } from 'lib/mixins/i18n';
+
+// TODO - create PaymentMethod like https://github.com/Automattic/wp-calypso/blob/29984e5677960bdba231577a253cf864d4f17796/client/my-sites/upgrades/checkout/stored-card.jsx
+// or https://github.com/Automattic/wp-calypso/tree/29984e5677960bdba231577a253cf864d4f17796/client/my-sites/upgrades/checkout
```
DanReyLop: ^^ this ^^

client/components/payment-method/style.scss
```diff
@@ -0,0 +1,66 @@
...
+		white-space: nowrap;
+	}
+}
+
+input[type=radio].payment-method__radio.form-radio {
+	margin-top: 25px;
+}
+
+.payment-method.visa {
+	background-image: url('/calypso/images/upgrades/cc-visa-disabled.svg');
```
DanReyLop: If you can't get Webpack to copy the assets to the bundle folder, you should consider inlining the SVG here, using a loader. SVGs are very small and can be embedded as plain text (no base64 overhead), I think that would give you less headaches :)

client/lib/form-base/actions.js
```diff
@@ -0,0 +1,82 @@
+// SET_FORM_DATA_VALUE is used to update a form field's underlying setting, e.g. selected_payment_method_id
```
DanReyLop: I don't know how I feel about this file (and the reducers one). If it's specific to a part of the application, I think it should go into that sub-application directory tree. Trying to over-generalize didn't give us good results the last time :)

client/lib/form-base/reducer.js
```diff
@@ -0,0 +1,45 @@
...
+	const data = Object.assign( {}, state.data );
+	const meta = Object.assign( {}, state.meta, { [ action.key ]: action.value } );
+	return {
+		meta: meta,
+		data: data,
+	};
+};
+
+const reducer = ( state, action ) => {
+	if ( 'undefined' === typeof state ) {
```
DanReyLop: You can declare this function like this instead, and remove the `undefined` check:
`const reducer = ( state = initialState, action ) => {`
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Examples

client/lib/form-base/reducer.js
```diff
@@ -0,0 +1,45 @@
...
+		pristine: true,
+	},
+	data: {
+	},
+};
+
+const reducers = {};
+
+reducers[ SET_FORM_DATA_VALUE ] = ( state, action ) => {
+	const data = Object.assign( {}, state.data, { [ action.key ]: action.value } );
```
DanReyLop: For an alternative to this `Object.assign` syntax, see: http://redux.js.org/docs/recipes/UsingObjectSpreadOperator.html
Example in our codebase: https://github.com/Automattic/woocommerce-connect-client/pull/491/files#diff-a145a70f9fe0565fb7752d16400d20cbR25
Of course, to use one syntax or another is entirely personal :)

client/shared-settings/index.js
```diff
@@ -0,0 +1,39 @@
...
+	getReducer() {
+		return combineReducers( {
+			form: reducer,
+			notices,
+		} );
+	},
+
+	getHotReducer() {
+		return combineReducers( {
+			form: reducer,
```
DanReyLop: You need to `require()` the reducer again, or this hot-reloading won't work (anyways, I think our Webpack hot-reloading is mostly broken).

classes/class-wc-connect-payment-methods-store.php
allendav: Copy pasta! Will fix

allendav:
@DanReyLop - I knew I rushed this to review :)  Thank you for putting up with this.  Will fix all these things.

A big oops is `SharedSettingsRootView` really needs to be `PaymentMethodsRootView` since the packages view will be separate.

As for `lib/form-base`, I'd like to keep and I'd like us to consider sharing between these mini-apps.

@robobot3000 - would the form-base be useful for packages? That was my thinking. If not, I'll move it inside my container's folder

robobot3000:
@allendav I think form-base might be useful for me. But I don't mind if it gets removed. I also had some bad experiences with trying to generalize things too much

classes/class-wc-connect-payment-methods-store.php
allendav: They are only used for the debug log

classes/class-wc-connect-payment-methods-store.php
allendav: I would prefer to keep it. If we ever made a mistake on the server that caused one of these fields to be absent, I would not want to overwrite the option on the client with incomplete payment method data.

classes/class-wc-connect-api-client.php
allendav: Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/488/commits/8ebe257dc446a63b65a6a909757e8a51ee1eb35a

classes/class-wc-connect-payment-methods-store.php
allendav: Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/488/commits/8ebe257dc446a63b65a6a909757e8a51ee1eb35a

classes/class-wc-connect-payment-methods-store.php
allendav: Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/488/commits/8ebe257dc446a63b65a6a909757e8a51ee1eb35a

classes/class-wc-connect-settings-view.php
allendav: Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/488/commits/8ebe257dc446a63b65a6a909757e8a51ee1eb35a

client/lib/form-base/actions.js
allendav: Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/488/commits/8ebe257dc446a63b65a6a909757e8a51ee1eb35a

client/components/payment-method-selector/index.js
allendav: Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/488/commits/8ebe257dc446a63b65a6a909757e8a51ee1eb35a

client/components/payment-method/index.js
allendav: Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/488/commits/8ebe257dc446a63b65a6a909757e8a51ee1eb35a

client/components/payment-method/index.js
allendav: Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/488/commits/8ebe257dc446a63b65a6a909757e8a51ee1eb35a

client/shared-settings/index.js
allendav: Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/488/commits/8ebe257dc446a63b65a6a909757e8a51ee1eb35a

classes/class-wc-connect-payment-methods-store.php
allendav: I'd rather we preselect one whenever possible - to avoid happiness engineering getting calls that could be avoided.

allendav:
@DanReyLop I ended up having to copy the svgs into assets and then I was able to get the svg loader to work. Let me know if you want to try to get it to work directly from calypso and I'll open an issue.

Let me know if this is good to go.

classes/class-wc-connect-settings-pages.php
```diff
@@ -97,13 +144,15 @@ public function page() {
...
 
 			?>
-				<h2>
-					<?php _e( 'Account', 'woocommerce' ); ?>
-				</h2>
 				<div class="wc-connect-admin-container" id="<?php echo esc_attr( $root_view ) ?>"></div>
 			<?php
 		}
 
+		public function output_packages_screen() {
```
DanReyLop: I know "double tab-bars" are used widely in the WP-Admin, but I find them to be one of the worst UX patterns imaginable. Totally off-topic here, but I think we should avoid that.

If we are marketing Connect as "a bunch of features", just like Jetpack, I don't think we *need* to group their configuration inside a tab. Instead we should put them where they make more sense. For example, Payment Selection would be under the "Checkout" tab, and the Packages Manager under the "Shipping" tab.

As I said, totally off-topic, but maybe we should give this a thought? ( @kellychoffman )

client/account-settings/views/payment-method/index.js
```diff
@@ -4,10 +4,9 @@ import FormRadio from 'components/forms/form-radio';
 import { sprintf } from 'sprintf-js';
 import { translate as __ } from 'lib/mixins/i18n';
 
-// TODO - create PaymentMethod like https://github.com/Automattic/wp-calypso/blob/29984e5677960bdba231577a253cf864d4f17796/client/my-sites/upgrades/checkout/stored-card.jsx
-// or https://github.com/Automattic/wp-calypso/tree/29984e5677960bdba231577a253cf864d4f17796/client/my-sites/upgrades/checkout
-
 const PaymentMethod = ( { value, currentValue, onChange, name, cardType, cardDigits, expiry } ) => {
+	const selected = ( value === currentValue );
+	const containerClasses = selected ? 'payment-method__label selected' : 'payment-method__label';
```
DanReyLop: You could change this line to:
`const containerClass = classnames( 'payment-method__label', { selected } )`
Kind-of overkill for this, but that's what the `classnames` library is for and we've been using it all over the code.

DanReyLop:
The file `assets/images/google-apps-logo.png` is not used.

classes/class-wc-connect-payment-methods-store.php
DanReyLop: https://github.com/Automattic/woocommerce-connect-client/issues/447 and https://github.com/Automattic/woocommerce-connect-client/issues/496 should cover us in those cases.

What I'm worried about is that if a merchant has 2 credit cards, we would be essentially choosing one of them for him at random. That doesn't sound very good :(
Since this is a conscious decision, my job as "guy who points things out in PRs" is done, I just want you to consider it carefully :)

client/lib/form-base/reducer.js
DanReyLop: This comment got lost when you moved the file. In case you missed it, I find it quite easier to read (the JS community seems to agree). Same about the `Object.assign` comment, although that's more debatable.

DanReyLop:
I had no strong opinion one way or the other about the `form-base` stuff, maybe @robobot3000 can try to give it a go (even if it's on another subdirectory) for the `Packages Manager`, and if it works well for him we can move it back to being a generic lib.

> I ended up having to copy the svgs into assets and then I was able to get the svg loader to work. Let me know if you want to try to get it to work directly from calypso and I'll open an issue.

Totally fine for me to stay as it is. The icons for credit cards aren't gonna change any time soon.

About the `alert` that pops-up when you try to leave the page: That won't work after #491 is merged (see https://github.com/Automattic/woocommerce-connect-client/pull/491/files#diff-096103d8490b885f9d5037baea6aea80R33 ). We have a few options:
* Duplicate the `window.onbeforeunload` code for each application.
* Agree to use the `state.form.pristine` property for every application with a form.
* Change the `window.onbeforeunload` handler to use `state.form.meta.pristine` (that would require changing the structure of the data-driven state).
* Add a method in each sub-application (at the same level of `getReducer`, `getHotReducer`, etc) to return if the form is "dirty" or not, and make the `window.onbeforeunload` handler check that instead of `state.whatever.pristine`.
I'm fine either way, please create an issue with your preferred option :)

When you remove the PNG that is not needed, you can :shipit:

allendav:
Thank you @DanReyLop - I removed the superfluous file and opened https://github.com/Automattic/woocommerce-connect-client/issues/516

Issue #478, opened 2016-08-04
Shipping labels not loading: Update message


kellychoffman:
Right now, when there's an error and the labels won't display, we show this message:

<img width="302" alt="screen shot 2016-08-04 at 12 55 02 pm" src="https://cloud.githubusercontent.com/assets/5835847/17417539/1c68d60c-5a51-11e6-9586-f33c6816829b.png">

It works for the settings page, but should be updated to make more sense in the context of a metabox in the Orders page. 

I think it could say something like:

> Shipping labels not loading? Visit the debug page for troubleshooting steps.

DanReyLop:
Also we need to remember to change the rest of the places where we use the "WooCommerce Connect" name.

robobot3000:
Fixed in #489 - the new messages are: "Shipping labels not loading? Visit the status page for troubleshooting steps." and "Settings not loading? Visit the status page for troubleshooting steps."

PR #489, opened 2016-08-17
Changed the message appearing when settings/labels fail to load

robobot3000:
Fix for #478 

When labels fail to load the copy should read: "Shipping labels not loading? Visit the status page for troubleshooting steps."

When settings fail to load: "Settings not loading? Visit the status page for some troubleshooting steps."

CC @kellychoffman

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -164,7 +164,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 			wp_enqueue_style( 'wc_connect_admin' );
 
 			?>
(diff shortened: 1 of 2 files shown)


classes/class-wc-connect-shipping-label.php
```diff
@@ -164,7 +164,7 @@ public function meta_box( $post ) {
 			wp_enqueue_style( 'wc_connect_admin' );
 
 			?>
-			<div class="wc-connect-admin-container" id="<?php echo esc_attr( $root_view ) ?>"><span class="form-troubles" style="opacity: 0"><?php printf( __( 'Settings not loading? Visit the WooCommerce Connect <a href="%s">debug page</a> to get some troubleshooting steps.', 'woocommerce' ), $debug_page_uri ); ?></span></div>
```
kellychoffman: Minor: You got rid of `some` here, but not in the shipping-method file. I'd probably get rid of it there too to keep it consistent.

kellychoffman:
One super nitpicky change, but otherwise this is good to 🚢 .

PR #491, opened 2016-08-17
Replace tabbed labels UI with one-form approach

### LINKED ISSUE #480: Replace tabbed labels UI with one-form approach
adding-weight-packages-added

Full mocks: #501 

- [x] Dialog basic structure
- [x] Origin Address step
- [x] Destination Address step
- [x] Packages step
- [x] Rates step
- [x] Preview step
- [ ] Print the label & store in the order meta-data

Stretch tasks:
- [ ] Normalize Origin Address
- [ ] Normalize Destination Address
- [ ] Call the server box-packer if the order doesn't have packaging info
- [ ] Call the server shipping-rates endpoint to allow the merchant to choose the rates

DanReyLop:
Fixes #480

There is a lot still to be done, namely:
- [ ] Actually getting it to work (@jeffstieler I need you).
- [ ] Storing the label data (timestamp, tracking_id, label_id) in the order metadata.
- [ ] Supporting multiple labels in the same response.
- [ ] Normalizing both the addresses.
- [ ] Invoking the box packing algorithm if the order doesn't have it.
- [ ] Invoking the shipping rates endpoint to allow the merchant to select.
- [ ] A UX that actually makes sense.

@robobot3000 You can use this branch as a starting point to do the Refund flow. See `client/shipping-label/index.js` and `classes/class-wc-connect-shipping-label.php` (return some hard-coded data in `$admin_array['labelsData']` and render the Refund UI in the `client/shipping-label/index.js` `renderSuccessNotice()` function).

@jeffstieler I made a request that looks fine, but I receive an error like this:
`The element 'eVSRequest' has invalid child element 'Width'. List of possible elements expected: 'HoldForManifest, NineDigitRoutingZip, ShipInfo, ........., AltReturnAddress2, AltReturnAddress3, AltReturnAddress4, AltReturnAddress5, AltReturnAddress6, AltReturnCountry`

@allendav @nabsul @kellychoffman


DanReyLop:
I've decided to mark this as "Needs Review". This branch size is already quite big, and it provides a good starting point to keep iterating on this.

I couldn't get the endpoint to work, so I didn't implement the "store label data in the order meta-data" part of the flow.

nabsul:
Side note: I'd take the checklist you have up there and either copy it into an issue or break it down into several issues.

nabsul:
I was able to open the shipping label modal and play with the different fields there. However, I'm not sure what is supposed to be working and what is not... Can you clarify this with some test instructions?

classes/class-wc-connect-shipping-label.php
```diff
@@ -56,7 +56,7 @@ protected function get_packaging_data( WC_Order $order ) {
 			$shipping_methods = $order->get_shipping_methods();
 			$shipping_method = reset( $shipping_methods );
 			if ( ! $shipping_method || ! isset( $shipping_method[ 'wc_connect_packages' ] ) ) {
-				return $this->get_items_as_individual_packages( $order );
+				return array( FALSE, $this->get_items_as_individual_packages( $order ) );
```
nabsul: I had to look at how this method was used to understand what `TRUE` and `FALSE` where.
How about returning `array( 'is_packed' => FALSE, 'packages' => .... )` ?

classes/class-wc-connect-shipping-label.php
```diff
@@ -97,7 +97,7 @@ protected function get_selected_rates( WC_Order $order ) {
 		protected function get_form_data( WC_Order $order ) {
 			$form_data = array();
 
-			$form_data[ 'packages' ] = $this->get_packaging_data( $order );
+			list( $form_data[ 'is_packed' ], $form_data[ 'packages' ] ) = $this->get_packaging_data( $order );
```
nabsul: And then this code would be a lot simpler.

classes/class-wc-connect-shipping-label.php
```diff
@@ -107,12 +107,12 @@ protected function get_form_data( WC_Order $order ) {
 			$form_data[ 'destination' ] = array(
 				'name' => trim( $dest_address[ 'first_name' ] . ' ' . $dest_address[ 'last_name' ] ),
 				'company' => $dest_address[ 'company' ],
-				'address_1' => $dest_address[ 'address_1' ],
+				'address' => $dest_address[ 'address_1' ],
```
nabsul: Should we change the fieldname in our `formData` to `address_1`so that we maintain consistency with WooCommerce?

DanReyLop:
> Side note: I'd take the checklist you have up there and either copy it into an issue or break it down into several issues. Or copy the checklist to the new PR that will resolve the items :smile:

Normalizing the addresses and invoking the packaging and rates endpoints already have their own issues, and the "getting it to work" and "do something reasonable UX" are obvious :D . I''ve created an issue for the "multiple labels" thing, since that can be tricky.

> I was able to open the shipping label modal and play with the different fields there. However, I'm not sure what is supposed to be working and what is not... Can you clarify this with some test instructions?

UI must work. That means you can edit the origin and destination addresses (it's forced to US since we don't want to support international shipments for now), you can also edit the packages' weight. Validation works too, so you won't be able to leave any required field blank, and invalid values (such an invalid US postcode or negative weight in a package) will also be detected. The "Print" button will be enabled only if there are no errors in any field.

What doesn't work: Since we're not hitting the rates endpoint yet, only the "service_id" selected at check-out time for each package will be shown in the "Rates" step, with a simulated price ($9.99). Also, when you click the "Print" button, there is an error returned by the WCC server, so it won't do anything.

classes/class-wc-connect-shipping-label.php
DanReyLop: In this case, I think it's best to maintain consistency with the WCC server endpoints.

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -69,15 +69,34 @@ private function store_origin_address( $settings ) {
...
 			$this->store_origin_address( $settings );
 		}
 
-		return $response;
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		$labels_data = array();
+		foreach ( $response->labels as $index => $label_data ) {
```
nabsul: How do I test this?

classes/class-wc-connect-shipping-label.php
nabsul: Or maybe the server should also be changed to match WooCommerce?
Anyways, we can put this off for later.

classes/class-wc-rest-connect-shipping-label-controller.php
DanReyLop: You can't, since at this point I couldn't make a "valid" request to the `purchase-label` endpoint. I left this code here as a kind-of placeholder.

classes/class-wc-connect-shipping-label.php
DanReyLop: I don't know what I was thinking, this is really obfuscated code (maybe meant as temporal code?)

I've ended up breaking this method into 2, returning an array (of any kind) when you want a method to return 2 things doesn't feel like a good idea.

classes/class-wc-connect-shipping-label.php
```diff
@@ -52,14 +52,21 @@ protected function get_items_as_individual_packages( $order ) {
 			return $packages;
 		}
 
-		protected function get_packaging_data( WC_Order $order ) {
+		protected function get_packaging_metadata( WC_Order $order ) {
 			$shipping_methods = $order->get_shipping_methods();
 			$shipping_method = reset( $shipping_methods );
 			if ( ! $shipping_method || ! isset( $shipping_method[ 'wc_connect_packages' ] ) ) {
-				return $this->get_items_as_individual_packages( $order );
+				return FALSE;
```
allendav: We've been using the lowercase version of true and false (WP coding standards don't address this directly, but all the examples use lowercase too at https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/ )

classes/class-wc-connect-shipping-label.php
```diff
@@ -97,7 +104,8 @@ protected function get_selected_rates( WC_Order $order ) {
 		protected function get_form_data( WC_Order $order ) {
 			$form_data = array();
 
-			$form_data[ 'packages' ] = $this->get_packaging_data( $order );
+			$form_data[ 'is_packed' ] = FALSE !== $this->get_packaging_metadata( $order );
```
allendav: Same comment as above

classes/class-wc-connect-shipping-label.php
```diff
@@ -107,12 +115,12 @@ protected function get_form_data( WC_Order $order ) {
...
 				'name' => trim( $dest_address[ 'first_name' ] . ' ' . $dest_address[ 'last_name' ] ),
 				'company' => $dest_address[ 'company' ],
-				'address_1' => $dest_address[ 'address_1' ],
+				'address' => $dest_address[ 'address_1' ],
 				'address_2' => $dest_address[ 'address_2' ],
 				'city' => $dest_address[ 'city' ],
 				'state' => $dest_address[ 'state' ],
 				'postcode' => $dest_address[ 'postcode' ],
-				'country' => $dest_address[ 'country' ],
+				'country' => $dest_address[ 'country' ] ? $dest_address[ 'country' ] : 'US',
```
allendav: Under what circumstances would country be blank? Only in the case of a manually entered order?

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -69,15 +69,34 @@ private function store_origin_address( $settings ) {
 
 	public function update_items( $request ) {
 		$request_body = $request->get_body();
-		$settings = json_decode( $request_body, false, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
+		$settings = json_decode( $request_body, TRUE, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
```
allendav: Lowercase

client/components/country-dropdown/index.js
```diff
@@ -17,7 +17,6 @@ const CountryDropdown = ( props ) => {
 CountryDropdown.propTypes = {
 	id: PropTypes.string.isRequired,
 	countriesData: PropTypes.object.isRequired,
-	valuesMap: PropTypes.object.isRequired,
```
allendav: Why this change?

client/components/dropdown/index.js
```diff
@@ -15,6 +15,7 @@ const Dropdown = ( { id, valuesMap, title, description, value, updateValue, erro
 				name={ id }
 				value={ value }
 				onChange={ ( event ) => updateValue( event.target.value ) }
+				disabled={ Boolean( disabled ) }
```
allendav: If the prop type is required to be bool, why do we need Boolean here? Would having getDefaultProps instead be the more common pattern for React?

client/components/state-dropdown/index.js
```diff
@@ -27,7 +27,6 @@ const StateDropdown = ( props ) => {
 StateDropdown.propTypes = {
 	id: PropTypes.string.isRequired,
 	countriesData: PropTypes.object.isRequired,
-	valuesMap: PropTypes.object.isRequired,
```
allendav: Why this change?

client/main.js
```diff
@@ -30,7 +30,7 @@ const store = createStore(
 );
 
 window.addEventListener( 'beforeunload', ( event ) => {
-	if ( store.getState().form && store.getState().form.pristine ) {
+	if ( ! store.getState().form || store.getState().form.pristine ) {
```
allendav: Thank you - this will help my PR :)

client/shipping-label/state/actions.js
```diff
@@ -1,15 +1,99 @@
...
 };
 
-export const exitPrintingFlow = () => ( dispatch ) => {
-	dispatch( { type: EXIT_PRINTING_FLOW } );
-	dispatch( FormActions.resetFlow() );
+export const exitPrintingFlow = () => {
+	return { type: EXIT_PRINTING_FLOW };
+};
+
+export const updateAddressValue = ( group, name, value ) => {
```
allendav: What does `group` represent in the form?

client/shipping-label/state/actions.js
```diff
@@ -1,15 +1,99 @@
...
+		}
+	};
+	const setIsSaving = ( saving ) => {
+		if ( saving ) {
+			dispatch( { type: PURCHASE_LABEL_REQUEST } );
+		} else {
+			dispatch( { type: PURCHASE_LABEL_RESPONSE, response, error } );
+			if ( error ) {
+				dispatch( NoticeActions.errorNotice( error.toString(), {
+					duration: 7000,
```
allendav: Let's not auto-fade label purchase error notices, so 1) the merchant doesn't miss it and 2) can refer to the exact error message in help requests.

client/shipping-label/state/actions.js
```diff
@@ -1,15 +1,99 @@
...
+		}
+	};
+
+	const form = getState().shippingLabel.form;
+	const formData = {
+		origin: form.origin.values,
+		destination: form.destination.values,
+		label_size: form.preview.values.paper_size,
+		ship_date: moment().add( 1, 'days' ).format( 'YYYY-MM-DD' ),
+		payment_method_id: '123456789',
```
allendav: Please add an issue to connect this properly.

client/shipping-label/state/actions.js
```diff
@@ -1,15 +1,99 @@
...
+		if ( saving ) {
+			dispatch( { type: PURCHASE_LABEL_REQUEST } );
+		} else {
+			dispatch( { type: PURCHASE_LABEL_RESPONSE, response, error } );
+			if ( error ) {
+				dispatch( NoticeActions.errorNotice( error.toString(), {
+					duration: 7000,
+				} ) );
+			} else {
+				printDocument( 'data:application/pdf;base64,' + response[ 0 ].image ); // TODO: Figure out how to print multiple PDFs
```
allendav: Are you missing a dispatch to EXIT_PRINTING_FLOW after printDocument? (i.e. to close the label ui modal automatically)

client/shipping-label/state/actions.js
```diff
@@ -1,15 +1,99 @@
...
+			}
+		}
+	};
+
+	const form = getState().shippingLabel.form;
+	const formData = {
+		origin: form.origin.values,
+		destination: form.destination.values,
+		label_size: form.preview.values.paper_size,
+		ship_date: moment().add( 1, 'days' ).format( 'YYYY-MM-DD' ),
```
allendav: Please check with @jeffstieler if easypost has a preference about shipdate setting

client/shipping-label/views/index.js
```diff
@@ -25,10 +26,10 @@ const ShippingLabelRootView = ( props ) => {
 		<p className="wcc-metabox-button-container">
 			<GlobalNotices id="notices" notices={ notices.list } />
 			<PrintLabelDialog
-				{ ...( props.shippingLabel ) }
+				{ ...props.shippingLabel }
 				{ ...props }
 			/>
-			{ props.shippingLabel.success ? renderSuccessNotice() : renderPrintLabelFlow() }
+			{ props.shippingLabel.labels.purchased_time ? renderSuccessNotice() : renderPrintLabelFlow() }
```
allendav: What does this do? Would the merchant ever see a success notice? Or would we ideally close the modal after the printing starts?

client/shipping-label/views/steps/address/index.js
```diff
@@ -0,0 +1,50 @@
...
+import FoldableCard from 'components/foldable-card';
+import { translate as __ } from 'lib/mixins/i18n';
+import AddressFields from './fields';
+import ExpandButton from 'shipping-label/views/expand-button';
+
+const Origin = ( props ) => {
+	return (
+		<FoldableCard
+			header={ __( 'Origin address' ) }
+			summary={ __( '' ) }
```
allendav: Could probably just pass '' here and on the next line

client/shipping-label/views/steps/preview/index.js
```diff
@@ -0,0 +1,32 @@
...
+			header={ __( 'Preview' ) }
+			summary={ __( '' ) }
+			expandedSummary={ __( '' ) }
+			clickableHeader={ true }
+			actionButton={ <ExpandButton/> }
+			expanded={ false } >
+			<Dropdown
+				id="paper_size"
+				title={ __( 'Paper size' ) }
+				valuesMap={ { a4: 'A4', '4x6': '4"x6"' } }
```
allendav: Please replace A4 with Letter, as A4 is an unlikely size for US merchants' printers.

client/shipping-label/views/steps/rates/index.js
```diff
@@ -0,0 +1,37 @@
...
+import FoldableCard from 'components/foldable-card';
+import { translate as __ } from 'lib/mixins/i18n';
+import ShippingRates from './list';
+import ExpandButton from 'shipping-label/views/expand-button';
+
+const RatesStep = ( { form, storeOptions, labelActions, errors } ) => {
+	return (
+		<FoldableCard
+			header={ __( 'Rates' ) }
+			summary={ __( '' ) }
```
allendav: Same comment as above - can probably just do '' here and on the next line

client/shipping-label/views/dialog.js
```diff
@@ -1,14 +1,73 @@
...
+	const canPurchase = ! props.form.isSubmitting && ! hasErrors( props.errors );
 	return (
 		<Dialog
 			isVisible={ props.showDialog }
 			onClose={ props.labelActions.exitPrintingFlow }
-			additionalClassNames="wcc-modal wcc-shipping-label-dialog"
-		>
-			<span>Not implemented yet</span>
+			additionalClassNames="wcc-modal wcc-shipping-label-dialog" >
+			<AddressStep.Origin
```
allendav: Some markup is missing.  As part of this PR, let's include the markup for
* header, body and footer containers
* the dialog title "Create shipping labels" in the header container
* the left and right hand containers/"columns" (preview goes in the right hand one and all of the other steps inside the left) inside the body container
* the action buttons inside the footer container

Or something like that. We can fine tune styles later, but the markup defining those containers should be in this PR imho

Please also increase the dialog width by 50%

classes/class-wc-connect-shipping-label.php
DanReyLop: Yes, only for a manually entered order. In #498 I implemented that the "Create Label" meta-box won't show for international orders, but it *will* show for orders with no country (and in those orders we must force "US").

client/components/country-dropdown/index.js
DanReyLop: Because the CountryDropdown component doesn't receive the `valuesMap` property, instead it gets the values from the list of `storeOptions.countriesData`. This PropType definition was mistakingly kept from a previous refactor.

client/components/dropdown/index.js
DanReyLop: Because it's not required, so it can be `true`, `false` or `undefined` (which needs to be interpreted as `false`). To use `getDefaultProps`, this component would need to be a `class` component.

client/components/state-dropdown/index.js
DanReyLop: Same as above, the `StateDropdown` component generates its own `valuesMap` from the `countriesData` object.

client/shipping-label/state/actions.js
DanReyLop: `group` would be either `origin` or `destination`, so to change the origin postcode this action would be called like `updateAddressValue( 'origin', 'postcode', '12345' )`

client/shipping-label/state/actions.js
DanReyLop: Yes, I just added this here as a quick hardcoded value to make Jeff's endpoint swallow this request

client/shipping-label/state/actions.js
DanReyLop: Created: https://github.com/Automattic/woocommerce-connect-client/issues/504

client/shipping-label/views/index.js
DanReyLop: This `successNotice` would be renered *instead of* the "Create Label" button. A couple of weeks from now, instead of a generic "Congratulations, you printed a label" message, the "Refund" button will be there.

client/shipping-label/views/steps/address/index.js
DanReyLop: Yep, I left this here as a placeholder. According to the latest mock-ups, the summaries will have plenty of content :)

client/shipping-label/views/dialog.js
kellychoffman: I concur! Once the markup is in place, I will fine tune the styles before we merge this.

kellychoffman:
As per commit checklist, this should be mobile ready so I've included a mobile design mock at the top of #501.

client/shipping-label/state/actions.js
DanReyLop: For the record ( @kellychoffman ) EasyPost doesn't require ShipDate, so I'm removing this and you don't have to include in the UI (at least until/if we integrate with USPS directly)

jeffstieler:
I see the following console notice on this branch:
```/~/moment/lang/readme.md
Module parse failed: /node_modules/moment/lang/readme.md Unexpected character '#' (1:0)
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected character '#' (1:0)```

**Comment by DanReyLop** [comment]
@jeffstieler That's weird... seems like webpack is choking trying to bundle `moment.js`. Anyway, since we don't need `ShipDate`, I've removed the `moment.js` dependency entirely (it was overkill anyway).

@kellychoffman I've created the markup (and did some quick&dirty styling so make the genral layout similar to the mocks). Please tell me if you can pick it from here :)

**Comment by kellychoffman** [comment]
> Please tell me if you can pick it from here :)

Absolutely. Marking this as `in-progress` and getting my 🎨 on.

**Comment by kellychoffman** [comment]
Phew, okay I think this is good to go. ✨ And its responsive.✨

File: assets/stylesheets/shipping-label.scss [hunk 23]
```diff
@@ -7,26 +7,54 @@
 .wcc-shipping-label-dialog__content {
 	display: flex;
 	flex-direction: column;
-	height: 100%;
+	max-height: 100%;
```
DanReyLop: With this style, the scroll bar can show even if the modal has enough height. See this (Chrome in Windows):
scroll

Maybe in Mac it's not noticeable because of the invisible scrollbars :)
Putting back `height: 100%` here solves it (I don't know if it's some kind of CSS heresy though)

assets/stylesheets/style.scss
```diff
@@ -171,16 +171,41 @@
...
+			left: 12.5%;
+			right: 12.5%;
+			width: 75%;
 		}
 
 		.dialog__content {
 			height: 100%;
 		}
+
+		.form-buttons-bar {
```
DanReyLop: There's a rule for `.form-buttons-bar` in `shipping-label.scss` too. I suggest you move this there (or vice-versa).

DanReyLop:
Oh my god it looks sooo good on mobile! :star2: :sparkles: :crystal_ball: 
2 very minor comments and then you can :ship:

kellychoffman:
Fixed the two issues you raised. Got caught up in a deep rabbit hole of trying to make sure both this and the add-packages modal weren't conflicting with each other. Its not perfect, but its closer than it was.

DanReyLop:
Just a super-minor thing, either I missed it before (most likely) or it wasn't present. When the viewpoer is narrow (less than 480px), the action buttons are one bellow the other:
buttons

If you can't see it, try re-installing the client. I had to nuke `node_modules` because some weird Windows file permissions issue, so that may be related to an update in Calypso or something?

PR #494, opened 2016-08-18
Fixed the i18n task

DanReyLop:
Small fix. With the change on Calypso i18n structure (extracting it to a new module) the i18n task stopped working. This change fixes that.

To test, just run `npm run i18n` and check that it doesn't :boom: and it generates a valid `i18n/strings.php` file.

@robobot3000 @allendav @nabsul

### DIFF
--- a/tasks/i18n.js
+++ b/tasks/i18n.js
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-'use strict';
+'use strict'; /* eslint strict: 0 */
 
 /**
  * Module dependencies/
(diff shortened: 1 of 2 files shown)


allendav:
screen shot 2016-08-18 at 7 01 53 am

DanReyLop:
My bad, I didn't know we had that rule. Turns out, declaring variables with `let` is not allowed outside "strict mode", but everything that's compiled with Babel is considered "strict mode" by default, so we don't need to put it. Except for the tasks, which are plain Node.

I've added "use strict" to both the tasks (so the next dev that tries to use a strict-only feature doesn't panic). Ironically, I also removed the only usage of `let` in the i18n task because it's cleaner this way :)

robobot3000:
Ran `npm run i18n` and it generated a valid `i18n/strings.php` file. No lint warnings. I think it's ok to 🚢

PR #495, opened 2016-08-19
Package Settings - Assuming empty string if the value is undefined

### LINKED ISSUE #484: Warning: FormTextInput is changing an uncontrolled input of type text to be controlled
Occurs in the service instance settings when adding a new package and you start typing in the package name. This is likely due to this commit: https://github.com/Automattic/woocommerce-connect-client/commit/0f03bd3e48aa25b2915040fbca50b01ee5420a34

See also https://github.com/Automattic/woocommerce-connect-client/issues/371

robobot3000:
Fixes #484 

It's a bit of a tricky one and depends on the React/Redux practices. The React warning was caused by setting the `value` field to `undefined` on the form elements and then allowing the user to edit them. Now the setting to undefined/null happens all across the code in places such as reducer or preset packages. Rather than initialize default values there and set, for example, box weight to an empty string on the model, I assumed it's better to do it in the rendering step.

I'm happy to change it if necessary.

### DIFF
--- a/client/settings/views/packages/add-package.js
+++ b/client/settings/views/packages/add-package.js
@@ -121,7 +121,7 @@ const AddPackageDialog = ( props ) => {
 					id="name"
 					name="name"
 					placeholder={ __( 'The customer will see this during checkout' ) }
-					value={ name }
+					value={ name || '' }
 					onChange={ updateTextField }
 					isError={ modalErrors.name }
 				/>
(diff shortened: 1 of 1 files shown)


jeffstieler:
I'm indifferent to doing it this way, or initializing the `packages` state tree with empty values.

If anyone else has a strong opinion, they should voice it here, otherwise: :shipit:.

robobot3000:
Merged #495

PR #498, opened 2016-08-21
Display the "Create Label" meta-box only for qualifying orders

### LINKED ISSUE #432: [Labels] Only show the Create Label button when the order qualifies

DanReyLop:
Fixes #432

Hide the "Create Label" meta-box if we can't create a label for that order. The conditions are:
* The origin address (the WooCommerce "Base Location") must be in the US.
* The order shipping address must be in the US too.
* The order must contain at least one shippable product (if all the products are virtual/downloadable, a shipping label makes no sense).

To test, just create a bunch of orders and see if the "Create Label" meta-box is displayed. I've tried with an order with only downloadable products, an order for the UK, changing the WooCommerce Base Location to the UK, and in all those scenarios the meta-box must **not** appear.

@robobot3000 @allendav @nabsul @jeffstieler

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -132,6 +132,42 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 			return $result;
 		}
 
+		public function should_show_meta_box() {
+			global $post, $theorder;
+
+			if ( ! is_object( $post ) ) {
+				return FALSE;
+			}
+
+			if ( ! is_object( $theorder ) ) {
+				$theorder = wc_get_order( $post->ID );
+			}
+
+			if ( ! $theorder ) {
+				return FALSE;
+			}
+
+			// TODO: return TRUE if the order has already label meta-data
+
+			$base_location = wc_get_base_location();
+			if ( 'US' !== $base_location[ 'country' ] ) {
+				return FALSE;
+			}
+
+			$dest_address = $theorder->get_address( 'shipping' );
+			if ( $dest_address[ 'country' ] && 'US' !== $dest_address[ 'country' ] ) {
+				return FALSE;
+			}
+
+			foreach( $theorder->get_items() as $item ) {
+				$product = $theorder->get_product_from_item( $item );
+				if ( $product && $product->needs_shipping() ) {
+					return TRUE;
+				}
+			}
+			return FALSE;
+		}
+
 		public function meta_box( $post ) {
 			global $theorder;
 
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -133,47 +133,35 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		}
 
 		public function should_show_meta_box() {
-			global $post, $theorder;
+			$order = wc_get_order();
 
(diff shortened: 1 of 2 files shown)


classes/class-wc-connect-shipping-label.php
```diff
@@ -132,6 +132,42 @@ protected function get_states_map() {
...
+				return FALSE;
+			}
+
+			if ( ! is_object( $theorder ) ) {
+				$theorder = wc_get_order( $post->ID );
+			}
+
+			if ( ! $theorder ) {
+				return FALSE;
+			}
```
jeffstieler: Can we just use `wc_get_order()`, without any arguments, instead of checking global variables?

classes/class-wc-connect-shipping-label.php
```diff
@@ -132,6 +132,42 @@ protected function get_states_map() {
...
+
+			foreach( $theorder->get_items() as $item ) {
+				$product = $theorder->get_product_from_item( $item );
+				if ( $product && $product->needs_shipping() ) {
+					return TRUE;
+				}
+			}
+			return FALSE;
```
jeffstieler: Nitpick - WP core uses lowercase `true` and `false`.

classes/class-wc-connect-shipping-label.php
```diff
@@ -132,6 +132,42 @@ protected function get_states_map() {
...
+
+			// TODO: return TRUE if the order has already label meta-data
+
+			$base_location = wc_get_base_location();
+			if ( 'US' !== $base_location[ 'country' ] ) {
+				return FALSE;
+			}
+
+			$dest_address = $theorder->get_address( 'shipping' );
+			if ( $dest_address[ 'country' ] && 'US' !== $dest_address[ 'country' ] ) {
```
jeffstieler: `$dest_address[ 'country' ]` can result in a PHP Notice (undefined index).

Please use `isset()` here.

DanReyLop: Even in a newly created order, all the address fields are initialized with an empty string, so it will never be an `undefined index` (but it will be "falsey")

classes/class-wc-connect-shipping-label.php
```diff
@@ -132,13 +132,37 @@ protected function get_states_map() {
...
+
+			if ( ! $order ) {
+				return false;
+			}
+
+			// TODO: return true if the order has already label meta-data
 
-			if ( ! is_object( $theorder ) ) {
-				$theorder = wc_get_order( $post->ID );
+			$base_location = wc_get_base_location();
```
jeffstieler: Shouldn't we be using the origin set in the shipping method instance for this?

I realize that we don't currently have enough information to retrieve that based on the order..

DanReyLop: In the shipping method instance we only require the postcode. Plus, since labels and shipping rates are independent, it can be that no shipping methods are configured. The only "reliable" way of getting the warehouse's country is reading the `Base Location` of WooCommerce.

jeffstieler:
Code looks good and tests well. :shipit:

PR #514, opened 2016-08-26
Add REST controller for retrieving shipping rates from Connect.

### LINKED ISSUE #513: Add REST endpoint for rates retrieval
To be used by the Labels UI. This shouldn't require a cart, etc.

jeffstieler:
Fixes #513.

To test:
* Use a REST client to make a `POST` request to `/wc-api/v1/connect/shipping-rates`
* Sample data below

```javascript
{
    "origin": {
        "address": "2100 Park Ave",
        "address_2": "Unit 684212",
        "country": "US",
        "city": "Park City",
        "postcode": "84068",
        "state": "UT"
    },
    "destination": {
        "address": "1003 1st St",
        "address_2": "",
        "country": "US",
        "city": "Snohomish",
        "postcode": "98290",
        "state": "WA"
    },
    "carrier": "usps",
    "packages": [
        {
            "id": "box_1",
            "template": "medium_flat_box_top",
            "weight": 5
        },
        {
            "id": "box_2",
            "weight": 3,
            "length": 10,
            "width": 10,
            "height": 10
        }
    ]
}
```

### DIFF
--- a/classes/class-wc-rest-connect-shipping-rates-controller.php
+++ b/classes/class-wc-rest-connect-shipping-rates-controller.php
@@ -53,88 +53,19 @@ class WC_REST_Connect_Shipping_Rates_Controller extends WP_REST_Controller {
 	}
 
 	/**
-	 * Expected in request body:
-	 * array(
-	 *	'contents' => array(
-	 *		array(
-	 * 			'product_id' => 403,
-	 *			'height' => 10,
-	 *			'length' => 10,
-	 *			'width' => 10,
-	 *			'weight' => 10,
-	 *			'quantity' => 1,
-	 *		),
-	 *		...
-	 * 	),
-	 *	'origin' => array(
-	 *		'address' => '132 Hawthorne St',
-	 *		'address_2' => '',
-	 *		'city' => 'San Francisco',
-	 *		'state' => 'CA',
-	 *		'postcode' => '94107',
-	 *		'country' => 'US',
-	 *	),
-	 *	'destination' => array(
-	 *		'address' => '1550 Snow Creek Dr',
-	 *		'address_2' => '',
-	 *		'city' => 'Park City',
-	 *		'state' => 'UT',
-	 *		'postcode' => '84060',
-	 *		'country' => 'US',
-	 *	),
-	 * )
 	 *
-	 * @param WP_REST_Request $request
+	 * @param WP_REST_Request $request - See WC_Connect_API_Client::get_label_rates()
 	 * @return array
 	 */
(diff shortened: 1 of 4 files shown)


DanReyLop:
Minor thing: When you have the `variation_id` of a product, that can act as its own ID, the actual `product_id` is not needed. So we can use across all our code `product_id`, even when that means a concrete variation if it's a variable product.

Not-so-minor thing: This approach relies on having a USPS instance configured. For printing a label, that isn't enforced. Instead, I think a better approach would be to send the origin address (or at least, ZIP code) in the request, and having USPS return *all* the available rates.
In the labels UI we have the origin address, so that should be doable.

classes/class-wc-rest-connect-shipping-rates-controller.php
```diff
@@ -0,0 +1,148 @@
...
+
+			foreach ( (array) $instance->rates as $rate_idx => $rate ) {
+				$rate_id    = WC_Connect_Shipping_Method::format_rate_id( $instance->id, $instance->instance, $rate_idx );
+				$rate_label = WC_Connect_Shipping_Method::format_rate_title( $rate->title );
+
+				$processed_rates[] = array(
+					'id'         => $rate_id,
+					'label'      => $rate_label,
+					'cost'       => $rate->rate,
+					'method_id'  => 'usps',
```
DanReyLop: You can change this 'usps' to `$instance->id`. We are hardcoding stuff and that's not ideal, so at least let's minimize the impact :)

DanReyLop:
Works as expected, :shipit:

jeffstieler:
This needs to be changed once a new endpoint is built for retrieving rates for the purpose of label purchasing.

classes/class-wc-rest-connect-shipping-rates-controller.php
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 *
+	 * @param WP_REST_Request $request - See WC_Connect_API_Client::get_label_rates()
+	 * @return array
+	 */
+	public function update_items( $request ) {
+		$request_body = $request->get_body();
+		$payload      = json_decode( $request_body, true, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
+
+		$response = $this->api_client->get_label_rates( $payload );
```
DanReyLop: This call can potentially fail, you should add the usual `if ( is_wp_error( $response ) ) return new WP_Error(...)`

DanReyLop:
:shipit:

PR #520, opened 2016-08-29
Implemented the address steps, including address normalization

### LINKED ISSUE #481: Normalize origin and destination addresses in the labels UI
Replace the Hawk St business with calls to our address normalization endpoint on the server

DanReyLop:
Fixes #481 

All the logic is contained in the reducers (although a little bit spilled to the actions), so it's quite easy to tweak. This is the current logic:
* When the Labels dialog is opened, both addresses are immediately validated **iff**:
  * The address appears correct (no empty required fields)
  * The address is not already normalized
* The validation can have one of 3 outcomes:
  * The address normalized is exactly the same as the original address (unlikely). The address form is marked with a green checkmark.
  * The address is successfully normalized. The address form is marked with a warning sign, and inside it the "pick this or that address" component is shown, with the "pick normalized address" pre-selected.
  * The address could not be validated. The address form is marked with an error icon, and the label can't be purchased until it's fixed.
* When the user clicks in the Purchase button, if the addresses need validation, they will be validated first, and only if they are valid the actual purchase will be made.
* If the user edits any field on an address, that address automatically becomes "unvalidated".

Also, when an origin address is correctly validated, the normalized version is stored in the merchant host for future use.

The format to store the origin address in the merchant host has changed. If you get an error printed in the label meta-box, follow these steps:
* Go to `classes/class-wc-connect-service-settings-store.php`
* Before line 102 ( `$stored_address_fields = get_option( 'wc_connect_origin_address', array() );` ) add this line: `update_option( 'wc_connect_origin_address', array() );`
* Reload the `Edit Order` page
* Revert the change you just did

@allendav @robobot3000 @nabsul @jeffstieler 

@kellychoffman I think it's ready for you. Have fun! :)


allendav:
The card titles seem a bit off center (low) on Chrome on OSX:

screen shot 2016-08-29 at 2 09 19 pm

allendav:
Let's put the zip code last

screen shot 2016-08-29 at 2 14 58 pm

allendav:
@jeffstieler - is the name field actually needed for **origin** address for shipping labels? i looked at a few examples on the web and it seems rarely used by merchants for labels

classes/class-wc-rest-connect-address-validation-controller.php
```diff
@@ -0,0 +1,96 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( class_exists( 'WC_REST_Connect_Address_Validation_Controller' ) ) {
+	return;
+}
+
+class WC_REST_Connect_Address_Validation_Controller extends WP_REST_Controller {
```
allendav: Such a lovely long descriptive classname :P

classes/class-wc-rest-connect-address-validation-controller.php
```diff
@@ -0,0 +1,96 @@
...
+			'success' => true,
+			'normalized' => $response->normalized,
+		);
+	}
+
+	/**
+	 * Validate the requester's permissions
+	 */
+	public function update_items_permissions_check( $request ) {
+		return true; // non-authenticated service
```
allendav: Why should this be an open endpoint on the merchant's host? I think it should be limited to manage_woocommerce

client/shipping-label/state/actions.js
```diff
@@ -5,17 +5,31 @@ import flatten from 'lodash/flatten';
...
-
+import getFormErrors from 'shipping-label/state/selectors/errors';
+import { hasNonEmptyLeaves } from 'lib/utils/tree';
+import validateAddress from './validate-address';
 export const OPEN_PRINTING_FLOW = 'OPEN_PRINTING_FLOW';
 export const EXIT_PRINTING_FLOW = 'EXIT_PRINTING_FLOW';
 export const UPDATE_ADDRESS_VALUE = 'UPDATE_ADDRESS_VALUE';
+export const ADDRESS_VALIDATION_IN_PROGRESS = 'ADDRESS_VALIDATION_IN_PROGRESS';
+export const ADDRESS_VALIDATION_COMPLETED = 'ADDRESS_VALIDATION_COMPLETED';
+export const PICK_NORMALIZED_ADDRESS = 'PICK_NORMALIZED_ADDRESS';
```
allendav: Instead of `PICK_` could we use the more common `SELECT_` ?

client/shipping-label/state/actions.js
```diff
@@ -5,17 +5,31 @@ import flatten from 'lodash/flatten';
...
+export const EDIT_ORIGINAL_ADDRESS = 'EDIT_ORIGINAL_ADDRESS';
```
allendav: Instead of `ORIGINAL_` which could be confused with `ORIGIN` could we just say `EDIT_ADDRESS` ?

client/shipping-label/state/actions.js
```diff
@@ -5,17 +5,31 @@ import flatten from 'lodash/flatten';
...
 export const PURCHASE_LABEL_REQUEST = 'PURCHASE_LABEL_REQUEST';
 export const PURCHASE_LABEL_RESPONSE = 'PURCHASE_LABEL_RESPONSE';
 
-export const openPrintingFlow = () => {
-	return { type: OPEN_PRINTING_FLOW };
+export const openPrintingFlow = () => ( dispatch, getState, { storeOptions, addressValidationURL, nonce } ) => {
+	const { origin, destination } = getState().shippingLabel.form;
+	const errors = getFormErrors( getState(), storeOptions );
+	if ( ! hasNonEmptyLeaves( errors.origin ) && ! origin.isValidated && ! origin.validationInProgress ) {
+		validateAddress( dispatch, origin.values, 'origin', addressValidationURL, nonce ).catch( noop );
```
allendav: Why the `catch( noop )`  here and below?

classes/class-wc-rest-connect-address-validation-controller.php
allendav: But... (sorry)... could we be consistent with using validation or normalization?

client/shipping-label/state/actions.js
```diff
@@ -70,18 +99,29 @@ export const purchaseLabel = () => ( dispatch, getState, { callbackURL, nonce, s
...
-		payment_method_id: '123456789',
-		carrier: 'usps',
-		packages: form.packages.values.map( ( pckg, index ) => ( {
-			...omit( pckg, [ 'items', 'id' ] ),
-			service_id: form.rates.values[ index ],
-			products: flatten( pckg.items.map( ( item ) => fill( new Array( item.quantity ), item.product_id ) ) ),
-		} ) ),
-	};
+	let form = getState().shippingLabel.form;
+	const addressesValidation = [];
```
allendav: Could we name this as `addressValidationQueue` or something that indicates it is a collection of promises?

client/shipping-label/state/actions.js
```diff
@@ -70,18 +99,29 @@ export const purchaseLabel = () => ( dispatch, getState, { callbackURL, nonce, s
...
+	if ( ! form.destination.isValidated ) {
+		addressesValidation.push( validateAddress( dispatch, form.destination.values, 'destination', addressValidationURL, nonce ) );
+	}
+
+	Promise.all( addressesValidation ).then( () => {
+		form = getState().shippingLabel.form;
+		const formData = {
+			origin: form.origin.pickNormalized ? form.origin.normalized : form.origin.values,
+			destination: form.destination.pickNormalized ? form.destination.normalized : form.destination.values,
+			payment_method_id: '123456789',
```
allendav: Please fetch from `WC_Connect_Service_Settings_Store ::get_selected_payment_method_id`

client/shipping-label/state/actions.js
```diff
@@ -70,18 +99,29 @@ export const purchaseLabel = () => ( dispatch, getState, { callbackURL, nonce, s
...
+			carrier: 'usps',
```
allendav: Please add an issue to save the carrier (ID) for the customer selected real-time rate to order meta during checkout (or fetch it from order meta instead of hardcoding here if it is being saved already.)

client/shipping-label/state/actions.js
```diff
@@ -70,18 +99,29 @@ export const purchaseLabel = () => ( dispatch, getState, { callbackURL, nonce, s
...
+			packages: form.packages.values.map( ( pckg, index ) => ( {
+				...omit( pckg, [ 'items', 'id' ] ),
+				service_id: form.rates.values[ index ],
+				products: flatten( pckg.items.map( ( item ) => fill( new Array( item.quantity ), item.product_id ) ) ),
+			} ) ),
+		};
 
-	saveForm( setIsSaving, setSuccess, noop, setError, callbackURL, nonce, submitMethod, formData );
+		saveForm( setIsSaving, setSuccess, noop, setError, purchaseURL, nonce, 'POST', formData );
+	} ).catch( noop );
```
allendav: Why the `catch( noop )`?

client/shipping-label/views/dialog.js
```diff
@@ -75,7 +62,7 @@ const PrintLabelDialog = ( props ) => {
 						isDisabled: ! canPurchase,
 						onClick: props.labelActions.purchaseLabel,
 						isPrimary: true,
-						label: __( 'Purchase' ) + ( canPurchase ? ' (' + currencySymbol + getTotalCost() + ')' : '' ),
+						label: __( 'Print & Purchase' ) + ( canPurchase ? ' (' + currencySymbol + getTotalCost() + ')' : '' ),
```
allendav: @kellychoffman what do you think about `Buy and Print 1 Label ($x.xx)` instead?

client/shipping-label/views/steps/address/fields.js
```diff
@@ -3,8 +3,24 @@ import { translate as __ } from 'lib/mixins/i18n';
...
-const AddressFields = ( { values, allowChangeCountry, group, labelActions, storeOptions, errors } ) => {
+const AddressFields = ( { values, isValidated, normalized, pickNormalized, allowChangeCountry, group, labelActions, storeOptions, errors } ) => {
+	if ( isValidated && normalized && ! isEqual( normalized, values ) ) {
+		return (
+			<AddressSuggestion
+				values={ values }
+				normalized={ normalized }
+				pickNormalized={ pickNormalized }
+				pickNormalizedAddress={ ( pick ) => labelActions.pickNormalizedAddress( group, pick ) }
+				editOriginalAddress={ () => labelActions.editOriginalAddress( group ) }
```
allendav: Another spot where `editAddress` would potentially be less likely to be confused with origin

allendav:
> Also, when an origin address is correctly validated, the normalized version is stored in the merchant host for future use.

What exactly do I need to do to toggle from readonly to edit mode (or cancel edit mode) or trigger a validation? I did not see either in the code nor as i played with it

jeffstieler:
> is the name field actually needed for origin address for shipping labels? i looked at a few examples on the web and it seems rarely used by merchants for labels

@allendav `name` or `company` must be specified for label purchase. (Easypost)

DanReyLop:
> The card titles seem a bit off center (low) on Chrome on OSX:

Yep, I did absolutely no CSS fine-tuning, the weird thing would be if they lined up perfectly :D

> What exactly do I need to do to toggle from readonly to edit mode (or cancel edit mode) or trigger a validation? I did not see either in the code nor as i played with it

If you see the address form fields, then you are in edit mode, and if you change anything it will void any validation that was performed on that address. So, if you have a green checkmark on your `Origin Address` (meaning that's the normalized version) and you edit a field, then the checkmark will disappear and when you press the "Purchase" button it will have to be validated again.

I followed my gut on most of this logic. As I said, it's easy to change (reducers, actions, yada yada), I just made a few decisions to at least have a good starting point (maybe @kellychoffman can provide some UX insight on this).

client/shipping-label/views/steps/address/index.js
```diff
@@ -1,48 +1,79 @@
...
+import Card from 'shipping-label/views/card';
+import isEqual from 'lodash/isEqual';
+
+const renderSummary = ( { values, isValidated, normalized, pickNormalized, storeOptions, errors } ) => {
+	if ( hasNonEmptyLeaves( errors ) || ( isValidated && ! normalized ) ) {
+		return __( 'Invalid address' );
+	}
+	const { countriesData } = storeOptions;
+	const { city, postcode, state, country } = ( normalized && pickNormalized ) ? normalized : values;
+	// Summary format: "city, postcode state, country"
```
DanReyLop: So, ZIP at the end (I have no idea how a US address is formatted), and only put the country in the destination, **and** only if it's different of the origin country. Makes sense :)

client/shipping-label/state/actions.js
DanReyLop: In Chrome, an unhandled rejected Promise spits an error in the console, so if we want to use Promises for this I have to stick this ugly workaround. `node.js` follows the Promise spec by the book, and silently ignores any unhandled rejection, so this particular behaviour surprised me (and annoyed me).

I considered **not** using Promises for this, but converting this `Promise.all(validations)` to something with callbacks would be an uglier mess.

client/shipping-label/state/actions.js
DanReyLop: [Done](https://github.com/Automattic/woocommerce-connect-client/issues/525)

client/shipping-label/state/actions.js
DanReyLop: Done in https://github.com/Automattic/woocommerce-connect-client/pull/523

classes/class-wc-rest-connect-address-validation-controller.php
DanReyLop: Yep, I figured it would be useful during checkout for example, but that was before I added the `update_origin_address` call. Good catch :)

classes/class-wc-rest-connect-address-validation-controller.php
DanReyLop: Sure, I've changed everything to "normalization" now (will push it shortly)

DanReyLop:
I fixed all your comments and I rebased the branch.

kellychoffman:
> do we really need to display the origin country? In fact, for destination, I wouldn't display it there either unless it differed from the origin country.

Agree, I don't think we need it unless different from origin.

client/shipping-label/views/dialog.js
kellychoffman: > what do you think about Buy and Print 1 Label ($x.xx) instead?

That sounds good, lets try that.

kellychoffman:
When I manually create an order, fill out the shipping details, and then go to create a label, I'm getting an `undefined, undefined` warning message in the Destination Address step. 

Is the address in that step supposed to be prefilled with the address in the Shipping Details section on the Order page?

kellychoffman:
Marking this in-progress while I style.

kellychoffman:
I tried this out on another manually created order that had all the required origin address fields filled out except for the zip code field. As soon as I enter that in, the red error messages go away but I never get a green check.

DanReyLop:
> Is the address in that step supposed to be prefilled with the address in the Shipping Details section on the Order page?

It **is** supposed to. Did you press `Save Order` after editing the address in the Order Page?

> When I manually create an order, fill out the shipping details, and then go to create a label, I'm getting an undefined, undefined warning message in the Destination Address step.

Could you give me a screenshot, or steps to reproduce? I created a manual order without address, and in the destination step it shows "Invalid address" with all the required fields marked in red (as it's expected). I rebased the branch yesterday because there were conflicts with `master`. Maybe you should fetch it again and run `npm start` again, just in case funny stuff keeps happening.

> I tried this out on another manually created order that had all the required origin address fields filled out except for the zip code field. As soon as I enter that in, the red error messages go away but I never get a green check.

Yep, that's expected. At that point the address has not been proven to be invalid, but it hasn't been validated with the server, so it's in a "limbo" state. If you click the "Purchase" button, that will trigger the address validation call. Maybe it should be automatically called when the step is closed, or the address form loses focus?

kellychoffman:
> Yep, that's expected. At that point the address has not been proven to be invalid, but it hasn't been validated with the server, so it's in a "limbo" state. If you click the "Purchase" button, that will trigger the address validation call. Maybe it should be automatically called when the step is closed, or the address form loses focus?

The idea of the green checkmark is to visually show the merchant which sections are good to go so they feel confident in moving on to the section and/or purchasing. The validation needs to happen at the same time that the error message goes away and before the step is closed. We don't want the merchant to have to reopen the section to 'fix' something again, especially when they thought they were done. Ideally there shouldn't be any in between state: an action needs to be taken or it is good to go. I do think it would be nice to let the merchant know that _something_ is happening when the form is being validated. We should add [a spinner](https://wpcalypso.wordpress.com/devdocs/design/spinner) in the same place as the warning/error/success gridicon and in place of the card summary it should say `Validating address...`, then if its good, the green checkmark appears. If they need to chose, the warning appears and they pick the correct address, confirm, and the green checkbox shows up.

kellychoffman:
> It is supposed to. 

> Maybe you should fetch it again and run npm start again, just in case funny stuff keeps happening.

👍  Its working as expected now!

DanReyLop:
> The validation needs to happen at the same time that the error message goes away and before the step is closed.

That's going to be technically hard to implement. If the user is typing a street name, we don't want to spam the validation endpoint with a request after each keystroke, and if we wait until the textfield loses focus, that can be the moment when the merchant closes the step and goes to the next. But I agree this would be the ideal solution, UX-wise. I'll see what I can do :)

> I do think it would be nice to let the merchant know that something is happening when the form is being validated. We should add a spinner in the same place as the warning/error/success gridicon and in place of the card summary it should say Validating address...

Oh nice, I didn't know about the spinner component. Right now it's implemented (pporly) with a "sync" Gridicon, I'll change it now.

> If they need to chose, the warning appears and they pick the correct address, confirm, and the green checkbox shows up.

There's no "confirm" button (as of this moment), it's just two options. Maybe if the merchant opens the step and closes it again we can mark it as "green"?

I'm glad this PR is spinning up this kind of discussion, this is much better seeing this in action than talking about mock-ups :)

DanReyLop:
@kellychoffman Take a look at it now, the spinner, confirmation button and all the new logic is in place :)

Also for the rest of you, this introduced a reasonable amount of code, so you'll need to take another :eyes: to this PR (eventually).

kellychoffman:
> The card titles seem a bit off center (low) on Chrome on OSX:

Fixed in 5f24421

assets/stylesheets/shipping-label.scss
```diff
@@ -66,4 +65,96 @@
...
+	margin-bottom: 8px;
+}
+
+.address__city,
+.address__state,
+.address__postal-code {
+	width: 100%;
+
+	@include breakpoint( ">960px" ) {
+		width: 33%;
```
DanReyLop: A small nitpick. For some countries there isn't a concept of "State", so for those countries the `State` fieldset simply isn't rendered. So instead of hard-coding here a `33%`, could you use something more general? `display: flex` would work better I think.

kellychoffman: ah, forgot about that. 

`display: flex` applies to the parent element. i'll update the child elements tho. thanks!

kellychoffman: fixed! thanks again

kellychoffman:
Resolved my conflicts thanks to @DanReyLop and was able to finish styling. 🎉 

Just one UX comment. In the original design post, I had mentioned that any of the steps that contained an error or warning would be open by default. That way its one less click to print a label and its more immediately apparent what you have to do first. 

This is what I'm thinking: 

open first error or warning msg

Basically, the only change is to have the first or only error/warning step be expanded. If there is multiple, just the first one would be open. You would fix the address for example by clicking the confirm button, which would collapse that step and the next error step (if any) would automatically expand.

Otherwise, I think this is good to go from a design perspective. Really excited about how this is working and looking!

DanReyLop:
Done. Please check it out, I think the logic for this is a little convoluted, but it's still easy enough to follow.

kellychoffman:
👍  Looks good to 🚢 . 

Just one thing we forgot to bring over, but it can go in another PR. The light blue highlighted change in the Suggested address step.

nabsul:
Looks great.  :shipit:

PR #523, opened 2016-08-29
Use the real selected payment method ID in the Purchase Label request

### LINKED ISSUE #504: Use the selected payment method in the request to purchase a label
Depends on https://github.com/Automattic/woocommerce-connect-client/issues/471

Once the payment method selector is finished, actually use the selected payment method ID in the "purchase label" request to the WCC server.

DanReyLop:
Fixes #504

Use the real selected payment method ID in the Purchase Label request, instead of a hardcoded one.

There's no real way to test it, since the Purchase Label request doesn't actually work.

@nabsul @jeffstieler @robobot3000 @allendav


jeffstieler:
Looks good. :shipit:

Issue #505, opened 2016-08-24
[Labels] Packages step UI (readonly)


DanReyLop:
For the case where we already have a packing solution from checkout (i.e. readonly)

Weight, however, should be editable

kellychoffman:
Design: https://github.com/Automattic/woocommerce-connect-client/issues/511

PR #531, opened 2016-09-05
Packages step UI (readonly)

nabsul:
#505

### DIFF
--- a/client/shipping-label/views/steps/packages/list.js
+++ b/client/shipping-label/views/steps/packages/list.js
@@ -22,9 +22,7 @@ const getPackages = ( dimensionUnit ) => {
 	return result;
 };
 
-const OrderPackages = ( props ) => {
-	const { packages, dimensionUnit, weightUnit, errors } = props;
-	console.dir( props );
+const OrderPackages = ( { packages, updateWeight, dimensionUnit, weightUnit, errors } ) => {
 	const renderPackageInfo = ( pckg, pckgIndex ) => {
 		const pckgErrors = errors[ pckgIndex ] || {};
 		return (
@@ -44,7 +42,7 @@ const OrderPackages = ( props ) => {
 					id={ `weight_${pckgIndex}` }
 					title={ __( 'Total Weight' ) }
 					value={ pckg.weight }
-					updateValue={ () => {} }
+					updateValue={ ( value ) => updateWeight( pckgIndex, value ) }
 					error={ pckgErrors.weight }
 				/>
 
--- a/client/shipping-label/views/steps/packages/list.js
+++ b/client/shipping-label/views/steps/packages/list.js
@@ -13,9 +13,9 @@ const OrderPackages = ( { packages, updateWeight, dimensionUnit, weightUnit, err
 		return (
 			<div key={ itemIndex } className="wcc-package-item">
 				<div className="item-name">
-					<p>
+					<span className="item-name__title">
 						<a href={ item.url } target="_blank">{ item.name }</a>
-					</p>
+					</span>
 					{ item.attributes && <p>{ item.attributes }</p> }
 				</div>
 				<div className="item-quantity">{ item.quantity }</div>
@@ -27,13 +27,8 @@ const OrderPackages = ( { packages, updateWeight, dimensionUnit, weightUnit, err
 		const pckgErrors = errors[ pckgIndex ] || {};
 		return (
 			<div key={ pckgIndex }>
-				<p>
+				<div className="wcc-package-package-number">
 					{ sprintf( __( 'Package %d (of %d)' ), pckgIndex + 1, packages.length ) }
-				</p>
-
-				<div>
-					<FormLegend>{ __( 'Package dimensions' ) }</FormLegend>
-					<span>{ renderPackageDimensions( pckg, dimensionUnit ) }</span>
 				</div>
 
 				<div>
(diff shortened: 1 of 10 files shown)


kellychoffman:
Marked as in progress while I style.

kellychoffman:
Okay, @DanReyLop I've styled this up and have 3 requests :) 

1. Can the dimensions have a label? `Package dimension` should suffice.
2. Can the Package Contents table have a container around the whole thing? I wasn't sure how to add that. 
3. Lets add the confirmation button to this step :) I know it wasn't in the design, but when styling this I went through the UX about 20+ times. And when I get to this step after the address steps, I miss having a button that confirms the change and automatically collapses the card and brings you to the next step.

DanReyLop:
@kellychoffman Done. Dor the most part, you can assume that everything inside the `JSX` syntax behaves exactly like HTML (except when it doesn't) :smile:

kellychoffman:
Thanks @DanReyLop! Finished styling based on your changes and its up for review again. Loving the UX flow so far!

allendav:
Let's _not_ include the post ID in the item title. Instead, I suggest we only use the title and, if the merchant has set one, the SKU for the product (available from Edit Product, Inventory, SKU), e.g. instead of

screen shot 2016-09-07 at 2 30 38 pm

do this:

screen shot 2016-09-07 at 2 31 35 pm

OK to punt this to a new issue

allendav:
We're missing a space between the weight and the units:

screen shot 2016-09-07 at 2 34 10 pm

allendav:
The weight in the expander header doesn't account for the combined weight of all packages:

screen shot 2016-09-07 at 2 54 48 pm

DanReyLop:
@allendav If the merchant has an SKU for the product, it will appear in place of the post ID. This is how the product names are usually formatted in the Admin Dashboard, so I think it's better to leave it as-is.

About the combined weight... yeah, don't expect it to work. Putting the combined weight there has no value to the merchant (why should he care if all the packages sum 30lbs?). We need to figure out what to put there if there are multiple packages, and that's out of this sprint's scope. Unless @kellychoffman (or you) can come up with something quick.

allendav:
> @allendav If the merchant has an SKU for the product, it will appear in place of the post ID. This is how the product names are usually formatted in the Admin Dashboard, so I think it's better to leave it as-is.

Agreed then.

> About the combined weight... yeah, don't expect it to work. Putting the combined weight there has no value to the merchant (why should he care if all the packages sum 30lbs?). We need to figure out what to put there if there are multiple packages, and that's out of this sprint's scope. Unless @kellychoffman (or you) can come up with something quick.

Hmmmm. Agreed. Let's _not_ display the weight when there are multiple packages but display the number of packages instead (e.g. "2 packages")

client/shipping-label/views/steps/packages/list.js
```diff
@@ -1,44 +1,65 @@
...
 
 const OrderPackages = ( { packages, updateWeight, dimensionUnit, weightUnit, errors } ) => {
+	const renderItemInfo = ( item, itemIndex ) => {
+		return (
+			<div key={ itemIndex } className="wcc-package-item">
+				<div className="item-name">
+					<span className="item-name__title">
+						<a href={ item.url } target="_blank">{ item.name }</a>
+					</span>
+					{ item.attributes && <p>{ item.attributes }</p> }
```
allendav: Neat syntax

client/shipping-label/views/steps/packages/list.js
```diff
@@ -1,44 +1,65 @@
...
 
 const renderPackageDimensions = ( pckg, dimensionUnit ) => {
 	return `${pckg.length} ${dimensionUnit} x ${pckg.width} ${dimensionUnit} x ${pckg.height} ${dimensionUnit}`;
 };
 
 const OrderPackages = ( { packages, updateWeight, dimensionUnit, weightUnit, errors } ) => {
+	const renderItemInfo = ( item, itemIndex ) => {
+		return (
+			<div key={ itemIndex } className="wcc-package-item">
+				<div className="item-name">
```
allendav: I think this className is dangerously generic. Let's do something like wcc-package-item__name or something

client/shipping-label/views/steps/packages/list.js
```diff
@@ -1,44 +1,65 @@
...
+					<span className="item-name__title">
```
allendav: Not as dangerous as the classname above, but let's do something like wcc-package-item__title

client/shipping-label/views/steps/packages/list.js
```diff
@@ -1,44 +1,65 @@
...
+						<a href={ item.url } target="_blank">{ item.name }</a>
+					</span>
+					{ item.attributes && <p>{ item.attributes }</p> }
+				</div>
+				<div className="item-quantity">{ item.quantity }</div>
```
allendav: Same comment as above

client/shipping-label/views/steps/packages/list.js
```diff
@@ -1,44 +1,65 @@
...
-
-				<ul>
+			<div key={ pckgIndex }>
+				<div className="wcc-package-package-number">
+					{ sprintf( __( 'Package %d (of %d)' ), pckgIndex + 1, packages.length ) }
+				</div>
+
+				<div>
+					<div className="wcc-package-items-header">
+						<FormLegend className="item-name">{ __( 'Package contents' ) }</FormLegend>
```
allendav: Same comment as above for this line and the next

allendav:
@DanReyLop - a couple class names need fixing - otherwise good to go

allendav:
:shipit:

Issue #476, opened 2016-08-04
Share custom packages across shipping method instances AND labels


allendav:
Right now custom packages are defined only within the context of a shipping method instance, i.e.:

![screen shot 2016-08-04 at 10 40 42 am](https://cloud.githubusercontent.com/assets/1595739/17412117/eebbc014-5a2f-11e6-88e8-2d57fd7cafb2.png)

If you add another instance, you have to add packages again.

#449 will also add another consumer of custom packages - the labels ui

We should come up with a means of having a single source of custom package truth that can be used by instances, the labels ui and beyond, so that merchants are not potentially having to enter custom package information repeatedly.

This may be mostly behind-the-scenes work - with the exception that the labels ui would need to add in something like what shipping method instances already have.

kellychoffman:
> We should come up with a means of having a single source of custom package truth that can be used by instances, the labels ui and beyond, so that merchants are not potentially having to enter custom package information repeatedly.

Yes! I was struggling with whole idea while designing the Packages step in the shipping label creation flow. Sharing custom packages across methods and labels makes complete sense to me. 

What if you are creating a label for the first time and haven't saved a custom package yet, but while you are going through the flow, you realize how many times you'll probably be doing this with the same few packages. I'm just not sure how much it makes sense to do this from a users perspective. Do they want this in the label creation flow?

For the packages step, I think we have three options:

- Let users add a package, with the ability to save and reuse later (we'd be creating two places to do the same thing)
- Add a link off to centralized place to do so (flow could get messy)
- Only allow you to enter the weight of the package (leave the package creation out of the label ui)

allendav:
Adding @kellychoffman for design

PR #532, opened 2016-09-06
Add shared packages

robobot3000:
Adds #476

*Note*: you will need to update both the client and the server (PR 444) to make this work.

There's a new page under WooCommerce -> Settings -> "WooCommerce Connect" -> Packages that reuses the already existing UI:
screen shot 2016-09-06 at 15 27 48

The service settings pages link to it:
screen shot 2016-09-06 at 15 28 59

The server returns the packages schema as a part of the services schema. The schema used to be: `{shipping:[..]}` and now is: `{shipping:[..],boxes:{..}}`

The packages are sent to the server alongside the service settings, not embedded in them, i.e. the request payload was: `{services:[..],...}` and now is: `{services:[..],boxes:[..],...}`

CC @kellychoffman @allendav


kellychoffman:
I am getting a blank Packages page with `woocommerce-connect-client.js:56993 Uncaught TypeError: Cannot read property 'items' of null` js console error. I wasn't sure if this was because I don't have any packages defined, but we should have a zero packages state: 

screen shot 2016-09-06 at 2 58 40 pm

This is the php log:

`[06-Sep-2016 21:21:44 UTC] boxes_not_object Malformed service schemas. 'boxes' is not an object. (fetch_service_schemas_from_connect_server)`

robobot3000:
@kellychoffman - my bad, I should have mentioned that you need to checkout the `add/shared-packages` branches for both the client and the server for this to work.

classes/class-wc-connect-service-settings-store.php
```diff
@@ -199,7 +199,9 @@ public function get_enabled_services() {
 		 * @return object|array
 		 */
 		public function get_service_settings( $service_id, $service_instance = false ) {
-			return get_option( $this->get_service_settings_key( $service_id, $service_instance ), array() );
+			$settings = get_option( $this->get_service_settings_key( $service_id, $service_instance ), array() );
+
+			return $settings;
```
DanReyLop: No reason to make this change (I guess to put some debugging code here and forgot to change it back :smile: )

classes/class-wc-connect-settings-pages.php
```diff
@@ -148,8 +155,37 @@ public function output_account_screen() {
...
+			wp_enqueue_script( 'wc_connect_admin' );
+			wp_enqueue_style( 'wc_connect_admin' );
+
+			// hiding the save button because the react container has its own
+			global $hide_save_button;
+			$hide_save_button = true;
+
 			?>
+			<div class="wc-connect-admin-container" id="<?php echo esc_attr( $root_view ) ?>">
+			</div>
```
DanReyLop: To make this consistent with the other React containers, please add the `Having trouble loading this? ...` text.

classes/class-wc-rest-connect-packages-controller.php
```diff
@@ -0,0 +1,78 @@
...
+				'permission_callback' => array( $this, 'update_items_permissions_check' ),
+			),
+		) );
+	}
+
+	public function update_items( $request ) {
+		$request_body = $request->get_body();
+		$packages = json_decode( $request_body, true, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
+
+		$result = $this->service_settings_store->update_packages( $packages );
```
DanReyLop: This call can't fail, so the `is_wp_error` check is not needed.

client/packages/state/reducer.js
```diff
@@ -83,7 +113,27 @@ reducers[ TOGGLE_OUTER_DIMENSIONS ] = ( state ) => {
...
+
+reducers[ SET_IS_SAVING ] = ( state, action ) => {
+	return Object.assign( {}, state, {
+		isSaving: action.isSaving,
+		pristine: ! action.isSaving, //set pristine after the form has been saved
+	} );
+};
+
 const packages = ( state = {}, action ) => {
+	if ( 'undefined' === typeof state ) {
```
DanReyLop: This is never going to be true, because you put `{}` as default value for the `state` argument. I think what you are trying to do is this:
`const packages = ( state = initialState, action ) => {`

client/packages/views/index.js
```diff
@@ -0,0 +1,66 @@
...
+				</FormFieldset>
+			</CompactCard>
+			<CompactCard className="save-button-bar">
+				<ActionButtons buttons={ buttons } />
+			</CompactCard>
+		</div>
+	);
+};
+
+Packages.propTypes = {
```
DanReyLop: There are some `PropTypes` missing from here (form, noticeActions...)

client/settings/views/wcc-settings-form/settings-item.js
```diff
@@ -102,20 +100,15 @@ const SettingsItem = ( {
...
-					packages={ fieldValue }
-					presets={ schema.definitions.preset_boxes }
-					dimensionUnit={ storeOptions.dimension_unit }
-					removePackage={ removeArrayItem }
-					savePackage={ savePackage }
-					weightUnit={ storeOptions.weight_unit }
-					errors={ errors }
-					packageSchema={ fieldSchema.items }
-				/>
+				<div dangerouslySetInnerHTML={ { __html: packagesMsg } } />
```
DanReyLop: I'm not a fan of using `dangerouslySetInnerHTML` when there's not really needed. What about:
```
<div>
    { __( 'Manage your packages on' ) } <a href="/wp-admin...">{ __( 'the package settings page' ) }</a>
</div>
```

This would make it weirder for the translators, so I don't have a strong opinion about it. Your call :)

Also, I would add `target="_blank"` to the link.

tests/php/test_wc-connect-services-validator.php
```diff
@@ -34,7 +34,8 @@ private static function get_golden_services() {
 						)
 					)
 				)
-			)
+			),
+			'boxes' => (object) array()
```
DanReyLop: What about `new stdClass()` instead of this weird cast?

client/settings/views/wcc-settings-form/settings-item.js
DanReyLop: This link gives a `404` in my machine, because I have WordPress installed in a sub-folder (`localhost/wordpress/wp-admin...`). The link should be just `admin.php?page=wc-settings&tab=connect&section=packages`

DanReyLop:
Works nicely. Except for the link issue, feel free to ignore the rest of the comments, they are mostly nit-picks.

robobot3000:
Thanks Daniel. I'm new at php, so even nitpicking is appreciated :)
I think I addressed all of your notes.

DanReyLop:
:shipit:

kellychoffman:
I want to review before you ship.

kellychoffman:
Looks good. 4 requests:

- We previously discussed naming the page Packaging manager. Can we change it to that? 
- Lets add a 'Packaging' header to the left of the card so that it[ matches the layout](https://cloud.githubusercontent.com/assets/800604/18275325/b07c6cf0-7446-11e6-825b-ec6d35169d75.png) of the USPS shipping settings page.
- The 'save changes' button should always be active, even if there is nothing to save, just like on the USPS shipping settings page. 
- On the USPS shipping settings page the new section should have a label above it:

**Saved packages**
Add and edit saved packages using the [Packaging manager](#).

allendav:
Needs a rebase :)

robobot3000:
I rebased and applied Kelly's notes

DanReyLop:
Every time you squash, God kiils a :cat: 
Rebasing when you have multiple commits is a bit harder because of conflict resolution, but at the very least you should have commited the new changes (Kelly's notes) independently so they are easier to review. Just FYI for the next time :smile:

DanReyLop:
The `Save Changes` button should be always enabled, **except** when the form is being submitted, since this is the only visual indication that _something_ is happening when you click.

Once you fix that, it's ready to go.

robobot3000:
Merged #532

PR #542, opened 2016-09-07
Added an error to the "address" field if USPS couldn't validate an address

### LINKED ISSUE #530: Error message in labels modal easy to miss
When I first tried out this UI and entered an invalid address, I completely failed to notice the error message at the top. I assumed that something was crashing in the background and nothing was actually happening.

I think the error message needs to be:
- a little longer (like: the address doesn't appear valid, please correct it)
- Either place the message closer to the "use this address" button or attract the user's attention to the top of that box somehow. With my eyes focused below, I completely missed the message.

image

What do you think?
@DanReyLop @kellychoffman

DanReyLop:
Fixes #530

To test, just input an address with all the required fields, but a totally fake street name.

After you try to validate the address, the step should be marked as "Error", and the `Address` field should have an error that says "This address was not recognized".

@kellychoffman @allendav @nabsul @jeffstieler

### DIFF
--- a/assets/stylesheets/shipping-label.scss
+++ b/assets/stylesheets/shipping-label.scss
@@ -121,7 +121,7 @@
 			}
 
 			.highlight {
-				background-color: $blue-light;
+				background-color: lighten( $blue-light, 15 );
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


kellychoffman:
👍  love this.

robobot3000:
Works and looks good! :shipit:

Issue #482, opened 2016-08-09
Display tracking information in Labels Meta Box in Edit Order


allendav:
For orders where a label(s) has already been purchased, display tracking information for each purchased label in the order.

( Split out here from #390 )

kellychoffman:
Originally posted https://github.com/Automattic/woocommerce-connect-client/issues/391#issuecomment-228394141

Was thinking it would be nice to expose the tracking number in the customer notes metabox:
![](https://cloud.githubusercontent.com/assets/5835847/16343885/ac6b9588-3a07-11e6-9185-2f1da90dcc17.png)

allendav:
@robobot3000 - let's put this on hold for the moment until we hear from marketing on whether this feature should be included in the MVP

I'm moving it from Sprint 4 to Sprint 6

cc @warrendholmes @avivapinchas

PR #544, opened 2016-09-08
Allow the merchant to refund and re-print a label

### LINKED ISSUE #433: [Labels] Allow the merchant to refund a label
If a label for an order was already printed, and it's still on time to be refunded, show a button in the Edit Order page to receive a refund.

We should have a separate server endpoint for this, first to check if a refund is possible (and how much money would the merchant get back), and then to actually request the refund.

DanReyLop:
Fixes #433


allendav:
I got this working with some local server hacks. The label is coming out too big. I was expecting 4" x 6" but its bigger than letter size:

screen shot 2016-09-08 at 9 12 34 am

allendav:
I noticed the following on the reprint dialog:

> If there was a printing error when you purchased the label, you can print it again. NOTE: If you already used the label in a package, printing it and using it again will be considered a felony.

This seems a bit harsh. Maybe we should say "is a violation of our terms of service and may result in criminal charges." When we add a TOS page in woo-land, we can link to it.

cc @clickysteve

DanReyLop:
@allendav Yep, I just put there the first thing that came to my mind. There was no design for that dialog (even if it's **that** simple), so @kellychoffman will have to tweak it anyway.

About the **big** label, as discussed in Slack: after this moment all future labels will be requested in `PDF` format. Right now that PDF is `Letter`-sized with the label centered on the sheet. Sadly, that means that a label printer won't be able to print that. EasyPost is going to change the returned PDF to be a `4x6"` PDF, and that **can** be printed in a normal Letter printer, label printer, etc.

kellychoffman:
> There was no design for that dialog (even if it's that simple), so @kellychoffman will have to tweak it anyway.

Yeah, that was my bad. Was so focused on refunding, I forgot about reprinting 😬

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -72,10 +83,20 @@ public function update_items( $request ) {
 				);
 			}
 			$labels_data[] = $label_data->label;
+			$labels_order_meta[] = array(
+				'label_id' => $label_data->label->label_id,
+				'provider_label_id' => $label_data->label->provider_label_id,
+				'refundable_amount' => $label_data->label->refundable_amount,
+				'created' => $label_data->label->created,
+				'carrier_id' => $settings[ 'carrier' ],
+				'service_name' => $settings[ 'carrier' ], // TODO: Get the used service name from the request
```
DanReyLop: I will implement this after the `Rates Step` PR is merged, I will need to touch code from there. The only real consequence is that the `Ship via` item in the meta-box will simply display "usps" instead of "USPS Priority blah blah".

allendav:
The following linting warning needs fixing

```
/Applications/MAMP/htdocs/wp-content/plugins/woocommerce-connect-client/client/shipping-label/state/reducer.js
  198:48  warning  Unnecessarily quoted property 'label_id' found  quote-props
  230:48  warning  Unnecessarily quoted property 'label_id' found  quote-props
```

allendav:
And the following errors too

```
ERROR in ./client/shipping-label/views/refund/index.js

/Applications/MAMP/htdocs/wp-content/plugins/woocommerce-connect-client/client/shipping-label/views/refund/index.js
  1:17  error  'PropTypes' is defined but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)


ERROR in ./client/shipping-label/views/reprint/index.js

/Applications/MAMP/htdocs/wp-content/plugins/woocommerce-connect-client/client/shipping-label/views/reprint/index.js
  1:17  error  'PropTypes' is defined but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)


ERROR in ./client/lib/utils/format-date.js

/Applications/MAMP/htdocs/wp-content/plugins/woocommerce-connect-client/client/lib/utils/format-date.js
  11:2  error  Missing semicolon  semi

✖ 1 problem (1 error, 0 warnings)
```

allendav:
@DanReyLop - the label is still oversize, but if you'd like we can fix that in a separate issue

screen shot 2016-09-08 at 2 51 26 pm

allendav:
The tracking number is hardcoded - this is a separate issue #482

DanReyLop:
@allendav My bad, I didn't check the `eslint` errors. I'll fix them and rebase.

> the label is still oversize, but if you'd like we can fix that in a separate issue

Once a label is purchased, the image is cached. To get a "good" label PDF, you'll need to make a new label purchase (with the most recent version of `master` in the server).

> The tracking number is hardcoded

No it's not :smile: 
Looks like the EasyPost test API always gives "8888888888888" as the tracking number, but it's not hardcoded on our side.

DanReyLop:
I've rebased, fixed a few things I encountered along the way (related with the rebase), implemented the `Shiped via: USPS Priority Mail Blah Blah` and fixed the `eslint` errors. Apart from the design, I would say **now** this is ready for review.

allendav:
>> the label is still oversize, but if you'd like we can fix that in a separate issue

> Once a label is purchased, the image is cached. To get a "good" label PDF, you'll need to make a new label purchase (with the most recent version of master in the server).

I did a completely new purchase and I still get a way too big label

screen shot 2016-09-08 at 5 37 27 pm

allendav:
I also noticed that when I look at an order that has had a label printed, I get this in my php error log

```
[09-Sep-2016 00:38:12 UTC] PHP Warning:  http_build_query(): Parameter 1 expected to be Array or Object.  Incorrect value given in /Applications/MAMP/htdocs/wp-includes/Requests/Transport/cURL.php on line 507
[09-Sep-2016 00:38:12 UTC] PHP Warning:  Cannot modify header information - headers already sent by (output started at /Applications/MAMP/htdocs/wp-includes/Requests/Transport/cURL.php:507) in /Applications/MAMP/htdocs/wp-includes/rest-api/class-wp-rest-server.php on line 1196
```

and this on my screen:

screen shot 2016-09-08 at 5 38 27 pm

allendav:
This is logged in Javascript console for orders with labels already printed:

```
woocommerce-connect-client.js:14660 Warning: Failed propType: Invalid prop `created` of type `string` supplied to `RefundDialog`, expected `number`. Check the render method of `ShippingLabelRootView`.
```

allendav:
The following is logged when the print dialog opens on my computer:

```
print-document.js:35 Resource interpreted as Document but transferred with MIME type image/png: "blob:http://localhost:8888/3e923bf3-e40d-4a6a-b761-7a42508fc670".
```

DanReyLop:
```
woocommerce-connect-client.js:14660 Warning: Failed propType: Invalid prop `created` of type `string` supplied to `RefundDialog`, expected `number`. Check the render method of `ShippingLabelRootView`.
```
Before, `created` was a timestamp (integer), now it's a string. I've changed the client code to fix it.

```
print-document.js:35 Resource interpreted as Document but transferred with MIME type image/png: "blob:http://localhost:8888/3e923bf3-e40d-4a6a-b761-7a42508fc670".
```
I'm using `URL.createObjectURL( Blob )` to load the image data into the `iframe`. I'm following more or less [this example](https://developer.mozilla.org/en-US/docs/Web/API/GlobalFetch/fetch#Example). It works, but it prints that warning (I don't know why). Anyway, it's just a "notice", not even a warning, I think we can live with it.

> I did a completely new purchase and I still get a way too big label

You're still geting a `PNG` label. You need to update your WCC server code so you have https://github.com/Automattic/woocommerce-connect-server/pull/443 . It was merged a couple of days ago.

allendav:
OK, updated everything and now the label fits on the page, but its full size - I think it should be 4" x 6" but let's tackle this in a subsequent issue so we can wrap this PR up.

I do, however, still get this in the php error logs which I think needs to be fixed before this PR can be committed:

```
[09-Sep-2016 20:50:50 UTC] PHP Warning:  http_build_query(): Parameter 1 expected to be Array or Object.  Incorrect value given in /Applications/MAMP/htdocs/wp-includes/Requests/Transport/cURL.php on line 507
[09-Sep-2016 20:50:50 UTC] PHP Warning:  Cannot modify header information - headers already sent by (output started at /Applications/MAMP/htdocs/wp-includes/Requests/Transport/cURL.php:507) in /Applications/MAMP/htdocs/wp-includes/rest-api/class-wp-rest-server.php on line 1196
```

and I suspect this is why we see this on the page after completing printing:

screen shot 2016-09-09 at 1 51 02 pm

DanReyLop:
Yep, this log and the `Unexpected token error` seem to be related.

About the PDF. I've tried on the Mac native print dialog:
mac_print
The radio-buttons say (roughly): `Adjust scale` (the good one) and `Adjust dimensions` (the one selected by default). You need to select `Adjust scale` and put `100%` for the label to be printed correctly. It's a bit convoluted, but at least it's possible (it **wasn't** possible when the label was a PNG).

Something similar happens in the Chrome print dialog, although it's simpler:
chrome_print
Just unchecking the `Adjust to page` option and you're good to go.

I'll play with a PDF handling library during the weekend, I hope I'll be able to offer a 1-click solution to this mess, I think it will come really handy when we implement multi-label printing.

allendav:
@DanReyLop - ok - let's move forward with this, especially since it looks like the error is caused by WordPress and not our code.

Ty!

:shipit:

allendav:
OK, as we chatted in slack, let's apply the design from the mocks for the refund and for the already-purchased-a-label meta box content.  Please also make the reprint text gentler :P

DanReyLop:
@allendav I've applied the design and made the text gentler as you suggested.

allendav:
Styles and text are much improved, ty! :shipit:

PR #555, opened 2016-09-16
Fix shipping rates without package data

jeffstieler:
This fixes an issue @kellychoffman had today testing the labels flow using a manually created order.

It also addresses the use case where an order does *not* have a rate from WCC during checkout.

To test:
* Ensure that the label flow works for orders made from the storefront
* Create a new order from the admin (or use one already existing)
* Ensure that the label flow still works for that "manual" order

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -170,7 +170,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 			$form_data = compact( 'is_packed', 'packages', 'origin', 'destination' );
 
 			$form_data[ 'rates' ] = array(
-				'selected'  => $selected_rates,
+				'selected'  => (object) $selected_rates,
 			);
 
 			$form_data[ 'order_id' ] = $order->id;
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -39,7 +39,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 				$name = html_entity_decode( $product->get_formatted_name() );
 
 				for ( $i = 0; $i < $item[ 'qty' ]; $i++ ) {
-					$packages[] = array(
+					$packages[ "weight_{$i}_individual" ] = array(
 						'height' => ( float ) $height,
 						'length' => ( float ) $length,
 						'weight' => ( float ) $weight,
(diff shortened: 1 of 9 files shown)


kellychoffman:
👍 

AMAZING. First time I've gotten through the entire flow and its all thanks to you. 👏 👏

allendav:
@jeffstieler - i did checkout with free shipping on the front side... i then went into edit order, purchase label, normalized the destination address and selected priority mail from the rates, but when I click on the (enabled) Buy and Print label button, nothing happens. No network requests or anything.

allendav:
@jeffstieler - i tracked it down further.  There is an exception occurring in shipping-label/state/actions.js purchaseLabel.  If you replace the noop in the catch with console logging you'll see it.

```
woocommerce-connect-client.js:61012 Uncaught (in promise) TypeError: form.packages.values.map is not a function(…)
```

classes/class-wc-connect-shipping-label.php
```diff
@@ -39,7 +39,7 @@ protected function get_items_as_individual_packages( $order ) {
 				$name = html_entity_decode( $product->get_formatted_name() );
 
 				for ( $i = 0; $i < $item[ 'qty' ]; $i++ ) {
-					$packages[] = array(
+					$packages[ "weight_{$i}_individual" ] = array(
```
DanReyLop: The `{}` are not needed. And if you put them (for clarity), the correct way is `"weight_${i}_individual"`

jeffstieler: You can do either, I'm not sure one way is more correct than the other.

See "Complex (curly) syntax" example here: https://secure.php.net/manual/en/language.types.string.php#language.types.string.syntax.double

DanReyLop: Oh well, PHP being PHP... sorry about that :D

client/shipping-label/state/reducer.js
```diff
@@ -130,11 +130,10 @@ reducers[ CONFIRM_ADDRESS_SUGGESTION ] = ( state, { group } ) => {
...
 
-reducers[ UPDATE_PACKAGE_WEIGHT ] = ( state, { packageIndex, value } ) => {
-	const newPackages = [ ...state.form.packages.values ];
-	newPackages[ packageIndex ] = { ...newPackages[ packageIndex ],
-		weight: parseFloat( value ),
-	};
+reducers[ UPDATE_PACKAGE_WEIGHT ] = ( state, { packageId, value } ) => {
+	const newPackages = { ...state.form.packages.values };
+
+	newPackages[ packageId ].weight = parseFloat( value );
```
DanReyLop: Remember that `...whatever` only does a shallow clone. You need to clone `newPackages[ packageId ]` if you're going to edit the weight (like how it was before).

client/shipping-label/views/purchase/steps/packages/index.js
```diff
@@ -7,15 +7,18 @@ import { sprintf } from 'sprintf-js';
 import StepContainer from '../../step-container';
 
 const PackagesStep = ( { values, storeOptions, labelActions, errors, expanded } ) => {
-	const isValid = 0 < values[ 0 ].weight;
+	const packageIds = Object.keys( values );
+	const firstPackageId = packageIds[ 0 ];
+	const firstPackage = values[ firstPackageId ];
+	const isValid = 0 < firstPackage.weight;
```
DanReyLop: You should check here if _any_ weight is invalid. I know, my bad, I implemented this supporting only 1 package, but since now it supports multiple, better to do it well.

client/shipping-label/views/purchase/steps/packages/list.js
```diff
@@ -44,7 +48,7 @@ const OrderPackages = ( { packages, updateWeight, dimensionUnit, weightUnit, err
 					className="wcc-package-weight"
```
DanReyLop: Better to change the ID of this field from `weight_${pckgIndex}` to `weight_${pckgId}`

allendav:
Thank you @jeffstieler ! - I am able to buy labels for orders that have both free shipping and orders that have a rate selected at checkout. If @DanReyLop is cool with the changes, I say :shipit:

DanReyLop:
:shipit:

PR #564, opened 2016-09-22
Only add body data to POST and PUT requests.

jeffstieler:
None of the server endpoints are looking for bodies on GET requests, and it isn’t technically correct to send them.

Before this change, all GET requests made by the client throw: `PHP Warning:  http_build_query(): Parameter 1 expected to be Array or Object.  Incorrect value given in wp-includes/Requests/Transport/cURL.php on line 507`. This has been occurring since the first GET request was added.

Correcting this behavior has the side effect of fixing the JSON error that occurs during label status retrieval. See below:

screen-shot-2016-09-19-at-3-01-10-pm


robobot3000:
Looks good :shipit:

PR #565, opened 2016-09-23
Packages - fixed the outer dimensions field remaining open after dismissing the modal

robobot3000:
During the Connectathon someone found this bug on the packages settings page:

- open the package dialog (either for edit or add new)
- show the outer dimensions field
- close the dialog by clicking the cancel button
- reopen the dialog as described in 1.
- the outer dimensions field will remain open, and there is no way to hide it

This PR fixes it.


robobot3000:
Merged #565

PR #594, opened 2016-09-29
Remove .catch(noop) to make errors in Promises be logged in the console

### LINKED ISSUE #556: Replace catch( noop ) with at least console logging in shipping-label/state/actions.js
In three places (getRates, submitAddressForNormalization and purchaseLabel) we have catch( noop ) which silently consumes exceptions (like what was happening here: https://github.com/Automattic/woocommerce-connect-client/pull/555#issuecomment-247690790 )

I think we should at least be console logging the exception. Otherwise debugging and happiness will have nothing to go on other than "I clicked the (enabled) button and nothing happens"

DanReyLop:
Fixes #556

The default behaviour (in Chrome) when a Promise is rejected and there is no `.catch` handler is to log the error in the console. So by removing the `.catch(noop)` the errors will be logged.

@allendav @nabsul @robobot3000 @jeffstieler

### DIFF
--- a/classes/class-wc-rest-connect-address-normalization-controller.php
+++ b/classes/class-wc-rest-connect-address-normalization-controller.php
@@ -65,6 +65,14 @@ class WC_REST_Connect_Address_Normalization_Controller extends WP_REST_Controlle
 		);
 		$response = $this->api_client->send_address_normalization_request( $body );
 
+		if ( is_wp_error( $response ) ) {
+			return new WP_Error(
+				$response->get_error_code(),
+				$response->get_error_message(),
+				array( 'message' => $response->get_error_message() )
+			);
+		}
+
 		if ( isset( $response->error ) ) {
 			return new WP_Error(
 				$response->error->code,
--- a/classes/class-wc-rest-connect-address-normalization-controller.php
+++ b/classes/class-wc-rest-connect-address-normalization-controller.php
@@ -34,9 +34,15 @@ class WC_REST_Connect_Address_Normalization_Controller extends WP_REST_Controlle
 	 */
 	protected $settings_store;
 
-	public function __construct( WC_Connect_API_Client $api_client, WC_Connect_Service_Settings_Store $settings_store ) {
+	/**
+	 * @var WC_Connect_Logger
+	 */
+	protected $logger;
+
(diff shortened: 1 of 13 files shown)


allendav:
Hmmm. I gave this issue kinda a bad title. It be better to replace the catch noop with a catch that presents an error in the form (and disables submittal or something) - that way the end user can report that error to us.

A lot of merchants will not know how / think to look in Dev Console and will just think the form is broken.

We should also log a detailed error in dev console for site devs and tech-savvy merchants.

DanReyLop:
Ok, so I made a few changes:
* Added error logging to all the REST endpoints. If the endpoint is going to return a `WP_Error`, then that error will be logged.
* Changed the `normalizeAddress` Promise slightly. Now, if the server returns an error, the Promise will **resolve** with a value of `false`, instead of being rejected. So now we can handle success (promise resolved), validation error (promise resolved with `false`), and catastrophic error (promise rejected).
* Every time a Promise is rejected, the error will be logged in the console and an Error Notice will appear with the errror. Once we solve the issue of "make the notices be *on top* of the dialog", this should be enough.

classes/class-wc-rest-connect-address-normalization-controller.php
```diff
@@ -65,12 +71,24 @@ public function update_items( $request ) {
...
+			return $error;
+		}
+
 		if ( isset( $response->error ) ) {
-			return new WP_Error(
+			$error = new WP_Error(
 				$response->error->code,
 				$response->error->message,
 				array( 'message' => $response->error->message )
 			);
```
jeffstieler: You could use [`WP_Error::add_data()`](https://developer.wordpress.org/reference/classes/wp_error/add_data/) here instead of creating a new `WP_Error`.

jeffstieler:
@DanReyLop can you list out your testing steps for this PR? Is there an easy way to get all the requests to trigger the error condition? (perhaps turning off a local server?)

DanReyLop:
@jeffstieler Yes, turning off the local server before a request is the easiest way to test this. Note that for some cases (such as purchasing a label) you need to have the server running to go through all the process, and turn it off before hitting the `Purchase` button.

jeffstieler:
I'm not crazy about our message for timeouts ("Error: The WooCommerce Connect server returned ( 0 ) and an empty response body."), but this looks good.

PR #597, opened 2016-10-01
Change the obscure "Syntax Error <" notice to something more readable when the JSON response is malformed

### LINKED ISSUE #554: Connect-a-thon: Gracefully handle 500s from the purchase label endpoint
Right now you get this below the label dialog, if the purchase fails:

screen shot 2016-09-13 at 11 52 24 am

If you dismiss the dialog, the error can be seen but is not user friendly

screen shot 2016-09-13 at 11 54 46 am

Adding logging to class-wc-rest-connect-shipping-label-controller.php and dumping the $response from $this->api_client->send_shipping_label_request( $settings ); reveals:

```
09-13-2016 @ 12:08:06 - WP_Error Object
(
    [errors] => Array
        (
            [wcc_server_error_response] => Array
                (
                    [0] => Error: The WooCommerce Connect server returned: Internal Server Error An internal server error occurred ( 500 )
                )

        )

    [error_data] => Array
        (
        )

)
 (shipping-label-controller)
```

DanReyLop:
Fixes #554 

When the result of an AJAX call fails in the PHP side, the returned error page won't be valid JSON, so the error shown to the user was the quite obscure `Syntax error: < ...`. This can happen even if PHP prints a `NOTICE` before actually printing the JSON.

I wrapped all the places where the result of an AJAX request is parsed and if it fails, show a more legible generic error (and log the whole thing in the console).

@allendav @nabsul @robobot3000 @jeffstieler

### DIFF
--- a/client/lib/utils/parse-json.js
+++ b/client/lib/utils/parse-json.js
@@ -1,0 +1,25 @@
+import { translate as __ } from 'lib/mixins/i18n';
+
+/**
+ * Parses a server response into a JSON object, providing a human-readable error if the
+ * JSON is invalid.
+ *
+ * @param {Response} response Response object, as obtained from a fetch call.
+ * @returns {Promise} Resolves with the parsed JSON object,
+ * or rejects with a human-readable error if the payload is not valid JSON
+ */
+export default ( response ) => {
+	if ( ! response instanceof Response ) {
+		console.error( 'Invalid Response object' );
+		return Promise.reject( __( 'Unexpected server error.' ) );
+	}
+	return response.text().then( ( text ) => {
+		try {
+			return JSON.parse( text );
+		} catch ( error ) {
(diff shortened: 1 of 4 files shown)


client/lib/utils/parse-json.js
```diff
@@ -0,0 +1,25 @@
...
+		console.error( 'Invalid Response object' );
+		return Promise.reject( __( 'Unexpected server error.' ) );
+	}
+	return response.text().then( ( text ) => {
+		try {
+			return JSON.parse( text );
+		} catch ( error ) {
+			console.error( error );
+			console.error( text );
+			throw __( 'Unexpected server error.' );
```
DanReyLop: I had writer's block here. Any other suggestions for a "Generic Error Message"?

Issue #590, opened 2016-09-28
Test with node 6.7.0


allendav:
Calypso is advancing to 6.7.0 soon. Let's make sure we aren't going to asplode.

Ref: https://github.com/Automattic/wp-calypso/pull/8297

Adding this to Sprint 6 by fiat :P 

cc @gwwar

allendav:
Was able to install modules (after blowing them away), build and run client just fine with 6.7.0

No adverse effect

PR #608, opened 2016-10-10
Update NodeJS version on TravisCI to 6.7.0

DanReyLop:
As of #590, we tested the client using nodejs `6.7.0`, but we didn't update the version running in our CI. This PR bumps that version.

This doesn't need to be tested. If the TravisCI passes, it's good to go.

@nabsul @jeffstieler @robobot3000 @allendav


nabsul:
Travis CI seems to be failing....

DanReyLop:
So, here is a recap:

CI was failing because `node-sass` is compatible with node 6 only from version `3.7.0` and up. The version we had in our `shrinkwrap` was `3.4.2`. We were locking a version of `wp-calypso` from mid-June, before *they* moved to node `6.7.0`. As a quick fix, I edited manually the shrinkwrap, and it works (it failed in one machine, but that's probably Travis fault).

Then, I fixed `shrinkwrap`, maybe once and for all. I made a new `npm` task: `npm run update-deps`. Look at what it does, it takes a while. Maybe it's overkill but I'm not taking any chances.

Hopefully we won't need to nuke and regenerate the `shrinkwrap` file any more. According to the NPM documentation, using `npm ... --save` to install/uninstall/update dependencies (as oposed to manually editing `package.json`) will update both `package.json` **and** `npm-shrinkwrap.json`. This was buggy until a while ago, so I recommend installing the latest version of NPM (`npm install npm --global`).

package.json
```diff
@@ -8,7 +8,8 @@
     "test": "karma start --single-run --no-auto-watch && npm run eslint",
     "test-watch": "karma start --auto-watch --no-single-run",
     "release": "node tasks/release.js",
-    "i18n": "npm run dist && node tasks/i18n"
+    "i18n": "npm run dist && node tasks/i18n",
+	"update-deps": "rm npm-shrinkwrap.json && rm -rf node_modules && npm install --no-optional && npm install --no-optional && npm prune && npm shrinkwrap --dev"
```
DanReyLop: Yeah, this is that `wp-calypso` does. I really don't want to learn why do they have to `npm install` twice, I prefer to live in ignorance.

Note that this is only needed to regenerate the `shrinkwrap` file, you **shouldn't** need to do this almost ever.

client/config.js
```diff
@@ -0,0 +1,16 @@
...
+
+export default ( key ) => {
+	if ( key in data ) {
+		return data[ key ];
+	}
+	throw new Error( 'config key `' + key + '` does not exist' );
+};
+
+export const isEnabled = () => false;
+export const anyEnabled = () => false;
```
jeffstieler: Can you explain the need for this file? Which Calypso components (that we're using) require this?

PR #610, opened 2016-10-11
Label tracking code comes back in its own property now, not “provider label id”.

jeffstieler:
To test:
* Purchase a label via Edit Order screen, or go to an order that already has a label
* Verify that the tracking code still appears and is linked


DanReyLop:
When purchasing a label, the result label data is stored in the order (see class `WC_REST_Connect_Shipping_Label_Controller`). You will need to change there `prodiver_label_id` => `tracking` too.

jeffstieler:
@DanReyLop I thought I caught that already? See: https://github.com/Automattic/woocommerce-connect-client/pull/610/files#diff-2de8d7a14a4e916c44429278e661a140R106

DanReyLop:
Yes, sorry about that, I got the names of the classes mixed up. :shipit:

PR #611, opened 2016-10-11
Make a real refund request from the Edit Order screen.

jeffstieler:


classes/class-wc-rest-connect-shipping-label-refund-controller.php
```diff
@@ -59,22 +59,20 @@ public function register_routes() {
...
-			'label_id' => ( int ) $request[ 'label_id' ],
-			'refunded_time' => time() * 1000,
-		);
+		$response = $this->api_client->send_shipping_label_refund_request( $request[ 'label_id' ] );
 
-		if ( is_wp_error( $response ) ) {
-			$error = new WP_Error(
-				$response->get_error_code(),
-				$response->get_error_message(),
-				array( 'message' => $response->get_error_message() )
```
DanReyLop: You need to keep this line, or else the client 💥 when receiving this error.

jeffstieler: It still happens on line 72.

PR #619, opened 2016-10-14
Naming: Find 'WooCommerce Connect', replace with 'Connect for WooCommerce'

kellychoffman:
Fixes: https://github.com/Automattic/woocommerce-connect-client/issues/592 in which we update this to its rightful name!

Server change: https://github.com/Automattic/woocommerce-connect-server/pull/558


allendav:
:shipit:

PR #623, opened 2016-10-17
The destination country for a manual order defaults to the country of origin instead of being blank

DanReyLop:
With some recent refactor, when you create an order from scratch, the destination address is completely empty. That means even the country, which defaults to the first one alphabetically (Aland Islands), and the country selector is disabled. Since we don't support shipping to wherever that is, this is a bug :)

To test:
* Create an empty order with a shippable item.
* Go to the Print Label Dialog.
* Notice that the (disabled) destination country selector has `United States (US)` selected.

@nabsul @allendav @jeffstieler @robobot3000


DanReyLop:
We don't allow international shipping yet. The `allowChangeCountry` forced to false is on purpose, we don't want someone changing the destination country to outside of the US and not being able to fill a Customs form.

PR #624, opened 2016-10-17
Add build support for requiring the whole lodash library

DanReyLop:
Using [this Babel plugin](https://github.com/lodash/babel-plugin-lodash) means we can now import the whole `lodash` library wherever we need it without making the bundle bigger.

We can use this syntax:
```
import _ from 'lodash';
_.map(_.omit(...))
```

Or this one:
```
import { map, omit } from 'lodash';
map(omit(...));
```

To test that this change doesn't affect our bundle size, I actually changed all our lodash usages with the `import _ from 'lodash'` import. Yes, I got carried away, but it was just half an hour, or even less :)

Our bundle is now 60KB smaller (from 695KB to 634KB), altough maybe that has something to do with using a more recent version of lodash.

I also changed all our dependencies to use the `^X.Y.Z` notation. Most of them were like that but we weren't being consistent about it. Anyway, I didn't modify the versions in the `shrinkwrap` file, so this change has no immediate effect.

@allendav @jeffstieler @nabsul @robobot3000


allendav:
Tests nicely. No collateral from what I could see. 633K JS for me as well. :shipit:

PR #625, opened 2016-10-18
Generate preview and printing label PDFs in the client

DanReyLop:
This PR includes:
* PDF generation client-side using the `pdfkit` library.
* UI changes needed (adding a `Paper size` selector to the Labels UI, and remembering that selection).
* Completely generate the PDFs client-side, both for Preview, Purchase and Reprint labels.
* "Smarts" built-in: automatically put 2 labels in 1 page if they fit.

This doesn't work in IE/Edge, and there is no way to make it work. Hopefully I'll be able to reuse most of the code for re-implementing this server-side.

Note: You'll need also the `try/pdf-manipulation` server branch to try this.

### DIFF
--- a/assets/stylesheets/shipping-label.scss
+++ b/assets/stylesheets/shipping-label.scss
@@ -86,7 +86,7 @@
 		justify-content:center;
 		align-items:center;
 
-		img {
+		iframe {
 			max-height: 100%;
 		}
 	}
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -103,6 +103,14 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 			return array_merge( $wc_address_fields, $stored_address_fields );
 		}
 
+		public function get_preferred_paper_size() {
+			return get_option( 'wc_connect_paper_size', '' );
+		}
+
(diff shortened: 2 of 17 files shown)


DanReyLop:
@allendav @robobot3000 @nabsul @jeffstieler 

I'm ok with you folks taking a small look, or ignoring this completely. Most of that code is going to be reused in the final solution (the actual PDF-handling will be in the server), so that PR is only useful for curiosity. I was genuinely surprised of all that can be accomplished client-side only (except for IE, damn IE).

DanReyLop:
For the record, the reason why this doesn't work on IE/Edge: The PDF generation works fine, and it produces a `Blob`. A `Blob` is a representation of arbitrary data, kinda like a `Buffer`. From a `Blob`, you can obtain an special URL pointing to that `Blob`, something like `blob:localhost/asdfafafdaasdf-asdfads-asdf-asdffd`. You can then use that URL as a `src` of an ``, `<iframe>`, etc.

**BUT**, Microsoft saw this as a security threat, so you **can't** use `Blob` URLs (or `data:base64` URLs) for `<iframe>`, `<object>`, `<embed>`, `window.open()`, etc. You can only use it for things like CSS of images. Oh, and you can also **download** a Blob in IE (so we can generate the PDF and trigger its "download", but that's not what we want). This hasn't changed even in the most recent version of Edge.

PR #633, opened 2016-10-19
Remove the "isError" prop for FormTextarea, since it isn't supported

### LINKED ISSUE #630: Unknown prop isError on textarea in status/self-help
screen shot 2016-10-19 at 1 36 24 pm

DanReyLop:
Fixes #630

The `isError` prop was being passed to the Calypso `<FormTextarea>` component, but since it was always `undefined`, it was being ignored (yay, Javascript!). With a recent fix that made all the `isError` props be `Boolean`, this warning slipped through.

To test:
* Go to the Self-Help page.
* Check there are no errors on the console and the Log textarea shows fine.

@allendav @jeffstieler @robobot3000 @nabsul

### DIFF
--- a/client/components/text-area/index.js
+++ b/client/components/text-area/index.js
@@ -19,7 +19,6 @@ const TextArea = ( { id, readonly, title, description, value, placeholder, updat
 				readOnly={ readonly }
 				value={ value }
 				onChange={ handleChangeEvent }
-				isError={ Boolean( error ) }
 			/>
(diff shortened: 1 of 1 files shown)


allendav:
It's not a sprint 7 item, but since its done :shipit:

DanReyLop:
I have no mental strenght at this hour to tackle any big task. Plus, I saw the issue pop-up in Slack and it took me less time to solve it than to scroll past it :)

PR #635, opened 2016-10-19
Only load the Babel polyfill if there hasn't already been loaded

### LINKED ISSUE #634: Plug-in doesn't load when there's another React app in the same page
In the console appears this error and the React Components don't load:
`"only one instance of babel-polyfill is allowed"`

To reproduce:
* Install and activate a React-based plug-in, such as @nabsul's [Sift Science](https://wordpress.org/plugins/fermiac-siftscience-for-woocommerce/)
* Go to an order page (with both the Sift Science meta-box and the Shipping Labels meta-box).

DanReyLop:
Fixes #634 

This issue is caused by a pre-condition check in `babel-polyfill` ([here](https://github.com/babel/babel/blob/de975b9660014a4533c90190b297352495636530/packages/babel-polyfill/src/index.js#L4)). Since the polyfill modifies the global scope (needed for adding methods to built-in types like `Array`, `Object`, etc), it has a check to make sure that it's loaded only once. Since virtually all React-based apps use babel + polyfill, that means 2 React apps can't coexist in the same page, at least without the work-around in this PR.

The risk of this PR is that the polyfill that runs first may be a different version that the one bundled in our app, so we would not be in total control of our environment, but that's always a risk when we develop for a platform like WordPress :)

To test:
* Install and activate a React-based plug-in, such as @nabsul's [Sift Science](https://wordpress.org/plugins/fermiac-siftscience-for-woocommerce/)
* Go to an order page (with both the Sift Science meta-box and the Shipping Labels meta-box).
* Check that both plug-ins load.

@allendav @nabsul @jeffstieler @robobot3000 
@coderkevin You should take a look too. I would argue that all React-based apps that are made to run in only part of a page should include this workaround.

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -1,3 +1,7 @@
+if ( ! global._babelPolyfill ) {
+	require( 'babel-polyfill' );
+}
+
 /*global wcConnectData */
 import React from 'react';
 import ReactDOM from 'react-dom';
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -19,7 +19,7 @@ module.exports = {
 	babelSettings,
 	cache: true,
 	entry: {
-		'woocommerce-connect-client': [ 'babel-polyfill', './client/main.js' ],
+		'woocommerce-connect-client': [ './client/main.js' ],
 	},
 	output: {
 		path: path.join( __dirname, 'dist' ),


coderkevin:
I wonder if there's a way to determine the version of `babel-polyfill` after it's added. It probably wouldn't fix the problem, but at least it could log an error.

DanReyLop:
> I wonder if there's a way to determine the version of babel-polyfill after it's added. It probably wouldn't fix the problem, but at least it could log an error.

No, no way. The `_babelPolyfill` is just set to `true` when the polyfill is loaded. Maybe it could be set to the polyfill **version** loaded (so it can still be "true" in a boolean sense).

> Could a better solution be to submit a PR to add a check to babel-polyfill?

I'll try. This is definitely something that can be useful in many situations. Having several full-blown React apps running in the same page is not the most common use case, but if we start promoting this kind of plug-in development, soon we won't be alone.

They've not been very receptive to this issue so far. See [here](https://github.com/babel/babel/issues/4019) or [here](https://github.com/babel/babel/issues/1943#issuecomment-119531679)

According to the Babel docs, another option would be to use [`transform-runtime`](http://babeljs.io/docs/plugins/transform-runtime/) instead of the full `polyfill`. `transform-runtime` can make work global objects like `Promise` and static methods like `Object.assign`, but since it doesn't modify the global scope, it can't polyfill instance methods like `Array.prototype.includes`. And I don't have the stomach to search the Calypso codebase to see if we're using those.

coderkevin:
I'm not a huge fan of the `transform-runtime` idea, especially since it's not a complete solution. It would be great if this is something that could be solved at the source from the `babel-polyfill` itself. Even though there's been debate in those issues, there are a lot of folks feeling the same pain as us here, so I think there's hope.

DanReyLop:
> I'm not a huge fan of the transform-runtime idea, especially since it's not a complete solution.

I agree. I've only mentioned it because it's the "recommended" solution to have the polyfill with no conflicts.

> Even though there's been debate in those issues, there are a lot of folks feeling the same pain as us here, so I think there's hope.

https://github.com/babel/babel/issues/4019 Is the second-most upvoted issue in the entire project, maybe that will get their attention :)
I think a reasonable solution would be technically simple, but the maintainers need to be onboard or else it will be rejected.

nabsul:
Plays nicely with my plugin. I'll have to steal this solution to make sure my plugin plays nice with others :-)

If you want to use this solution temporarily and create a separate issue for the better long-term solution :shipit:

DanReyLop:
I'll try to work this out with the `babel` folks. There's not much else we can do on our side. Maybe try to figure out if we can use `transform-runtime` instead of the `polyfill`, but that would require some work in the tooling here and in Calypso (to emit `eslint` errors, for example, if code is using a method that will **not** be polyfilled by `transform-runtime`).

Until then, this workaround is better than nothing. We will always have troubles like this in real-life WordPress, a plug-in can't live completely isolated :)

PR #637, opened 2016-10-20
Omitting the box_id field when purchasing a label

robobot3000:
Fixes the server's issue Automattic/woocommerce-connect-server#582 by handling the `box_id` field in the same way as the other unwanted fields.

CC @nabsul


nabsul:
Tested! :shipit:

robobot3000:
Merged #637

Issue #570, opened 2016-09-26
Allow the merchant to remove packages and items from a label purchase


DanReyLop:
As part of #449 , in the `Packages` step, allow the merchant to remove packages from the shipment.

What should we do if the merchant removes a package with several items in it? My 2 cents:
- If the package is _not_ empty, move its contents to individual packages.
- Items can be removed individually, and that means they are removed for good (at least until the user refreshes the page). That would allow the merchant to print labels only for parts of an order.

kellychoffman:
>  means they are removed for good

Where would they go?

DanReyLop:
> whats the use case for removing them all together?

This ties in with the "I want to print labels only for some of the items" scenario. So the merchant can print labels for a few of the items, but send other items offline (maybe they are extra-fragile and doesn't trust USPS, or they're above the weight limit and it requires an specialiced carrier).

> Where would they go?

They would disappear from the Labels dialog but they would be still in the order, so if the user refreshes the page they would re-appear. I think this would be the way to go, generally: the changes made in the Labels UI (Packaging, destination address, etc) should **not** overwrite the data that the order already has. But that's just my 2 cents.

kellychoffman:
That makes a ton of sense. I think this is the way to go as long as the UI is clear that the items aren't being deleted, we are just removing it from this particular shipping label order. But that should be fairly easy to do. Thanks!

Issue #569, opened 2016-09-26
Allow the merchant to manually add a new package to the shipment


DanReyLop:
As part of #449 , in the `Packages` step, allow the merchant to add a package to the shipment. That means, independently of what the packing algorithm decided, the merchant must be allowed to add a box (from the Package Manager) and put one or several of the order's items in that box.

Note: The UI must allow somehow to add a new type of package. We could simply put a link to the `Package Manager`, but then we must make sure that the packages list in the Labels UI are updated without reloading the page.

kellychoffman:
> Note: The UI must allow somehow to add a new type of package. We could simply put a link to the Package Manager, but then we must make sure that the packages list in the Labels UI are updated without reloading the page.

This brings up a good point about the shipping labels UIs for when there are zero saved packages. What about doing this?

<img width="589" alt="screen shot 2016-10-04 at 12 11 54 pm" src="https://cloud.githubusercontent.com/assets/5835847/19082373/e3fb4e34-8a2b-11e6-8d6d-116091f6482f.png">

DanReyLop:
Looks good. It would also be useful to have that link even if the user has saved packages. Something like having a `<select>` with the last option being `Ad a package...` that takes you to the Packaging Manager.

kellychoffman:
> a <select> with the last option being Add a package...

Great idea. Since it would be strange to have that option be the only option in the select, we should have 'Add a package...' be the last item when there is at least one option present, like in  https://github.com/Automattic/woocommerce-connect-client/issues/585#issuecomment-251386056.

Issue #568, opened 2016-09-26
Allow the merchant to move items between packages


DanReyLop:
As part of #449 , in the `Packages` step, allow the merchant to move the items between packages. Ideally using drag&drop. Only allow to move an item into a "real" package, not an "individual" box, since that wouldn't make sense.

Note: we must allow to move an item "outside any package", to send it individually.

PR #638, opened 2016-10-25
Editable packaging solution

robobot3000:
Adds #568 #569 #570 

Adds editable packing solution:
oct-25-2016 13-05-13

Left to do:
- [x] Use a dropdown to select a box type instead of a modal
- [x] Flat rate packages. Should we allow to select them?

CC @allendav @kellychoffman @DanReyLop

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -80,7 +80,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 			return sprintf( '%s - %s', $identifier, $product->get_title() );
 		}
 
-		protected function get_packages( WC_Order $order ) {
+		protected function get_selected_packages(WC_Order $order ) {
 			$packages = $this->get_packaging_metadata( $order );
 			if ( ! $packages ) {
 				return $this->get_items_as_individual_packages( $order );
@@ -105,6 +105,23 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 					}
 					$formatted_packages[ $package_id ][ 'items' ][ $item_index ] = $product_data;
 				}
+			}
+
+			return $formatted_packages;
+		}
+
+		protected function get_all_packages() {
+			$packages = $this->settings_store->get_packages();
+
+			if ( ! $packages ) {
+				return new stdClass();
+			}
+
+			$formatted_packages = array();
+
+			foreach( $packages as $package ) {
+				$package_id = $package[ 'name' ];
+				$formatted_packages[ $package_id ] = $package;
 			}
 
 			return $formatted_packages;
@@ -166,17 +183,18 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		}
 
 		protected function get_form_data( WC_Order $order ) {
-			$packages        = $this->get_packages( $order );
-			$is_packed       = ( false !== $this->get_packaging_metadata( $order ) );
-			$origin          = $this->get_origin_address();
-			$selected_rates  = $this->get_selected_rates( $order );
-			$destination     = $this->get_destination_address( $order );
+			$selected_packages = $this->get_selected_packages( $order );
+			$all_packages      = $this->get_all_packages();
+			$is_packed         = ( false !== $this->get_packaging_metadata( $order ) );
+			$origin            = $this->get_origin_address();
+			$selected_rates    = $this->get_selected_rates( $order );
+			$destination       = $this->get_destination_address( $order );
 
 			if ( ! $destination[ 'country' ] ) {
 				$destination[ 'country' ] = $origin[ 'country' ];
 			}
 
-			$form_data = compact( 'is_packed', 'packages', 'origin', 'destination' );
+			$form_data = compact( 'is_packed', 'selected_packages', 'all_packages', 'origin', 'destination' );
 
 			$form_data[ 'rates' ] = array(
 				'selected'  => (object) $selected_rates,
--- a/client/shipping-label/state/reducer.js
+++ b/client/shipping-label/state/reducer.js
@@ -22,8 +22,29 @@ import {
 	OPEN_REPRINT_DIALOG,
 	CLOSE_REPRINT_DIALOG,
 	CONFIRM_REPRINT,
+	OPEN_PACKAGE,
+	OPEN_ITEM_MOVE,
+	MOVE_ITEM,
+	CLOSE_ITEM_MOVE,
+	SET_TARGET_PACKAGE,
+	ADD_PACKAGE,
+	REMOVE_PACKAGE,
+	SET_PACKAGE_TYPE,
+	SAVE_PACKAGES,
+	OPEN_ADD_ITEM,
+	CLOSE_ADD_ITEM,
+	SET_ADDED_ITEM,
 } from './actions';
+import getBoxDimensions from 'lib/utils/get-box-dimensions';
 import _ from 'lodash';
+
+const generateUniqueBoxId = ( keyBase, boxIds ) => {
+	for ( let i = 0; i <= boxIds.length; i++ ) {
+		if ( -1 === boxIds.indexOf( keyBase + i ) ) {
+			return keyBase + i;
+		}
+	}
+};
 
 const reducers = {};
 
@@ -130,7 +151,7 @@ reducers[ CONFIRM_ADDRESS_SUGGESTION ] = ( state, { group } ) => {
 };
 
 reducers[ UPDATE_PACKAGE_WEIGHT ] = ( state, { packageId, value } ) => {
-	const newPackages = { ...state.form.packages.values };
+	const newPackages = { ...state.form.packages.selected };
 
 	newPackages[ packageId ] = {
 		...newPackages[ packageId ],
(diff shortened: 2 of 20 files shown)


robobot3000:
Spoke with @allendav - the flat rate packages are to be added in #642. This is ready for review.

kellychoffman:
This is what I see when I arrive at the Packages step:

screen shot 2016-10-26 at 4 21 04 pm

I click 'add package' and there's only one option, `Envelope`:
screen shot 2016-10-26 at 4 26 23 pm

Can we just show this by default?

One other issue here is the fact that the contents section is empty:

screen shot 2016-10-26 at 4 29 00 pm

I assume its because the item doesn't fit into this package. Could we display an error here? It allows me to move on to the last step and even purchase the label:

screen shot 2016-10-26 at 4 29 00 pm

kellychoffman:
The Packages step doesn't auto expand once you have confirmed the destination address.

kellychoffman:
When I remove all packages, I see a floating `Saved for later`:

screen shot 2016-10-26 at 4 59 48 pm

robobot3000:
You must have viewed an order with no packages or items in it. Not sure how that happened. Does the same order display any packages when you open it in a different branch?

> This is what I see when I arrive at the Packages step
> When I remove all packages, I see a floating Saved for later

As mentioned above, this shouldn't happen, but I added some messaging:
screen shot 2016-10-27 at 14 23 10
The floating `Saved for later` should be fixed also.

> One other issue here is the fact that the contents section is empty

I added a message for when the contents are empty:
screen shot 2016-10-27 at 14 23 46

>  I click 'add package' and there's only one option, Envelope

The one option must be happening because you only have `Envelope` configured in you packaging manager.

> The Packages step doesn't auto expand once you have confirmed the destination address.

We don't autoexpand other sections when the ones before them were closed. The destination address wouldn't be expanded if it didn't need to be normalized. If we add autoexpand, then we should probably start at the origin address and then proceed from there?

jeffstieler:
This might be beyond the scope of this issues being resolved but.. it is possible to remove all of the package contents, and then be stuck.

We should have a way to add items to empty packages.

jeffstieler:
I'm having to scroll in almost all of the package modals.

See:
screen shot 2016-10-27 at 12 13 32 pm

client/shipping-label/views/purchase/steps/packages/index.js
```diff
@@ -1,56 +1,129 @@
...
 
-const PackagesStep = ( { values, storeOptions, labelActions, errors, expanded } ) => {
-	const packageIds = Object.keys( values );
-	const firstPackageId = packageIds[ 0 ];
-	const firstPackage = values[ firstPackageId ];
-	const isValid = 0 < firstPackage.weight;
+const PackagesStep = ( { openedPackageId, selected, all, unpacked, storeOptions, labelActions, errors, expanded, showItemMoveDialog, movedItemIndex, targetPackageId } ) => {
+	const packageIds = Object.keys( selected );
+	const itemsCount = packageIds.reduce( ( result, pId ) => ( result + selected[ pId ].items.length ), 0 );
+	const totalWeight = packageIds.reduce( ( result, pId ) => ( result + selected[ pId ].weight ), 0 );
```
jeffstieler: Consider using lodash's [`sumBy()`](https://lodash.com/docs/4.16.4#sumBy) for this.

client/shipping-label/views/purchase/steps/packages/index.js
```diff
@@ -1,56 +1,129 @@
...
+	const isValidWeight = packageIds.reduce( ( result, pId ) => ( result && 0 < selected[ pId ].weight ), true );
```
jeffstieler: Can't this use the `totalWeight` value instead of a call to `.reduce()`?

client/shipping-label/state/reducer.js
```diff
@@ -140,7 +150,235 @@ reducers[ UPDATE_PACKAGE_WEIGHT ] = ( state, { packageId, value } ) => {
...
+reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
+	const box = state.form.packages.all[ boxTypeId ];
+	const newPackages = {...state.form.packages.selected};
+	const { length, width, height } = getBoxDimensions( box );
+	const oldPackage = newPackages[ packageId ];
+
+	newPackages[ packageId ] = {
+		...oldPackage,
+		height, length, width,
+		weight: box.box_weight + oldPackage.items.reduce( ( sum, item ) => ( sum + item.weight ), 0 ),
```
jeffstieler: Consider using lodash's [`sumBy()`](https://lodash.com/docs/4.16.4#sumBy) for this.

client/shipping-label/state/reducer.js
```diff
@@ -140,7 +150,235 @@ reducers[ UPDATE_PACKAGE_WEIGHT ] = ( state, { packageId, value } ) => {
...
+		form: {
+			...state.form,
+			packages: {
+				...state.form.packages,
+				selected: newPackages,
+				saved: false,
+			},
+			rates: {
+				...state.form.rates,
+				values: _.mapValues( newPackages, ( pckg ) => ( pckg.service || '' ) ),
```
jeffstieler: Same question about `rates.values` and `rates.available`.

client/shipping-label/state/test/reducer.js
```diff
@@ -0,0 +1,245 @@
...
+		},
+	},
+	openedPackageId: 'weight_0_custom1',
+};
+
+describe( 'Label purchase form reducer', () => {
+	afterEach( () => {
+		// make sure the state hasn't been mutated
+		// after each test
+		expect( initialState ).to.eql( {
```
jeffstieler: Could using [`deepFreeze()`](https://github.com/substack/deep-freeze) on `initialState` be a way to avoid this?

client/shipping-label/views/purchase/steps/packages/list.js
```diff
@@ -1,78 +1,104 @@
...
+	let boxCount = 0;
+	const packed = [];
+	const individual = [];
+
+	Object.keys( selected ).forEach( ( pckgId ) => {
+		const pckg = selected[ pckgId ];
+
+		if ( 'individual' === pckg.box_id ) {
+			individual.push(
+				<div className="wcc-packages-list__item" key={ pckgId }>
```
jeffstieler: Let's keep this DRY - can you write a render method for package list items? (This is copied almost verbatim below for boxes/envelopes)

client/shipping-label/views/purchase/steps/packages/list.js
```diff
@@ -1,78 +1,104 @@
...
+	}
 
-				<div className="wcc-package-package-dimension">
-					<FormLegend>{ __( 'Package dimensions' ) }</FormLegend>
-					<span className="wcc-package-package-dimension__unit">{ renderPackageDimensions( pckg, dimensionUnit ) }</span>
-				</div>
-			</div>
+	if ( individual.length ) {
+		individual.unshift(
+			<div className="wcc-packages-list__item wcc-packages-list__header" key="unpacked_header">{ __( 'Original packaging' ) }</div>
```
jeffstieler: Not a huge deal, but another DRY candidate ("boxed" header above).

client/shipping-label/views/purchase/steps/packages/move-item.js
```diff
@@ -0,0 +1,101 @@
...
+				onClose={ closeItemMove }
+				additionalClassNames="wcc-modal wcc-label-packages-dialog" >
+			<div className="wcc-label-packages-dialog__content">
+				<h1 className="form-section-heading">{ __( 'Move item' ) }</h1>
+				<div className="wcc-label-packages-dialog__body">
+					<p dangerouslySetInnerHTML={ { __html: desc } }/>
+					<p>{ __( 'Where would you like it moved?' ) }</p>
+					{ Object.values( _.mapValues( _.omitBy( selected, ( pckg, pckgId ) => ( pckgId === openedPackageId || 'individual' === pckg.box_id ) ), renderPackageOption ) ) }
+					{ ! openedPackage || 'individual' !== openedPackage.box_id ? renderRadioButton( 'individual', __( 'Ship in original packaging' ) ) : null }
+					{ '' !== openedPackageId ? renderRadioButton( '', __( 'Save for later' ) ) : null }
```
jeffstieler: These three lines are really hard to read.

client/shipping-label/views/purchase/steps/packages/package-info.js
```diff
@@ -0,0 +1,90 @@
...
+	};
+
+	const pckg = selected[ packageId ];
+
+	const renderPackageSelect = () => {
+		if ( ! all[ pckg.box_id ] ) {
+			return ( <div className="wcc-package__desc">Package - { renderPackageDimensions( pckg, dimensionUnit ) }</div> );
+		}
+
+		const packageOptions = Object.values( _.mapValues( all, renderPackageOption ) );
```
jeffstieler: You can use `_.map()` here instead of wrapping `_.mapValues()` with `Object.values()`.

classes/class-wc-connect-shipping-label.php
```diff
@@ -110,6 +110,23 @@ protected function get_packages( WC_Order $order ) {
 			return $formatted_packages;
 		}
 
+		protected function get_all_packages() {
+			$packages = $this->settings_store->get_packages();
+
+			if ( ! $packages ) {
+				return new stdClass();
```
robobot3000: The empty array will be serialized to json's `[]` which will lead to a React error, because it expects an object. Not sure if there's a way of working around that.

client/shipping-label/state/test/reducer.js
robobot3000: The purpose of this is to test if the reducer methods don't mutate the state, so I actually want the `initialState` to not be frozen if this happens. But I removed the code duplication now, so this should be a bit easier to read.

robobot3000:
Thanks for reviewing @jeffstieler. I made the changes you suggested.
> I'm having to scroll in almost all of the package modals.
I tweaked the styling a bit, so the modal should be a bit taller, and its contents centered

> I just realized I can go to "saved for later" to find the items, but I found that by chance. There was nothing indicating to me that any items were "saved".
I think @kellychoffman is going to have a look at this once she gets her local branch to show packages and items.

classes/class-wc-connect-shipping-label.php
nabsul: I wonder if we should just start using stdClass more consistently in our PHP code to avoid this issue altogether?

kellychoffman:
> Package 1

In the Package manager, you can name your package. We don't seem to expose this anywhere. Instead of saying `Package 1` `Package 2` etc here, we should write out the name of the package so it matches what is in the dropdown and in the Package manager.

kellychoffman:
In the item contents and 'move' dialog, the product appears to be linked, but doesn't go anywhere. It should link to the edit product page.

kellychoffman:
If I add a new box with the intention of moving an item to that, I should be able to add it from the new box. This would be possible if we added a button next to the 'There are no items in this package.' message: 

screen shot 2016-10-28 at 2 55 46 pm

kellychoffman:
> EDIT: I just realized I can go to "saved for later" to find the items, but I found that by chance. There was nothing indicating to me that any items were "saved".

Can we add a `count` to the right of the Saved packages link?

kellychoffman:
Question: why can't I change the box on a package with items, like I can on a package without items? This seems like the most convenient way to bulk move items from one package to another.

robobot3000:
> In the Package manager, you can name your package. Instead of saying Package 1 Package 2 or Package - [dimensions here]
> Question: why can't I change the box on a package with items, like I can on a package without items? This seems like the most convenient way to bulk move items from one package to another.

Both questions are about the same thing: it seems the default packages selected by the customer were the flat rate packages. The labels UI displays `Package` when it can't match the preselected package to the saved shared packages. This could also happen when a custom package is selected, but later deleted. Maybe it should show `Unknown` instead?

> In the item contents listing and 'move' dialog, the product appears to be linked, but doesn't go anywhere. It should link to the edit product page.

It links to the product page for me. Looking at the code, this should always link to somewhere. What is the value of the `href` attribute on these links for you?

robobot3000:
I also added the `Add Item` button.

kellychoffman:
> It links to the product page for me. Looking at the code, this should always link to somewhere. What is the value of the href attribute on these links for you?

Its just `<a target="_blank">` for me. Tried on a few different orders and product items and they are all empty.

kellychoffman:
> Its just <a target="_blank"> for me. Tried on a few different orders and product items and they are all empty.

I _do_ get links if I create a new product, purchase it, and then create an order.

jeffstieler:
Sorry to nitpick - but I think we should change the logic that calculates the "Box N" naming.

It should be calculated off packages only - right now it includes items in their original packaging. This is confusing if you start putting individually packages items into a box. You'll see `Box 4` for instance, and then as you move individual items to it, the number decrements.

screen shot 2016-11-04 at 10 53 13 am

screen shot 2016-11-04 at 10 53 32 am

robobot3000:
This is now fixed. I also tweaked the rates step to show the same package names as in the packages step

PR #646, opened 2016-10-27
Allow layout schema to provide Text component body.

jeffstieler:
Allow layout schema to provide `Text` component body.

This avoids having to add non-settings to the form schema, which is how the Self Help page `Text` component is currently rendered.


jeffstieler:
> There is a text component exactly for this (it's used only in the self-help page currently).

Thanks for the heads up! It looks like I can make a small change to this, and we're good.

jeffstieler:
@DanReyLop this is ready for another pass. :)

PR #647, opened 2016-10-27
Add connectforwoocommerce text domain to all strings and load it

### LINKED ISSUE #474: Connect needs its own text domain / textdomain
Since we're not going to merge into woocommerce anymore, we need to update everything to use a new connect specific text domain.

allendav:
Fixes #474

To test:
- Delete `i18n/strings.php`
- `npm run i18n`
- Make sure `i18n/strings.php` gets regenerated with all the strings
- Open each of checkout, settings, self-help and a shipping method instance and make sure strings appear and no errors are logged in php errors nor the javascript console


allendav:
@DanReyLop - while working this PR I noticed that the i18n task (`npm run i18n`) broke between August and now.

The first problem was that the i18n-calypso dependency was never added - I fixed that, but then I noticed it was trying to extract strings from uglified JS.

I fixed that too, and now it still doesn't find the translate/__ calls in the webpacked JS - I tried to fix that by tweaking the keyword passed to Xgettext in tasks/i18n.js but it still doesn't work (it just generates a mostly empty file in i18n/strings.php

Before I spend anymore time trying to get this to work, do you have any suggestions? Maybe I need a different keyword in the call to Xgettext?

Also cc'ing @aduth

DanReyLop:
Back in the day, it worked even with uglified JS. I'll take a look at this, give me until the Apple event starts and I'll get back to you :)

DanReyLop:
I've fixed the problem. There was a version discrepancy between the `i18n-calypso` dependency and the `xgettext-js` dependency. I've pushed the fix to this branch, along with the generated `strings.php` file (take it as proof that this works, at least in my machine).

I had to use `npm run update-deps` (the thing that nukes `shrinkwrap` and everything and regenerates it from scratch) because it was misbehaving. In the future, *never* add, remove or edit any entry from `dependencies` or `devDependencies`. If you do those operations using the `npm` command (`npm install --save PCKG`, `npm uninstall --save PCKG`, etc) both `package.json` and `npm-shrinkwrap.json` will be updated automatically.

allendav:
Good to know - ty!

allendav:
@DanReyLop - good to go now?

PR #658, opened 2016-11-02
Support printing multiple labels at once

### LINKED ISSUE #500: Support printing multiple shipping labels
An order can have multiple packages, and so multiple shipping labels will need to be printed. The client must allow to do it easily. Questions:
- Should the WCC server return a multi-page PDF, or the client should somehow manage stiching together multiple labels?
- What happens if one shipping label purchase fails? Should the entire "transaction" be rolled back?
- Should be allow the merchant to print a label for only one package, even if the order has multiple packages?

DanReyLop:
Fixes #500 

To test this, you will need also the `try/pdf-manipulation` server-side branch.

This PR:
* Makes the `PREVIEW` box in the Labels UI a bit more useful. The merchant can select a paper size and preview how the PDF he'll have to print looks.
* Adds a `Paper Size` dropdown to the Labels UI. The selected value is remembered forever, so the merchant will only have to change it once (and if he buys a new printer).
* Adds the same `Paper Size` dropdown to the `Reprint` dialog.
* The rest of the changes are "under the hood", on the way the PDF is actually generated (a GET request to the merchant host that is transformed into a POST request to the server).

How to test:
* Create an order with multiple items.
* Experiment with the `Paper size` dropdown in the Labels UI and how that affects the Preview.
* Buy labels for all those items.
* Check that the PDF to print has the expected page size and number of labels.
* Try to reprint a newly purchased label.

@nabsul @robobot3000 @jeffstieler @allendav @kellychoffman


nabsul:
When I have multiple labels and the page is loading, I see this:
image

Is there anyway we could replace that with one spinner?

DanReyLop:
@coderkevin Jump in if you want. This PR has a bigger, most complex server-side counterpart, but taking a look only at the client-side of things can be useful too :)

jeffstieler:
The `GET` to `POST` juggling seems pretty hacky to me - are there any alternatives? I've seen a technique used to set a form's target to an iframe so that the submission response is rendered in the iframe. Do you think that would work here?

client/shipping-label/state/actions.js
```diff
@@ -229,22 +233,52 @@ export const updatePackageWeight = ( packageId, value ) => {
...
-export const confirmPackages = () => ( dispatch, getState, { getRatesURL, storeOptions, nonce } ) => {
+const refreshPreview = ( dispatch, getState, context ) => {
+	if ( ! canPurchase( getState(), context.storeOptions ) ) {
+		return;
+	}
+	const state = getState().shippingLabel;
+	const { form, paperSize } = state;
+	let pckgIndex = 1;
+	const labels = _.map( form.packages.values, () => ( {
+		caption: sprintf( __( 'Package %d (of %d)' ), pckgIndex++, Object.keys( form.packages.values ).length ).toUpperCase(),
```
jeffstieler: Why `.toUpperCase()` here?

client/shipping-label/state/actions.js
```diff
@@ -265,8 +299,12 @@ export const purchaseLabel = () => ( dispatch, getState, { purchaseURL, addressN
 				console.error( error );
 				dispatch( NoticeActions.errorNotice( error.toString() ) );
 			} else {
-				// TODO: Figure out how to print multiple labels
-				printDocument( sprintf( labelImageURL, response[ 0 ].label_id ), nonce )
+				let pckgIndex = 1;
+				const labels = response.map( ( label ) => ( {
+					caption: sprintf( __( 'Package %d (of %d)' ), pckgIndex++, response.length ).toUpperCase(),
```
jeffstieler: Same question about `.toUpperCase()`.

client/shipping-label/state/actions.js
```diff
@@ -229,22 +233,52 @@ export const updatePackageWeight = ( packageId, value ) => {
...
 };
 
-export const confirmPackages = () => ( dispatch, getState, { getRatesURL, storeOptions, nonce } ) => {
+const refreshPreview = ( dispatch, getState, context ) => {
+	if ( ! canPurchase( getState(), context.storeOptions ) ) {
+		return;
+	}
+	const state = getState().shippingLabel;
+	const { form, paperSize } = state;
+	let pckgIndex = 1;
```
jeffstieler: You can avoid this counter by using the second argument passed to `_.map()` callbacks.

client/shipping-label/state/actions.js
```diff
@@ -265,8 +299,12 @@ export const purchaseLabel = () => ( dispatch, getState, { purchaseURL, addressN
 				console.error( error );
 				dispatch( NoticeActions.errorNotice( error.toString() ) );
 			} else {
-				// TODO: Figure out how to print multiple labels
-				printDocument( sprintf( labelImageURL, response[ 0 ].label_id ), nonce )
+				let pckgIndex = 1;
```
jeffstieler: You can avoid this counter by using the second argument passed to `Array.prototype.map` callbacks.

DanReyLop:
> The GET to POST juggling seems pretty hacky to me - are there any alternatives? I've seen a technique used to set a form's target to an iframe so that the submission response is rendered in the iframe. Do you think that would work here?

Cool, I didn't know about that technique, TIL! :)

Still, that feels even more hacky to me. Instead of just setting an `<iframe>` with an URL, the client code would need to create a `<form>`, attach it to the `DOM`, submit it and remove it afterwards. I haven't tried it, but something tells me that submitting a form without direct user input could cause trouble in some or all browsers.

We are used to the merchant host being kind of a "proxy", but that doesn't need to be the case at all. In this case, it makes sense to the client to call the REST endpoint with a `GET`, and the server expects a `POST`, so I don't see the problem having the "middle layer" doing the conversion.

nabsul:
Tests well with the server. Does Travis need to be re-run to get it to pass? 
I'll leave it to Jeff for the final sign off.

jeffstieler:
Just one comment regarding building the label preview/purchase URLs. Other than that - :ship: it.

allendav:
Please open as a new issue - we should have Letter size selected as default

screen shot 2016-11-07 at 11 13 02 am

allendav:
Please open as a new normal priority issue - can we 1) get rid of the PREVIEW heading inside the preview box or 2) at least the background scrolling

preview-header-and-scrolling

allendav:
@DanReyLop - has this been tested on Microsoft Edge?

allendav:
Please open as a new normal priority issue - it is a little worse in Firefox

firefox-pdf-labels

allendav:
Please open as a new normal priority issue - also kinda wonky in Safari

safari-pdf-labels

allendav:
@DanReyLop - needs rebasing and a few issues need writing up - no other comments for me - please proceed when @jeffstieler is happy

DanReyLop:
> has this been tested on Microsoft Edge?

@allendav It was tested in Edge and it's the best-looking case of the lot. Like Chrome, but it doesn't show the "preview" bar, so pretty much ideal. Every browser, OS, and PDF-viewing plugin will do this differently (in fact, my Chrome preview looks different than yours). I wanted to get this merged as soon as possible, but I agree we should aim to improve it in the near future. That will probably involve a serious amount of hacking.

> Please fix or open as a new high priority issue - we should have Letter size selected as default

This was easy enough, no need for a new issue :)

> Since we're leaving the GET-to-POST juggle in, please consider using Node's [querystring.stringify()](https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options) method here.

Oh boy, I didn't realize that was even possible (using core Node.JS modules in the client). Webpack is so awesome 😄 . And here I was, doing string concatenation like a caveman...

DanReyLop:
I've created #675 for the PDF rendering issues, and I've rebased the branch.

PR #659, opened 2016-11-02
Generate anonymous UUID

### LINKED ISSUE #651: Generate a "unique" id when connect is activated for the first time and add it to every request
To facilitate anonymous usage data collection, maybe using the RFC4122 technique atom does

atom-2
atom-1

nabsul:
Closes #651 

This PR includes code to create and store a GUID as a store option, and send that GUID with all API requests to the server. This is needed to collect anonymized stats on the server side.

To test:
- Test with pending server PR 607 if it's not merged yet.
- Open up the settings page or anything that requires communication with the server
- Check that there is now a `wc_connect_store_guid` field in the database
- Check that the field is sent with API requests
- Check that if the option is deleted, it is automatically regenerated
- Check that the field is not regenerated if it already exists

@allendav @jeffstieler @DanReyLop @robobot3000

### DIFF
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -366,6 +366,7 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 
 			// Add interesting fields to the body of each request
 			$body['settings'] = wp_parse_args( $body['settings'], array(
+				'store_guid' => $this->get_guid(),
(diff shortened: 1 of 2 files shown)


allendav:
Tests nicely. :shipit: after https://github.com/Automattic/woocommerce-connect-server/pull/607

nabsul:
Cool. Will merge this after the server side gets deployed. Otherwise, the server will start failing.

Issue #670, opened 2016-11-07
Flat Rate Packages Manager: client side 


robobot3000:
On the Packaging Manager page:
* show a list of predefined packages per service (currently only flat rates for USPS)
* allow the merchant to select which packages they have available
On the Shipping Instance settings page:
* do not show the flat rate services that use the packages unavailable to the merchant
When requesting rates:
* make sure that the flat rate services with disabled packages are not requested

CC @allendav

PR #672, opened 2016-11-07
Added UI for managing flat rate packages

robobot3000:
Adds #670 

Should be tested with `add/centralized-flat-rates` checked out on the server-side.

To test:
- Check that you can see the predefined packages in the Packaging Manager:
screen shot 2016-11-07 at 14 03 49
- Check that disabling the packages hides the corresponding services from the shipping instance settings:
screen shot 2016-11-07 at 14 05 39
- Check that no flat rates are returned to the customers when flat rate packages are disabled in the packaging manager

### DIFF
--- a/client/packages/views/index.js
+++ b/client/packages/views/index.js
@@ -4,6 +4,8 @@ import FormSectionHeading from 'components/forms/form-section-heading';
 import ActionButtons from 'components/action-buttons';
 import FormFieldset from 'components/forms/form-fieldset';
 import FormButton from 'components/forms/form-button';
+import FoldableCard from 'components/foldable-card';
+import Gridicon from 'components/gridicon';
 import PackagesList from './packages-list';
 import AddPackageDialog from './add-package';
 import { translate as __ } from 'lib/mixins/i18n';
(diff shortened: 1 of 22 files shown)


allendav:
@kellychoffman @robobot3000 - to be consistent with our other UX, how about checkboxes on the left instead?

screen shot 2016-11-07 at 11 44 28 am

allendav:
The number of options for USPS flat rate packaging is overwhelming. Since all domestic flat rate boxes can be used for international flat rate (with one exception), let's not have separate checkboxes for domestic and international for a given size.

The one exception is the APO/FPO/DPO flat rate box. APO/FPO/DPO addresses are for US armed forces overseas, but they are treated like domestic addresses.  Supporting this flat rate box will probably require further code changes (i.e. to detect APO/FPO/DPO addresses), so defer this particular box to a separate issue.

allendav:
Let's also not bother with Gift Card Flat Rate Envelopes at this time.  Please exclude them from the UI.

allendav:
Let's not bother with the DVD Box nor the Large Video Box at this time. Please exclude them from the UI.

allendav:
Instead of a single group of "USPS Flat Rate Packages" lets break this up a bit so that merchants have an easier time finding what they need.  I propose the following grouping (cc'ing @kellychoffman  for her insights)

Let's also take care to name these exactly as shown below (both groups and packages) since these are the exact names on the USPS website and the names the merchant is most likely to understand

(Note that there are two Window Flat Rate Envelopes - I've included the dimensions in parentheses, but the name should not include the dimensions since we display them separately)

screen shot 2016-11-07 at 12 35 58 pm

allendav:
Neither "Post Card" nor "Letter" are flat rate packaging. Please do not include in the UI.

screen shot 2016-11-07 at 12 40 59 pm

allendav:
I could not find this flat rate packaging on the USPS web site at all. Maybe it doesn't exist / we entered it in error?

screen shot 2016-11-07 at 12 42 18 pm

allendav:
@robobot3000 - if combining domestic and international for each package is going to substantially blow up the work for this PR, then it is OK for us to have this PR NOT do that - in that case we should 1) write a new issue to combine domestic and international for each package and 2) have six groups (three domestic groups and three international groups).  The group names should be something like

USPS Priority Mail Flat Rate Envelopes
USPS Priority Mail Flat Rate Boxes
USPS Priority Mail Express Flat Rate
USPS Priority Mail International Flat Rate Envelopes
USPS Priority Mail International Flat Rate Boxes
USPS Priority Mail Express International Flat Rate

robobot3000:
Most of the requested changes had to be added on the server side in Automattic/woocommerce-connect-server#619. I disabled the services that you requested to hide from the UI (gift cards, videos etc). I also added groups and merged the domestic and international flat rates. See the server PR for more info

client/packages/views/index.js
```diff
@@ -34,7 +84,12 @@ const Packages = ( props ) => {
 			<CompactCard className="settings-group-card">
 				<FormSectionHeading className="settings-group-header">{ __( 'Packaging' ) }</FormSectionHeading>
```
DanReyLop: If the other section is called `Prefecined packages`, I would call this one something more consistent, like `Custom packages`.

client/packages/views/packages-list.js
```diff
@@ -16,18 +18,28 @@ const noPackages = () => {
...
+			return null;
+		}
+
+		const allPackageIds = packages.map( ( def ) => def.id );
+		const selectedAll = 0 === _.difference( allPackageIds, selected ).length;
+		return <FormLegend className="package-actions"><FormCheckbox checked={ selectedAll } onChange={ () => toggleAll( serviceId, groupId ) }/></FormLegend>;
+	};
 
 	const renderPackageListItem = ( pckg, idx ) => {
+		const isSelected = selected && -1 !== selected.indexOf( pckg.id );
```
DanReyLop: `Array.prorotype.includes()` is much cleaner than the `-1` thing. Let's make good use of babel :) [Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

client/settings/views/services/group.js
```diff
@@ -63,14 +64,18 @@ const ShippingServiceGroup = ( props ) => {
 				</label>
 			</div>
 
-			{ services.map( ( service, idx ) => (
-				<ShippingServiceEntry
+			{ services.map( ( service, idx ) => {
+				if ( service.predefined_package && -1 === predefinedPackages.indexOf( service.predefined_package ) ) {
```
DanReyLop: Same comment here about `Array.prototype.includes`

allendav:
@robobot3000 - this looks much better!  But I get the following error when checking a box:

screen shot 2016-11-08 at 12 51 42 pm

robobot3000:
This happens only if the packages were never defined (the option doesn't exist in the database). It's now fixed

allendav:
Thank you @robobot3000 - it tests nicely now

client/packages/views/index.js
```diff
@@ -13,8 +15,56 @@ import * as PackagesActions from '../state/actions';
...
+		const elements = [];
+
+		_.forEach( props.form.predefinedSchema, ( servicePackages, serviceId ) => {
+			const serviceSelected = props.form.packages.predefined[ serviceId ] || [];
+
+			_.forEach( servicePackages, ( predefGroup, groupId ) => {
+				const groupPackages = predefGroup.definitions;
+				const groupPackageIds = groupPackages.map( ( def ) => def.id );
+				const diffLen = _.difference( groupPackageIds, serviceSelected ).length;
+				const summary = 0 === diffLen ? __( 'All packages selected' ) : sprintf( __( '%d packages selected' ), groupPackageIds.length - diffLen );
```
allendav: When only 1 package is selected, this displays "1 packages selected" - let's handle that case better ("1 package selected")

allendav:
I think we can 🚢 this ... @DanReyLop do you concur?

see also https://github.com/Automattic/woocommerce-connect-server/pull/619

DanReyLop:
Yes, OK to 🚢 for me

PR #678, opened 2016-11-09
Fix label Preview

DanReyLop:
With one of the Packaging PRs, the state tree structure changes, so `form.packages.values` doesn't exist anymore, and to iterate on the selected packages we must use `form.packages.selelcted` instead. This was broken during a rebase (or maybe a merge).

@nabsul


nabsul:
Looks good. :shipit:

PR #682, opened 2016-11-10
Add/669 store shipping details

### LINKED ISSUE #669: Send Order Shipping Information with Label Rate Request
To allow for pre-selecting label rates based on the rate the customer paid for, we'll need to send this information to the server when label rates are requested.

nabsul:
Closes #669 

This PR includes the following changes:
- Pass `service_id` field to connect server
- Check for `is_selected` in label rate response and use it if it's there
- Minor refactoring of a really long line of code

To test: 
- Test with corresponding server PR 
- Buy an item as a customer and select a shipping rate that is recognized in the mapping https://github.com/Automattic/woocommerce-connect-server/blob/master/lib/shipping/usps/easypost-service-mapping.js#L1
- Log in as an admin and open the order page
- Open the modal for printing a shipping label, and observe that the rate is pre-selected
- Server should also continue to work as usual with current client

@allendav @jeffstieler @robobot3000 @DanReyLop

### DIFF
--- a/classes/class-wc-rest-connect-shipping-rates-controller.php
+++ b/classes/class-wc-rest-connect-shipping-rates-controller.php
@@ -71,7 +71,7 @@ class WC_REST_Connect_Shipping_Rates_Controller extends WP_REST_Controller {
 		$payload[ 'carrier' ] = 'usps';
 
 		// Exclude extraneous package fields
-		$whitelist = array_fill_keys( array( 'length', 'width', 'height', 'weight', 'template' ), true );
+		$whitelist = array_fill_keys( array( 'length', 'width', 'height', 'weight', 'template', 'service_id' ), true );
 		$formatted_packages = array();
 
 		foreach ( $payload[ 'packages' ] as $package_id => $package ) {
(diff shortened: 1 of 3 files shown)


classes/class-wc-rest-connect-shipping-rates-controller.php
```diff
@@ -71,7 +71,7 @@ public function update_items( $request ) {
 		$payload[ 'carrier' ] = 'usps';
 
 		// Exclude extraneous package fields
-		$whitelist = array_fill_keys( array( 'length', 'width', 'height', 'weight', 'template' ), true );
+		$whitelist = array_fill_keys( array( 'length', 'width', 'height', 'weight', 'template', 'service_id' ), true );
```
robobot3000: Will this handle the case when a package has been changed, or blow up the server's request validation?

I also moved this logic to JS in my PR: https://github.com/Automattic/woocommerce-connect-client/pull/683/files#diff-d8b04016a3889c977c51e0ed0ae5a2f7 . It was basically the same when requesting rates and when purchasing labels. Just FYI to avoid merge conflicts.

robobot3000: Nevermind, I just noticed that `service_id` is optional on the server side.

Issue #671, opened 2016-11-07
Flat Rate Packages Manager: show the predefined packages in the label's editable packaging


robobot3000:
TBD when #638 is merged in.

Show the flat rate packages in the editable packaging UI. Selecting a flat rate package should show the flat rate price in the rates step.

PR #683, opened 2016-11-10
Add flat rate packages to labels

robobot3000:
Adds #671

Test with https://github.com/Automattic/woocommerce-connect-server/pull/625 

To test:
* open the label print UI
* verify that the flat rate packages appear in the packaging editor
  * at the moment all flat rate packages are shown, regardless of what has been enabled/disabled in the packaging manager
* select one or more flat rate packages
* verify that the prices that appear are for flat rate deliveries
* check that purchasing a label works

### DIFF
--- a/classes/class-wc-connect-service-schemas-store.php
+++ b/classes/class-wc-connect-service-schemas-store.php
@@ -195,5 +195,16 @@ if ( ! class_exists( 'WC_Connect_Service_Schemas_Store' ) ) {
 
 			return ( object ) $predefined_packages;
 		}
+
+		public function get_predefined_packages_schema_for_service( $service_id ) {
+			$package_schemas = $this->get_predefined_packages_schema();
+
+			if ( is_null( $package_schemas )
+			     || ! property_exists( $package_schemas, $service_id ) ) {
+				return array();
+			}
+
+			return $package_schemas->{ $service_id };
+		}
 	}
 }
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -14,9 +14,15 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		 */
 		protected $settings_store;
 
-		public function __construct( WC_Connect_API_Client $api_client, WC_Connect_Service_Settings_Store $settings_store ) {
+		/**
+		 * @var WC_Connect_Service_Schemas_Store
+		 */
+		protected $service_schemas_store;
+
+		public function __construct( WC_Connect_API_Client $api_client, WC_Connect_Service_Settings_Store $settings_store, WC_Connect_Service_Schemas_Store $service_schemas_store ) {
 			$this->api_client = $api_client;
 			$this->settings_store = $settings_store;
+			$this->service_schemas_store = $service_schemas_store;
 		}
 
 		protected function get_items_as_individual_packages( $order ) {
@@ -111,20 +117,42 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		}
 
 		protected function get_all_packages() {
-			$packages = $this->settings_store->get_packages();
-
-			if ( ! $packages ) {
-				return new stdClass();
-			}
+			$custom_packages = $this->settings_store->get_packages();
 
 			$formatted_packages = array();
 
-			foreach( $packages as $package ) {
+			foreach( $custom_packages as $package ) {
 				$package_id = $package[ 'name' ];
 				$formatted_packages[ $package_id ] = $package;
 			}
 
+			$service_id = 'usps'; //TODO: remove hardcoding
+			$predefined_packages_schema = $this->service_schemas_store->get_predefined_packages_schema_for_service( $service_id );
+			$enabled_predefined_packages = $this->settings_store->get_predefined_packages_for_service( $service_id );
+
+			foreach ( $predefined_packages_schema as $group ) {
+				foreach ( $group->definitions as $package ) {
+					if ( ! in_array( $package->id, $enabled_predefined_packages ) ) {
+						continue;
+					}
+
+					$formatted_packages[ $package->id ] = $package;
+				}
+			}
+
 			return $formatted_packages;
+		}
+
+		protected function get_flat_rate_packages_groups() {
+			$service_id = 'usps'; //TODO: remove hardcoding
+			$predefined_packages_schema = $this->service_schemas_store->get_predefined_packages_schema_for_service( $service_id );
+			$groups = array();
+
+			foreach ( $predefined_packages_schema as $group_id => $group ) {
+				$groups[ $group_id ] = $group->title;
+			}
+
+			return $groups;
 		}
 
 		protected function get_selected_rates( WC_Order $order ) {
@@ -185,6 +213,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		protected function get_form_data( WC_Order $order ) {
 			$selected_packages = $this->get_selected_packages( $order );
 			$all_packages      = $this->get_all_packages();
+			$flat_rate_groups  = $this->get_flat_rate_packages_groups();
 			$is_packed         = ( false !== $this->get_packaging_metadata( $order ) );
 			$origin            = $this->get_origin_address();
 			$selected_rates    = $this->get_selected_rates( $order );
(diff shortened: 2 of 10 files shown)


robobot3000:
Also, note that you will need automattic/woocommerce-connect-server#625 checked out to test a following scenario:
* as a customer, place an order and select a flat rate package
* as a merchant, purchase a label for that order without making any changes to the packaging

allendav:
@robobot3000 wrote:

> at the moment all flat rate packages are shown, regardless of what has been enabled/disabled in the packaging manager

Is that a separate issue? I could not find it

robobot3000:
> Is that (only displaying in the label UI what has been enabled in the packaging manager) a separate issue? I could not find it

It's not, but I assumed this is what we want, since the rates shown in the labels UI are also not filtered.

allendav:
> It's not, but I assumed this is what we want, since the rates shown in the labels UI are also not filtered.

Hmmm. From the merchant's perspective, I think we want to filter out flat rate services whose packaging is not enabled in the packaging manager. Here's my rationale: as a merchant, when I am in that UI, I expect to be able to print a label immediately and stick it on the box and ship. For carrier provided packaging, like flat rate boxes, I'm going to only want to see boxes that I have on hand (out of the dozen flat rate boxes, a merchant is likely to NOT have all of them and may only have a few of a kind.)

So, I think we should filter the flat rate services in the labels UI based on the flat rate packaging enabled in packaging manager.  What do you think?

allendav:
I get the following linting warning:

```
WARNING in ./client/shipping-label/state/actions.js

/Applications/MAMP/htdocs/wp-content/plugins/woocommerce-connect-client/client/shipping-label/state/actions.js
  470:36  warning  Multiple spaces found before ''items''  no-multi-spaces
```

classes/class-wc-connect-shipping-label.php
```diff
@@ -111,22 +117,38 @@ protected function get_selected_packages(WC_Order $order ) {
...
 
 			$formatted_packages = array();
 
-			foreach( $packages as $package ) {
+			foreach( $custom_packages as $package ) {
 				$package_id = $package[ 'name' ];
 				$formatted_packages[ $package_id ] = $package;
 			}
 
+			$flat_rate_schema = $this->service_schemas_store->get_predefined_packages_schema_for_service( 'usps' );
```
allendav: If we filter on enabled services only, could we avoid hardcoding `usps` here and make this support predefined packages for other carriers, like FedEx One Rate boxes?

classes/class-wc-connect-shipping-label.php
```diff
@@ -111,22 +117,38 @@ protected function get_selected_packages(WC_Order $order ) {
...
+
+			foreach ( $flat_rate_schema as $group ) {
+				foreach ( $group->definitions as $package ) {
+					$package_id = property_exists( $package, 'id' ) ? $package->id : $package->name;
```
allendav: Under what circumstances do packages not have ids?

classes/class-wc-connect-shipping-label.php
```diff
@@ -111,22 +117,38 @@ protected function get_selected_packages(WC_Order $order ) {
...
+					$formatted_packages[ $package_id ] = $package;
+				}
+			}
+
 			return $formatted_packages;
 		}
 
+		protected function get_flat_rate_packages_groups() {
+			$flat_rate_schema = $this->service_schemas_store->get_predefined_packages_schema_for_service( 'usps' );
```
allendav: We will be adding non flat rate predefined packages next (like these: https://store.usps.com/store/browse/uspsProductDetailMultiSkuDropDown.jsp?categoryNavIds=shipping-supplies%3afree-shipping-supplies&categoryNav=false&navAction=push&navCount=0&atg.multisite.remap=false&categoryId=free-shipping-supplies&productId=P_O_BOX4 )

Please filter the packages returned by `get_predefined_packages_schema_for_service ` for flat rate packages only or a TODO comment here to fix this when we add non flat rate predefined packages.

client/shipping-label/state/actions.js
```diff
@@ -467,14 +467,17 @@ export const purchaseLabel = () => ( dispatch, getState, context ) => {
 			return;
 		}
 		const state = getState().shippingLabel;
+		const packageParamBlacklist = [  'items', 'id', 'box_id' ];
+		const templateParamBlacklist = [ ...packageParamBlacklist, 'width', 'length', 'height' ];
```
allendav: This seems backwards compared to the way package and template whitelists are defined above?

allendav: Nevermind - i see it is a blacklist :)

robobot3000:
The packages should be now filtered depending on their state in the packaging manager.

> Please filter the packages returned by get_predefined_packages_schema_for_service for flat rate packages only or a TODO comment here to fix this when we add non flat rate predefined packages.

I don't think this needs any more logic to return flat rates only. If non-flat rate predefined packages are added to the service schema, they should still be shown, and handled on the server-side if necessary.

> could we avoid hardcoding usps here

The entire labels UI is hardcoded to use USPS. Changing this might be out of scope of this PR. But I added a TODO for this string.

classes/class-wc-connect-service-schemas-store.php
```diff
@@ -195,5 +195,16 @@ public function get_predefined_packages_schema() {
 
 			return ( object ) $predefined_packages;
 		}
+
+		public function get_predefined_packages_schema_for_service( $service_id ) {
+			$package_schemas = $this->get_predefined_packages_schema();
+
+			if ( ! isset( $package_schemas )
```
jeffstieler: Since `get_predefined_packages_schema()` returns `null` or `object`, doesn't a `is_null()` check make more sense here?

classes/class-wc-connect-service-schemas-store.php
```diff
@@ -195,5 +195,16 @@ public function get_predefined_packages_schema() {
...
+			     || ! property_exists( $package_schemas, $service_id ) ) {
+				return array();
+			}
+
+			return $package_schemas->$service_id;
```
jeffstieler: Just a matter of style, but I'd prefer to see curly braces surrounding variable object properties. e.g. `$package_schemas->{$service_id}`.

classes/class-wc-connect-shipping-label.php
```diff
@@ -111,22 +117,44 @@ protected function get_selected_packages(WC_Order $order ) {
...
+
 			return $formatted_packages;
 		}
 
+		protected function get_flat_rate_packages_groups() {
+			$service_id = 'usps'; //TODO: remove hardcoding
+			$predefined_packages_schema = $this->service_schemas_store->get_predefined_packages_schema_for_service( $service_id );
+			$groups = array();
+
+			foreach ( $predefined_packages_schema as $groupId => $group ) {
```
jeffstieler: Please use snake case for PHP.

client/shipping-label/state/actions.js
```diff
@@ -102,6 +102,10 @@ export const submitStep = ( stepName ) => ( dispatch, getState, { storeOptions }
 	expandFirstErroneousStep( dispatch, getState, storeOptions, stepName );
 };
 
+const convertToApiPackage = ( pckg ) => {
+	return _.pick( pckg, [ 'id', 'box_id', 'service_id', 'length', 'width', 'height', 'weight' ] );
+};
+
```
jeffstieler: Just curious - why move this to the JS client side layer?

client/shipping-label/views/purchase/steps/packages/package-info.js
```diff
@@ -68,11 +68,29 @@ const PackageInfo = ( { packageId, selected, all, unpacked, dimensionUnit, weigh
...
+		const groups = _.reduce( flatRateGroups, ( result, groupTitle, groupId ) => {
+			result[ groupId ] = { title: groupTitle, definitions: {} };
+			return result;
+		}, {
+			custom: { title: __( 'Custom Packages' ), definitions: {} },
+		} );
+
+		_.forEach( all, ( groupPckg, groupPckgId ) => {
+			const group = groupPckg.group_id ? groups[ groupPckg.group_id ] : groups.custom;
+			group.definitions[ groupPckgId ] = groupPckg;
```
jeffstieler: Consider using `_.filter()` to build group definitions, and bring it into the above `_.reduce()` call. You should be able to determine group packages by: `_.filter( all, { group_id: groupId } )`.

jeffstieler:
I'd like to add - this tested really well!

client/shipping-label/state/actions.js
robobot3000: This functionality existed in both PHP (purchasing labels) and JS (requesting rates), so it's basically code deduplication. I kept the JS bit, as we seem to put more logic stuff on the JS side.

robobot3000:
Thanks for reviewing. I cleaned up the code, rebased, and fixed one issue.

PR #693, opened 2016-11-11
Add a "Phone Number" field to both the origin and destination address in the Labels UI

DanReyLop:
Fixes https://github.com/Automattic/woocommerce-connect-client/issues/574

You need to test it with the `add/phone_address` branch in the server.

**Background:**
EasyPost uses internally different APIs for ["regular" mail](https://www.usps.com/business/web-tools-apis/delivery-confirmation-domestic-shipping-label-api.htm) and for ["Priority Express" mail](https://www.usps.com/business/web-tools-apis/priority-mail-express-api.htm).

The normal API doesn't require the parameter `FromPhone` or `ToPhone`, but the Express API does (10-digit number, without punctuation). The solution is to add a `Phone number` field to the `Origin address` and `Destination address` form. This will make the phone field mandatory even for services that don't require it, but it's the simplest solution.

**Changes:**
* WooCommerce by default requires the _billing_ phone field to be filled, but there's no **shipping** phone. I've added that field.
* For orders done prior to this change, or for the ones that don't have an explicit *shipping address* (they only filled the Billing details) the phone number entered in the Billing Address is used instead.
* The cached merchant address for the Labels UI will no longer be valid (since it doesn't have a phone number), so the next time the merchant opens the Purchase Label UI he will have to input his phone (just once).
* The phone number validation is client-only. It uses [`libphonenumber`](https://github.com/seegno/google-libphonenumber) to pretty-print, normalize and validate the phone numbers.

**Possible things to improve**
* Right now the phone validator assumes that the phone number the merchant is entering will be of the same country as the address. This is a safe assumption for now, since the `USPS` API only accepts US-based phone numbers anyway. Do you think this could be a problem? Like, you're shipping from UK but your phone number is Canadian?
* The phone validation WooCommerce applies at checkout is really weak, it only checks that the entered string is a bunch of numbers. So it's possible that the merchant opens the Purchase Label UI and the entered phone is just garbage (and let's remember that's a required field). I guess that could also happen with an address...
* With this change, the phone number is required for every possible service. That means that if the merchant only wants to send items using `Priority Mail`, that field would not be needed at all. Maybe we should make it optional? Or maintain a list of services that require it? Or send the origin/destination phone along with the address in the `getRates` call and have some logic in the server to **not** return `Express` services if the phone is not present?

**To test**
* Buy something and go to the checkout.
* Check that you can fill the "Phone" field in the "Shipping Address".
* Go to the "Edit Order" page.
* Check that the phone number you entered appears on the Order Details.
* The "Origin Address" step should be marked as invalid, with the "Phone" field in red.
* Enter a phone number. Try with different phones, invalid phones, vanity phones, etc.
* Check that the phone you entered on the checkout appears in the "Destination Address" step.
* On the Rate selection step, choose a `Priority Express` service, and purchase the label(s).

@nabsul @allendav @robobot3000 @jeffstieler

### DIFF
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -291,14 +291,13 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 				return $headers;
 			}
 
-			set_time_limit( 0 ); // Disable PHP timeout (30 seconds default)
 			$args = array(
 				'headers' => $headers,
 				'method' => $method,
 				'body' => $body,
 				'redirection' => 0,
 				'compress' => true,
-				'timeout' => 60,
+				'timeout' => 20,
 			);
 			$args = apply_filters( 'wc_connect_request_args', $args );
 
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -291,13 +291,14 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 				return $headers;
 			}
 
+			set_time_limit( 0 ); // Disable PHP timeout (30 seconds default)
 			$args = array(
 				'headers' => $headers,
 				'method' => $method,
 				'body' => $body,
 				'redirection' => 0,
 				'compress' => true,
-				'timeout' => 20,
+				'timeout' => 60,
 			);
 			$args = apply_filters( 'wc_connect_request_args', $args );
 
--- a/client/lib/utils/phone-format.js
+++ b/client/lib/utils/phone-format.js
@@ -1,0 +1,27 @@
+import { PhoneNumberUtil, AsYouTypeFormatter } from 'google-libphonenumber';
+
+export const isValidPhone = ( phoneNumber, country ) => {
+	const phoneUtil = PhoneNumberUtil.getInstance();
+	try {
+		const parsedPhone = phoneUtil.parse( phoneNumber, country );
+		return phoneUtil.isValidNumber( parsedPhone );
+	} catch ( e ) {
+		return false;
+	}
+};
+
+export const getPlainPhoneNumber = ( phoneNumber, countryCode ) => {
+	if ( ! isValidPhone( phoneNumber, countryCode ) ) {
+		return phoneNumber;
+	}
+	return '' + PhoneNumberUtil.getInstance().parse( phoneNumber, countryCode ).getNationalNumber();
+};
+
+export const formatPhoneForDisplay = ( rawNumber, countryCode ) => {
+	const formatter = new AsYouTypeFormatter( countryCode );
+	let phoneNumber = '';
+	rawNumber.split( '' ).forEach( ( char ) => {
+		phoneNumber = formatter.inputDigit( char );
+	} );
+	return phoneNumber;
+};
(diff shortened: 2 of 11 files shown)


client/lib/utils/phone-format.js
```diff
@@ -0,0 +1,27 @@
+import { PhoneNumberUtil, AsYouTypeFormatter } from 'google-libphonenumber';
+
+export const isValidPhone = ( phoneNumber, country ) => {
+	const phoneUtil = PhoneNumberUtil.getInstance();
```
jeffstieler: Is it possible to hang on to an instance of this instead?

client/lib/utils/phone-format.js
```diff
@@ -0,0 +1,27 @@
...
+	} catch ( e ) {
+		return false;
+	}
+};
+
+export const getPlainPhoneNumber = ( phoneNumber, countryCode ) => {
+	if ( ! isValidPhone( phoneNumber, countryCode ) ) {
+		return phoneNumber;
+	}
+	return '' + PhoneNumberUtil.getInstance().parse( phoneNumber, countryCode ).getNationalNumber();
```
jeffstieler: Same instance question.

client/shipping-label/state/selectors/errors.js
```diff
@@ -10,15 +11,19 @@ const getAddressErrors = ( { values, isNormalized, normalized, selectNormalized
...
+	const requiredFields = [ 'name', 'phone', 'address', 'city', 'postcode', 'country' ];
 	const errors = {};
 	requiredFields.forEach( ( field ) => {
 		if ( ! values[ field ] ) {
 			errors[ field ] = __( 'This field is required' );
 		}
 	} );
 
+	if ( ! isValidPhone( phone, country ) ) {
+		errors.phone = __( 'Invalid phone number' );
```
jeffstieler: What do you think about adding country to this message? In testing I tried a UK number, but it formatted as US, and then said it was invalid.

classes/class-wc-connect-api-client.php
```diff
@@ -291,13 +291,14 @@ protected function request( $method, $path, $body = array() ) {
...
 
+			set_time_limit( 0 ); // Disable PHP timeout (30 seconds default)
 			$args = array(
 				'headers' => $headers,
 				'method' => $method,
 				'body' => $body,
 				'redirection' => 0,
 				'compress' => true,
-				'timeout' => 20,
+				'timeout' => 60,
```
DanReyLop: The `set_time_limit()` option only affects the current request. What happened is that I put the "request timeout" to 60 seconds, but since the default PHP timeout (in my machine) was 30 seconds, the process was killed anyways, so the `set_time_limit()` call was needed. Do you think it would be better if I put an explicit PHP timeout, like `set_time_limit(60)` instead of infinite?

This change wasn't needed for this PR, but testing those changes I run into that issue (maybe EasyPost was having a bad day, or maybe the "Express" USPS API is slower than the regular one). I can put it in a separate PR if you want.

client/lib/utils/phone-format.js
DanReyLop: No need for that. I've checked the source and it looks like it's using the "typical" Java Singleton pattern: https://github.com/google/closure-library/blob/969c7e53decf7b6ee3b2dd14011663d0762a1e82/closure/goog/base.js#L755

So after the first time `getInstance()` is called, the subsequent calls are fast, the `PhoneNumberUtil` object isn't being re-initialized again and again.

classes/class-wc-connect-api-client.php
DanReyLop: Done: https://github.com/Automattic/woocommerce-connect-client/pull/703

PR #698, opened 2016-11-12
Added an spinner to the "Purchase" button while the request is being processed

### LINKED ISSUE #603: Connect-a-thon: Unsatisfying response to clicking on Buy & Print button
A few people in connect-a-thon remarked that when they clicked on the Buy & Print button, they were expecting something more since nothing (other than disabling the button) happens for **8 seconds**.  We might need an animation or something there?

8-seconds-of-silence

DanReyLop:
Fixes #603 

When the Purchase Label form is submitting, instead of just disabling the "Purchase" button, change its text and show a progress indicator (a Spinner).

Screenshot:
submitting

To test, just go purchase a Label.

The only doubt I have about this one is maybe using `Buying...` instead of `Purchasing...`, since the original button label is `Buy & Print N Labels`.

@nabsul @allendav @jeffstieler @robobot3000

### DIFF
--- a/client/shipping-label/views/purchase/index.js
+++ b/client/shipping-label/views/purchase/index.js
@@ -14,7 +14,7 @@ const PrintLabelDialog = ( props ) => {
 	const currencySymbol = props.storeOptions.currency_symbol;
 
 	const getPurchaseButtonLabel = () => {
-		if ( true || props.form.isSubmitting ) {
+		if ( props.form.isSubmitting ) {
 			return (
 				<div>
 					<Spinner size={ 24 } className="wcc-shipping-label-dialog__button-spinner" />
--- a/client/shipping-label/views/purchase/index.js
+++ b/client/shipping-label/views/purchase/index.js
@@ -14,11 +14,11 @@ const PrintLabelDialog = ( props ) => {
 	const currencySymbol = props.storeOptions.currency_symbol;
 
 	const getPurchaseButtonLabel = () => {
-		if ( props.form.isSubmitting ) {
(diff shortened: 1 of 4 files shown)


client/shipping-label/views/purchase/index.js
```diff
@@ -13,6 +14,14 @@ const PrintLabelDialog = ( props ) => {
 	const currencySymbol = props.storeOptions.currency_symbol;
 
 	const getPurchaseButtonLabel = () => {
+		if ( props.form.isSubmitting ) {
+			return (
+				<div>
```
kellychoffman: Can we expand on this and change add an additional className to the button?

kellychoffman:
I found a work around for the button style issue. This is good to 🚢  and I can open an issue if this ends up not being enough.

DanReyLop:
@kellychoffman So, are you OK with the label being `Purchasing...`? As I said in Slack, `Buying...` didn't feel right either. One suggestion was to move the Spinner to the right and make the label be `Buy & Print N labels` (without the price) so the button width and arrangment doesn't change much.

Also, I went ahead and made a couple minor adjustments to your code, hope you don't mind. We use `autoprefixer` so the `-webkit` prefixes are not needed, and you forgot the `true || ` you added while testing (I did the exact same thing 😄 )

kellychoffman:
> So, are you OK with the label being Purchasing...? 

Yeah, I think this is the clearest label to use for this state.

> forgot the true || you added while testing

Ooops! Thanks for fixing!

PR #702, opened 2016-11-14
Status page: Make the Jetpack error more clear

kellychoffman:
Found this while testing the plugin on one of my live sites. I didn't have Jetpack and there was no clear indication that I needed to. I was linked to the Status page and saw this error:

> Please connect Jetpack to WordPress.com

It doesn't make much sense to someone who doesn't have Jetpack (or maybe hasn't heard of it). I updated the message to say: 

> Jetpack is not connected to WordPress.com. Make sure the Jetpack plugin is installed, activated, and connected.

### DIFF
--- a/classes/class-wc-connect-help-view.php
+++ b/classes/class-wc-connect-help-view.php
@@ -123,7 +123,7 @@ if ( ! class_exists( 'WC_Connect_Help_View' ) ) {
 					'jetpack_indicator',
 					'notice',
 					'indicator-error',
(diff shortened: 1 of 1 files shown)


allendav:
> Should we add a link to the Jetpack plug-in page?

Kinda - let's add a link that puts them on the Add Plugin page on their own site with Jetpack as the search.  Let's also do this as a separate issue and ship this one :)  https://github.com/Automattic/woocommerce-connect-client/issues/710

PR #703, opened 2016-11-15
Put a higher remote request timeout (1 minute) in the server, disable default PHP script timeout.

DanReyLop:
As surfaced in this discussion: https://github.com/Automattic/woocommerce-connect-client/pull/693#discussion_r87910303

We need to raise the configured REST request timeout (from merchant host to WCC server) because 20 seconds is too low for complex operations like multiple label purchasing. The problem is, the default PHP timeout (for a whole PHP script execution) is 30 seconds in my machine (I assume I don't have a very exotic configuration), and that can be too low too, so we need to at least try to increase the PHP timeout.

@jeffstieler said: I'm not sure we want to mess with PHP timeout

My reponse:
The `set_time_limit()` option only affects the current request. What happened is that I put the "request timeout" to 60 seconds, but since the default PHP timeout (in my machine) was 30 seconds, the process was killed anyways, so the `set_time_limit()` call was needed. Do you think it would be better if I put an explicit PHP timeout, like `set_time_limit(60)` instead of infinite?

Either way, it makes sense to separate this issue into a different PR.

@jeffstieler @robobot3000 @allendav @nabsul

### DIFF
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -291,14 +291,15 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 				return $headers;
 			}
 
-			set_time_limit( 0 ); // Disable PHP timeout (30 seconds default)
+			$http_timeout = 60; // 1 minute
+			wc_set_time_limit( $http_timeout + 10 );
 			$args = array(
 				'headers' => $headers,
 				'method' => $method,
 				'body' => $body,
 				'redirection' => 0,
 				'compress' => true,
-				'timeout' => 60,
+				'timeout' => $http_timeout,
 			);
 			$args = apply_filters( 'wc_connect_request_args', $args );
 
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -291,13 +291,14 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 				return $headers;
 			}
 
+			set_time_limit( 0 ); // Disable PHP timeout (30 seconds default)
 			$args = array(
 				'headers' => $headers,
 				'method' => $method,
 				'body' => $body,
 				'redirection' => 0,
 				'compress' => true,
-				'timeout' => 20,
+				'timeout' => 60,
 			);
 			$args = apply_filters( 'wc_connect_request_args', $args );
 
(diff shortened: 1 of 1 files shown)


DanReyLop:
Relevant: [The docs for `set_time_limit`](http://php.net/manual/en/function.set-time-limit.php)

Even if it fails (because lack of permissions, or whatever), it doesn't crash the process, it just returns `FALSE`, so it works as a "best-effort" solution.

allendav:
As I read set_time_limit, it sets the time limit for execution of the entire PHP script, not just the next request.  We may need it AND we will probably also need to do something with wp_remote_post options ( https://codex.wordpress.org/Function_Reference/wp_remote_post )

See also http://wordpress.stackexchange.com/questions/154248/error-timed-out-with-succesfull-wp-remote-post

DanReyLop:
> As I read set_time_limit, it sets the time limit for execution of the entire PHP script, not just the next request

Yes, that's what I meant. `set_time_limit` sets the limit for the currently running PHP script (I'm just used to call PHP executions "requests").

The `wp_remote_*` options stuff is already taken care of here: https://github.com/Automattic/woocommerce-connect-client/pull/703/files#diff-57dabbbbd8e057f1082af23aa58d3f5aR301
Not sure what is the difference between this and what you found in StackOverflow, but both solutions work :)

allendav:
OMG @DanReyLop - I'm sorry - i missed the 60 somehow - ty. Will try this now.

allendav:
Let's use woocommerce's set_time_limit wrapper - https://github.com/woocommerce/woocommerce/commit/a40fb29eca050436ba6757e92fcc1f7fd59ccd8d

allendav:
This does work pretty sweetly - if you add a 45 second delay into the server master blows up but this PR works :)

Let's just switch to woocommerce's set_time_limit wrapper and then this is good to go.

classes/class-wc-connect-api-client.php
```diff
@@ -291,13 +291,15 @@ protected function request( $method, $path, $body = array() ) {
 				return $headers;
 			}
 
+			$http_timeout = 60; // 1 minute
+			wc_set_time_limit( $http_timeout + 10 );
```
allendav: Since WC could be deactivated independently of us, please test for the function before calling it

DanReyLop: Hummm... you're right. I've been assuming everywhere that WooCommerce is available. I wonder in how many ways our plugin 💥 when you deactivate WooCommerce.

allendav:
:shipit:

PR #704, opened 2016-11-15
Since all the dates stored in the server and the merchant host are UTC-based, add the "UTC" timezone before parsing them for display

### LINKED ISSUE #687: Label purchase date display appears to be using incorrect timezone
Observed on our test pressable site using the staging server.  I double checked and my time zone is correctly set in my WordPress.com preferences to Los Angeles.  I expected the purchase date to say 4:19 PM and not 12:19 AM (on the next day - basically suggesting UTC+0/GMT)

screen shot 2016-11-10 at 4 21 00 pm

DanReyLop:
Fixes #687 

All the dates stored in the WCC server and returned by it are in the `YYYY-MM-DD HH:MM:SS` format. Note that there's no timezone information in the date string, so `UTC` is assumed by default when the date is serialized, but local timezone is assumed when parsing a date without an explicit timezone set.

We have 2 options:
1) Making sure that in the future, dates are stored with their corresponding timezone information.
2) Keep assuming that all dates are UTC, and make sure they are treated accordingly just for display purposes.

Solution `2)` turned out to be far easier (plus, it's retroactive). In the client, before parsing a date string to be displayed, it's enough to add the ` UTC` suffix so that date is interpreted as UTC instead of local time.

To test:
* Make sure you are in a non-UTC timezone. If you are in a UTC timezone, catch a plane and go somewhere else. Alternatively, you can change the timezone setting in your computer and refresh the page.
* Purchase a label.
* In the meta-box, check that the `Purchase date` shown is correct (it matches the local time when you purchased it).

Fun fact: I live in the `UTC` timezone. I would have never spotted this issue 😄 

@jeffstieler @nabsul @allendav @robobot3000

### DIFF
--- a/client/lib/utils/format-date.js
+++ b/client/lib/utils/format-date.js
@@ -1,7 +1,7 @@
 export default ( date ) => {
 	// Note: Safari ignores the 'locale' and 'options' argument. It will display
 	// the full date (with timezone and everything) in the browser's locale (instead of WP lang)
-	return new Date( date ).toLocaleDateString( document.documentElement.lang, {
+	return new Date( date + ' UTC' ).toLocaleDateString( document.documentElement.lang, {
 		day: 'numeric',
 		month: 'long',
 		year: 'numeric',


allendav:
Hey @DanReyLop - this doesn't seem to be working correctly. I get the following:

screen shot 2016-11-15 at 3 54 55 pm

We store the time in order meta in milliseconds, and JS gets that too to milliseconds, e.g.:

screen shot 2016-11-15 at 3 58 32 pm

Whatcha think?

allendav:
@DanReyLop - bump

DanReyLop:
Yes, I didn't forget about this. I've been thinking on and off about the best solution in the long run. I think we should standarize all our date representations to use timestamps instead of date strings, and only "stringify" them for display. With the database already populated, it won't be a trivial change. I'll keep trying to figure this out while I look at other issues.

DanReyLop:
I'm closing this, since I've implemented a server-side only solution.

PR #716, opened 2016-11-17
Fixed the incorrect service counts when predefined packages are disabled

### LINKED ISSUE #715: Enabled services count incorrect after changing predefined packages in package manager
screen shot 2016-11-16 at 12 17 58 pm

To reproduce:
* Select all USPS Priority Mail services in one shipping zone instance
* Save
* In Package Manager, only enable "large flat rate box" under Priority
* Go back to shipping zone instance settings form
* Uncheck one service
* Verify that count is higher than the amount of visible selected services

robobot3000:
Fixes #715 

I fixed this by still showing the flat rate services when the corresponding flat rate package has been disabled, but instead marking them as disabled as well:
screen shot 2016-11-17 at 13 30 45
The disabled rows have a `title` attribute that reads: `You can manage this service after enabling the corresponding package in the Packaging Manager`
CC @allendav to approve this change. I can always revert back to the simple fix for the service count

CC @jeffstieler

### DIFF
--- a/client/settings/views/services/entry.js
+++ b/client/settings/views/services/entry.js
@@ -4,6 +4,7 @@ import FormSelect from 'components/forms/form-select';
 import Gridicon from 'components/gridicon';
 import classNames from 'classnames';
 import NumberInput from 'components/number-field/number-input';
+import { translate as __ } from 'lib/mixins/i18n';
 
 const ShippingServiceEntry = ( props ) => {
 	const {
@@ -11,6 +12,7 @@ const ShippingServiceEntry = ( props ) => {
 		updateValue,
 		errors,
 		service,
+		isManageable,
 	} = props;
 
 	const {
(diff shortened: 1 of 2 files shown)


allendav:
I think this is a great idea cc @kellychoffman

jeffstieler:
Does it make sense to change the "X service(s) selected" label now? In your example @robobot3000, 10 are selected, but only 5 are "active"...

robobot3000:
@jeffstieler I pushed a change that fixes the count to only include the manageable services, and the "Select all" checkbox only affects the manageable services as well.

client/settings/views/services/entry.js
```diff
@@ -23,23 +25,24 @@ const ShippingServiceEntry = ( props ) => {
 	const hasError = errors[ service.id ];
 
 	return (
-		<div className={ classNames( 'wcc-shipping-service-entry', { 'wcc-error': hasError } ) }>
+		<div className={ classNames( 'wcc-shipping-service-entry', { 'wcc-error': hasError } ) } title={ disabled ? __( 'You can manage this service after enabling the corresponding package in the Packaging Manager' ) : '' }>
 			<label className="wcc-shipping-service-entry-title">
 				<FormCheckbox
-					checked={ enabled }
+					checked={ enabled && ! disabled }
```
DanReyLop: Please come up with more descriptive variable names. I realize that they are technically correct, but this line hurts my brain.

client/settings/views/services/group.js
```diff
@@ -32,14 +32,16 @@ const ShippingServiceGroup = ( props ) => {
...
-	const summary = summaryLabel( services );
 	const actionButton = (
 		<button className="foldable-card__action foldable-card__expand" type="button">
 			<span className="screen-reader-text">{ __( 'Expand Services' ) }</span>
 			<Gridicon icon="chevron-down" size={ 24 } />
 		</button>
 	);
-	const allChecked = _.every( services, ( service ) => service.enabled );
+	//filter out the services that have a disabled corresponding predefined package
+	const manageableServices = _.filter( services, ( service ) => ! service.predefined_package || predefinedPackages.includes( service.predefined_package ) );
```
DanReyLop: No need to use `lodash` if `services` is an array. Just use `Array.prototype.filter()`.

PR #717, opened 2016-11-17
Logic to auto-accept address normalizations if the response has a special flag

### LINKED ISSUE #600: Connect-a-thon: If address normalization changes are trivial, auto accept the changes
If the difference between the unnormalized address and normalized address is any of the following, automatically accept the address without requiring merchant review.
- capitalization
- ZIP/ZIP+4 (so long as the 5 digit ZIP code matches, let's not worry about the 4 digit ZIP+4 being added)
- common PUB 28 Appendix C suffixes (Avenue/AVE, Boulevard/BLVD, Circle/CIR, Court/CT, Drive/DR, Lane/LN, Place/PL, Road/RD and Street/ST) (when they are at the end of the address only, not mid-string)

Feel free to add others that make sense

ref: http://pe.usps.com/cpim/ftp/pubs/Pub28/pub28.pdf

DanReyLop:
Fixes #600 

This PR is totally backwards and forwards-compatible, but you won't be able to test it properly without the `update/auto-accept-address-normalization` branch on the server.

When sending an address to normalize to the server, if the response has the `is_trivial_normalization` property set to `true`, then the normalized address will be automatically accepted without user intervention. As its name indicates, that flag will be set when the changes made by the address normalizator were trivial, such as adding the `+4` portion to a ZIP code, or changing capitalization, or changing `street` to `st` for example.

To test:
* Go purchase a label.
* Make sure the origin address is normalized (the step has the green checkmark icon).
* Make *only* trivial changes, such as:
  * Change some things to lowercase or mixed case.
  * Remove the `+4` part of the ZIP code.
  * Change the address suffix ('RD', 'ST', etc) to its non-abbreviated form ('road', 'street', etc).
* Click the "Use this address" button and check that the form auto-skips to the next step, and the address was corrected automatically.
* Make a different change, such as mispelling the City name, and check that you're prompted to select the original address or the corrected address.

@nabsul @jeffstieler @allendav @robobot3000

### DIFF
--- a/classes/class-wc-rest-connect-shipping-label-controller.php
+++ b/classes/class-wc-rest-connect-shipping-label-controller.php
@@ -61,6 +61,8 @@ class WC_REST_Connect_Shipping_Label_Controller extends WP_REST_Controller {
 	public function update_items( $request ) {
 		$request_body = $request->get_body();
 		$settings = json_decode( $request_body, true, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
+		$this->settings_store->update_origin_address( $settings[ 'origin' ] );
+
 		$order_id = $settings[ 'order_id' ];
 
 		$settings[ 'payment_method_id' ] = $this->settings_store->get_selected_payment_method_id();
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -123,7 +123,7 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 		}
 
 		public function update_origin_address( $address ) {
-			return update_option( 'wc_connect_origin_address', ( array ) $address );
+			return update_option( 'wc_connect_origin_address', $address );
 		}
 
 		protected function sort_services( $a, $b ) {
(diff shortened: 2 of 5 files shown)


classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -61,6 +61,8 @@ public function register_routes() {
 	public function update_items( $request ) {
 		$request_body = $request->get_body();
 		$settings = json_decode( $request_body, true, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
+		$this->settings_store->update_origin_address( $settings[ 'origin' ] );
```
jeffstieler: Why did you move the call to update the store's origin address to the purchase step?

DanReyLop: Before:
* Merchant enters his address.
* Submit address for normalization.
* Server returns normalized address, and **that** address is stored in the settings.
* Merchant is a rebel and picks the un-normalized address anyway.
* Merchant purchases a label.

In that case, the stored address would be the normalized one, even though the merchant explicitly selected the other one (as Kelly told me, it happens that sometimes the address normalization services give you an address that's plain wrong). The next time the merchant opens the Labels UI modal, the `Origin` step would be collapsed and with a ✔️ so he won't realise that the address is "wrong".

It's better to only store the address when the merchant is actually purchasing a label for that address. Not 100% related with this issue but I noticed that odd behaviour while implementing it.

allendav:
Tests well. :shipit: 

@jeffstieler - if you like we can re-open concerning your comment

PR #719, opened 2016-11-21
Fix the Add Package dialog height

### LINKED ISSUE #711: Add (custom) package dialog layout is very borked
screen shot 2016-11-15 at 4 40 01 pm

cc @robobot3000 @kellychoffman

robobot3000:
Fixes #711

In the first commit I fixed the heights of the modals to properly render the Add Package dialog.

In the second commit I used the CSS3 `calc` method to always add a scrollbar to the dialogs when their contents are too high. I checked all the dialogs that we render and it works fine on Chrome, Safari, Opera and Firefox.

CC @allendav @kellychoffman


robobot3000:
The height is controlled by the position properties. The problem affects a bunch of other modals, especially in the editable packaging UI. With some CSS3 this could be handled a bit better per dialog, but I think it's out of scope of this PR

PR #728, opened 2016-11-23
"0" and "0.0" dimensions are treated differently for shipping rate requests

### LINKED ISSUE #724: "0" and "0.0" dimensions are treated differently for shipping rate requests
I've noticed if you set the length of a product to "0" and add that item to a cart and try to get rates, you don't get any rates returned. However, if you set the length of the product to "0.0", the rate request goes out to the server and causes an error on the server side (also resulting in no results).

image

nabsul:
Closes #724 

Something fishy is going on in the WooCommerce product object. the function objects `get_length()`, `get_width()`, `get_height()`are returning an empty string when the value of the dimension is "0". This doesn't seem to happen if the value is "0".

To mitigate this issue I'm forcing whatever these functions return into a float using `floatval()`. I thought about putting some error checking and logging in place for this situation, but decided to just go with the simplest solution.

I've submitted a report to WooCommerce Core:
https://github.com/woocommerce/woocommerce/issues/12455

To test:
- Try setting a product dimension to "0" from some non-zero value (this seems to be important)
- Add that product to the cart and try requesting shipping rates for that product
- Observe that the current client from master returns a 400
- Observe that with this change rates are successfully fetched

@allendav @DanReyLop @jeffstieler @robobot3000

### DIFF
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -97,25 +97,25 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 						);
 					}
 
-					$weight = floatval( $product->get_weight() );
+					$weight = $product->get_weight();
 
 					$height = 0;
 					$length = 0;
 					$width = 0;
 
 					if ( $product->has_dimensions() ) {
-						$height = floatval( $product->get_height() );
-						$length = floatval( $product->get_length() );
-						$width  = floatval( $product->get_width() );
+						$height = $product->get_height();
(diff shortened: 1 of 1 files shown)


allendav:
Would there be value in making our float approach consistent? We cast elsewhere ( e.g. https://github.com/Automattic/woocommerce-connect-client/blob/c9ac106dc5b9e8592c8f8795bcc8058be11cbb54/classes/class-wc-connect-shipping-label.php#L49 )

allendav:
:shipit:

PR #732, opened 2016-11-25
Tweak the WebPack config so the bundle builds faster

DanReyLop:
There's no issue for this because this was part of my "exploring Nightwatch.js" work.

Tweaked and cleaned-up the `webpack.config.js` file.
Clean-ups:
* Move `eslint` to a preLoader
* Be consistent in the use of `path.resolve`
* Remove unneeded properties
* Make the `karma.conf.js` file use the whole `webpack` config, and only change the bits that are needed for Karma

Improvements:
* Only run `eslint` in our own code
* Only use `Babel` in the code that needs it (our code and Calypso's). This wasn't working before because the loader's `include` property matches against the **absolute** path of a file, so that means that if you cloned the repo in `/Users/Whatever/Projects/woocommerce-connect-client`, **all** the possible files, even the ones in `node_modules`, match the RegExp `/(client|wp-calypso)/`.

Incidentally, now the unit tests work. Before, at least in Windows, there was an error related with `browserify-crypto`.

To test:
* Run `npm start`
* Notice that everything works, and that the `npm start` command executed faster
* Run `npm test`, the test should execute
* Run `npm run dist`, the bundles should be generated and should be correct

@nabsul @robobot3000 (All the other ones are digesting 🦃)


DanReyLop:
I cleaned-up a little more. We were using `isparta-instrument-loader` for code coverage, but that package is no longer maintained. I changed it for `isparta-loader`, with some tweaks to make code coverage work even for files that are not tested at all.

@allendav @jeffstieler

robobot3000:
Looks good, :shipit:

Issue #563, opened 2016-09-19
Dev: Refine Purchase Label Meta Box


allendav:
Implement the designs from #562 once they become available.

allendav:
Needs estimate first

robobot3000:
Assigning an estimate of 5 - looking at the designs this could require some server work.

Issue #562, opened 2016-09-19
Design: Refine Purchase Label Meta Box


allendav:
@kellychoffman - I was looking for mocks for the purchase label meta box and could not find them - if they exist please paste them here and then we'll close this issue.

The Shipping Label meta box should include:
- a Create Label button if a label has not yet been purchased for the order
- a Reprint Label button for each label purchased in the order
- a Refund Label button for each label purchased in the order and the label was not purchased more than 30 days ago
- a warning that refunds will take at least 14 days to process (this idea was proposed during Happiness Academy)
- a reverse chronologically sorted action log of create, reprint, and refund actions with date and username of action initiator
- the state of any refund label request (pending, refunded, declined)
- a spot for tracking information for each label that supports it ( see also #482 )

The design will take multiple packages/labels into account and make it easy to ensure you are taking the action on the specific label you want to.

The ideal design will hide the complete action log and just show the most recent actions - with some means of expanding to show the complete log.

allendav:
Please also include price paid for each label

allendav:
Please also consider how this might work with order notes, either as part of order notes or by mimicking/improving the UX of order notes

jameskoster:
Here's my first pass at this. 

Default meta box is basically the same as it is now;

![meta-box](https://cldup.com/7yhFiWGGkX.png)

Here's how it looks with content:

![meta-box-content](https://cldup.com/XTlNn9_wS4.png)

'Label Identifier' is a placeholder. Not sure what text we'd use there?

I decided to separate the log into a separate box as it seems kinda situational. This way it can be hidden entirely or moved somewhere less prominent if desired.

I assume that refunded labels cannot be reprinted? Could be wrong though xD

Definitely keen for feedback on this @kellychoffman

kellychoffman:
> Here's how it looks with content:

Very cool. One concern I have is with the relative size and importance of the print action. Since this is the view for labels that have already been purchased and printed, I think we should make this less prominent. Maybe just as a button labeled 'Reprint'? Or 'Print' if there is there a scenario in which the label never printed, but the 'Create new label' modal has been closed. cc-ing @allendav here for feedback here. Any other scenarios I'm missing? 

> 'Label Identifier' is a placeholder. Not sure what text we'd use there?

Would this link you back to the label? Maybe we could get rid of the label identifier all together and just rely on the Reprint button. 

[...]

PR #734, opened 2016-11-28
Add refined label box

robobot3000:
Adds #563 

Please test with the latest `master` server checked out.

* Visual changes
* More info added to the label boxes, including the refund status, label service, package and its contents
* Added a history meta box logging purchases, refunds and reprints

CC @allendav @kellychoffman @DanReyLop @nabsul @jeffstieler


robobot3000:
Screenshot:
screen shot 2016-11-29 at 14 22 30

Test tips:
* purchase a bunch of labels
* request a refund
  * _some refunds return an error saying that the parcel has been shipped or the refund has already been requested. @jeffstieler - is this something to be worried about? This PR doesn't change the body of the refund request, and I'm assuming this is because the label IDs are reused in the test environment_
* in the server DB, modify the rows in `wcc_shipping_label_refund` table for your pending refunds. Set the `status` value to 1 for pending, 3 for rejected, 6 for complete (+ set the `refund_date` value)
* refresh the page and you should see the status
  * note that only the status for labels that are marked as pending is refreshed. If the refund has been marked as complete or rejected once, it won't be updated from the server
* confirm that everything works as expected

jeffstieler:
> some refunds return an error saying that the parcel has been shipped or the refund has already been requested. @jeffstieler - is this something to be worried about? This PR doesn't change the body of the refund request, and I'm assuming this is because the label IDs are reused in the test environment

@robobot3000 the "parcel has been shipped" will occur randomly, and is expected. EasyPost just updates any refunds in the test enviroment with random statuses.

You should *not* be seeing "refund already requested" messages for new labels. That is based on the label ID being found in the refunds table. Can you reliably reproduce this?

robobot3000:
> You should not be seeing "refund already requested" messages for new labels. That is based on the label ID being found in the refunds table. Can you reliably reproduce this?

It happens only when I request a refund right after getting the "parcel has been shipped" message. I assume it's to be expected as well

jeffstieler:
> It happens only when I request a refund right after getting the "parcel has been shipped" message. I assume it's to be expected as well

Yep - that's expected if you're requesting a refund again for the same label.

jeffstieler:
Just in testing (haven't gotten into code yet):

1. I tried to refund a label, and got the error that the package had already been shipped. The label history still showed that I refunded the label.

Can we check the response of the refund request before adding to history? It might still make sense to add an item to the history, but it should indicate that the refund was *not* successful.

2. When I take an action that gets added to the history, the immediate UI update just says "User" instead of my username:

screen shot 2016-12-01 at 10 36 59 am

robobot3000:
0. I added label IDs to the UI, so it's easier to read the history modal and see which labels had actions performed on them
1. I changed the history entry to appear as `<user> unsuccessfully requested a refund of Label #95` when the request errors out, and similar when the request was successful
2. The reprint history entry should now display the name/url properly without the need for page refresh

classes/class-wc-connect-service-settings-store.php
```diff
@@ -122,6 +122,71 @@ public function update_label_order_meta_data( $order_id, $new_label_data ) {
...
+			return $this->decorate_history_entry( $entry );
+		}
+
+		public function get_label_history( $order_id ) {
+			$raw_history_data = get_post_meta( ( int ) $order_id, 'wc_connect_label_history', true );
+			if ( $raw_history_data ) {
+				$history_data = json_decode( $raw_history_data, true, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
+			} else {
+				$history_data = array();
+			}
```
jeffstieler: This looks like a candidate for DRY. (it's repeated from above)

classes/class-wc-connect-shipping-label.php
```diff
@@ -301,27 +301,34 @@ public function meta_box( $post ) {
...
 			) );
 
-			$store_options = $this->settings_store->get_store_options();
 			$root_view = 'wc-connect-create-shipping-label';
 			$admin_array = array(
 				'purchaseURL'             => get_rest_url( null, '/wc/v1/connect/label/purchase' ),
 				'addressNormalizationURL' => get_rest_url( null, '/wc/v1/connect/normalize-address' ),
 				'getRatesURL'             => get_rest_url( null, '/wc/v1/connect/shipping-rates' ),
-				'labelStatusURL'          => get_rest_url( null, '/wc/v1/connect/label/' . $order->id . '-%d' ),
+				'labelStatusURL'          => get_rest_url( null, '/wc/v1/connect/label/' . $order->id . '-%d-%d' ),
```
jeffstieler: It looks like you're following the existing pattern, but why are these endpoint parameters part of the path? Why aren't we using a query string here, or posting some JSON?

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -101,21 +109,48 @@ public function update_items( $request ) {
...
+				$label_meta[ 'package_name' ] = __( 'Individual packaging', 'connectforwoocommerce' );
+			} else if ( isset( $package_lookup[ $box_id ] ) ) {
+				$label_meta[ 'package_name' ] = $package_lookup[ $box_id ][ 'name' ];
+			} else {
+				$label_meta[ 'package_name' ] = __( 'Unknown package', 'connectforwoocommerce' );
+			}
+
+			$product_names = array();
+			foreach ( $package[ 'products' ] as $product_id ) {
+				$product = $product_factory->get_product( $product_id );
```
jeffstieler: You should be able to use [`wc_get_product()`](https://docs.woocommerce.com/wc-apidocs/function-wc_get_product.html) here.

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -89,8 +91,14 @@ public function update_items( $request ) {
...
 
+		$label_ids = array();
 		$labels_order_meta = array();
-		$labels_data = array();
+		$existing_labels_data = get_post_meta( $order_id, 'wc_connect_labels', true );
+		if ( $existing_labels_data ) {
+			$labels_order_meta = json_decode( $existing_labels_data, true, WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH );
+		}
+		$package_lookup = $this->settings_store->get_package_lookup_for_service( $carrier );
+		$product_factory = new WC_Product_Factory();
```
jeffstieler: This should be unnecessary, see comment below.

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -101,21 +109,48 @@ public function update_items( $request ) {
...
 			);
+
+			$package = $settings[ 'packages' ][ $index ];
+			$box_id = $package[ 'box_id' ];
+			if ( 'individual' === $box_id ) {
+				$label_meta[ 'package_name' ] = __( 'Individual packaging', 'connectforwoocommerce' );
+			} else if ( isset( $package_lookup[ $box_id ] ) ) {
+				$label_meta[ 'package_name' ] = $package_lookup[ $box_id ][ 'name' ];
+			} else {
+				$label_meta[ 'package_name' ] = __( 'Unknown package', 'connectforwoocommerce' );
```
jeffstieler: Can you explain the circumstances that lead to an "unknown" package?

classes/class-wc-connect-service-settings-store.php
```diff
@@ -122,6 +122,71 @@ public function update_label_order_meta_data( $order_id, $new_label_data ) {
...
+		private function decorate_history_entry( $entry ) {
+			$user_id = $entry[ 'user_id' ];
+
+			$entry[ 'user_name' ] = $this->get_user_name( $user_id );
+			$entry[ 'user_url' ] = get_edit_user_link( $user_id );
+
+			return $entry;
+		}
+
+		public function update_label_order_history( $order_id, $label_ids, $entry_type, $time ) {
```
jeffstieler: All uses of this pass in `$time = time() * 1000`. Perhaps this can just be an optional argument with a default?

Additionally, what do you think about handling both single `$label_id` and an array of `$label_ids`? That way, function calls don't have to wrap `array( $a_single_label_id )`.

classes/class-wc-rest-connect-shipping-label-status-controller.php
```diff
@@ -22,7 +22,7 @@ class WC_REST_Connect_Shipping_Label_Status_Controller extends WP_REST_Controlle
 	 *
 	 * @var string
 	 */
-	protected $rest_base = 'connect/label/(?P<order_id>\d+)-(?P<label_id>\d+)';
+	protected $rest_base = 'connect/label/(?P<order_id>\d+)-(?P<label_id>\d+)-(?P<get_refund>\d+)';
```
jeffstieler: Same question about path vs. query string.

client/shipping-label/views/history.js
```diff
@@ -0,0 +1,79 @@
...
+	const renderUserName = ( entry ) => {
+		if ( entry.user_name ) {
+			return <a href={ entry.user_url }>{ entry.user_name }</a>;
+		}
+
+		return <span>{ __( 'User' ) }</span>;
+	};
+
+	const renderHistoryEntry = ( entry, index ) => {
+		const entryMoment = CalypsoI18n.moment( entry.time ).fromNow();
```
jeffstieler: Are we using Calypso's `moment()` anywhere else? I'm not sure we should be here.. from what I understand, `i18n-calypso` exposes a `moment` instance with a locale already set, but if we aren't using `i18n-calypso` for all translations in the project - can't this be a mismatch?

Why not use `moment` directly?

client/shipping-label/views/history.js
```diff
@@ -0,0 +1,79 @@
...
+import Gridicon from 'components/gridicon';
+import formatDate from 'lib/utils/format-date';
+import { translate as __ } from 'lib/mixins/i18n';
+import { sprintf } from 'sprintf-js';
+const LabelHistory = ( { labelsHistory } ) => {
+	if ( ! labelsHistory || ! labelsHistory.length ) {
+		document.getElementById( 'woocommerce-order-label-history' ).style.display = 'none';
+		return null;
+	}
+	document.getElementById( 'woocommerce-order-label-history' ).style.display = 'block';
```
jeffstieler: _This comment applies to both uses of `style.display =` above_

Shouldn't this be applied with a class that gets dynamically set based on the value of `labelHistory`?

client/shipping-label/views/index.js
```diff
@@ -1,112 +1,222 @@
...
+							{ label.product_names.map( ( productName, productIdx ) => <li key={ productIdx }>{ productName }</li> ) }
+						</ul>
+					</div>
+				</Tooltip>
+			</span>
+		);
+	}
+
+	renderLabel( label, index ) {
+		const purchased = CalypsoI18n.moment( label.created ).fromNow();
```
jeffstieler: Same question about using `i18n-calypso` here.

client/shipping-label/views/index.js
```diff
@@ -1,112 +1,222 @@
...
+
+		this.needToFetchLabelsStatus = true;
+
+		this.state = {
+			showTooltips: props.shippingLabel.labels.map( () => false ),
+		};
+	}
+
+	componentDidMount() {
+		this.historyMetaBox = document.getElementById( 'wc-connect-shipping-label-history' );
```
jeffstieler: Is there a reason this needs to be a separate element from the label purchase metabox?

(This is the first occurrence of `document.getElementById()` and `ReactDOM.render()` outside `main.js`, so it stood out to me)

client/shipping-label/views/index.js
```diff
@@ -1,112 +1,222 @@
...
 		);
-	};
+	}
+
+	renderRefundLink( label ) {
+		const today = new Date();
+		const thirtyDaysAgo = new Date().setDate( today.getDate() - 30 );
+		if ( ( label.used_date && label.used_date < today.getTime() ) || ( label.created_date && label.created_date < thirtyDaysAgo ) ) {
+			return null;
+		}
```
jeffstieler: Are we sure we want to enforce "refundability" logic in this component? Does it make more sense on the server perhaps?

If we do - I think a label having a `used_date` at all makes it ineligible for refund.

client/shipping-label/views/index.js
```diff
@@ -1,112 +1,222 @@
...
+			case 'pending':
+				className = ( 'wcc-metabox-label-item__refund-pending' );
+				text = __( 'Refund pending' );
+				break;
+			case 'complete':
+				className = ( 'wcc-metabox-label-item__refund-complete' );
+				text = sprintf( __( 'Refunded on %s' ), formatDate( label.refund.refund_date ) );
+				break;
+			case 'rejected':
+				className = ( 'wcc-metabox-label-item__refund-rejected' );
```
jeffstieler: Nitpick: why the surrounding parentheses?

client/shipping-label/views/index.js
```diff
@@ -1,112 +1,222 @@
...
-				</div>
+			<span className={ className } ><Gridicon icon="help-outline" size={ 18 }/> { text }</span>
+		);
+	}
+
+	renderReprint( label ) {
+		const todayTime = new Date().getTime();
+		if ( ( label.used_date && label.used_date < todayTime ) || ( label.expiry_date && label.expiry_date < todayTime ) ) {
+			return null;
+		}
```
jeffstieler: It makes sense to inquire about where this logic should live, given my comments about refund links..

woocommerce-connect-client.php
```diff
@@ -747,9 +748,15 @@ public function hide_wc_connect_package_meta_data( $hidden_keys ) {
...
 		function hide_wc_connect_order_meta_data( $protected, $meta_key, $meta_type ) {
-			if ( 'wc_connect_labels' === $meta_key ) {
-				$protected = true;
+			switch ( $meta_key ) {
+				case 'wc_connect_labels':
+					$protected = true;
+					break;
+				case 'wc_connect_label_history':
+					$protected = true;
+					break;
```
jeffstieler: You can stack `switch` `case`s that have the same resulting block:
```javascript
switch ( $meta_key ) {
	case 'wc_connect_labels':
	case 'wc_connect_label_history':
		$protected = true;
		break;
}
```

Basically, if you remove the `break;`, the execution path just "falls through" to the next case.

jeffstieler:
Sorry for the delay on the review - this is a big PR!

allendav:
Needs some styling tweaks - the fonts are a bit bigger than the design, spacing too

cc @kellychoffman 

i.e. (for each pairing, the top is the design, the bottom this PR)

screen shot 2016-12-07 at 1 17 49 pm

allendav:
Let's suppress outline on the tracking number link, i.e. get rid of this effect after clicking:

screen shot 2016-12-07 at 1 27 19 pm

classes/class-wc-rest-connect-shipping-label-controller.php
robobot3000: It's kind of an edge case, but it's possible to print a label for a box that has been removed from a packaging manager.

client/shipping-label/views/history.js
robobot3000: This has to hide the whole meta box, which is not controlled by React, but by PHP. I could use a class, but it is easier and more cross-browser compatible to just set the CSS `display` directly.

client/shipping-label/views/index.js
robobot3000: The only reason is the design - it shows the history in a separate metabox. I think visually it's cleaner, but unfortunately, this is what needs to be done in order to achieve it in the code.

client/shipping-label/views/index.js
robobot3000: You're right. It makes sense for this to live in the Connect server. I opened Automattic/woocommerce-connect-server#711 and I think it could be handled in a separate PR.

client/shipping-label/views/index.js
robobot3000: See above: Automattic/woocommerce-connect-server#711

robobot3000:
Thanks for having a look @jeffstieler @allendav 

I pushed some code and style changes.

client/lib/utils/time-ago.js
```diff
@@ -0,0 +1,48 @@
+import { translate as __ } from 'lib/mixins/i18n';
+import { sprintf } from 'sprintf-js';
+
```
jeffstieler: I should have been more clear - I wasn't against the use of `moment()`, I just questioned bringing in all of Calypso's `i18n` module to use it.

Don't feel like you have to remove this now though.

client/shipping-label/views/history.js
jeffstieler: I think we should be hiding the history metabox by default (in our stylesheet) and overriding the `display` property if there's history to be shown. I'm willing to concede the class-based approach, but are you *sure* that using a class to set `display` isn't cross browser compatible?

See: https://developer.mozilla.org/en-US/docs/Web/CSS/display#Browser_compatibility

I'd also like to see this logic moved to where the `LabelHistory` component is actually hooked into the DOM. (in the `ShippingLabelRootView` component)

(cc: @allendav @DanReyLop @v18 @nabsul for discussion)

client/lib/utils/time-ago.js
robobot3000: I decided to create my own - from what I read it is difficult to get `moment` to use the correct locale when it's installed with npm, and it seemed that we are not using large parts of that library.

DanReyLop: I suggest you take a look at `date-fns`: https://date-fns.org/docs/distanceInWordsToNow
I18N in `date-fns`: https://date-fns.org/docs/I18n

It's being advertised as "similar to `moment`, but you can just use individual functions". It certainly looks like a better solution than reimplementing it yourself :)

client/shipping-label/views/index.js
DanReyLop: The other meta-box should be another completely different React "application", but currently we rely on a global variable `wcConnectData` so it's not possible. Tomorrow I have a ✈️ , I'll give this a shot and see if I can make 2 applications live in the same page.

client/shipping-label/views/history.js
robobot3000: The history metabox is not empty - it contains markup generated by `add_meta_box`. But I added CSS to hide it by default and some code to show it.

By browser compatibility I meant JavaScript's `classList.add()`, but the only browser that doesn't support it is IE7, so it turns out everything is good.

kellychoffman:
Why are there two sections of label history? 

screen shot 2016-12-08 at 12 20 12 pm

1. We should get rid of the second meta box below, as it is redundant and not as informative as the listing above.
2. Add a 'Shipping labels history' header above the history listing that only displays when there is history?

kellychoffman:
When a refund is pending on a label, you shouldn't be able to reprint the label. Let's remove that link. 

screen shot 2016-12-08 at 12 50 59 pm

robobot3000:
I hid the reprint button for the labels with refund requested.

Re: two meta boxes:

This is what has been requested by @allendav in #562:
> a reverse chronologically sorted action log of create, reprint, and refund actions with date and username of action initiator

And what has been added to the design by @jameskoster:


Are you sure you want me to remove it?

kellychoffman:
Ah, I was going off of the latest design in that thread: https://github.com/Automattic/woocommerce-connect-client/issues/562#issuecomment-256627258

I think the double history listing is too confusing.

allendav:
@robobot3000 - I see the disconnect about the second meta box - i did not even notice it yesterday because it appears way below the fold on my Edit Order pages.

As I re-read my description in the original issue I can see how the reverse chronology could be interpreted as a distinct, separate component but I didn't mean it that way - my apologies for the misunderstanding

Let's go ahead and remove the 2nd metabox and keep just the richer top container

allendav:
Thank you @kellychoffman for the style updates

@robobot3000 - delete the Shipping Labels History meta box and then this is good to go

kellychoffman:
👍 Looks good.

Issue #684, opened 2016-11-10
Add non-flat-rate pre-defined packaging support


allendav:
- [x] Add Priority Mail and Priority Mail Express non-flat-rate pre-defined packaging as two new groups in the pre-defined section of the packaging manager
- [x] Include them (if selected) in rates requests
- [x] Expose them (if selected) as packaging options in the labels UI

allendav:
cc @robobot3000

PR #739, opened 2016-12-05
Sending a list of enabled preset boxes to the server

robobot3000:
Adds #684 

Test steps outlined in Automattic/woocommerce-connect-server#704


jeffstieler:
I'm unable to interact with the service foldable cards and I'm getting this error in the console when looking at my shipping method instance settings after enabling all non-flat-rate Priority packages (regional boxes):

`Uncaught Error: removeComponentAsRefFrom(...): Only a ReactOwner can have refs. You might be removing a ref to a component that was not created inside a component's 'render' method, or you have multiple copies of React loaded`

The error gives this link for more info: https://fb.me/react-refs-must-have-owner

robobot3000:
Looks like the same problem appears when I change client branch to `master`. I investigated a bit and it is caused by rendering `TextField` in `settings-item.js`. I will look into it more.

robobot3000:
This should be testable now

PR #749, opened 2016-12-08
Allow payment methods to be managed in jetpack dev mode

### LINKED ISSUE #748: Come up with a means of getting a payment method ID when testing locally
As a result of result changes, the payment methods screen now expects you to have a working Jetpack connection before it attempts to fetch a payment method.

This means that new local installs of the client (which cannot connect to Jetpack), can't make any purchases of labels when testing.

As a temporary workaround, add something like this to force the test payment method ID for your dummy auth user (don't use 88888888 - use your actual payment method id #):

```
function allendav_admin_init() {
        update_option( 'wc_connect_account_settings', array( 'selected_payment_method_id' => 88888888 ) );
}

add_action( 'admin_init', 'allendav_admin_init' );
```

allendav:
Fixes #748 

To test, try browsing payment methods on a dev mode site and a non dev mode site

Adds a note below the container to warn you if you are in dev mode, since the only payment methods you can see then are based on local server configuration, e.g.:

screen shot 2016-12-08 at 10 34 36 am

### DIFF
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -406,3 +406,12 @@
 		}
 	}
 }
+
+.wc-connect-admin-dev-notice {
+	width: 700px;
+
+	p {
+		font-style: italic;
+		color: $gray;
+	}
+}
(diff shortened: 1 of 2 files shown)


robobot3000:
Looks good :shipit:

PR #752, opened 2016-12-09
Add id and box_id when preparing labels from a free shipping order

### LINKED ISSUE #751: Labels UX blocked by packages (400) error if the order used Free Shipping
Reported in https://wordpress.org/support/topic/free-shipping-17/

> Error: The Connect for WooCommerce server returned: Bad Request child “packages” fails because [“packages” at position 0 does not match any of the allowed types] ( 400 )

I was able to reproduce this on a fresh install with 0.9.3.

Edit: Important: Do not use Connect for WooCommerce to get rates during checkout - e.g. do a free shipping flow instead

allendav:
Fixes #751

To test, complete an order for a shippable item with free shipping as the shipping method.  Make sure you do not get a "packages" error in the labels ui. Make sure you are prompted to select from various shipping options, e.g.:

screen shot 2016-12-09 at 3 16 35 pm

cc @robobot3000  or @DanReyLop for review since you're most familiar with this part of the code

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -45,7 +45,10 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 				$name = html_entity_decode( $product->get_formatted_name() );
 
 				for ( $i = 0; $i < $item[ 'qty' ]; $i++ ) {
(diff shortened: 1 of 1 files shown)


allendav:
NOTE: After merging this we should roll a new version of the client, deploy it or .org and ask the impacted user to update and try again

allendav:
ty!

Issue #758, opened 2016-12-14
Labels - show a message about the packaging manager when there are no packages configured 


marcinbot:
Link to the packaging manager when the merchant hasn't configured any packages. Also, show a warning icon next to the the packaging section in the labels ui if this is the case

jeffstieler:
This message was added and later removed - closing.

PR #763, opened 2016-12-16
Labels - Added a message when there are no packages configured

robobot3000:
Adds #758 

To test:
* remove all the packages from the packaging manager
* place a new order
* verify that when you open the Labels modal, the packages section is automatically opened, and shows a message

### DIFF
--- a/client/shipping-label/views/purchase/steps/packages/index.js
+++ b/client/shipping-label/views/purchase/steps/packages/index.js
@@ -6,6 +6,7 @@ import MoveItemDialog from './move-item';
 import AddItemDialog from './add-item';
 import Unpacked from './unpacked';
 import FormButton from 'components/forms/form-button';
+import Notice from 'components/notice';
 import { hasNonEmptyLeaves } from 'lib/utils/tree';
 import { sprintf } from 'sprintf-js';
 import StepContainer from '../../step-container';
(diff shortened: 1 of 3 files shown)


client/shipping-label/views/purchase/steps/packages/index.js
```diff
@@ -30,25 +31,58 @@ const PackagesStep = ( {
...
 
-		if ( 1 === packageIds.length ) {
-			return sprintf( __( '%d items in 1 package: %f %s total' ), itemsCount, totalWeight, storeOptions.weight_unit );
+		return { isSuccess: true, summary };
+	};
+
+	const renderWarning = () => {
+		if ( ! hasAnyPackagesConfigured ) {
+			const packagesMsg = sprintf(
+				__( 'There are no packages configured. The items have been packed individually. You can add or enable packages using the <a href="%(url)s">Packaging Manager</a>.' ),
```
DanReyLop: I wonder if all these URLs that take you out of the printing flow should have `target="_blank"`. @kellychoffman I'll trust you blindly on this one, I just wanted to bring it to your attention :)

client/shipping-label/views/purchase/steps/packages/index.js
```diff
@@ -106,11 +140,10 @@ const PackagesStep = ( {
 	return (
 		<StepContainer
 			title={ __( 'Packages' ) }
-			isSuccess={ isValidWeight && isValidPackages }
-			isError={ ! isValidWeight || ! isValidPackages }
-			summary={ renderSummary() }
+			{ ...getContainerState() }
 			expanded={ expanded }
 			toggleStep={ () => labelActions.toggleStep( 'packages' ) } >
+			{ renderWarning() }
```
DanReyLop: When some piece of HTML is gonna render conditionally, I usually prefer to put that condition here, like this:
```{ ! hasAnyPackagesConfigured && renderWarning() }```
Or something similar with the ternary operator, instead of having to go into the function to check if it always render something.

Personal preference, feel free to ignore :)

robobot3000:
`target="_blank"` has been removed from other parts of the code by someone, I think it was @kellychoffman 

The problem is that after the packages have been updated on the packaging manager page, the order page needs to be refreshed to reflect these changes. It is possible to implement a check for any settings changes, but it could also be a bit of a rabbit hole when trying to resolve conflicts between the current view/model and the new settings.

DanReyLop:
Makes sense then, :shipit:

PR #766, opened 2017-01-05
Fix Chrome MIME type warning

### LINKED ISSUE #718: MIME type warning for label preview
The following appears in the dev console during label preview:

screen shot 2016-11-17 at 3 43 55 pm

robobot3000:
Fixes #718

I replaced `iframe` with `embed` element. Tested on Chrome, Opera, Firefox and Safari.

### DIFF
--- a/assets/stylesheets/shipping-label.scss
+++ b/assets/stylesheets/shipping-label.scss
@@ -148,7 +148,7 @@
 		justify-content:center;
 		align-items:center;
 
-		iframe {
+		embed {
 			max-height: 100%;
 			height: 100%;
 			width: 100%;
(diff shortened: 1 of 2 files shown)


DanReyLop:
As a side note: I wouldn't stress much about this issue. If it's not an easy fix, I would simply close it. We are loading a PDF document inside an `<iframe>` (downloaded with its correct `pdf` MIME type), and it's loading fine. We aren't doing anything wrong. The fact that it spits some warning into the console on a particular browser shouldn't be considered a bug in our part.

robobot3000:
I agree with @DanReyLop. I thought that using `embed` would be a quick fix, but I wasn't aware of the not re-rendering issue (tested different paper sizes on Firefox and it worked fine).

We should avoid hacking the DOM and the issue should be waived.

Issue #674, opened 2016-11-07
Package data determined at checkout shown as JSON blob on Edit Order screen


jeffstieler:
![screen shot 2016-11-07 at 10 07 37 am](https://cloud.githubusercontent.com/assets/63922/20067403/35ec75ea-a4d2-11e6-968a-6ff025623050.png)

robobot3000:
This happens when a purchase has been made with Connect active, and then the order is viewed without the Connect plugin (or if TOS wasn't accepted)

jeffstieler:
> This happens when a purchase has been made with Connect active, and then the order is viewed without the Connect plugin (or if TOS wasn't accepted)

Hmm.. I wonder if prefixing the `meta_key` with an underscore would resolve this? They're usually treated as "private" meta values.

allendav:
What version of WooCommerce are you seeing this with, @jeffstieler ?

DanReyLop:
Related code: https://github.com/Automattic/woocommerce-connect-client/blob/86c98f069d0358a446e131c92a78194d8cf01a68/woocommerce-connect-client.php#L720

Not really the underscore trick (I don't know if that would work), but pretty close.

robobot3000:
The underscore trick doesn't work with WooCommerce 2.6. In fact, the only way to hide a metadata key is to add it to the `woocommerce_hidden_order_itemmeta` filter, which we already do.

Unfortunately the downside of that approach is that the metadata is shown when the plugin is disabled.

It will work for WooCommerce 2.7, though, so we can "future-proof" the code: https://github.com/woocommerce/woocommerce/blob/c29d597986f626a1718a67727ac98b1f80822301/includes/class-wc-order-item.php#L184

jeffstieler:
I'm going to icebox this for now. The 2.7 compatibility issue is more pressing.

jeffstieler:
@marcinbot is there anything to be done here, or can we close this?

marcinbot:
There's still #767, but it has been parked.

jeffstieler:
Closing this until it's reported by a user.

PR #767, opened 2017-01-06
Prefixed packages metadata key with an underscore

marcinbot:
Addresses #674

Test steps:
* place an order
* checkout this branch
* verify that the order and labels UI are shown correctly in the orders admin panel
* place an order and verify that it is shown correctly to the merchant as well
* disable Connect
* make a backup copy of your WooCommerce plugin dir
* update WooCommerce to 2.7 (either by pulling the latest from git, or here: https://github.com/woocommerce/woocommerce/releases/tag/2.7.0-beta-1)
* open the order placed after this branch has been checked out
* verify that the package metadata is not shown
* restore version 2.6 of WooCommerce and enable Connect

See https://github.com/Automattic/woocommerce-connect-client/issues/674#issuecomment-270886618 for more info

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -73,11 +73,19 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		protected function get_packaging_metadata( WC_Order $order ) {
 			$shipping_methods = $order->get_shipping_methods();
 			$shipping_method = reset( $shipping_methods );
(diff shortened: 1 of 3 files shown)


jeffstieler:
@robobot3000 have you tested placing an order in WC 2.7? I just did, and I don't think any metadata is being stored.

robobot3000:
@jeffstieler - I can't figure out why the metadata is not being saved, on top of other problems with 2.7. 

FYI, the underscore prefix is not causing this. I now remember it has been a known problem before.

Should this be dropped?

Issue #770, opened 2017-01-10
Display an 'updated TOS' dismissible notice on demand


v18:
The design conversation is happening in #736 

# Requirements:
- get notified by the server that a notice needs to be displayed
- display an entirely custom message, or a message with custom info:
  - TOS link
  - TOS GH changeset link
  - date
  - product name
- notice can be dismissed
- record when the date/time the notice is dismissed
- let the server know when the notice was dismissed

# Things we want to log (this is also related to the server)
- when a notice is sent to a site (server-related)
- when the site receives the notice (client-related)
- when the notice is dismissed (client-related)

# Possible approaches to receiving the updated TOS notice (no digging has yet been done on these):
1. Client asks server periodically if there are any notices to display. Every X days?
2. Client asks server if there are any notices to display based on user action (going to WCC screen, printing label, etc)
3. Server always sends notice-related data with all responses

# Possible approach to recording notice status:
- use a WordPress option called something like `wc_connect_updated_tos_notice` (since we already have `wc_connect_tos_accepted`), with this info:
  - the contents of message
  - a message slug or other ID
  - date of acceptance
- delete when notice dismissal successfully logged on the server

v18:
Q: What happens when there are two or more notices in a row (unlikely scenario)?
A: probably the best move is to add to `wc_connect_updated_tos_notice` (in case we still have not finished recording the dismissal of the previous one) and display only the latest

Q: does accepting TOS depend on having a working JP connection?

jeffstieler:
> Q: What happens when there are two or more notices in a row (unlikely scenario)?
A: probably the best move is to add to wc_connect_updated_tos_notice (in case we still have not finished recording the dismissal of the previous one) and display only the latest

Thinking about this like a generic notice system.. multiple simultaneous notices could be possible. (More so than a strictly ToS implementation)

Keeping with the generic idea.. perhaps not all notices need to have their dismissal/acceptance logged, and don't all need to be displayed on the same admin screens. We can certainly start with a "show all the notices everywhere we can" approach, but let's keep in mind a potential future where we're more selective about display.

> Q: does accepting TOS depend on having a working JP connection?

Currently, the banner is shown without checking for an active Jetpack. You can accept the ToS without JP, but we won't be able to track that action. Definitely an area we can focus some NUX love.

v18:
[...]

PR #774, opened 2017-01-13
Show 'updated TOS' and other notices

v18:
This is an in-progress PR for #770. Currently it:
- has a fake API call that returns some dummy data
- saves the data to wp_options
- displays a notice with the data

## Next Steps:
- [ ] decide on object structure for list notices received & saved
- [ ] display only the latest notice from TOS
- [ ] save the last retrieved notice id in wp_options
- [ ] let user dismiss notice
- [ ] updating the wp_option with new data
- [ ] log: when notice is received 
- [ ] log: when notice is dismissed

## About the notices

Here are things we care about for each notice.
- id
- title
- body
- button text
- should_log_response
- ^ type: info, success, etc
- ^ subject: TOS, other
- ^ pages it should be displayed on ('all' for TOS)
- ^ people it should be displayed to ('store owner' for TOS)

^ means that those items will not be added now, but may be added later as needed

What will not be built in this iteration
- logic for showing multiple types of notices (TOS at the same time as non-TOS, for example, and non-TOS with other non-TOS)
- notice types: info/success, and tos/non-tos
- pages to display notice on

The object received will probably look like:
```
{
  tos: [
    {
      id: 1,
      title: '',
      body: '',
      should_log_dismissal: true,
      can_display_with_others: false,
      type: 'info',
      pages: 'all',
      people: ['store-owner']
    }
  ],
  others: [...]
}
```

### DIFF
--- a/classes/class-wc-connect-remote-notices-store.php
+++ b/classes/class-wc-connect-remote-notices-store.php
@@ -21,10 +21,15 @@ if ( ! class_exists( 'WC_Connect_Remote_Notices_Store' ) ) {
 
 		public function __construct( WC_Connect_API_Client $api_client ) {
 			$this->api_client = $api_client;
+			add_action( 'admin_notices', array( $this, 'display_notices' ) );
 		}
 
 		public function set_notices( $notices ) {
 			update_option( 'wc_connect_remote_notices', $notices );
+		}
+
+		public function get_notices() {
+			return get_option( 'wc_connect_remote_notices', null );
 		}
 
 		public function fetch_notices_from_connect_server() {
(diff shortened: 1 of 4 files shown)


v18:
Will have a look at the failing checks tomorrow :)

classes/class-wc-connect-remote-notices-store.php
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		public function set_notices( $notices ) {
+			update_option( 'wc_connect_remote_notices', $notices );
+		}
+
+		public function fetch_notices_from_connect_server() {
+			$response = $this->api_client->get_notices( 1, 'en-us' );
+			// TODO: save more than one notice
+			$notices = (object) [
```
jeffstieler: Use `array()` here instead of `[]` (since it's PHP 5.4.x and up).

uninstall.php
```diff
@@ -0,0 +1,7 @@
+<?php
+// if uninstall.php is not called by WordPress, die
+if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
+	die;
+}
+
+delete_option( 'wc_connect_remote_notices' );
```
jeffstieler: Nice! This is an area we're currently deficient in. See: https://github.com/Automattic/woocommerce-connect-client/issues/776

classes/class-wc-connect-remote-notices-store.php
v18: Thanks! I also see this is the reason the automated tests failed for 5.3 and 5.2. Fixed, rebased, and pushed.

v18:
Not working on this for the foreseeable future, so going to close the PR for now.

PR #775, opened 2017-01-13
Make sure plugin is compatible with WooCommerce 3.0

### LINKED ISSUE #653: Shipping labels: Package errors when using woocommerce-2.7.0 dev version
I'm using the latest dev version of woocommerce. I can receive rates at checkout no problem. The problem comes in when I go to the order and create a label. Here's what happens:
1. Create a label
2. This error message appears in the console: 
   https://cloudup.com/izaVN_XvCJk
3. Fill out and confirm address
4. Skips past an errored Packages step, opens up rates step instead
5. Manually open packages step, see message that says there are no items or packages (even though there is both)
   screen shot 2016-10-28 at 10 08 43 am
6. Add package big enough for item to fit into, see that there is still no item
   screen shot 2016-10-28 at 10 09 24 am
7. 😭 
8. (╯°□°）╯︵ ┻━┻
   ....
9. Delete the dev version and replace with woocommerce 2.6.7 
10. 😀  🎉

nabsul:
Closes #653

I've tested this plugin against 2.6 and 2.7, and tried various things like: Getting rates at checkout, printing labels, etc. More testing would be great though.

- Use get_id() instead of ->id because it's become private in 3.0
- Fixed some phpDocs with typos and missing parameters
- Fixed some minor php issues with undefined variables
- add some phpDoc to clarify expected variable types
- Add logic to correctly use $order->id vs. $ $order->get_id();
- Pass order_id to get_form_data to avoid duplicate order_id statement
- Use PHPStorms code analysis to correct various issues
- Add a compatibility class for managing different versions of WC

To test: 
- Try out this branch with the latest WooCommerce 3.0
- Also make sure nothing is broken in previous versions of WooCommerce

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,24 +26,28 @@ env:
 matrix:
   include:
    - php: 5.2
-     env: WP_VERSION=latest
+     env: WP_VERSION=latest WC_VERSION=2.6.11
    - php: 5.2
-     env: WP_VERSION=4.4
+     env: WP_VERSION=4.4 WC_VERSION=2.6.11
    - php: 5.3
-     env: WP_VERSION=latest
+     env: WP_VERSION=latest WC_VERSION=2.6.11
    - php: 5.3
-     env: WP_VERSION=4.4
+     env: WP_VERSION=4.4 WC_VERSION=2.6.11
    - php: 5.4
-     env: WP_VERSION=latest
+     env: WP_VERSION=latest WC_VERSION=2.6.11
    - php: 5.4
-     env: WP_VERSION=4.4
+     env: WP_VERSION=4.4 WC_VERSION=2.6.11
    - php: 5.5
-     env: WP_VERSION=latest
+     env: WP_VERSION=latest WC_VERSION=2.6.11
    - php: 5.5
-     env: WP_VERSION=4.4
+     env: WP_VERSION=4.4 WC_VERSION=2.6.11
+   - php: 5.5
+     env: WP_VERSION=latest WC_VERSION=3.0.3
      # 5.6 / master already included above as first build.
    - php: 5.6
-     env: WP_VERSION=4.4
+     env: WP_VERSION=4.4 WC_VERSION=2.6.11
+   - php: 5.6
+     env: WP_VERSION=4.4 WC_VERSION=3.0.3
 
 # Cache the NPM dependencies
 cache:
@@ -61,7 +65,7 @@ install:
 
 # Clones WordPress and configures our testing environment.
 before_script:
-    - bash tests/bin/install-wc-tests.sh wordpress_test travis '' localhost $WP_VERSION
+    - bash tests/bin/install-wc-tests.sh wordpress_test travis '' localhost $WP_VERSION $WC_VERSION
 
 script: phpunit && npm test
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,26 +26,26 @@ env:
 matrix:
   include:
    - php: 5.2
-     env: WP_VERSION=latest WC_VERSION=2.6.11
+     env: WP_VERSION=latest WC_VERSION=2.6.14
    - php: 5.2
-     env: WP_VERSION=4.4 WC_VERSION=2.6.11
+     env: WP_VERSION=4.4 WC_VERSION=2.6.14
    - php: 5.3
-     env: WP_VERSION=latest WC_VERSION=2.6.11
+     env: WP_VERSION=latest WC_VERSION=2.6.14
    - php: 5.3
-     env: WP_VERSION=4.4 WC_VERSION=2.6.11
+     env: WP_VERSION=4.4 WC_VERSION=2.6.14
    - php: 5.4
-     env: WP_VERSION=latest WC_VERSION=2.6.11
+     env: WP_VERSION=latest WC_VERSION=2.6.14
    - php: 5.4
-     env: WP_VERSION=4.4 WC_VERSION=2.6.11
+     env: WP_VERSION=4.4 WC_VERSION=2.6.14
    - php: 5.5
-     env: WP_VERSION=latest WC_VERSION=2.6.11
+     env: WP_VERSION=latest WC_VERSION=2.6.14
    - php: 5.5
-     env: WP_VERSION=4.4 WC_VERSION=2.6.11
+     env: WP_VERSION=4.4 WC_VERSION=2.6.14
    - php: 5.5
      env: WP_VERSION=latest WC_VERSION=3.0.3
      # 5.6 / master already included above as first build.
    - php: 5.6
-     env: WP_VERSION=4.4 WC_VERSION=2.6.11
+     env: WP_VERSION=4.4 WC_VERSION=2.6.14
    - php: 5.6
      env: WP_VERSION=4.4 WC_VERSION=3.0.3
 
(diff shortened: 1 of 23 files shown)


woocommerce-connect-client.php
```diff
@@ -632,6 +632,7 @@ public function admin_enqueue_scripts() {
 			wp_register_script( 'wc_connect_admin', $wc_connect_base_url . 'woocommerce-connect-client.js', array(), false, true );
 
 			require_once( plugin_basename( 'i18n/strings.php' ) );
+			// TODO: Where is $i18nStrings defined?
```
DanReyLop: Here: https://github.com/Automattic/woocommerce-connect-client/blob/master/woocommerce-connect-client.php#L634

I don't know why PHPStorm chokes on it, though.

nabsul:
I've done some initial basic testing of this code and it currently works on both WC2.7 and and WC2.6 with now errors or warnings thrown in the debug log. 

The saving of metadata is still an issue though, and I'm waiting on https://github.com/woocommerce/woocommerce/issues/12850 to figure out how to best proceed.

classes/class-wc-connect-help-view.php
```diff
@@ -525,8 +531,9 @@ protected function get_form_schema() {
 						}
 
 						foreach ( $fieldsetitem->items as $item ) {
+							/** var WC_Order_Item $item */
```
jeffstieler: I don't think this is correct.

nabsul: My bad.

classes/class-wc-connect-compatibility.php
```diff
@@ -0,0 +1,66 @@
...
+		public static function instance() {
+			if( self::$singleton === null ) {
+				self::$singleton = new WC_Connect_Compatibility();
+			}
+
+			return self::$singleton;
+		}
+
+		public function __construct() {
+		}
```
jeffstieler: The singleton pattern can be useful if an instance is required, but if it isn't, a class with static methods works just as well.

`WC_Connect_Compatibility::get_item_product( $order, $item );` vs. `WC_Connect_Compatibility::instance()->get_item_product( $order, $item );`.

What do you think?

classes/class-wc-connect-shipping-method.php
```diff
@@ -274,6 +275,7 @@ public function calculate_shipping( $package = array() ) {
 					),
 					__FUNCTION__
 				);
+				return;
```
jeffstieler: This already returns on line 270.

classes/class-wc-connect-compatibility.php
nabsul: The singleton is more future-proof compared to a static class. Right now the class is fairly simple, and either one would work, but in the future I can see this class getting bigger, perhaps needing a constructor to initialize values, or even an interface to load different versions of a base class depending on the version of WooCommerce being used.

classes/class-wc-connect-shipping-method.php
nabsul: I opted to keep the second return statement since technically this function isn't supposed to return any value.

nabsul:
Needs testing against latest changes: https://github.com/woocommerce/woocommerce/pull/13284

classes/class-wc-connect-compatibility.php
```diff
@@ -0,0 +1,53 @@
...
+
+		public function get_order_id( WC_Order $order ) {
```
v18: I wonder if we should add a comment about the changes from 2.6->3.0 that require this? Or is that overkill ;)?

v18:
👍. I tested some things:

Tested on 2.6
- TOS acceptance
- add USPS shipping method
- checkout using calculated rate
- fallback rate at checkout
- buy label 
- print label
- add Canada Post shipping method
- calculate Canada Post shipping cost

Tested on ~~2.7~~ 3.0
- TOS acceptance
- add USPS shipping method
- add custom package
- fallback rate at checkout
- checkout using calculated rate
- buy label 
- print
- add Canada Post shipping method
- calculate Canada Post shipping cost
- uninstall

classes/class-wc-connect-compatibility.php
nabsul: Actually, the more I think about this the more I think splitting this up into an interface and two classes would be better.

Performance-wise it would mean a single if-statement instead of one per function call, and it would automatically clarify what you just described without a comment.

nabsul:
I intend to complete this PR some time next week. Last I checked with the latest and greatest WooCommerce RC, a few more things broke. That and the merge conflicts need to be resolved.

DanReyLop:
Is there anything blocking this PR? Turns out, the currently released plugin is not as "compatible" with 3.0 as we assumed. In the `Packages` step of the labels UI dialog, things are pretty broken in `master` (but fixed in this branch). See https://github.com/Automattic/woocommerce-services/issues/982#issuecomment-294438660

nabsul:
I've added unit tests and different versions of WC to Travis. This is ready for some manual testing.

classes/class-wc-connect-shipping-label.php
```diff
@@ -352,12 +357,14 @@ public function meta_box( $post ) {
...
+			$order_id = WC_Connect_Compatibility::instance()->get_order_id( $order );
 			$admin_array = array(
 				'purchaseURL'             => get_rest_url( null, '/wc/v1/connect/label/purchase' ),
 				'addressNormalizationURL' => get_rest_url( null, '/wc/v1/connect/normalize-address' ),
 				'getRatesURL'             => get_rest_url( null, '/wc/v1/connect/shipping-rates' ),
-				'labelStatusURL'          => get_rest_url( null, '/wc/v1/connect/label/' . $order->id . '-%d' ),
-				'labelRefundURL'          => get_rest_url( null, '/wc/v1/connect/label/' . $order->id . '-%d/refund' ),
+				'labelStatusURL'          => get_rest_url( null, '/wc/v1/connect/label/' . $order_id . '-%d' ),
+				'labelRefundURL'          => get_rest_url( null, '/wc/v1/connect/label/' . $order_id . '-%d/refund' ),
+				'labelsPreviewURL'        => get_rest_url( null, '/wc/v1/connect/labels/preview' ),
```
DanReyLop: This URL is no longer needed, I guess you carried it from a rebase/merge.

.travis.yml
```diff
@@ -26,24 +26,28 @@ env:
...
-     env: WP_VERSION=4.4
+     env: WP_VERSION=4.4 WC_VERSION=2.6.14
+   - php: 5.5
+     env: WP_VERSION=latest WC_VERSION=3.0.3
      # 5.6 / master already included above as first build.
    - php: 5.6
-     env: WP_VERSION=4.4
+     env: WP_VERSION=4.4 WC_VERSION=2.6.14
+   - php: 5.6
+     env: WP_VERSION=4.4 WC_VERSION=3.0.3
```
jeffstieler: This was originally meant to test the latest and previous versions of WP - can you update `4.4` to `4.6.4`?

Also, why aren't we testing WC `3.0.3` against PHP versions < 5.5?

v18: Should we also test PHP 7.0 and 7.1? Related: p6q8Tx-BE-p2

DanReyLop: There's an ~app~ issue for that: https://github.com/Automattic/woocommerce-services/issues/780

We can discuss in that issue the combinations (WC/WP/PHP) we should need to test (more than 10 would seem overkill for me, specially if we aren't supporting WordPress all the way back to 4.4).

Let's not block this issue any longer :)

Issue #792, opened 2017-01-21
[Labels UI] Add a success message about the PURCHASE


DanReyLop:
After the purchase is completed, we should get a visual confirmation.

Options:
* Success notice (let it show for a couple of seconds before the PDF printing dialog is opened. [Not]
* Change the copy of the "Purchase" button into something like "Success" while the PDF is being retrieved. That would take care of #771 too. [Yes, this one]

DanReyLop:
After talking this though, we agreed that this is sufficiently fixed by #797.

DanReyLop:
After some back and forth, this was fully fixed by #797

PR #797, opened 2017-01-21
	Make the label UI submit button keep being disabled while the PDF is being downloaded

### LINKED ISSUE #771: "Buy & Print" Button available just before printing a label
When I click on "buy & print" to purchase a label, it shows a spinner while the processing is happening. However, when the process is complete, the button looks like it's available again for clicking. This last for a second or so until the print dialog pops up. 

Perhaps would could at least make sure the button is shaded in a way that makes it clear it's disabled?

DanReyLop:
Fixes #771 

To test:
* Go purchase a label.
* While the purchase button is in the "Purchasing..." style, note that you can't close the dialog or click the button.
* Note that the button keeps being in that state until the PDF print dialog appears.

### DIFF
--- a/client/shipping-label/state/actions.js
+++ b/client/shipping-label/state/actions.js
@@ -450,16 +450,16 @@ export const purchaseLabel = () => ( dispatch, getState, context ) => {
 				console.error( error );
 				dispatch( NoticeActions.errorNotice( error.toString() ) );
 			} else {
-				const labelsToPrint = labels.map( ( label, index ) => ( {
-					caption: sprintf( __( 'PACKAGE %d (OF %d)' ), index + 1, labels.length ),
+				const labelsToPrint = response.map( ( label, index ) => ( {
+					caption: sprintf( __( 'PACKAGE %d (OF %d)' ), index + 1, response.length ),
 					labelId: label.label_id,
 				} ) );
 				const state = getState().shippingLabel;
 				printDocument( getPrintURL( state.paperSize, labelsToPrint, context ) )
 					.then( () => {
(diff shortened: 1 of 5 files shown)


client/shipping-label/index.js
```diff
@@ -28,6 +28,7 @@ export default ( { formData, labelsData, paperSize, storeOptions } ) => ( {
 				labels: labelsData || [],
 				paperSize,
 				form: {
+					isSuccess: true,
```
DanReyLop: This is a convenience so the "Success" button is always shown when you open the dialog, no need to go through the flow. Delete this line after design.

kellychoffman:
Agree. We shouldn't split the two actions up: the success criteria of this action is the print dialog appearing.

client/shipping-label/state/actions.js
```diff
@@ -452,14 +450,19 @@ export const purchaseLabel = () => ( dispatch, getState, context ) => {
...
 					labelId: label.label_id,
 				} ) );
 				const state = getState().shippingLabel;
 				printDocument( getPrintURL( state.paperSize, labelsToPrint, context ) )
-					.then( () => dispatch( exitPrintingFlow() ) )
+					.then( () => {
+						const noticeText = 1 === response.length
+								? __( 'Your shipping label was purchased successfully' )
+								: sprintf( __( 'Your %d shipping labels were purchased successfully' ), response.length );
+						dispatch( NoticeActions.successNotice( noticeText ) );
```
DanReyLop: I've ended up implementing the notice suggested by #792 . It doesn't go away after any time, but it's dismissable.

PR #805, opened 2017-01-21
Always fetching the label refund status

### LINKED ISSUE #789: Fetch label status (refund status) from server every time
We're currently caching refund status on the store. So if an HE refunds, the client will never know.

robobot3000:
Fixes #789

To test:
* purchase a new label in the client, do not request a refund on it
* in the server database, add a new row for that label's refund in `wcc_shipping_label_refund`
  * alternatively use the support tools to refund a label
* refresh the order page
* verify that the refund status has been updated


jeffstieler:
Code looks good and tests well. 🚢 it.

PR #809, opened 2017-01-21
Only validate (instance) form fields that have been changed

### LINKED ISSUE #795: Labels UI - first open shouldn't be riddled with errors
Only auto validate origin after first save

jeffstieler:
Fixes #795.

To test:
* Add a new USPS shipping zone instance
* Verify that the initial form load (with empty origin zip) does not have an error
* Alter the "name" field
* Verify the "origin" has not been validated
* Save the form
* Verify that "origin" has an error (from server)
* Refresh the page - *do not save*
* Enter a single character into "origin"
* Verify that "origin" has an error (from client)

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -15,6 +15,7 @@ import Settings from './settings';
 import ShippingLabel from './shipping-label';
 import AccountSettings from './account-settings';
 import Packages from './packages';
+import every from 'lodash/every';
 
 ( () => {
 	if ( ! global.wcConnectData ) {
@@ -47,7 +48,7 @@ import Packages from './packages';
 	window.addEventListener( 'beforeunload', ( event ) => {
 		const state = store.getState();
 
-		if ( ! state.form || state.form.pristine || ( state.form.meta && state.form.meta.pristine ) ) {
+		if ( ! state.form || ( state.form.meta && state.form.meta.pristine ) || every( state.form.pristine ) ) {
 			return;
 		}
 		const text = __( 'You have unsaved changes.' );
--- a/client/main.js
+++ b/client/main.js
@@ -15,7 +15,7 @@ import Settings from './settings';
 import ShippingLabel from './shipping-label';
 import AccountSettings from './account-settings';
 import Packages from './packages';
-import every from 'lodash/every';
+import _ from 'lodash';
 
 ( () => {
 	if ( ! global.wcConnectData ) {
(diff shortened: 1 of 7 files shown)


client/settings/state/reducer.js
```diff
@@ -1,14 +1,15 @@
 import { SET_FORM_PROPERTY } from './actions';
 import values from './values/reducer';
 import * as formValueActions from './values/actions';
+import mapValues from 'lodash/mapValues';
```
DanReyLop: We can use the full `import _ from 'lodash';` now :)

jeffstieler: Fixed in https://github.com/Automattic/woocommerce-connect-client/pull/809/commits/ca0d444ba7d11b22d6a9a504355ba6dad2e396d9.

client/main.js
```diff
@@ -15,6 +15,7 @@ import Settings from './settings';
 import ShippingLabel from './shipping-label';
 import AccountSettings from './account-settings';
 import Packages from './packages';
+import every from 'lodash/every';
```
DanReyLop: Same thing about importing the whole `lodash`

DanReyLop:
> Verify that "origin" has an error (from server)

That's not ideal. The "ideal" sulition would be:
* Let the user click the Save button
* Remove all the "pristine" flags in the form
* Recalculate the errors
* If there are errors, give an error notice (client-side only)
* If not, submit to the server for the real validation

That's not as complex as it sounds. Maybe 5-10 lines on the "submit" redux action, but maybe it's not worth it, so I'm ok if you don't do it. You have my blessing to :shipit: otherwise.

jeffstieler:
@DanReyLop - added the changes you requested

client/settings/state/values/actions.js
```diff
@@ -62,7 +63,16 @@ export const submit = ( schema, silent ) => ( dispatch, getState, { callbackURL,
 		}
 	};
 
-	const coercedValues = coerceFormValues( schema, getState().form.values );
+	const coercedValues = coerceFormValues( formSchema, getState().form.values );
```
DanReyLop: `schema` was already being passed to this function as a parameter, so no need to use the `thunk` property.

jeffstieler: Fixed in https://github.com/Automattic/woocommerce-services/pull/809/commits/57e2960ea57ae406c27a4469a2d820caf221073e.

PR #812, opened 2017-01-21
Added fallback rates to the client-side

### LINKED ISSUE #807: Show fallback rates (if available) if the server returns an error
With Automattic/woocommerce-connect-server#748 the server should return the configured fallback rate when there are no rates returned from the service.

However, if the server is unreachable for any reason, the client should use the same service setting to return the fallback when there are no rates available.

robobot3000:
Fixes #807

Automattic/woocommerce-connect-server#748 will be useful during testing

To test:
* Configure the fallback rate in the shipping settings
* Stop the server (to trigger a request error)
  * Verify that you can get the fallback rate at the checkout
* Check that you still get the fallback rate if the server is running, not returning an error, but also not returning any rates
* Place an order and check that everything is ok on the admin order page

### DIFF
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -242,7 +242,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 
 			$rate_to_add = array(
 				'id'        => self::format_rate_id( 'fallback', $this->id, 0 ),
-				'label'     => self::format_rate_title( __( 'Fallback rate' ) ),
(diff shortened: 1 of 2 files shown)


classes/class-wc-connect-shipping-method.php
```diff
@@ -339,9 +358,14 @@ public function calculate_shipping( $package = array() ) {
 					);
 
 					$this->add_rate( $rate_to_add );
+					$added_rates_count++;
 				}
 			}
 
+			if ( 0 === $added_rates_count ) {
```
jeffstieler: We should be able to just `count( $this->rates )` here - what do you think?

Issue #811, opened 2017-01-21
[Labels UI] You can trigger a rates request even with bogus addresses


DanReyLop:
In the Purchase Label dialog, even if some of the addresses steps is wrong (empty ZIP, for example), you can trigger a rates request, which will fail with a nasty error:

![error](https://cloud.githubusercontent.com/assets/1715800/22178434/4dff4304-e00c-11e6-90d9-cd56e6120508.png)

For example, you can do that by clicking the "Use this package" button on the Packages step. Look for usages of `client/shipping-label/state/actions.js#getLabelRates`.

jeffstieler:
Fixed in https://github.com/Automattic/woocommerce-services/pull/881.

PR #815, opened 2017-01-22
When origin address is empty, don't immediately validate its fields

### LINKED ISSUE #813: Only auto validate origin address after first save
We need to avoid the "Everything is red!!!" experience when a merchant opens the Purchase Label UI.

If the merchant hasn't stored an origin address, he shouldn't see errors right away.

DanReyLop:
Fixes #813 

To avoid the overwhelming "everything is red!!!1!" experience when the merchant opens the Purchase Label dialog without an stored Origin address, I've tweaked the validation logic to only kick in when a particular field is edited.

To test:
* Remove the origin address from your settings. That's the `wc_connect_origin_address` option. You can just also go to `classes/class-wc-connect-service-settings-store.php`, function `get_origin_address()`, put this in at the start: `update_option( 'wc_connect_origin_address', array() );` and reload the Order page once.
* Go print a label. You should see the "Origin address" step expanded but no visible errors.
* Once you edit a field such as the ZIP code, you see the error (if any) on that field, but not on the rest.
* You can click the "Use this address" button any time, and that will do a full validation (only local validation if there are invalid fields).

Because #811, the UX on this isn't ideal yet, but it's still an improvement.

To the reviewers: I'm OK with the solution but not with the variable names. My brain can't think on anything better than `ignoreValidation`. Plz help, thx.

### DIFF
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -98,6 +98,7 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 			$wc_address_fields[ 'address_2' ] = '';
 			$wc_address_fields[ 'city' ] = '';
 			$wc_address_fields[ 'postcode' ] = '';
+			$wc_address_fields[ 'phone' ] = '';
 
 			$stored_address_fields = get_option( 'wc_connect_origin_address', array() );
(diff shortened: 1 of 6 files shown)


DanReyLop:
> Please add a unit test to cover the changes to the reducer.

@jeffstieler Done :)

nabsul:
:shipit:

PR #826, opened 2017-01-23
Flat rate services - checkboxes always enabled

### LINKED ISSUE #818: Flat rate shipping - Enable checkboxes, and if user selects them, auto add that predefined box

marcinbot:
Fixes #818

This does the following things:
* removes the flat rate packages from the packaging manager screen
* the flat rate services are always enabled
* when enabling a flat rate service the corresponding package is treated as enabled automatically

To test:
* checkout the server branch at Automattic/woocommerce-connect-server#750
* verify that it's possible to enable flat rate services
* request rates and verify that the flat rates are coming back
* place an order and verify that the flat rate packages are still visible in the packaging manager


nabsul:
@marcinbot Would it be hard to make the box packer "select-all" checkboxes visible even when the list is not expanded? This would make it consistent with the service list.

nabsul:
Looks like you'll need to rebase this.

nabsul:
:shipit:

PR #831, opened 2017-01-23
Don't allow merchants to enter label dialog before selecting a payment method

### LINKED ISSUE #549: Trying to buy a shipping label with no payment method selected
I tried to buy a shipping label before selecting my payment method and the error message wasn't clear about that.

jeffstieler:
Fixes #549.

To test:
* With a credit card selected for payment - go to an order detail page
* Verify that an info notice shows the last 4 digits of the card selected for labels
* Verify that the "manage cards" link goes to the correct settings page
* With **no** credit card selected for payment - go to an order detail page
* Verify that an info notice states that you need to select a card to print labels
* Verify that the "add card" link goes to the correct settings page 
* Verify that no "create label" button is shown

### DIFF
--- a/client/shipping-label/views/index.js
+++ b/client/shipping-label/views/index.js
@@ -84,8 +84,8 @@ class ShippingLabelRootView extends Component {
 				<PurchaseLabelDialog
 					{ ...this.props.shippingLabel }
 					{ ...this.props } />
-				<Button className="wcc-metabox__new-label-button" onClick={ this.props.labelActions.openPrintingFlow } >
 				{ this.renderPaymentInfo() }
+				<Button className="wcc-metabox__new-label-button" disabled={ ! paymentMethod } onClick={ this.props.labelActions.openPrintingFlow } >
 					{ __( 'Create new label' ) }
 				</Button>
 			</div>
--- a/client/shipping-label/views/index.js
+++ b/client/shipping-label/views/index.js
@@ -63,7 +63,7 @@ class ShippingLabelRootView extends Component {
 							cardDigits: `<strong>${paymentMethod}</strong>`,
 						}
 					) } }></p>
-					<p><a href={ '#' }>Manage cards</a></p>
+					<p><a href="admin.php?page=wc-settings&tab=connect">{ __( 'Manage cards' ) }</a></p>
 				</Notice>
 			);
 		}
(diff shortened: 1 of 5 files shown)


jeffstieler:
@kellychoffman - can you give this some style love?

kellychoffman:
Updated styles. Also updated the 'no credit card' state to be an invitation instead of an error. Can we remove the `Create new label` button in this instance?

client/shipping-label/views/index.js
```diff
@@ -49,15 +52,49 @@ class ShippingLabelRootView extends Component {
 		this.setState( { showTooltips } );
 	}
 
+	renderPaymentInfo() {
+		const paymentMethod = this.props.shippingLabel.paymentMethod;
+
+		if ( paymentMethod ) {
+			return (
+				<Notice isCompact={ true } showDismiss={ false } className="wcc-metabox-label-payment inline">
+					<p dangerouslySetInnerHTML={ { __html: sprintf(
```
DanReyLop: No need to use `dangerouslySetInnerHTML` here, you can just print the `<strong>${digits}</strong>` outside of the `sprintf` string.

client/shipping-label/views/index.js
```diff
@@ -49,15 +52,49 @@ class ShippingLabelRootView extends Component {
 		this.setState( { showTooltips } );
 	}
 
+	renderPaymentInfo() {
+		const paymentMethod = this.props.shippingLabel.paymentMethod;
+
+		if ( paymentMethod ) {
+			return (
+				<Notice isCompact={ true } showDismiss={ false } className="wcc-metabox-label-payment inline">
```
DanReyLop: Nitpick: You can omit the `={ true }`, `<Component propName>` is the same as `<Component propName={ true }>`

client/shipping-label/views/index.js
jeffstieler: String translations need to contain tokens for string substitutions so they can be translated for RTL languages. If we append the card digits, we break that.

v18:
Works for me. There is an (really) edgecase issue where the dialogue will open for someone who loaded the edit order page, and _then_ removed their credit card can still open the print label modal. This seems like not an easy fix, but we should probably check for it in the `openPrintingFlow` action at some point.

DanReyLop:
> Works for me. There is an (really) edgecase issue where the dialogue will open for someone who loaded the edit order page, and then removed their credit card can still open the print label modal. This seems like not an easy fix, but we should probably check for it in the openPrintingFlow action at some point.

If the merchant is playing with multiple tabs, the purchase will 💥 eventually, I think that's enough for now.

Code looks good and @v18 says it works fine, go for it :shipit:

jeffstieler:
> > Works for me. There is an (really) edgecase issue where the dialogue will open for someone who loaded the edit order page, and then removed their credit card can still open the print label modal. This seems like not an easy fix, but we should probably check for it in the openPrintingFlow action at some point.

> If the merchant is playing with multiple tabs, the purchase will 💥 eventually, I think that's enough for now.

Let's keep these scenarios in mind, perhaps spiking on a live-refresh functionality. 👍

PR #836, opened 2017-01-23
Added a dotted underline to the "Label #N" title on the labels list, to make it clearer that there is a tooltip

DanReyLop:
Added a dotted underline to the "Label #N" title on the labels list, to make it clearer that there is a tooltip there.

dot

Right now, `text-decoration` is only well supported in Firefox (see [support table](http://caniuse.com/#feat=text-decoration)). In other browsers, it will just display a solid black line, not a dotted dark-gray line.

The other option is using a `border-bottom: 1px dotted $gray-dark;`, but that feels too hacky. @kellychoffman what do you think?

### DIFF
--- a/assets/stylesheets/shipping-label.scss
+++ b/assets/stylesheets/shipping-label.scss
@@ -33,7 +33,6 @@
 				font-weight: bold;
 				font-size: 14px;
 				color: $gray-dark;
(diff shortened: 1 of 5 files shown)


nabsul:
:shipit:

PR #839, opened 2017-01-23
Added a class for managing NUX notices

marcinbot:
This manages the notices shown on specific pages as part of the NUX. I added an example labels notice.

The intention is for this class to also host other NUX-related PHP code.

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -31,7 +31,8 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 		}
 
 		public function show_labels_notice() {
-			$dismiss_url = add_query_arg( 'wc-connect-nux', 'labels' );
+			return;
+			/*$dismiss_url = add_query_arg( 'wc-connect-nux', 'labels' );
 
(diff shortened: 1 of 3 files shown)


v18:
Haven't read through the code, but going through the nux I have two notes so I don't forget later:

- the previous purple/cloud style is missing from the TOS notice
- dismissing the TOS notice should take us to the shipping zones page (not required in this PR, but we should be able to do this later)

classes/class-wc-connect-nux.php
```diff
@@ -0,0 +1,66 @@
...
+			if ( ! isset( $_GET['wc-connect-nux'] ) ) {
+				return;
+			}
+
+			$notice = $_GET['wc-connect-nux'];
+			$notices = $this->get_notice_states();
+			$notices[ $notice ] = true;
+			update_option( 'wc_connect_nux_notices', $notices );
+
+			wp_safe_redirect( remove_query_arg( 'wc-connect-nux' ) );
```
DanReyLop: Do we need this redirect? It would be cleaner (and faster) without it.

PR #843, opened 2017-01-24
Add shipping zone pointer

v18:
This PR adds a pointer to the shipping zone page. The pointer is dismissible, and multiple pointers can eventually be shown on the same page in sequence, and on other pages.

The pointer does not point up from the '+' button like this:

original

Instead, it looks like this:

current-pointers

cc @kellychoffman  ^

The reason is because the zones are added via js in WC, and I didn't see a way to hook in after they're added to avoid a race condition which would potentially make our pointers not show up at all.

Should we add styles that will match the pointers to our admin notices?

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -4,6 +4,10 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 
 	class WC_Connect_Nux {
 
+		function __construct() {
+			$this->init_pointers();
+		}
+
 		private function add_shared_notice_styles() {
 			$wc_connect_base_url = defined( 'WOOCOMMERCE_CONNECT_DEV_SERVER_URL' ) ? WOOCOMMERCE_CONNECT_DEV_SERVER_URL : plugins_url( 'dist/', __FILE__ );
 			wp_enqueue_style( 'wc_connect_banner', $wc_connect_base_url . 'woocommerce-connect-client-banner.css' );
@@ -60,6 +64,59 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 			$notices[ $notice ] = true;
 			update_option( 'wc_connect_nux_notices', $notices );
 		}
-	}
 
+		private function init_pointers() {
+			add_filter( 'wc_services_pointer_woocommerce_page_wc-settings', array( $this, 'register_add_service_to_zone_pointer' ) );
+		}
+
+		public function show_pointers( $hook ) {
+			/* Get admin pointers for the current admin page.
+			 *
+			 * @since 0.9.6
+			 *
+			 * @param array $pointers Array of pointers.
+			 */
+			$pointers = apply_filters( 'wc_services_pointer_' . $hook, array() );
+
+			if ( ! $pointers || ! is_array( $pointers ) ) {
+				return;
+			}
+
+			$dismissed_pointers = explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) );
+			$valid_pointers = array();
+
+			if( isset( $dismissed_pointers ) ) {
+				foreach ( $pointers as $pointer ) {
+					if ( ! in_array( $pointer['id'], $dismissed_pointers ) ) {
+						$valid_pointers[] =  $pointer;
+					}
+				}
+			} else {
+				$valid_pointers = $pointers;
+			}
+
+			if ( empty( $valid_pointers ) ) {
+				return;
+			}
+
+			wp_enqueue_style( 'wp-pointer' );
+			wp_localize_script( 'wc_services_admin_pointers', 'wcSevicesAdminPointers', $valid_pointers );
+			wp_enqueue_script( 'wc_services_admin_pointers' );
+		}
+
+		public function register_add_service_to_zone_pointer( $pointers ) {
+			$pointers[] = array(
+				'id' => 'wc_services_add_service_to_zone',
+				'target' => 'th.wc-shipping-zone-methods',
+				'options' => array(
+					'content' => sprintf( '<h3>%s</h3><p>%s</p>',
+						__( 'Add WooCommerce Shipping Service to a Zone' ,'woocommerce-services' ),
+						__( 'Choose a zone that is applicable to USPS or Canada Post.', 'woocommerce-services' )
+					),
+					'position' => array( 'edge' => 'right', 'align' => 'left' ),
+				)
+			);
+			return $pointers;
+		}
+	}
 }
--- a/client/admin-pointers.js
+++ b/client/admin-pointers.js
@@ -1,0 +1,35 @@
+/*global wcSevicesAdminPointers, ajaxurl */
+import jQuery from 'jquery';
+
+jQuery( document ).ready( function( $ ) {
+	function show_pointer( pointers, i ) {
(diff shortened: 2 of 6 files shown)


classes/class-wc-connect-nux.php
```diff
@@ -94,8 +94,10 @@ public function show_pointers( $hook ) {
 			}
 
 			wp_enqueue_style( 'wp-pointer' );
-			wp_enqueue_script( 'wc_services_admin_pointers', plugins_url( 'assets/js/admin-pointers.js', dirname(__FILE__) ), array( 'wp-pointer', 'jquery' ) );
+			$wc_connect_base_url = defined( 'WOOCOMMERCE_CONNECT_DEV_SERVER_URL' ) ? WOOCOMMERCE_CONNECT_DEV_SERVER_URL : plugins_url( 'dist/', __FILE__ );
+			wp_register_script( 'wc_services_admin_pointers', $wc_connect_base_url . 'woocommerce-services-admin-pointers.js', array( 'wp-pointer', 'jquery' ), false, true );
```
DanReyLop: If you look at `woocommerce-services.php#admin_enqueue_scripts()`, there all the scripts/styles are registered in the same place, but they are only required (`wp_enqueue_script`) when they are actually needed. I would prefer to maintain that pattern, but I don't have a strong opinion about it (the `banner.css` file doesn't follow it already).

client/admin-pointers.js
```diff
@@ -0,0 +1,26 @@
+/*global wcSevicesAdminPointers, ajaxurl */
+import jQuery from 'jquery';
```
DanReyLop: jQuery code built by Webpack. Now that's something I didn't expect to see today. I like it 😄

client/admin-pointers.js
```diff
@@ -0,0 +1,26 @@
...
+				} );
+				if ( pointers[ i + 1 ] ) {
+					show_pointer( i + 1 );
+				}
+			},
+		} );
+		$( pointer.target ).pointer( options ).pointer( 'open' );
+	}
+
+	if ( pointers[ 0 ] ) {
```
DanReyLop: Can `pointers` be `null` / `undefined`? Better be deffensive here IMHO

DanReyLop:
> The reason is because the zones are added via js in WC, and I didn't see a way to hook in after they're added to avoid a race condition which would potentially make our pointers not show up at all.

You dodged a bullet there. In WC 2.7, we don't even have the `+` button, so maintaining that code would have been a pain 😄

classes/class-wc-connect-nux.php
```diff
@@ -60,6 +64,55 @@ public function check_notice_dismissal() {
...
+				'target' => 'th.wc-shipping-zone-methods',
+				'options' => array(
+					'content' => sprintf( '<h3>%s</h3><p>%s</p>',
+						__( 'Add WooCommerce Shipping Service to a Zone' ,'woocommerce-services' ),
+						__( 'Choose a zone that is applicable to USPS or Canada Post.', 'woocommerce-services' )
+					),
+					'position' => array( 'edge' => 'right', 'align' => 'left' ),
+				)
+			);
+			return $pointers;
```
jeffstieler: If only [`WP_Internal_Pointers::print_js()`](https://developer.wordpress.org/reference/classes/wp_internal_pointers/print_js/) was public. 😛

classes/class-wc-connect-nux.php
v18: I agree with you, @DanReyLop - and I made this change in https://github.com/Automattic/woocommerce-services/pull/843/commits/1fca8276d3731d95c6fbd0752bf6b2a9a216230c. We may need to update `class-wc-connect-nux.php#add_shared_notice_styles()` later as well, @marcinbot.

client/admin-pointers.js
v18: Technically not, since if there are no notices to display, WordPress won't run the script. However, I agree about being defensive so I added this & other checks in f50a15e5db729b04978bd1ca532bd12b38bc0de7

v18:
> You dodged a bullet there. In WC 2.7, we don't even have the + button, so maintaining that code would have been a pain 😄

You make a great point about 2.7 @DanReyLop. Here's what it currently looks like in 2.7 (without any version checking / changes):

shipping-zone-pointer-2 7

From what I can tell, the shipping zone methods can now only be edited under the shipping zone settings page, which like this:

shipping-zone-edit-settings-page

I think we'll want to re-consider the pointer we show for 2.7 in this case. What do you think @kellychoffman  & @jeffstieler ?

classes/class-wc-connect-nux.php
```diff
@@ -60,6 +64,55 @@ public function check_notice_dismissal() {
...
+				$valid_pointers = $pointers;
+			}
+
+			if ( empty( $valid_pointers ) ) {
+				return;
+			}
+
+			wp_enqueue_style( 'wp-pointer' );
+			$wc_connect_base_url = defined( 'WOOCOMMERCE_CONNECT_DEV_SERVER_URL' ) ? WOOCOMMERCE_CONNECT_DEV_SERVER_URL : plugins_url( 'dist/', __FILE__ );
+			wp_register_script( 'wc_services_admin_pointers', $wc_connect_base_url . 'woocommerce-services-admin-pointers.js', array( 'wp-pointer', 'jquery' ), false, true );
```
v18: Nice! I made the change in 30f64a7ae69556d1bcc73fce429f3c08fc6b142c, also moving wc_connect_base_url to a class property. I defined in in `__construct()` - let me know if there's a better place, though.

client/admin-pointers.js
```diff
@@ -0,0 +1,26 @@
+/*global wcSevicesAdminPointers, ajaxurl */
+import jQuery from 'jquery';
+
+jQuery( document ).ready( function( $ ) {
+	const pointers = wcSevicesAdminPointers;
+
+	function show_pointer( i ) {
```
v18: The index is required in order to show the next available pointer. I didn't want to use an external counter since the call to `show_pointer()` to show the next pointer seemed intuitive enough. However, if that's not clear, I can make the change.

DanReyLop:
> I think we'll want to re-consider the pointer we show for 2.7 in this case.

I'm OK shelving this until after launch. It's not like this will break horribly in 2.7. Maybe we should tweak the wording so it applies to the old and the new UI.

v18:
> I'm OK shelving this until after launch. It's not like this will break horribly in 2.7. Maybe we should tweak the wording so it applies to the old and the new UI.

Agreed :) Not a blocker.

classes/class-wc-connect-nux.php
```diff
@@ -60,6 +64,53 @@ public function check_notice_dismissal() {
...
-	}
 
-}
\ No newline at end of file
+		private function init_pointers() {
+			add_filter( 'wc_services_pointer_woocommerce_page_wc-settings', array( $this, 'register_add_service_to_zone_pointer' ) );
+		}
+
+		public function show_pointers( $hook ) {
+			$pointers = apply_filters( 'wc_services_pointer_' . $hook, array() );
```
jeffstieler: Let's document our hooks (note: I'm not too sure we've been great about this so far) using the WP standards: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

v18:
Rebased + pushed!

Issue #776, opened 2017-01-14
Clean up plugin options when uninstalled


jeffstieler:
We should remove all stored `wp_options` when the plugin is uninstalled.

See: https://developer.wordpress.org/plugins/the-basics/uninstall-methods/

PR #859, opened 2017-02-02
Tracking the "opt in" event only after the TOS has been accepted

### LINKED ISSUE #665: woocommerceconnect_opted_in tracked before opt-in
With WP_DEBUG on, the following is logged even before accepting our TOS

> [03-Nov-2016 22:01:22 UTC] Tracked the following event: woocommerceconnect_opted_in

It shouldn't be, at least not until they actually opt-in :)

marcinbot:
Closes #665

I also added the TOS update on plugin deactivation, so that if the plugin is deactivated and activated again, the TOS notice will always show.

To test:
* you will need to have the plugin activated and enable debugging on the System Status page
* enable WP_DEBUG (I also enabled logging to file, as I couldn't get the events to show in the html)
* enable the plugin
* verify that `Tracked the following event: woocommerceconnect_opted_in` is not logged yet, and that the TOS notice appears
* accept the TOS
* verify that the event has been tracked now

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -159,7 +159,6 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		}
 
 		static function plugin_deactivation() {
-			update_option( 'wc_connect_tos_accepted', false );
 			$tracks = self::load_tracks_for_activation_hooks();
 			$tracks->opted_out();
 			wp_clear_scheduled_hook( 'wc_connect_fetch_service_schemas' );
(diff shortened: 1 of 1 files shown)


woocommerce-services.php
```diff
@@ -158,12 +158,8 @@ static function load_tracks_for_activation_hooks() {
 			return new WC_Connect_Tracks( $logger );
 		}
 
-		static function plugin_activation() {
-			$tracks = self::load_tracks_for_activation_hooks();
-			$tracks->opted_in();
-		}
-
 		static function plugin_deactivation() {
+			update_option( 'wc_connect_tos_accepted', false );
```
DanReyLop: Are we sure we want this? Can be annoying for no good reason.

jeffstieler: Let's remove this change in favor of handling it in https://github.com/Automattic/woocommerce-services/issues/776.

PR #863, opened 2017-02-03
Only Jetpack owner accepts TOS

### LINKED ISSUE #777: Only allow Jetpack connection owner (i.e. the "master" user) accept ToS
The terms of service notice should only be shown to the user that own the store's Jetpack connection. Currently it is shown to everyone.

In addition, the functions that check for acceptance/declination should only work for admins - adding a nonce to the URL and using [`check_admin_referer()`](https://developer.wordpress.org/reference/functions/check_admin_referer/)

nabsul:
Fixes #777 

The TOS notice is now only rendered if the current user is the owner of the jetpack token. I also rearranged the flow of TOS and notice dismissal checks to make the code more readable.

To Test: Unfortunately, you can't really test this in Jetpack's dev mode (in which case the TOS is always shown for any admin). You'll need to setup an internet-facing machine and really connect Jetpack.

- To reset the flow, simply delete the `wc_connect_tos_accepted` option from your database
- Try out the dismissal; it should disable the plugin and show a notice
- Try logging in as an admin that is not the jetpack token owner
- Try accepting the TOS and make sure the plugin gets enabled correctly

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -341,7 +341,8 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		 * @codeCoverageIgnore
 		 */
 		public function init() {
-			if ( ! $this->check_tos() ) {
+			if ( ! get_option( 'wc_connect_tos_accepted', false ) ) {
+				add_action( 'admin_init', array( $this, 'admin_tos_notice' ) );
 				return;
 			}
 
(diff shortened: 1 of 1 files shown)


woocommerce-services.php
```diff
@@ -679,19 +680,55 @@ public function get_active_shipping_services() {
...
+				return true;
+			}
+
+			if ( ! class_exists( 'Jetpack_Data' ) ) {
+				return false;
+			}
+
+			$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
+			if ( $user_token && is_object( $user_token ) && isset( $user_token->external_user_id ) ) {
+				return get_current_user_id() === $user_token->external_user_id;
```
jeffstieler: Just curious - is this method of checking for the JP master user better than what we already had here: https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-settings-pages.php#L132 ?

Should we standardize around one method?

PR #864, opened 2017-02-03
Added indeterminate checkboxes

marcinbot:
Added indeterminate checkboxes for when only a subset of service/package checkboxes is selected:
screen shot 2017-02-03 at 22 09 20

To test:
* navigate to the service settings or package manager
* verify that everything works fine (selecting all through the main checkbox, selecting only a few boxes, selecting all the boxes manually)
* also check the console for any React errors
* check that the settings save as expected

CC @jeffstieler @kellychoffman

### DIFF
--- a/client/components/indeterminate-checkbox/index.js
+++ b/client/components/indeterminate-checkbox/index.js
@@ -1,0 +1,27 @@
+import React, { PropTypes } from 'react';
+import Gridicon from 'components/gridicon';
+
+const IndeterminateCheckbox = ( { selectedCount, allCount, onToggle } ) => {
+	const allSelected = selectedCount === allCount;
+	const someSelected = 0 < selectedCount && allCount > selectedCount;
+
+	const onClick = ( event ) => {
+		const checked = ! ( someSelected || allSelected );
+		onToggle( event, checked );
+	};
+
+	return (
+		<span className="wcc-indeterminate-container" onClick={ onClick }>
+			<input type="checkbox" checked={ allSelected } readOnly />
+			{ someSelected ? <Gridicon className="bulk-select__some-checked-icon" icon="minus-small" size={ 18 }/> : null }
+		</span>
+	);
+};
+
+IndeterminateCheckbox.propTypes = {
+	selectedCount: PropTypes.number.isRequired,
+	allCount: PropTypes.number.isRequired,
+	onToggle: PropTypes.func,
+};
+
+export default IndeterminateCheckbox;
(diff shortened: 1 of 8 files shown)


client/components/indeterminate-checkbox/index.js
```diff
@@ -0,0 +1,27 @@
+import React, { PropTypes } from 'react';
+import Gridicon from 'components/gridicon';
+
+const IndeterminateCheckbox = ( { selectedCount, allCount, onToggle } ) => {
+	const allSelected = selectedCount === allCount;
+	const someSelected = 0 < selectedCount && allCount > selectedCount;
+
+	const onClick = ( event ) => {
+		const checked = ! ( someSelected || allSelected );
```
DanReyLop: I feel like this logic should be in the reducer, instead of passing down one extra param.

marcinbot: It copies the logic from Calypso's `BulkSelect` and also enforces consistent behaviour. Before, clicking the main checkbox in services while it was indeterminate would uncheck all, and in packages it would select all.

client/settings/views/services/group.js
```diff
@@ -1,27 +1,25 @@
...
 	const format = ( 1 === numSelected ) ? __( '%d service selected' ) : __( '%d services selected' );
 	return sprintf( format, numSelected );
 };
 
-const updateAll = ( event, updateValue, services ) => {
+const updateAll = ( event, checked, updateValue, services ) => {
+	event.stopPropagation();
 	services.forEach( ( service ) => {
-		updateValue( [ service.id, 'enabled' ], event.target.checked );
+		updateValue( [ service.id, 'enabled' ], checked );
```
DanReyLop: Same here. The reducer has enough information to know if a toggle will turn on/off all the services, this extra param is redundant.

client/components/indeterminate-checkbox/index.js
DanReyLop: Good point about consistent behaviour. I still think this would be better in the reducers though, having a trail of `checked` parameters passing through half the app just seems pointless. What about we let the next reviewer break the tie? I don't have any more comments about this PR and I don't want to hold it for a small thing like this.

jeffstieler:
This is a total nitpick, but I'm more bothered by the naming of `IndeterminateCheckbox` than I am the "some checked" logic living outside the reducer.. can we name the component something else? It isn't *always* indeterminate after all. :)

client/components/indeterminate-checkbox/index.js
jeffstieler: I'll tie break - let's keep it as is for now.

kellychoffman:
👍 I approve.

PR #865, opened 2017-02-03
[DO NO MERGE] Reinstall the plugin from the correct repo

### LINKED ISSUE #855: Release a new version in the old plugin slug with a notice to migrate to the new one
We shouldn't abandon the brave users that installed our plugin when it was in BETA. Since we moved to a different plugin slug (`woocommerce-services`), the old one (`connect-for-woocommerce`) won't receive any updates, or notices, or anything.

We should:
* Branch from the last `0.9.X` release (or from `master`, I don't have a strong opinion about this).
* Implement an admin notice telling the user to uninstall this plugin and install the new one. As annoying as possible.
* Release that to the old `connect-for-woocommerce` plugin slug.

nabsul:
Closes #855

This branch should NOT get checked into master. It is custom code for the old repo at https://wordpress.org/plugins/connect-for-woocommerce/. Here's what it does:

- Displays a notice that the plugin has moved to another repo
- Allows the user to dismiss that notice (or should we make it annoying and stay forever?)
- If the user clicks "reinstall", the current plugin is deactivated, deleted, and the new one is installed and activated.


woocommerce-services.php
```diff
@@ -346,7 +346,8 @@ public function load_textdomain() {
 		 * @codeCoverageIgnore
 		 */
 		public function init() {
-			if ( ! $this->check_tos() ) {
+			if ( ! get_option( 'wc_connect_tos_accepted', false ) ) {
```
DanReyLop: This notices refactor **should** get merged into `master`, right?

woocommerce-services.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Plugin Name: WooCommerce Services
+ * Plugin Name: WooCommerce Services (deprecated)
```
DanReyLop: Maybe keep here `Connect for WooCommerce (deprecated)` to make this transition less confusing?

woocommerce-services.php
```diff
@@ -769,10 +771,116 @@ public function show_tos_notice() {
 			<?php
 		}
 
+		public function reinstall_notice() {
+			if ( get_option( 'wc_connect_dismiss_deprecation', false ) ) {
+				return;
```
DanReyLop: I don't think we want to make this notice dismissable. I would just keep nagging the merchant, there's absolutely no reason to not upgrade.

woocommerce-services.php
```diff
@@ -769,10 +771,116 @@ public function show_tos_notice() {
...
+		public function show_reinstall_notice() {
+			$accept_url = add_query_arg( array( 'wc-connect-deprecate' => 'reinstall' ) );
+			$decline_url = admin_url( 'plugins.php?wc-connect-deprecate=dismiss' );
+
+			?>
+			<div class="notice wcc-admin-notice">
+				<h1><?php _e( 'WooCommerce Services Plugin has Moved' ) ?></h1>
+				<a href="<?php echo esc_url( $decline_url ); ?>" style="text-decoration: none;" class="notice-dismiss" title="<?php esc_attr_e( 'Dismiss and deactivate the plugin', 'woocommerce-services' ); ?>"></a>
+				<p>
+					<b>The repository from which this plugin has been installed is deprecated. But don't worry, updating is painless!</b>
```
DanReyLop: Missing i18n

woocommerce-services.php
```diff
@@ -769,10 +771,116 @@ public function show_tos_notice() {
...
+				return;
+			}
+
+			add_action( 'admin_notice', array( $this, 'reinstall_success' ) );
+		}
+
+		public function reinstall_failed() {
+			?>
+			<div class="notice notice-error is-dismissible">
+				<p><?php _e( 'Failed to reinstall WooCommerce Services. You may need to manually reinstall the plugin.', 'woocommerce-services' ); ?></p>
```
DanReyLop: At the very least, show here a link to the Plugin Install page. Ideally, redirect the user to that page directly (with this notice showing).

woocommerce-services.php
nabsul: For a retired repo that we'll never update with i18n strings? ¯\_(ツ)_/¯

DanReyLop:
My bad. I downloaded a ZIP with this branch and did a clean install, and that obviously worked. I didn't realize that a plugin update doesn't search again for the plugin's main file (it didn't even cross my mind that install and update would be operations with different consequences at all).

nabsul:
good catch @jeffstieler

jeffstieler:
This has been deployed to WP.org - https://wordpress.org/plugins/connect-for-woocommerce/developers/

I cherry picked the reinstall functions into the existing plugin file - adding the NUX classes, etc, seemed overkill.

PR #871, opened 2017-02-07
Show origin and destination fields if they're not normalized on open print flow

### LINKED ISSUE #870: When opening printing flow, address fields should be open if they aren't being verified
Here's what it currently looks like when opening the modal:

before

v18:
Fixes #870

Before, the origin and destination fields would not open until the normalization was finished:

before

Now, they'll both open if there are still things for the user to do, i.e. there are errors, or we're ignoring validation:

after

The unit tests are not complete. They are missing the cases where:
1. There are errors for either origin or destination address - this is because mocking `getFormErrors` was taking too long. I think the best way forward is to add it as a param to `openPrintingFlow`
2. The fields should _not_ open - this is because the toggle action will still be dispatched in `expandStepAfterAction` on line 202. This happens because the `getState` function will not actually return the new state with the `expanded` props. Thinking on the best way around this. Ideas welcome.

In the interest of shipping (our deadline is Feb 7), I think we can skip the extra unit tests, so I'm making this as review-ready.

FAQ
Q: Why are you using sinon 2 (i.e. `sinon@next`) instead of the current version 1.17.7? 
A: So it's compatible with webpack

Q: Why aren't you using sinon-chai to make use of the beautiful `have.been.calledWith()`
A: It's incompatible with sinon 2

### DIFF
--- a/client/lib/calypso-boot/index.js
+++ b/client/lib/calypso-boot/index.js
@@ -1,4 +1,4 @@
-import ReactInjection from 'react/lib/ReactInjection';
+import ReactClass from 'react/lib/ReactClass';
 import i18n from '../mixins/i18n';
 
 export default function boot() {
@@ -11,7 +11,7 @@ export default function boot() {
 
 	i18n.initialize( i18nLocaleStringsObject );
 
-	ReactInjection.Class.injectMixin( i18n.mixin );
+	ReactClass.injection.injectMixin( i18n.mixin );
 }
 
 boot();
--- a/client/shipping-label/state/actions.js
+++ b/client/shipping-label/state/actions.js
@@ -140,19 +140,39 @@ const refreshPreview = ( dispatch, getState, context ) => {
 	} );
 };
 
-export const openPrintingFlow = () => ( dispatch, getState, context ) => {
+export const openPrintingFlow = () => (
+	dispatch,
+	getState,
+	context,
+	getErrors = getFormErrors
+) => {
 	const { storeOptions, addressNormalizationURL, getRatesURL, nonce } = context;
 	let form = getState().shippingLabel.form;
 	const { origin, destination } = form;
-	let errors = getFormErrors( getState(), storeOptions );
+	let errors = getErrors( getState(), storeOptions );
 	const promisesQueue = [];
 
-	if ( ! origin.ignoreValidation && ! hasNonEmptyLeaves( errors.origin ) && ! origin.isNormalized && ! origin.normalizationInProgress ) {
+	if ( ! origin.ignoreValidation &&
+		! hasNonEmptyLeaves( errors.origin ) &&
+		! origin.isNormalized &&
+		! origin.normalizationInProgress ) {
 		promisesQueue.push( normalizeAddress( dispatch, origin.values, 'origin', addressNormalizationURL, nonce ) );
 	}
 
-	if ( ! destination.ignoreValidation && ! hasNonEmptyLeaves( errors.destination ) && ! destination.isNormalized && ! destination.normalizationInProgress ) {
+	if ( origin.ignoreValidation || hasNonEmptyLeaves( errors.origin ) ) {
+		dispatch( toggleStep( 'origin' ) );
+	}
+
+	if ( ! destination.ignoreValidation &&
+		! hasNonEmptyLeaves( errors.destination ) &&
+		! destination.isNormalized &&
+		! destination.normalizationInProgress ) {
 		promisesQueue.push( normalizeAddress( dispatch, destination.values, 'destination', addressNormalizationURL, nonce ) );
+	}
+
+	if ( destination.ignoreValidation ||
+			hasNonEmptyLeaves( errors.destination ) ) {
+		dispatch( toggleStep( 'destination' ) );
 	}
 
 	waitForAllPromises( promisesQueue ).then( () => {
--- a/client/shipping-label/state/test/actions.js
+++ b/client/shipping-label/state/test/actions.js
@@ -1,0 +1,192 @@
+import { expect } from 'chai';
+import { openPrintingFlow } from '../actions';
+import sinon from 'sinon';
+
+const context = {
+	storeOptions: {
+		countriesData: {
+			US: '',
+		},
+	},
+};
+
+function createGetStateFn( newProps = { origin: {}, destination: {} } ) {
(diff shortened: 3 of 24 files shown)


client/shipping-label/state/test/actions.js
```diff
@@ -0,0 +1,76 @@
...
+				} ) ).to.equal( true );
+			} );
+			it( 'opens print flow', () => {
+				expect( spy.calledWith( {
+					type: 'OPEN_PRINTING_FLOW',
+				} ) ).to.equal( true );
+			} );
+		} );
+	} );
+} );
```
jeffstieler: Tests!! 🎉

client/shipping-label/state/test/actions.js
```diff
@@ -0,0 +1,76 @@
...
+				},
+			},
+		},
+	};
+};
+
+describe( 'Shipping label Actions', () => {
+	describe( '#openPrintingFlow', () => {
+		describe( 'when ignoring validation for addresses', () => {
+			let spy = sinon.spy();
```
jeffstieler: What do you think about renaming this to reflect that it's a `dispatch()` spy?

Also, if this were called with an unexpected action (not the 3 below) - what happens?

client/shipping-label/state/actions.js
```diff
@@ -147,14 +147,28 @@ export const openPrintingFlow = () => ( dispatch, getState, context ) => {
...
+	}
+
+	if ( ! destination.ignoreValidation &&
+		! hasNonEmptyLeaves( errors.destination ) &&
+		! destination.isNormalized &&
+		! destination.normalizationInProgress ) {
 		promisesQueue.push( normalizeAddress( dispatch, destination.values, 'destination', addressNormalizationURL, nonce ) );
 	}
 
+	if ( origin.ignoreValidation || hasNonEmptyLeaves( errors.origin ) ) {
```
v18: Yes, it should :) My tests were also not catching this, so I fixed/expanded them, too.

client/shipping-label/state/test/actions.js
v18: Sure - renamed to `dispatchSpy`.

> Also, if this were called with an unexpected action (not the 3 below) - what happens?

The tests will still pass if it's called with other actions. I also added some safeguards to make sure it's not called with any of those those steps using a different (untested) code path.

v18:
> don't forget to update the npm shrinkwrap

Ah, yes, thanks! I updated it, and it required updating a couple of things in-depth:
- upgrading to React 15.4
- moving Gridicons from a `wp-calypso` component to its own npm module

client/lib/calypso-boot/index.js
```diff
@@ -11,7 +11,7 @@ export default function boot() {
 
 	i18n.initialize( i18nLocaleStringsObject );
 
-	ReactInjection.Class.injectMixin( i18n.mixin );
+	ReactClass.injection.injectMixin( i18n.mixin );
```
DanReyLop: Well, this was easier than expected :D

client/shipping-label/state/actions.js
```diff
@@ -140,21 +140,41 @@ const refreshPreview = ( dispatch, getState, context ) => {
 	} );
 };
 
-export const openPrintingFlow = () => ( dispatch, getState, context ) => {
+export const openPrintingFlow = () => (
+	dispatch,
+	getState,
+	context,
+	getErrors = getFormErrors
```
DanReyLop: I'm not a fan of changing client code just to better suit the tests. Can you instead mock `getFormErrors` in the test? Or is it too messy?

DanReyLop:
> Ah, yes, thanks! I updated it, and it required updating a couple of things in-depth

You got into a 🐰 🕳  there :)
The best way to install dependencies is `npm install packageName --save` or `npm install packageName --save-dev`, that way the `shrinkwrap` **should be** automatically updated.
But I'm glad you figured out how to get us to the latest React version :)

client/shipping-label/state/actions.js
jeffstieler: @v18 can you explain the aliasing of this method name? I'm not readily seeing the point of this.

DanReyLop: This method didn't receive a `getErrors` param before, now it can receive one. If it doesn't, it uses `getFormErrors`, so it works as before. **But**, in the tests, this method can be called with a custom `getErrors` function, which makes it easier to "mock" a validator.

That's why I brought this to attention: this is code that only serves the tests. If there's a less intrusive way to mock the validator, we should do that instead.

v18:
@DanReyLop - t

client/shipping-label/state/actions.js
v18: Yeh, I struggled with this one :) @DanReyLop - I tried to stub it, but honestly it was a lot of extra effort and at the end I didn't get it working. Figured better to test the error condition than not, hence the extra param.

@jeffstieler - the aliasing is to avoid the `no-shadow` lint error. Happy to change it to an ignore comment instead, don't have a strong preference on how to make eslint happy here.

DanReyLop:
> That's what I tried, but the shrinkwrap file was not updated automatically.

Welcome to the "why I want to switch to Yarn" stage of grief.

> Maybe because by default it doesn't update with --save-dev

That's not it. Our `shrinkwrap` is generated with the `--dev` flag so it has the dev dependencies. Maybe when you installed the package, `shrinkwrap` gave you an error and it refused to be updated (those errors are easy to miss). If it happens to you again, check the console output for errors and ping me :)

v18:
The notices now look ok, but still need some style tweaks. I also checked all of the uses of Gridicon, and they're all fine. Still to do:

- [ ] update notices style
- [ ] rebase
- [ ] ask for another look over

v18:
I updated the notices to look better. Here's what they look like now:

new

And with more text:

new-example

And on a small screen:

wc-small-screen

It differs a bit from our previous style (now the Gridicon is visible):

wc-small-screen-old

The reason is that it use the new wp-calypso notice styles:

calypso-small-screen

This PR could use another look-over with these changes, cc @DanReyLop and @jeffstieler

v18:
Fixed the inline notices in the shipping labels modal to be full width:

full-width-3
full-width-1
full-width-2

I'll also fix those minor alignment issues with the text and icon next.

PR #874, opened 2017-02-07
NUX: Shipping zones: Update copy to be more clear

kellychoffman:
The current copy isn't quite clear and the title and content are a bit redundant. It also doesn't take into account new users, who haven't set up any zones yet.

Currently:

screen shot 2017-02-07 at 1 20 56 pm

This PR changes the copy to:

To ship products to customers using USPS or Canada Post, you will need to add them as a shipping method to an applicable zone. If you don't have any zones, add one first.

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -75,8 +75,8 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 				'target' => 'th.wc-shipping-zone-methods',
 				'options' => array(
 					'content' => sprintf( '<h3>%s</h3><p>%s</p>',
-						__( 'Add WooCommerce Shipping Service to a Zone' ,'woocommerce-services' ),
-						__( 'Choose a zone that is applicable to USPS or Canada Post.', 'woocommerce-services' )
+						__( 'Add a WooCommerce shipping services to a Zone' ,'woocommerce-services' ),
+						__( 'To ship products to customers using USPS or Canada Post, you will need to add them as a shipping method to an applicable zone. If you don't have any zones, add one first.', 'woocommerce-services' )
 					),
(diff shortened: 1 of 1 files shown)


classes/class-wc-connect-nux.php
```diff
@@ -75,8 +75,8 @@ public function register_add_service_to_zone_pointer( $pointers ) {
 				'target' => 'th.wc-shipping-zone-methods',
 				'options' => array(
 					'content' => sprintf( '<h3>%s</h3><p>%s</p>',
-						__( 'Add WooCommerce Shipping Service to a Zone' ,'woocommerce-services' ),
-						__( 'Choose a zone that is applicable to USPS or Canada Post.', 'woocommerce-services' )
+						__( 'Add a WooCommerce shipping services to a Zone' ,'woocommerce-services' ),
```
jeffstieler: Should "services" be plural here?

classes/class-wc-connect-nux.php
```diff
@@ -75,8 +75,8 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+						__( 'To ship products to customers using USPS or Canada Post, you will need to add them as a shipping method to an applicable zone. If you don't have any zones, add one first.', 'woocommerce-services' )
```
jeffstieler: The apostrophe in "don't" needs to be escaped since the string is single quoted. Like: `don\'t`.

classes/class-wc-connect-nux.php
kellychoffman: Thanks! Fixed in c7090f6

classes/class-wc-connect-nux.php
kellychoffman: Thanks! Fixed in c7090f6

PR #876, opened 2017-02-07
Restore tracks event when users accept the Terms of Services.

jeffstieler:
Restore tracks event when users accept the Terms of Services.

This was errantly removed in https://github.com/Automattic/woocommerce-services/commit/8612b81cfe02967316e1b8f5b8c33cb6386a0325


v18:
👍 looks good to me!

PR #880, opened 2017-02-10
Cleaning up plugin options on uninstall

marcinbot:
Adds #776 

### Careful when testing this one!
Removing the plugin will delete all of the directory contents, including local git repo and node_modules. This happens even if you use symlinks. Obvious, but worth pointing out.

Best checkout this repo separately into a fresh WP/WooCommerce installation, run `npm dist`, update some settings, then delete the plugin and verify that all of the options are gone as well.

### DIFF
--- a/classes/class-wc-connect-options.php
+++ b/classes/class-wc-connect-options.php
@@ -2,21 +2,21 @@
 
 if ( ! class_exists( 'WC_Connect_Options' ) ) {
 	class WC_Connect_Options {
-		public static $tos_accepted = 'wc_connect_tos_accepted';
-		public static $store_guid = 'wc_connect_store_guid';
-		public static $error_notice = 'wc_connect_error_notice';
-		public static $debug_enabled = 'wc_connect_debug_logging_enabled';
-		public static $payment_methods = 'wc_connect_payment_methods';
-		public static $service_schemas = 'wc_connect_services';
-		public static $last_schema_update = 'wc_connect_services_last_update';
-		public static $last_heartbeat = 'wc_connect_last_heartbeat';
-		public static $account_settings = 'wc_connect_account_settings';
-		public static $paper_size = 'wc_connect_paper_size';
-		public static $origin_address = 'wc_connect_origin_address';
-		public static $custom_packages = 'wc_connect_packages';
-		public static $predefined_packages = 'wc_connect_predefined_packages';
-		public static $last_rate_request = 'wc_connect_last_rate_request';
-		public static $nux_notices = 'wc_connect_nux_notices';
+		const TOS_ACCEPTED = 'wc_connect_tos_accepted';
+		const STORE_GUID = 'wc_connect_store_guid';
+		const ERROR_NOTICE = 'wc_connect_error_notice';
+		const DEBUG_ENABLED = 'wc_connect_debug_logging_enabled';
+		const PAYMENT_METHODS = 'wc_connect_payment_methods';
+		const SERVICE_SCHEMAS = 'wc_connect_services';
+		const LAST_SCHEMA_UPDATE = 'wc_connect_services_last_update';
+		const LAST_HEARTBEAT = 'wc_connect_last_heartbeat';
+		const ACCOUNT_SETTINGS = 'wc_connect_account_settings';
+		const PAPER_SIZE = 'wc_connect_paper_size';
+		const ORIGIN_ADDRESS = 'wc_connect_origin_address';
+		const CUSTOM_PACKAGES = 'wc_connect_packages';
+		const PREDEFINED_PACKAGES = 'wc_connect_predefined_packages';
+		const LAST_RATE_REQUEST = 'wc_connect_last_rate_request';
+		const NUX_NOTICES = 'wc_connect_nux_notices';
 
 		/**
 		 * Based on the service id and optional instance, generates the options key that
@@ -70,20 +70,11 @@ if ( ! class_exists( 'WC_Connect_Options' ) ) {
 				return;
 			}
 
-			delete_option( self::$tos_accepted );
-			delete_option( self::$store_guid );
-			delete_option( self::$error_notice );
-			delete_option( self::$debug_enabled );
-			delete_option( self::$payment_methods );
-			delete_option( self::$service_schemas );
-			delete_option( self::$last_schema_update );
-			delete_option( self::$last_heartbeat );
-			delete_option( self::$account_settings );
-			delete_option( self::$paper_size );
-			delete_option( self::$origin_address );
-			delete_option( self::$custom_packages );
-			delete_option( self::$predefined_packages );
-			delete_option( self::$last_rate_request );
+			$reflect = new ReflectionClass( get_called_class() );
+			$constants = $reflect->getConstants();
+			foreach ( $constants as $constantName => $constantValue ) {
+				delete_option( $constantValue );
+			}
 
 			self::delete_all_shipping_methods_options();
 		}
--- a/classes/class-wc-connect-options.php
+++ b/classes/class-wc-connect-options.php
@@ -16,6 +16,7 @@ if ( ! class_exists( 'WC_Connect_Options' ) ) {
 		public static $custom_packages = 'wc_connect_packages';
 		public static $predefined_packages = 'wc_connect_predefined_packages';
 		public static $last_rate_request = 'wc_connect_last_rate_request';
+		public static $nux_notices = 'wc_connect_nux_notices';
 
 		/**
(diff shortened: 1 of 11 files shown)


classes/class-wc-connect-options.php
```diff
@@ -0,0 +1,105 @@
+<?php
+
+if ( ! class_exists( 'WC_Connect_Options' ) ) {
+	class WC_Connect_Options {
+		public static $tos_accepted = 'wc_connect_tos_accepted';
```
DanReyLop: What about using [constants](http://php.net/manual/en/language.oop5.constants.php) instead of static variables? The behaviour is pretty much the same but the intention is clearer.

classes/class-wc-connect-options.php
```diff
@@ -0,0 +1,105 @@
...
+			delete_option( self::$payment_methods );
+			delete_option( self::$service_schemas );
+			delete_option( self::$last_schema_update );
+			delete_option( self::$last_heartbeat );
+			delete_option( self::$account_settings );
+			delete_option( self::$paper_size );
+			delete_option( self::$origin_address );
+			delete_option( self::$custom_packages );
+			delete_option( self::$predefined_packages );
+			delete_option( self::$last_rate_request );
```
DanReyLop: I think it would be better to loop through the constants instead of manually having to add them here too (this will be prone to error, I guarantee you). If you change the static variables to constants, you can do something like this:
```php
$reflect = new ReflectionClass(get_called_class());
$constants = $reflect->getConstants();
foreach($constants as $constantName => $constantValue) {
    delete_option($constantValue);
}
```

jeffstieler: Use of the Reflection classes isn't always the most intuitive thing to run into in PHP code, and isn't a common WordPress pattern.

I know seeing a bunch of `delete_option()` calls isn't pleasing, and does require vigilance when options are added - but I think that's more of a case for combining our plugin's options into a single object rather than introducing some clever code.

DanReyLop: > Use of the Reflection classes isn't always the most intuitive thing to run into in PHP code, and isn't a common WordPress pattern.

I know it's a bit tricky, but it's the "most straight-forward" way to simulate an Enum. It's just a couple of lines of Reflection-related code.

> I know seeing a bunch of delete_option() calls isn't pleasing

I don't have a problem with the "unpleasing" part, but I think when adding more options to this file, we will sometimes forget to add the corresponding `delete_option` and our plugin won't uninstall cleanly. For example, right now this code doesn't delete the `$nux_notices` option (I don't know if it was intentional).

I won't push for this solution anymore, but at least I brought to light an alternative solution :)

v18: > does require vigilance when options are added

We could add some tests to make sure that all options are accounted for :) Im thinking: keep track of all the extra options added after install and make sure they're all gone after uninstall. Maybe there's also another better way I'm not thinking of.

jeffstieler: > We could add some tests to make sure that all options are accounted for :) Im thinking: keep track of all the extra options added after install and make sure they're all gone after uninstall. Maybe there's also another better way I'm not thinking of.

I think this is a good compromise. If class constants are used only for option keys, the unit test could use the reflection API to ensure that all options are deleted after uninstall.

v18: Hmm, giving my test idea another thought (install and make sure all options are removed after uninstall) may not work if not all the options are added within the test run.

Another suggestion (I heard from someone else): linting tests instead of unit tests. Have not read the code to see if that's even possible, so putting it here so I don't forget.

> class constants are used only for option keys

That may make both the test ideas possible :D

marcinbot: I'm siding with @DanReyLop on this one. Using class constants and reflection seems to be the most proper way to go.

My worry with the tests is that we don't have a 100% test coverage. In the ideal situation where we would, having a test that checks if all options have been removed would make sense.

jeffstieler: > My worry with the tests is that we don't have a 100% test coverage. In the ideal situation where we would, having a test that checks if all options have been removed would make sense.

I don't think the lack of complete test coverage is ever a good argument for not writing tests. :) How else will we get there?

This feels like too much of a back and forth too late, but I can't seem to shake that this could all be simplified:

> I think that's more of a case for combining our plugin's options into a single object

I'm not sure if this was missed - I didn't get a response. With the addition of the `WC_Connect_Options` class, we have an opportunity to refactor how our plugin options work. If we follow Jetpack's options example, we would only have a single options key to delete on plugin uninstall.

What do you think?

classes/class-wc-connect-order-item-meta.php
```diff
@@ -0,0 +1,20 @@
...
+		public static function delete_all_plugin_order_item_meta() {
+			if ( defined( 'WOOCOMMERCE_CONNECT_DEV_SERVER_URL' ) ) {
+				return;
+			}
+
+			global $wpdb;
+			$reflect = new ReflectionClass( get_called_class() );
+			$constants = $reflect->getConstants();
+			foreach ( $constants as $constantName => $constantValue ) {
+				$wpdb->query( "DELETE FROM {$wpdb->prefix}woocommerce_order_itemmeta WHERE meta_key='{$constantValue}'" );
```
jeffstieler: This should use [`$wpdb->prepare()`](https://developer.wordpress.org/reference/classes/wpdb/prepare/).

classes/class-wc-connect-options.php
```diff
@@ -0,0 +1,96 @@
...
+		 * Deletes all options created by WooCommerce Services
+		 */
+		public static function delete_all_options() {
+			if ( defined( 'WOOCOMMERCE_CONNECT_DEV_SERVER_URL' ) ) {
+				return;
+			}
+
+			$reflect = new ReflectionClass( get_called_class() );
+			$constants = $reflect->getConstants();
+			foreach ( $constants as $constantName => $constantValue ) {
```
jeffstieler: Please use underscores instead of camel case in PHP. 👍 (`constant_name`)

classes/class-wc-connect-order-item-meta.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+if ( ! class_exists( 'WC_Connect_Order_Item_Meta' ) ) {
+	class WC_Connect_Order_Item_Meta {
```
DanReyLop: Hold on. Do we really want to do this? Wipe all existing order packaging meta-data and labels history when the plugin is uninstalled? Seems too destructive to me.

classes/class-wc-connect-options.php
DanReyLop: > I think that's more of a case for combining our plugin's options into a single object

I didn't miss it, I thought you were referring to having all the *logic* that handles options in the same object/class, and that's exactly what this PR already has implemented :)

> If we follow Jetpack's options example, we would only have a single options key to delete on plugin uninstall.

I'm ok with that. Intuitively, I think performance would take a hit, because on each request we would need to fetch from MySQL and then parse a "massive" JSON object to get just a small part of it. But if it's good enough for Jetpack, it will be good enough for us.

classes/class-wc-connect-order-item-meta.php
jeffstieler: I had the same thought - I inquired in Slack about it, and core WC will only delete options/data if that's opted in.

A very unscientific sampling of a few shipping extensions did not yield examples of total data annihilation on uninstall.

The issue this PR addresses only mentions `wp_options` in scope.

classes/class-wc-connect-options.php
jeffstieler: > I'm ok with that. Intuitively, I think performance would take a hit, because on each request we would need to fetch from MySQL and then parse a "massive" JSON object to get just a small part of it. But if it's good enough for Jetpack, it will be good enough for us.

I believe in this case, a single option containing a large(r) JSON object will be more performant. Each `wp_option` (when WP is running without cache) is automatically queried for with every WP load. We currently have multiple options that require JSON decoding, so we'd be going from `X` queries and `Y` decodes to `1` of each.

jeffstieler:
> * simultaneous updates - Can we be affected by a race condition, where multiple admins save settings, or the plugin updates an option, and some changes are immediately overwritten? I think Jetpack addresses this by combining options that are read often, but rarely updated, while the often-updated options are still stored separately. I will propose the split for us below.

👍 

> * backwards compatibility - the existing plugin users will lose the options that were combined into a single object.

Can't we put some back-compat logic into `WC_Connect_Options::get_option()`?

PR #881, opened 2017-02-10
Prevent label rates requests when packaging data is errant

jeffstieler:
This is a fix for https://github.com/Automattic/woocommerce-connect-server/issues/775, which was thought to be a server side issue.

To test:
* Manually create an order with a product that has invalid or missing weight/dimensions info, but with a destination address that will automatically pass normalization
* Open create label dialog
* Verify that the Packaging step is open with errors, and rates are not requested
* Reopen the Destination step, click "use this address"
* Verify that the Packaging step is open with errors, and rates are not requested
* Reopen the Destination step, enter in an address that will trigger a non-trivial normalization result
* Click "use this address"
* Verify that the Packaging step is open with errors, and rates are not requested

### DIFF
--- a/client/shipping-label/state/actions.js
+++ b/client/shipping-label/state/actions.js
@@ -239,8 +239,12 @@ export const confirmAddressSuggestion = ( group ) => ( dispatch, getState, { sto
 
 	const errors = getFormErrors( getState(), storeOptions );
 
-	// Make sure packaging is error free before fetching new rates after address confirmation
-	if ( hasNonEmptyLeaves( errors.packages ) ) {
+	// If all prerequisite steps are error free, fetch new rates
+	if (
+		hasNonEmptyLeaves( errors.origin ) ||
+		hasNonEmptyLeaves( errors.destination ) ||
+		hasNonEmptyLeaves( errors.packages )
+	) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


client/shipping-label/state/actions.js
```diff
@@ -236,6 +237,13 @@ export const confirmAddressSuggestion = ( group ) => ( dispatch, getState, { sto
 		expandFirstErroneousStep( dispatch, getState, storeOptions, group );
 	};
 
+	const errors = getFormErrors( getState(), storeOptions );
+
+	// Make sure packaging is error free before fetching new rates after address confirmation
+	if ( hasNonEmptyLeaves( errors.packages ) ) {
```
DanReyLop: You should check here than origin & destination are valid too. This function is the result of the user "confirming" a validated address, but the other one could still be invalid.

DanReyLop:
This PR will solve https://github.com/Automattic/woocommerce-services/issues/811 too

client/shipping-label/state/actions.js
jeffstieler: Fixed in https://github.com/Automattic/woocommerce-services/pull/881/commits/3fd8ab429cc76e0b0fa3b89c112676cff6a1edf1.

PR #883, opened 2017-02-14
Include JSON parse error in service settings update repsonse

jeffstieler:
Include the encountered JSON error in the response when service settings update fails.

See: https://wordpress.org/support/topic/shipping-zones-not-being-accepted/

Initially this started with just adding the numeric JSON decoding error to the response, but after finding the WP REST API classes builtin handling of JSON errors, it switched to a refactoring of our REST API usage.

This PR seeks to add the `Content-Type: application/json` header to WP REST API requests (in order to invoke the built in handling), and remove the manual `json_decode()`ing of those request bodies.

To test:
* Send bad JSON to any WP REST API endpoint added by WCS
* Verify that an error is returned
* Verify that the error and request info is logged
* Verify that all existing functionality works in the /wp-admin/ (packages save, settings save, account settings save, address normalize, label rates request)

### DIFF
--- a/classes/class-wc-rest-connect-services-controller.php
+++ b/classes/class-wc-rest-connect-services-controller.php
@@ -96,7 +96,10 @@ class WC_REST_Connect_Services_Controller extends WP_REST_Controller {
 		if ( empty( $settings ) ) {
 			$error = new WP_Error( 'bad_form_data',
 				__( 'Unable to update service settings. The form data could not be read.', 'woocommerce-services' ),
-				array( 'status' => 400 )
+				array(
+					'status'     => 400,
+					'error_code' => json_last_error(),
+				)
 			);
 			$this->logger->debug( $error, __CLASS__ );
 			return $error;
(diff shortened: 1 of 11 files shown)


classes/class-wc-rest-connect-services-controller.php
```diff
@@ -96,7 +96,10 @@ public function update_item( $request ) {
 		if ( empty( $settings ) ) {
 			$error = new WP_Error( 'bad_form_data',
 				__( 'Unable to update service settings. The form data could not be read.', 'woocommerce-services' ),
-				array( 'status' => 400 )
+				array(
+					'status'     => 400,
+					'error_code' => json_last_error(),
+				)
 			);
 			$this->logger->debug( $error, __CLASS__ );
```
nabsul: Two questions:
1) Is this worth logging whether or not debug is enabled?
2) Would it be worth surfacing as an admin notice, or at least re-use the existing notice that directs the user to the logs?

jeffstieler: > 1. Is this worth logging whether or not debug is enabled?

Yes, I think it is - I'll add that.

> 2. Would it be worth surfacing as an admin notice, or at least re-use the existing notice that directs the user to the logs?

The existing notice is generic enough. 👍

nabsul:
:shipit:

PR #884, opened 2017-02-14
Fix IE11 and MS Edge errors

DanReyLop:
Our plugin was completely broken on IE11 and Edge. This is what was happening:

`require()` statements are executed **after** `import` statements, even if they are before in the file. That means that the Polyfill wasn't executed until all the `import`ed JS modules were parsed and executed. This broke IE11 pretty bad. The solution is to include the polyfill using a regular `import`. This will break if there is another React-based plugin in the same page, but for now it's an acceptable tradeoff. It will be fixed when we start using `babel-preset-env` (spiking on it this week).

Also, MS Edge breaks if `fetch()` is called with a request with `body === null`, for some reason. I've changed it to include the `body` property only when there is actually a body.

### DIFF
--- a/client/lib/save-form/index.js
+++ b/client/lib/save-form/index.js
@@ -10,8 +10,10 @@ const saveForm = ( setIsSaving, setSuccess, setFieldsStatus, setError, url, nonc
 		headers: {
 			'X-WP-Nonce': nonce,
 		},
-		body: formData ? JSON.stringify( formData ) : null,
 	};
+	if ( formData ) {
+		request.body = JSON.stringify( formData );
+	}
 
 	return fetch( url, request ).then( ( response ) => {
 		setError( null, null );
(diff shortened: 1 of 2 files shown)


jeffstieler:
> The solution is to include the polyfill using a regular import. This will break if there is another React-based plugin in the same page, but for now it's an acceptable tradeoff.

I'm still waiting for an IE 11 VM to download (yay!) so I can't test this idea, but can we just add the polyfill to the webpack output like so?

`'woocommerce-services': [ 'babel-polyfill', './client/main.js' ],`

> It will be fixed when we start using `babel-preset-env`

Is that because it turns:

`import "babel-polyfill";`

into:

`import "core-js/modules/es7.string.pad-start";`
`import "core-js/modules/es7.string.pad-end";`

?

DanReyLop:
> can we just add the polyfill to the webpack output like so?

It's the same thing essentially. It would also break if there's another Babel-based bundle in the same page.

> Is that because it turns ... into ...?

Yep. Only the `babel-polyfill` package has the check that makes it fail if there is another polyfill already in the page. The individual `core-js` modules don't have that check.

PR #888, opened 2017-02-18
Change the bundles URLs on each new release to avoid cache issues

### LINKED ISSUE #887: Change the bundles file names on each new release to avoid cache issues
See https://github.com/Automattic/woocommerce-connect-support/pull/32 and https://github.com/Automattic/woocommerce-connect-support/pull/33

It's pretty normal that "static" files such as JS and CSS are given really long cache times, like 1 year. We need to make sure that when a user updates the plugin he is using the latest assets.

The simplest way to do that is to append something like `?version={PLUGIN_VERSION}` to our JS and CSS files when they are appended to the page. Note that WordPress already does something similar (except with the version of WordPress, not the plugin version), so maybe there can be some conflict there.

DanReyLop:
Fixes #887

Turns out, this is an already built-in WP functionality. By default it appends `?ver={WP_VERSION}` to all the JS/CSS, but you can specify a custom string. I've used the plugin version.

To test:
* Open any WCS page (Order screen, Self-help page, etc)
* Check in the Network Inspector that all WCS-related files were requested with `?ver=1.1.1` at the end

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -341,6 +341,8 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		 * @codeCoverageIgnore
 		 */
 		public function init() {
+			add_action( 'admin_init', array( $this, 'admin_enqueue_scripts' ) );
+
 			if ( ! get_option( 'wc_connect_tos_accepted', false ) ) {
 				add_action( 'admin_init', array( $this, 'admin_tos_notice' ) );
 				return;
@@ -425,7 +427,6 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 				add_action( 'woocommerce_shipping_zone_method_status_toggled', array( $this, 'shipping_zone_method_status_toggled' ), 10, 4 );
 			}
 
-			add_action( 'admin_enqueue_scripts', array( $this, 'admin_enqueue_scripts' ) );
 			add_action( 'rest_api_init', array( $this, 'rest_api_init' ) );
 			add_action( 'woocommerce_settings_saved', array( $schemas_store, 'fetch_service_schemas_from_connect_server' ) );
 			add_action( 'wc_connect_fetch_service_schemas', array( $schemas_store, 'fetch_service_schemas_from_connect_server' ) );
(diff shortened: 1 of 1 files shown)


woocommerce-services.php
```diff
@@ -341,6 +341,8 @@ public function load_textdomain() {
 		 * @codeCoverageIgnore
 		 */
 		public function init() {
+			add_action( 'admin_init', array( $this, 'admin_enqueue_scripts' ) );
```
DanReyLop: `admin_enqueue_scripts` only `registers` the bundles, doesn't do any actual work, so it's safe to execute always.

woocommerce-services.php
```diff
@@ -651,12 +652,17 @@ public function woocommerce_payment_gateways( $payment_gateways ) {
 		 * Registers the React UI bundle
 		 */
 		public function admin_enqueue_scripts() {
-			wp_register_style( 'noticons', plugins_url( 'assets/stylesheets/noticons.css', __FILE__ ), array(), '20150727' );
-			wp_register_style( 'dashicons', plugins_url( 'assets/stylesheets/dashicons.css', __FILE__ ), array(), '20150727' );
+			// Note: This will break outside of wp-admin, if/when we put user-facing JS/CSS we'll have to figure out another way to version them
```
DanReyLop: WooCommerce core uses a hard-coded `define( 'WC_VERSION', '2.7.0' )` constant for this, but as long as we only need to be inside wp-admin, there's no need for that.

jeffstieler: We could just hardcode a date-based version string for each registered script and style. For example, a version of `20170218` or `1487454198`.

This has worked well for me in the past, and allows us to update versions only for scripts and styles that change between releases. It also doesn't require plugin file header parsing on every admin request.

What do you think?

DanReyLop: > ... allows us to update versions only for scripts and styles that change between releases.

That's not a big deal. Realistically speaking, we are going to update the main JS bundle on every release.

> It also doesn't require plugin file header parsing on every admin request.

That's not a performance problem, something in the wp-admin code already parses *all* the plugin headers on every request.

> We could just hardcode a date-based version string for each registered script and style. For example, a version of 20170218 or 1487454198.

The problem with that is that we will forget. Using the same strategy as WooCommerce (`define 'WC_VERSION', '1.2.3')`) is a little less error-prone because it's just one value. Even then, I would prefer to fully automate this while possible (that is, until we start servin JS/CSS in non-admin pages), and parsing the plugin info is the most straight-forward way to do that.

jeffstieler:
Fix the merge conflicts and :shipit:.

PR #889, opened 2017-02-18
Make our JS bundle and plugin ZIP smaller

### LINKED ISSUE #886: [Spike] Make our JS bundle and plugin ZIP smaller
Currently, our JS bundle alone is 1.5MB (!!!) minified (but uncompressed). That's A LOT for a mediocre computer/browser to handle efficiently. Plus, our plugin ZIP file is much bigger than it should.

Proposed actions (from the top of my head):
* Use some babel transformations (remove PropTypes, babel-preset-env, etc)
* Try to look for lighter libraries (libphonenumber is MASSIVE and we are using it for a simple textfield, I assume full responsibility for that)
* Use Webpack 2 with its nice tree-shaking
* Look for unused CSS (are we actually using dashicons/noticons?)
* Don't commit the source files to the .org repo, only the needed files to run the plugin

DanReyLop:
Fixes #886 

This PR includes all the efforts to make both our JS/CSS assets and the whole plugin ZIP file smaller. I'll track 3 metrics:
* Minified but uncompressed JS bundle. Baseline: **1,537,077 bytes**
* Minified but uncompressed CSS bundle. Baseline: **268,952 bytes**
* Plugin ZIP size. Baseline: **693,860 bytes**

For CSS & JS, even if they will (probably) be `gzipped` when going over the network, the most important metric is the *uncompressed* size because that's the amount of code the browser will have to actually parse & execute.


DanReyLop:
Final numbers:
* Minified but uncompressed JS bundle: **766,520 bytes**
* Minified but uncompressed CSS bundle: **187,927 bytes**
* Plugin ZIP size: **357,325 bytes**

I'd call this a success :)

Over the next few months, `wp-calypso` will replace all its `require` / `module.exports` statements with `import` / `export`, so we will be able to use Webpack2 to its full potential and save another 50KB of JS bundle or so.

Also, on Webpack's roadmap is to implement scope hoisting, which will greatly help projects with a great number of really small modules, such as this one. Now that we are using the latest WebPack version, these improvements will come for free, just by updating the dependency.

This PR is massive, I don't recommend reviewing all of it at once. I've kept my commits short and focused, and in each commit message it's explained the rationale of each change, so reviewing commit by commit is the best way to do it.

webpack.production.config.js
```diff
@@ -12,9 +12,17 @@ config.plugins.push( new webpack.optimize.OccurrenceOrderPlugin( false ) );
 
 config.plugins.push( new webpack.optimize.UglifyJsPlugin( {
 	compress: {
+		screw_ie8: true,
```
jeffstieler: :rofl:

DanReyLop:
Done. `dashicons` aren't used at all, and `noticons` are only used for the `checked` state of checkboxes, absolutely nowhere else. I guess I didn't have any checked checkboxes in my entire UI, I 100% didn't expect it to break in that way :(

PR #890, opened 2017-02-20
Fix/451 method ids

### LINKED ISSUE #451: Having WooCommerce Connect Client active breaks WooThemes USPS Extension
With WCC activated, the (legacy) USPS option does not appear in navigation:

screen shot 2016-07-06 at 11 40 44 am

But once you deactivate WCC, it appears and is accessed just fine:

screen shot 2016-07-06 at 11 40 07 am

Possibly related to WCC using `usps` as the method_id

marcinbot:
Fixes #451 

The CL does two things:
* switches to using `method_id` instead of `id` to identify shipping zone methods
* on the first plugin load it migrates the existing services to also use `method_id`

To test:
* checkout Automattic/woocommerce-connect-server#785
* verify that every service-related request (saving and loading settings, requesting rates) works as expected
* also the WooCommerce USPS plugin should work now side by side with WooCommerce Services

### DIFF
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -237,12 +237,12 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 
 				$new_method_id = $service_schema->method_id;
 
-				$query = $wpdb->prepare(
-					"UPDATE {$wpdb->prefix}woocommerce_shipping_zone_methods " .
-					"SET method_id=%s " .
-					"WHERE instance_id=%s AND method_id=%s",
-					$new_method_id, $instance_id, $service_id);
-				$wpdb->query( $query );
+				$wpdb->update(
+					"{$wpdb->prefix}woocommerce_shipping_zone_methods",
+					array( 'method_id' => $new_method_id ),
+					array( 'instance_id' => $instance_id, 'method_id' => $service_id ),
+					array( '%s' ),
+					array( '%d', '%s' ) );
 
 				//update the migrated service settings
 				WC_Connect_Options::update_shipping_method_option( 'form_settings', $service_settings, $new_method_id, $instance_id );
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -214,8 +214,9 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 		 * Checks if the shipping method ids have been migrated to the "wc_services_*" format and migrates them
 		 */
 		public function migrate_legacy_services() {
-			if ( WC_Connect_Options::get_option( 'shipping_methods_migrated', false )
-				|| ! $this->service_schemas_store->get_all_shipping_method_ids() ) {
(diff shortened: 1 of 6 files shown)


jeffstieler:
On first page load I saw the following PHP warnings:

`Warning: Invalid argument supplied for foreach() in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/woocommerce-services.php on line 629`

`Warning: Invalid argument supplied for foreach() in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/woocommerce-services.php on line 645`

classes/class-wc-connect-service-settings-store.php
```diff
@@ -206,7 +208,49 @@ public function get_enabled_services() {
...
+				}
+
+				$new_method_id = $service_schema->method_id;
+
+				$query = $wpdb->prepare(
+					"UPDATE {$wpdb->prefix}woocommerce_shipping_zone_methods " .
+					"SET method_id=%s " .
+					"WHERE instance_id=%s AND method_id=%s",
+					$new_method_id, $instance_id, $service_id);
+				$wpdb->query( $query );
```
jeffstieler: Can [`$wpdb->update()`](https://developer.wordpress.org/reference/classes/wpdb/update/) be used here instead?

woocommerce-services.php
```diff
@@ -623,8 +624,10 @@ protected function get_service_object_by_id( $class_name, $service_id ) {
 		 * @return mixed
 		 */
 		public function woocommerce_shipping_methods( $shipping_methods ) {
-
-			$shipping_service_ids = $this->get_service_schemas_store()->get_all_service_ids_of_type( 'shipping' );
+			$shipping_service_ids = $this->get_service_schemas_store()->get_all_shipping_method_ids();
+			if ( ! $shipping_service_ids ) {
+				return $shipping_methods;
+			}
```
jeffstieler: See: https://github.com/Automattic/woocommerce-services/pull/890/files#r102328474 (applies to all below as well)

jeffstieler:
The changes look good - but I've come up with another question (sorry :D)

Is there any way to use a combination of `id` and `method_id` in the shipping method so we don't have to add support for two additional endpoints on the server?

Asked another way - why can't we still `POST /services/usps/settings`?

classes/class-wc-connect-service-settings-store.php
```diff
@@ -206,7 +208,49 @@ public function get_enabled_services() {
...
 			return $enabled_services;
+		}
+
+		/**
+		 * Checks if the shipping method ids have been migrated to the "wc_services_*" format and migrates them
+		 */
+		public function migrate_legacy_services() {
+			if ( WC_Connect_Options::get_option( 'shipping_methods_migrated', false ) //check if the method have already been migrated
+				|| ! $this->service_schemas_store->fetch_service_schemas_from_connect_server() //ensure the latest schemas are fetched
+				|| ! $this->service_schemas_store->get_all_shipping_method_ids() ) { //verify that the schemas contain the method_id field
```
DanReyLop: This check isn't needed. We can safely release the server before the client, server's changes are backwards-compatible. It doesn't hurt though, so leave it if you want.

PR #901, opened 2017-02-24
Fixed nux notice dismissal on the service settings page

marcinbot:
To test:
* remove `wc_connect_nux_notices` from the `wp_usermeta` table
* open shipping settings and dismiss the nux notice
* refresh
* verify the notice is gone for good


jeffstieler:
Can you write some steps to reproduce the issue this fixes? Is it perhaps a bug we introduced since the last release?

I tested on 1.1.1 and didn't have the notice reappear.

PR #902, opened 2017-02-24
Labels: Don't initially show red error highlights for missing destination address.

jeffstieler:
Apply same “ignore validation” logic to the destination shipping address during label creation.

For manual orders that may not have a shipping address yet, don’t scare the merchant with a “wall of red”.

To test:
* Edit an order to remove the shipping address, or manually create an order
* Click "create label"
* Verify that the destination step opens, but does not show errors until interacted with


v18:
:shipit:

Issue #819, opened 2017-01-22
Shipping labels: promote savings to merchant 


kellychoffman:
What benefit is it serving? Its too small to see, and the space could be better utilized. We can add a price summary there. Idea:

<img width="495" alt="screen shot 2017-01-22 at 1 20 21 pm" src="https://cloud.githubusercontent.com/assets/5835847/22184319/a24b4d10-e0a5-11e6-8e4d-3fee9909df81.png">

DanReyLop:
Can we get an agreement on this one?

Originally the preview was implemented because "people want to see what they are about to buy", even if it's fake. Now, with the paper size selection it now shows how the labels fit into a page and how many pages it will be. I'm just stating what purpose does it serve, I'm not saying we should neccesarily keep it :)

If we axe this, we can get rid on some technical debt on this component: https://github.com/Automattic/woocommerce-services/issues/598 https://github.com/Automattic/woocommerce-services/issues/718 and specially https://github.com/Automattic/woocommerce-services/issues/675

cc/ @jeffstieler @nabsul @v18 @marcinbot

kellychoffman:
Well, you know my opinion :)

> Now, with the paper size selection it now shows how the labels fit into a page and how many pages it will be

Good point, but this can also be served with the print dialog, which has a preview. And I would guess people aren't switching back and forth between different paper sizes per label. We could add that to a global setting that you set up initially, but one decision at a time 😀

jeffstieler:
> Originally the preview was implemented because "people want to see what they are about to buy", even if it's fake.

I remember this being the motivation, but I don't remember it being backed by anything other than gut feel. Even our Stamps.com extension doesn't show a preview: https://docs.woocommerce.com/document/woocommerce-shipping-stamps/

The only tangible value I can come up with for a preview is to do a test print, but we aren't really facilitating that right now.

What about removing the "preview" from the label modal, and adding a "print test label" feature to the System Status page?

kellychoffman:
That sounds like a plan to me.

jeffstieler:
Giving this issue a kick 😄 ..

![](https://cloud.githubusercontent.com/assets/5835847/22184319/a24b4d10-e0a5-11e6-8e4d-3fee9909df81.png)

This reimagined use of the preview real estate is great - but we can't deliver all of it now, and there are some nuances to discuss.

1. We can't offer insurance (yet)
1. Not all shipping services can be offered to our customers at a discount (including USPS services)
[...]

PR #908, opened 2017-02-25
Fix printing labels in different browsers

### LINKED ISSUE #885: Invoking the native "Print" dialog is broken in several browsers
Welcome to the old pains of cross-browser development.

Currently, the way labels PDFs are printed is as follows:
* User clicks button.
* PDF URL is constructed and loaded into an invisible `<iframe>`.
* When that PDF finishes loading, invoke `print()` in that `<iframe>`.
* Profit!

That doesn't work in all browsers (Firefox, IE or Edge), only in Chrome (and maybe Safari).

DanReyLop:
Fixes #885 

Sorry for the wall of text, but this issue had a lot of traps and nuances, I needed a place to dump my brain.

# Browser differences
## Chrome

An `<iframe>` can have it's `print()` method invoked any time, at least if it's from the same origin (which it is). Thus, the previous code worked fine here.

## Safari

Haven't tried yet.

## Firefox

Due to [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=911444) (I'm following it so I'll know if it eventually gets fixed), PDFs loaded into a webpage internally use a separate `<iframe>` with a different domain, so Cross-Domain restrictions apply and the `print()` method can't be invoked at all.

The best way to "fallback" in this case is to open the PDF into a new tab, using `window.open()`. **But**, this must be done as a direct response to an user interaction or it will be blocked by the browser's built-in popup blocker.

## IE11

IE11 doesn't come with a built-in PDF viewer. Thus, when trying to load a PDF into an `<iframe>`, the `Download/Open` pop-up is shown.
With `Acrobat Reader`, you can call `print()` on a PDF loaded inside an `<iframe>`, but for some reason it just prints a blank page.
**NOTE**: When enabling/disabling Acrobat Reader on IE11, restart the browser before testing anything or funky stuff will happen.

## MS Edge

The `<iframe>` must be visible. If it has `display: none` it won't load. It can be tricked though, just by rendering it visible but off-screen.

-----

There are 2 printing-related use cases: label purchase and label re-print.

# Label Purchase
## Previous behaviour
* User clicks `Buy` button.
* AJAX request is done to the server.
* Server replies with `success` and a list of labels.
* A PDF URL is constructed and loaded into an invisible `<iframe>`.
* When that PDF is loaded (`iframe.onload` handler), the native `Print` dialog is triggered by calling `iframe.contentWindow.print()`.

### Chrome, MS Edge (and I hope Safari)
Same as before, this is working fine.

### IE11 (without Acrobat Reader)
* User clicks `Buy` button.
* AJAX request is done to the server.
* Server replies with `success` and a list of labels.
* A PDF URL is constructed and loaded into an invisible `<iframe>`.
* Immediately, the native `Download / Open` popup is displayed. The name of the file is `print.pdf`. The user can then download it and print it himself.

### IE11 (with Acrobat Reader), and Firefox
* User clicks `Buy` button.
* AJAX request is done to the server.
* Server replies with `success` and a list of labels.
* The `Buy` button becomes a `Print` button.
* User clicks `Print` button.
* A PDF URL is constructed and loaded in a new tab using `window.open()`.
* In that tab, the PDF will be full-screen, with the Acrobat toolbar (which has a big Print button).

# Label reprint
## Previous behaviour
* User clicks `Print` button inside the `Reprint` dialog.
* A PDF URL is constructed and loaded into an invisible `<iframe>`.
* When that PDF is loaded (`iframe.onload` handler), the native `Print` dialog is triggered by calling `iframe.contentWindow.print()`.

### Chrome, MS Edge (and I hope Safari)
Same as before, this is working fine.

### IE11 (without Acrobat Reader)
* User clicks `Print` button inside the `Reprint` dialog.
* A PDF URL is constructed and loaded into an invisible `<iframe>`.
* Immediately, the native `Download / Open` popup is displayed. The name of the file is `print.pdf`. The user can then download it and print it himself.

### IE11 (with Acrobat Reader), and Firefox
* User clicks `Print` button inside the `Reprint` dialog.
* A PDF URL is constructed and loaded in a new tab using `window.open()`.
* In that tab, the PDF will be full-screen, with the Acrobat toolbar (which has a big Print button).

----

# Testing done

- [x] Chrome - Purchase Label (prints automatically)
- [x] Chrome - Reprint Label (prints automatically)
- [x] IE11 - Purchase Label (`Download` pop-up)
- [x] IE11 - Reprint Label (`Download` pop-up)
- [x] IE11 with Acrobat Reader - Purchase Label (`Purchase` button changes to `Print`, then opens PDF in new tab)
- [x] IE11 with Acrobat Reader- Reprint Label (opens PDF in new tab)
- [x] MS Edge - Purchase Label (prints automatically)
- [x] MS Edge - Reprint Label (prints automatically)
- [x] Firefox - Purchase Label (`Purchase` button changes to `Print`, then opens PDF in new tab)
- [x] Firefox - Reprint Label (opens PDF in new tab)
- [x] Safari - Purchase Label (should print automatically)
- [x] Safari - Reprint Label (should print automatically)

Can anyone please test it in Safari?


DanReyLop:
It's specially important to test purchasing a label in Firefox, because the UI changed there. After the purchase is completed, the `<Spinner> Purchasing...` button changes to `Print` and it must be clicked again. Maybe that's too subtle?

v18:
I tested with Safari 10.0.3 on Mac 10.11.6, and both printing & reprinting works.

I also tested with Chrome 56.0.2924.87 and I ran into an issue where the label will not print (either on original print or re-print):

screen shot 2017-03-01 at 8 28 31 am

The error I saw was: 

> You tried to return focus to null but it is not in the DOM anymore

client/lib/utils/print-document.js
```diff
@@ -1,27 +1,68 @@
...
+					return true;
+				} );
+
+		case 'addon':
+			// window.open will be blocked by the browser if this code isn't being executed from a direct user interaction
+			if ( isUserInteraction ) {
+				window.open( url );
+				return Promise.resolve( true );
+			}
+			return Promise.resolve( false );
```
v18: Should this be `reject` with an error?

client/lib/utils/print-document.js
```diff
@@ -1,27 +1,68 @@
...
+/**
+ * Opens the native printing dialog to print the given URL.
+ * Falls back to opening the PDF in a new tab if opening the printing dialog is not supported.
+ * @param {string} url URL of the document to print
+ * @param {boolean} isUserInteraction True if this code is being executed as a result of a direct user interaction,
+ * such as the click of a button. False otherwise (if this was triggered by an AJAX response, for example
+ * @returns {Promise} Promise that resolves with true if the printing dialog (or equivalent) was correctly
+ * invoked. It resolves with false if it's required to call this function again as a result of direct user interaction.
+ */
+export default ( url, isUserInteraction ) => {
```
v18: We always use this function assuming it returns a Promise. However, we don't always return a Promise (like on line 51 - `return true`. Can we return a Promise here from the start like `loadDocumentInFrame` above, and resolve and reject as needed within?

client/lib/utils/print-document.js
```diff
@@ -1,27 +1,68 @@
...
+	switch ( pdfSupport() ) {
+		case 'native':
+			// Happy case where everything can happen automatically. Supported in Edge, Chrome and Safari
+			return loadDocumentInFrame( url )
+				.then( () => {
+					iframe.contentWindow.print();
+					return true;
+				} );
+
```
v18: What happens if `loadDocumentInFrame` rejects?

client/lib/utils/print-document.js
```diff
@@ -1,27 +1,68 @@
...
+		case 'addon':
+			// window.open will be blocked by the browser if this code isn't being executed from a direct user interaction
+			if ( isUserInteraction ) {
+				window.open( url );
+				return Promise.resolve( true );
+			}
+			return Promise.resolve( false );
+
+		default:
+			// If browser doesn't support PDFs at all, this will trigger the "Download" pop-up.
```
v18: Is there an example of a browser that doesn't support PDFs? Want to try this out.

client/lib/utils/print-document.js
v18: Can we move the boolean `isUserInteraction` out into a separate function (for example `printOnUserInteraction`) or add an object parameter instead? It's a bit hard to read the function calls otherwise.

client/shipping-label/state/actions.js
```diff
@@ -514,13 +515,18 @@ export const purchaseLabel = () => ( dispatch, getState, context ) => {
...
+					.then( ( success ) => {
 						const noticeText = 1 === response.length
 								? __( 'Your shipping label was purchased successfully' )
 								: sprintf( __( 'Your %d shipping labels were purchased successfully' ), response.length );
 						dispatch( NoticeActions.successNotice( noticeText ) );
-						dispatch( exitPrintingFlow( true ) );
+						if ( success ) {
+							dispatch( exitPrintingFlow( true ) );
+						} else {
+							dispatch( { type: SHOW_PRINT_CONFIRMATION, printUrl } );
```
v18: `SHOW_PRINT_CONFIRMATION` seems to only depend on the browser behaviour defined in `printDocument`. Can we update the `purchaseLabel` flow instead to decide which button to show? Maybe using a component prop.

client/lib/pdf-label-utils/index.js
```diff
@@ -1,6 +1,7 @@
 import { translate as __ } from 'lib/mixins/i18n';
 import querystring from 'querystring';
 import _ from 'lodash';
+import pdfSupport from 'lib/utils/pdf-support';
```
v18: Can we rename this to `getPdfSupport` or `getPdfSupportType`?

assets/stylesheets/style.scss
```diff
@@ -40,6 +40,14 @@
...
         padding: 5px 14px 7px;
+
+		.spinner {
+			margin-bottom: -8px;
+
+			.spinner__border {
+				fill: transparent;
+			}
+		}
     }
```
v18: There are some spacing (spaces vs tabs) issues right around this block.

nabsul:
I testing printing a new label and reprinting in Firefox, Chrome, Edge and IE11 in Windows and they all behave as expected.

Out of curiosity I also tested IE 10, 9, 8 (in IE11's emulation mode). CSS starts to break down in 9 and 10, but it still works. IE8 Doesn't render the react app at all :-)

client/lib/pdf-label-utils/index.js
```diff
@@ -45,5 +46,5 @@ export const getPrintURL = ( paperSize, labels, { labelsPrintURL, nonce } ) => {
 };
 
 export const getPreviewURL = ( paperSize, labels, { labelsPreviewURL, nonce } ) => {
-	return _getPDFURL( paperSize, labels, labelsPreviewURL, nonce );
+	return pdfSupport() ? _getPDFURL( paperSize, labels, labelsPreviewURL, nonce ) : null;
```
DanReyLop: Yes, I added this to not break everything in IE11 right now, but this line of code will soon get removed by #819. `pdfSupport()` is used in more places.

client/lib/utils/print-document.js
DanReyLop: IE11. That's before you install the Adobe Acrobat plugin for it, of course :)

DanReyLop:
> IE8 Doesn't render the react app at all :-)

React doesn't support IE8 at all, so I'm not surprised :)

client/lib/utils/print-document.js
v18: On second thought, we're all set here since we're not always doing something async, and returning a value is the same as returning `Promise.resolve(value)`. Having a different behaviour for the default case is still a bit hard to digest, though.

DanReyLop: This function is always returning a Promise chain. In line 48 it returns a Promise, that eventually will resolve to `true`. Sometimes this function is sync, sometimes it's async, the beauty about this is that the calling code doesn't need to care. You weren't here when we started using Promises, these concepts can be hard to grasp at first :)

client/lib/utils/print-document.js
DanReyLop: I usually only use `rejections` for real errors, such as a `404 Not Found` when loading the PDF or things like that. This case just means "Couldn't print, but it you do it again with `isUserInteraction` set to `true` it will work". It **could** be a special kind of error, but I'm not a fan of modeling "normal" flows around what are basically `try... catch` constructs.

client/lib/utils/print-document.js
DanReyLop: The rejection will propagate through the Promise chain to the caller of `printDocument` (like [here](https://github.com/Automattic/woocommerce-services/pull/908/files#diff-d8b04016a3889c977c51e0ed0ae5a2f7R665)) and that will show an error notice.

client/lib/utils/print-document.js
DanReyLop: After the AJAX request for purchasing a label is completed, the function `printDocument(url, false)` is called, regardless of the browser capabilities. If the browser needs a direct user interaction, the Promise is resolved with `false`, that result is used [here](https://github.com/Automattic/woocommerce-services/pull/908/files/c224461151150ed30ac092db3f0e8888aec7a6ca#diff-d8b04016a3889c977c51e0ed0ae5a2f7R525), that code ultimately changes the `Purchasing...` button to `Print`. Since I had to explain how it works, that's a synthom that it's too complicated, I'll try to simplify.

client/lib/utils/print-document.js
v18: It looks like the rejection would be uncaught in the `default` case for printDocument. Not sure why I added my note on this line ;). If it comes back to me I'll reply again.

DanReyLop: Oh, yeah, that one. Since in the `default` case nothing will be loaded in the `<iframe>` (because the browser doesn't support PDFs), that Promise would never resolve or reject, so it's just ignored.

client/shipping-label/state/actions.js
```diff
@@ -514,13 +515,18 @@ export const purchaseLabel = () => ( dispatch, getState, context ) => {
...
+						}
 					} )
 					.catch( ( err ) => {
 						console.error( err );
```
DanReyLop: At this point, it's better to just let the user refresh the page. We don't have logic in place to correctly recover from an error in this flow. For example, the EasyPost `shipment_id` is not reset, so another try at purchasing would give an EasyPost error for sure.

client/shipping-label/state/actions.js
DanReyLop: We still need a way to indicate the purchase dialog "Purchase has completed, but don't close yourself yet". That has to go into the redux state tree anyway, so at least I prefer to make it explicit here.

DanReyLop:
> I also tested with Chrome 56.0.2924.87 and I ran into an issue where the label will not print (either on original print or re-print):

This is gonna sound incredibly arrogant but... it's not my code, my code is amazing, it's a bug in Chrome 🏆 
But seriously, it is. I've just tried and I can't print *anything* in Chrome 56 (I could on Chrome 55). Just try to print any webpage to double-check. I'll see if there's a bug filled already.

The `You tried to return focus to null but it is not in the DOM anymore` warning is unrelated, caused by Calypso changing its custom `<Dialog>` implementation to use `react-modal` ([here](https://github.com/Automattic/wp-calypso/pull/10819)). It's harmless though, everything works fine.

assets/stylesheets/style.scss
DanReyLop: Fixed on https://github.com/Automattic/woocommerce-services/pull/908/commits/70351831062649961cec8c67f80006acb7aea30c

client/lib/utils/print-document.js
DanReyLop: Simplified on https://github.com/Automattic/woocommerce-services/pull/908/commits/70351831062649961cec8c67f80006acb7aea30c

client/shipping-label/state/actions.js
```diff
@@ -648,7 +660,7 @@ export const confirmReprint = () => ( dispatch, getState, context ) => {
 	dispatch( { type: CONFIRM_REPRINT } );
 	const state = getState().shippingLabel;
 	const labelId = state.reprintDialog.labelId;
-	printDocument( getPrintURL( getState().shippingLabel.paperSize, [ { labelId } ], context ) )
+	printDocument( getPrintURL( getState().shippingLabel.paperSize, [ { labelId } ], context ), true )
 		.then( () => dispatch( closeReprintDialog() ) )
 		.catch( ( error ) => dispatch( NoticeActions.errorNotice( error.toString() ) ) );
```
DanReyLop: Fixed on https://github.com/Automattic/woocommerce-services/pull/908/commits/70351831062649961cec8c67f80006acb7aea30c

client/lib/utils/pdf-support.js
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * This function result is cached for performance reasons, since browser capabilities won't change.
+ *
+ * @returns {string|false} "native" if the browser can display PDFs and code can reliably call JS
+ * functions on those PDFs (such as print()). "addon" if the browser can display PDFs, but there's
+ * no guarantee that they will respond to a DOM-like API. false if the browser can't display
+ * PDFs at all.
+ */
+export default _.memoize( () => {
+	if ( navigator.userAgent.includes( 'Firefox' ) ) {
```
jeffstieler: Is there any alternative to using `navigator.userAgent`?

MDN lists it as a deprecated feature that will be removed eventually. See: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/userAgent

DanReyLop: `navigator.userAgent` is here since the beginning of JS, and virtually every page in the world relies on it. It may not be considered "best practice" to rely on it, but no browser is going to remove it. AFAIK, there's no other reliable way to check if the browser is `Firefox`.

v18:
@DanReyLop - I'm still not 100% sold on add the logic for `getPurchaseButtonLabel` in the state. It makes a it a bit difficult to reason about which button will show up. From what I can tell there isn't a reason to try to print the PDF to see if it doesn't work first (and then dispatch the `SHOW_PRINT_CONFIRMATION` action) - let me know if I missed something, though.

I think we can rely on `getPDFSupport()` directly in `PrintLabelDialog`.

DanReyLop:
> From what I can tell there isn't a reason to try to print the PDF to see if it doesn't work first

I've changed that. Now the AJAX purchase request is made, and **after** that, the `SHOW_PRINT_CONFIRMATION` action is dispatched if needed. That's what's needed to transition from the `Purchasing...` spinner to the `Print` button confirmation (or to automatically close the dialog if it's Chrome/Edge/Safari).

> I think we can rely on getPDFSupport() directly in PrintLabelDialog.

The dialog still needs to know somehow that it's on the "The purchase completed but this dialog must **not** be closed yet". One thing is to know the browser capabilities, and other is to know the current state of the purchase. For showing the `Print` button you need to know at least the state.

I hope it makes a little more sense now. I get that it's harder to follow than before, but I honestly don't know how to simplify it more. This PR adds complexity because the problem it solves is tricky, there's n way around that.

v18:
@DanReyLop yeh, I agree the new `'addon' === getPDFSupport()` check makes it much clearer :).
I missed that the `printDocument` is coupled string with the `labelActions.purchaseLabel` so that makes it difficult to update the button without using the state right now. Generally, I'd say it would be worth de-coupling those, but it's not a blocker for this PR. 🚢

PR #909, opened 2017-02-28
Fix Purchase Label Dialog cards not resizing to fill the width of the dialog on IE11

### LINKED ISSUE #906: Purchase Label dialog layout is not responsive in IE11
ie11-layout

DanReyLop:
Fixes #906

CSS `initial` value for properties isn't supported in IE. I've changed the `initial` usages for the default values (`0` for `min-width` and `none` for `max-width`), that's exactly equivalent but works in IE :)

To test:
* Open the Purchase Label UI in IE11 in a wide monitor.
* See that the steps take the whole available horizontal space.

### DIFF
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -310,8 +310,8 @@
 
 	// WP style conflict resets
 	.card {
-		min-width: initial;
-		max-width: initial;
+		min-width: 0;
+		max-width: none;
 	}
 
 	.wp-admin & {


nabsul:
I testing printing a new label and reprinting in Firefox, Chrome, Edge and IE11 in Windows and they all behave as expected.

Out of curiosity I also tested IE 10, 9, 8 (in IE11's emulation mode). CSS starts to break down in 9 and 10, but it still works. IE8 Doesn't render the react app at all :-)

nabsul:
I was not able to repro this issue. From master and on this PR, the elements expand to the full width of the modal. However, since the change is trivial, I'll take your word for it :shipit:

DanReyLop:
> I was not able to repro this issue.

I guess the emulation mode only can get you so far... I wouldn't trust it for this kind of stuff :)

PR #910, opened 2017-02-28
Ensure rates prop exists before searching it

v18:
To test:

1. Follow the steps on this server PR to get a rates response with an error:
https://github.com/Automattic/woocommerce-connect-server/pull/802

Or, use a fake response (listed at the end of this description) on the server.

2. Open the printing flow to get a rates request

3. See that there is no error in the console (without this fix, you'd see `Uncaught (in promise) TypeError: Cannot read property 'find' of undefined`)

4. See this for rates:

rates-with-errors

And here's what happens when clicking on the drop-down:

rates-with-errors-no-selections-available

# Communicating unavailable rates

I think this needs a bit of design help (not necessarily a blocker though), what do you @kellychoffman? At the very least, I was thinking of removing the drop-downs and explaining that there is an error:

screen shot 2017-02-28 at 5 11 54 pm

# Using a fake response

Above line 169 in lib/shipping/handler.js on the server, add this fake response code:

```
const mockUspsRates = {
  shipment_id: 'rates_weight_0_individual',
  rates: [ {
    title: 'USPS - Priority Mail',
    rate_id: 'rate_123',
    carrier_id: 'usps',
    rate: 2.87,
    service_id: 'priority',
  } ],
};
const mockFedexError = {
  code: 'ERR2',
  message: 'There was a fedex error!',
};
const fakeResponse = {
  rates: {
    weight_0_individual: {
      shipment_id: 'abcdef',
      rates: [ Object.assign( {}, mockUspsRates.rates[ 0 ], { title: 'USPS - Priority Mail translated' } ) ],
      errors: [ Object.assign( { carrier: 'fedex' }, mockFedexError ) ],
    },
    weight_1_individual: {
      shipment_id: 'abcdef',
      rates: [ Object.assign( {}, mockUspsRates.rates[ 0 ], { title: 'USPS - Priority Mail translated' } ) ],
      errors: [ Object.assign( { carrier: 'fedex' }, mockFedexError ) ],
    },
  },
};
const fakeResponse2 = {
  rates: {
    weight_0_individual: {
      errors: [ Object.assign( { carrier: 'fedex' }, mockFedexError ) ],
    },
    weight_1_individual: {
      errors: [ Object.assign( { carrier: 'fedex' }, mockFedexError ) ],
    },
  },
};
```

Then, on line  169, use `reply( fakeResponse );` You may need to change the package IDs.


nabsul:
"There isn't a rate right now....." 

I wonder if we can make this clearer? Which of the following are we trying to say?

- Something went wrong on our server side, please try again later.
- Please try tweaking your settings or the address above and try again.
- If this problem persists, please contact support.

nabsul:
Oh! My bad @v18 . I misread the description and thought you already implemented the last screenshot in there. Nevermind.

v18:
@nabsul thanks for the review!

> "There isn't a rate right now....."
> I wonder if we can make this clearer? Which of the following are we trying to say?

I think a good alternative is:

"Something went wrong on our server side, please try again later. If this problem persists, please contact support."

There is also an error send back with each response, like so:

```
{  
   "success":true,
   "rates":{  
      "weight_0_individual":{  
         "error":{  
            "code":"UNAUTHORIZED",
            "message":"Unable to access the requested resource, authorization failed."
         }
      }
   }
}
```

We can add our own pretty error message to go along with it, and show that by default. I'll make a list of possible errors we get back from easypost to see how we'd want to handle each.

kellychoffman:
> "Something went wrong ~on our server side~, please try again later. If this problem persists, please contact support."

I'm not sure "server side" would mean much to our merchants.

I think your solution good, especially getting rid of the dropdowns as you can't interact with them at all. I'd take it a step further and remove that section all together and just have the error message.

v18:
The changes in this PR now include:
- making sure the rate prop exists on `rates` before using it
- the rate errors are shown on the rates step (as per the designs above)

I'm parking this PR while I'm away. Here are the things that still need to be done:
- [ ] make sure that the title for the rate step makes sense when there are errors
- [ ] use the correct design when there is an error and only one package
- [ ] factor out the titles and other text from `<Dropdown>` + add unit tests, possibly/probably using Enzyme
- [ ] does the error architecture (error obj vs string) make sense in other steps with inline errors, like the addresses step

v18:
Rebased on master.

v18:
@olaolusoga I updated the rates step to include any errors we get from our services, like Fedex or USPS. Here's what it looks like (some styles are not exactly as shown on the screenshots - waiting for your input before updating):

Scenario 1: Multiple packages, each with no rates found, and with one error each:
screen shot 2017-04-26 at 10 09 57 pm

Scenario 2: One package that has some rates returned for one service (USPS) but an error for another service (Fedex). It also has a validation error for the form ('choose a rate'):

screen shot 2017-04-26 at 10 20 30 pm

Scenario 3: One package with some rates and an error:

screen shot 2017-04-26 at 10 21 00 pm

I also would love some input on the heading of the step itself - showing "$0.00" in the title for now, but could have other things there depending on the scenario.

For example, we show 'Unsaved changes made to packages' there when there are changes in the Packages step that have not yet been saved:

screen shot 2017-04-26 at 11 59 07 pm copy

v18:
Dev notes:
- `Object.values` was switched to `_.values`. This hacky-ish change was made so that the tests would run, but it looks like it may also help with browser compatability (IE does not implement Object.values, and we support IE11). More digging is required for this, though, but I'd prefer to just use lodash's `values` if no one objects
- related to styles not yet being finished (description ^): I need to switch out the use of <FieldError /> for a new component
- feedback about how hard or easy the new unit tests to understand is very welcome & appreciated! ([GH link to file](https://github.com/Automattic/woocommerce-services/pull/910/files#diff-9b96f40913eaf6ba71d3ebfc642534c8))
- to run the new tests alone, you can use`describe.only()`

Notes to self:
- if we use lodash, then we also need to change one more instance of `Object.values`
- naming: server error vs form error. probably want something like form -> validation, server -> ??
- I need figure out how to run these new tests on circle ci
- probably don't need carrier from server: e.g. 'fedex'

DanReyLop:
> it may also help with browser compatability (IE does not implement Object.values, and we support IE11)

Don't let this factor into your decision, we are already including polyfills for that kind of stuff.

jeffstieler:
> > Object.values was switched to _.values. This hacky-ish change was made so that the tests would run
> 
> Don't let this factor into your decision, we are already including polyfills for that kind of stuff.

Does this illustrate that we're not using the polyfills when running the tests? Is that the root problem?

v18:
> Does this illustrate that we're not using the polyfills when running the tests? Is that the root problem?

Not necessarily, since the polyfills are probably not included for tests. Some more digging into webpack is required on my part here :)

olaolusoga:
>Scenario 1: Multiple packages, each with no rates found, and with one error each:

This looks great. Curious, do we know what's causing the error(no rates available)? If so, can we tell the merchant how to fix? or do we know how the can fix it?

>Scenario 2: One package that has some rates returned for one service (USPS) but an error for another service (Fedex). It also has a validation error for the form ('choose a rate'):

Can we use a smart default here and select the cheapest available rate (akin to scenarios 3)?

>Scenario 3: One package with some rates and an error:

If USPS is selected, why are we showing a fedex error? Is it just informing the user that they can't select fedex options in the dropdown? If so, can we just make the fedex options inactive, or remove.

>I also would love some input on the heading of the step itself - showing "$0.00" in the title for now, but could have other things there depending on the scenario. For example, we show 'Unsaved changes made to packages' there when there are changes in the Packages step that have not yet been saved:

Could you explain more? Or is there a way I can view the different scenarios + titles on local?

jeffstieler:
> If USPS is selected, why are we showing a fedex error?

I believe all of our production requests are currently returning a FedEx error because we have an incomplete configuration of the account. I'll look into removing that so we don't show these errors prematurely.

v18:
@olaolusoga thanks for the review :)

> do we know what's causing the error?

I don't have a list of all the possible errors. Some that come to mind are generic server errors (here's a list of [all the possible errors](https://www.easypost.com/errors-guide#error-codes)). 

> If so, can we tell the merchant how to fix? or do we know how they can fix it?

In the beginning we'll send down errors as they come in directly from the service, with an eye to update them to more readable language that explains how to fix it or what to do next.

> Can we use a smart default here and select the cheapest available rate (akin to scenarios 3)?

Sorry, I think I added too many cases in my screenshots without an explanation. In all possible scenarios, we'll show the rate that the customer chose at checkout. If that doesn't exist, we show a warning message about it & a drop-down with all possible rates.

I think that makes sense, or would you like to tweak it? We can change that behaviour, but in a separate PR.

> If USPS is selected, why are we showing a fedex error?
> Is it just informing the user that they can't select fedex options in the dropdown? If so, can we just make the fedex options inactive, or remove.

The way I understood the Fedex vs USPS rates to work is that if someone selected both of the services, they'd get to see all of the rates available for both services. This means that they can get a Fedex error, but get all the USPS rates without an error. In that case, I'm showing the error, and all the rates. Does that make sense to do? The fedex options are not showing up in the dropdown in that case. Should I also remove the fedex error since there were at least _some_ rates shown? Was thinking a store owner may want to know, but I'm not 100% sure about that.

> I also would love some input on the heading of the step itself - showing "$0.00"
> Could you explain more? Or is there a way I can view the different scenarios + titles on local?

Sure - here I'm referring to the heading we show for the rates step. Here's what I mean: 

screen shot 2017-05-01 at 4 45 20 pm

Here's the extra text I was talking about:

screen shot 2017-05-01 at 4 46 05 pm

You can see it by going to the Order Edit screen, and clicking the 'Create new label' button.

jeffstieler:
> Should I also remove the fedex error since there were at least some rates shown? Was thinking a store owner may want to know, but I'm not 100% sure about that.

I'd vote for removing the error when rates are shown. Eventually _(I think)_ we'd want to show a generic error in the case of rate retrieval failure that hints at possible causes if obvious. E.g. "There was a problem retrieving rates, please [try again](#) in a moment." and/or "One of your packages is missing weight", etc.

v18:
> I'd vote for removing the error when rates are shown. Eventually (I think) we'd want to show a generic error in the case of rate retrieval failure that hints at possible causes if obvious

This sounds like a good server-side change to make. I'll send down an error for each package, and the rates in general, and let the client display all of the ones that are received. And the server can contain the logic for which to send, keeping with our usual paradigm on where the logic should go.

v18:
I updated the commits on this branch to be more atomic & flow more logically. I didn't open new PRs for the different parts since there's already an open PR with comments. You can read each commit's changes on their own for ease of readability :)

# What this PR does

It shows any servers errors sent for rates. You can simulate the errors by using the matching PR on the server side.

Here are the changes made:
- housekeeping changes to `<PrintLabelDialog />` in 67844113aedef909f3b012a9384d79d941053d9c
- housekeeping changes to `<ShippingRates />` in d3400a37da15c0fd38798c9ff1ab914e2112d899 and 17593bf0b33b14bf1b9285ca061dd82c74f8cafe
- updates the shipping rates reducer to handle nonexistant rates prop in 09a03ca7138dc456d12c0c844a0b593b89919542
- adds Enzyme support for tests in 72648ad5e9167a914cef72f896149e5cf0e654d7
- updated FieldText to allow custom error types (not just form validation errors) in 03a7e34b05d88ec740f3b2d99a202508e81d0a6b
- shows the rates errors in 71af402759eefcb900b66580320d850b2d3e2194 and 273b5fd515a636e119b952ff5d895fdf8cce2f82

The visual changes have all been approved already above. Any other changes will be made in subsequent PRs.

v18:
Closing this in favour of a few smaller PRs:
https://github.com/Automattic/woocommerce-services/pull/997
https://github.com/Automattic/woocommerce-services/pull/998
https://github.com/Automattic/woocommerce-services/pull/999
https://github.com/Automattic/woocommerce-services/pull/1000
https://github.com/Automattic/woocommerce-services/pull/1001

PR #911, opened 2017-03-01
Remove noticons.css

DanReyLop:
We were including a whole CSS font just to get the "checkmark" icon for checkboxes.

I've made a custom `Checkbox` component (refactored a bit the `BulkCheckbox` component in the process) that uses a `Gridicon` instead of the Noticons font. After that, I've removed the font entirely form our plugin.

To test: Just go to the package manager or to a service configuration and play with the checkboxes there.


v18:
The error we're running into is:

```
client/components/checkbox/index.js
1:17  error  'PropTypes' is defined but never used  no-unused-vars
```

Looks like we're using these props:
- `checked` (boolean)
- `readOnly` (boolean)
- `onChange` (function)

client/components/checkbox/style.scss
```diff
@@ -0,0 +1,27 @@
...
+	.gridicon {
+		color: $blue-medium;
+		pointer-events: none;
+		position: absolute;
+		left: 1px;
+		top: 1px;
+	}
+
+	&.is-disabled .gridicon {
+		color: lighten( $gray, 10% );
```
v18: Looks like we never have disabled checkboxes? Also, the grey color doesn't work me in Chrome (blue is still showing).

client/components/bulk-checkbox/index.js
```diff
@@ -11,9 +12,9 @@ const BulkCheckbox = ( { selectedCount, allCount, onToggle } ) => {
 	};
 
 	return (
-		<span className="wcc-bulk-checkbox-container" onClick={ onClick }>
-			<input type="checkbox" checked={ allSelected } readOnly />
-			{ someSelected ? <Gridicon className="bulk-select__some-checked-icon" icon="minus-small" size={ 18 }/> : null }
+		<span className="form-checkbox bulk-checkbox" onClick={ onClick }>
+			<Checkbox checked={ allSelected } readOnly />
+			{ someSelected && <Gridicon icon="minus-small" size={ 16 } /> }
```
v18: Can we move the `someSelected` logic to `Checkbox` directly, so only one component takes care of the entire visual state (unselected / partially selected / fully selected)?

The way it's written now the logic is a bit hard to follow, and also the HTML nesting is different in the two cases, making it a bit harder to change the CSS later.

client/components/checkbox/index.js
```diff
@@ -0,0 +1,16 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+import Gridicon from 'gridicons';
+import _ from 'lodash';
+
+const Checkbox = ( props ) => {
+	const otherProps = _.omit( props, 'className' );
+	return (
+		<span className={ classNames( props.className, 'form-checkbox', { 'is-disabled': props.disabled } ) } >
```
v18: Looks like we'll no longer need the class `form-checkbox` - it relies on Calypso CSS, but can remove it if we're not relying on their FormCheckbox.

v18:
I just noticed this was not up for review yet! Sorry for the pre-mature review :)

DanReyLop:
> I just noticed this was not up for review yet! Sorry for the premature review :)

Oh, it was! But apparently I missed the very trivial `PropTypes` thing :)

client/components/checkbox/style.scss
DanReyLop: checkbox-disabled

Works for me. Are you sure you put the right prop when testing? Should look like this (in the `package-list-item.js` file:
`<Checkbox checked={ selected } onChange={ onToggle } disabled />`

We aren't using it right now, but it's a use case so basic that I prefer to support it out-of-the-box.

client/components/checkbox/index.js
DanReyLop: I've used that class to style the "Gridicon checkbox", look at `client/components/checkbox/style.scss`

client/components/bulk-checkbox/index.js
DanReyLop: You're totally right. I've consolidated all the "indeterminate" functionality into the Checkbox component. This is ready for review now :)

PR #920, opened 2017-03-08
Updating the destination address after it was normalized

### LINKED ISSUE #543: If we normalize the destination address and the merchant accepts the suggestion, update the address in the order

marcinbot:
Fixes #543 

Also removes the hardcoded `usps` from the address normalization code ( @DanReyLop )

To test:
* create an order with an address that can be trivially normalized
Example: `132 Hawthorne Street San Francisco CA 94107`
  * when opening the label print dialog, the address should be automatically normalized and the address will be automatically updated on the order
* create an order with non-trivial normalization address
Example: `132 Hawthorne Street San Fransisco CA 94107` (note the Fransisco in the city name)
  * after selecting and confirming the normalized address in the label dialog the address should update on the order
* create an order with unrecognizable adress
Example: `132 Hawthorne Streat San Fransisco CA 94107`
  * after updating the address, it should get normalized and follow the logic above


classes/class-wc-rest-connect-address-normalization-controller.php
```diff
@@ -96,11 +105,29 @@ public function update_items( $request ) {
...
+		);
+	}
+
+	public function accept_suggestion( $request ) {
+		$data     = $request->get_json_params();
+		$order_id = $data[ 'orderId' ];
+		$type     = $data[ 'type' ];
+		$address  = $data[ 'address' ];
+
+		$this->update_address( $order_id, $type, $address );
```
DanReyLop: What about doing this when the actual purchase is made, or in the `Rates` request? 1 less endpoint, 1 less round-trip to the server, less complexity.

client/shipping-label/index.js
```diff
@@ -46,8 +46,8 @@ export default ( { formData, labelsData, paperSize, storeOptions, paymentMethod
 					},
 					destination: {
 						values: formData.destination,
-						isNormalized: false,
-						normalized: null,
+						isNormalized: ! ! formData.destination_normalized,
```
DanReyLop: We prefer `Boolean(value)` instead of `!!` (old discussion, I don't know how strict have we been so far).

client/shipping-label/state/normalize-address.js
```diff
@@ -31,6 +31,10 @@ export default ( dispatch, address, type, addressNormalizationURL, nonce ) => {
 				setTimeout( () => resolve( ! error ), 0 );
 			}
 		};
-		saveForm( setIsSaving, setSuccess, _.noop, setError, addressNormalizationURL, nonce, 'POST', { address, type } );
+		saveForm( setIsSaving, setSuccess, _.noop, setError, addressNormalizationURL, nonce, 'POST', { orderId, address, type } );
 	} );
 };
+
+export const acceptAddress = ( dispatch, orderId, address, type, addressAcceptURL, nonce ) => {
+	saveForm( _.noop, _.noop, _.noop, _.noop, addressAcceptURL, nonce, 'POST', { orderId, address, type } );
```
DanReyLop: This kind of function call to `saveForm` hurts my brain (all those noop...). This is a good candidate for a refactor in a near future. Maybe refactor `saveForm` to use Promises, or use something like Node's `EventEmitter` pattern.

classes/class-wc-rest-connect-address-normalization-controller.php
```diff
@@ -117,4 +144,22 @@ public function update_items_permissions_check( $request ) {
...
+		}
+
+		$order = wc_get_order( $order_id );
+		$wc_address = $order->get_address( 'shipping' );
+
+		$new_address = array_merge( array(), ( array ) $wc_address, ( array ) $api_address );
+		//rename address to address_1
+		$new_address[ 'address_1' ] = $new_address[ 'address' ];
+		//remove api-specific fields
+		unset( $new_address[ 'address' ], $new_address[ 'name' ] );
```
DanReyLop: Why are you removing `name`?

classes/class-wc-rest-connect-address-normalization-controller.php
```diff
@@ -117,4 +144,22 @@ public function update_items_permissions_check( $request ) {
 		return true; // non-authenticated service for the 'destination' address
 	}
 
+	private function update_address( $order_id, $type, $api_address ) {
+		if ( ! $order_id || 'destination' != $type ) {
```
DanReyLop: Nitpick: Use `!==` instead of `!=`

classes/class-wc-rest-connect-address-normalization-controller.php
```diff
@@ -56,20 +56,29 @@ public function register_routes() {
 				'permission_callback' => array( $this, 'update_items_permissions_check' ),
 			),
 		) );
+
+		register_rest_route( $this->namespace, '/' . $this->rest_base . '/accept', array(
```
DanReyLop: This is going to conflict hard with https://github.com/Automattic/woocommerce-services/pull/925 . If my PR is merged first I'll give you a hand on fixing the conflicts :)

classes/class-wc-connect-shipping-label.php
```diff
@@ -230,19 +230,24 @@ protected function get_destination_address( WC_Order $order ) {
...
+			$origin                 = $this->get_origin_address();
+			$selected_rates         = $this->get_selected_rates( $order );
+			$destination            = $this->get_destination_address( $order );
 
 			if ( ! $destination[ 'country' ] ) {
 				$destination[ 'country' ] = $origin[ 'country' ];
 			}
 
-			$form_data = compact( 'is_packed', 'selected_packages', 'all_packages', 'flat_rate_groups', 'origin', 'destination' );
+			$destination_normalized = get_post_meta( $order->id, 'wc_connect_destination_normalized', true );
```
DanReyLop: You can use casting:
```php
$destination_normalized = (bool) get_post_meta( $order->id, 'wc_connect_destination_normalized', true );
```

classes/class-wc-rest-connect-address-normalization-controller.php
marcinbot: We would need to handle the non-trivial normalization in the checkout UI and keep the admin normalization in case the address needs to be edited by the merchant. I think this would only add endpoints and requests to the server. But if we decide this is a good idea, it would probably be out of the scope of this PR.

classes/class-wc-rest-connect-address-normalization-controller.php
marcinbot: Thanks for noticing. I rebased and resolved the conflict, so the code is more in line with what was in `master`

classes/class-wc-rest-connect-address-normalization-controller.php
marcinbot: WooCommerce is using two fields: `first_name` and `last_name`, that we merge into `name` for our API. The purpose of this function is to revert all the changes made in https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-shipping-label.php#L199

classes/class-wc-rest-connect-address-normalization-controller.php
DanReyLop: I was referring to the `Rates` request on the Purchase Label dialog. The point is, use one of the endpoints that are already there and are hit in the course of a label purchase, instead of creating a new one.

PR #925, opened 2017-03-10
Fix compatibility with WP Super Cache

### LINKED ISSUE #438: Remove rest controller class check
At https://github.com/Automattic/woocommerce-connect-client/blob/b624e73afebd33df1c6539c0da4850a269946695/woocommerce-connect-client.php#L271

We don't need it anymore now that https://github.com/woothemes/woocommerce/pull/10435 has landed.

DanReyLop:
Following up on [this user report](https://wordpress.org/support/topic/labelprintpdf-is-only-ascii/), I discovered that printing labels doesn't work with a plugin (WP Super Cache) that millions of sites use. That's bad.

The problem is that WP Super Cache messes with the response headers, they are converted to `text/html`, so when trying to render a PDF, that header means it will be rendered as text, thus rendering binary garbage. `WP Super Cache` [has a PR](https://github.com/Automattic/wp-super-cache/pull/121) to address this (I think) but it's been in progress for months.

The real fix is on https://github.com/Automattic/woocommerce-services/commit/19a50ad185849f2656e37eb49af656c42fe6c99f but I decided to do a little refactor of our REST classes to make it easier. The 3 commits are independent to make review easier.

Also, fixed #438 while I was there.

To test:
* Install and enable `WP Super Cache`
* Enable the `WP Super Cache` option `Caching on (recommended)`
* Go through any label flow that involves a PDF. Either purchasing (and you should see the Preview, and the real label if you go through the purchase) or reprinting a label.
* On `master`, you should get a garbage text file. On this PR, you should see the PDFs correctly.

Note: Notify the user (ticket #543126) when this is released.


classes/class-wc-rest-connect-base-controller.php
```diff
@@ -0,0 +1,71 @@
...
+			array(
+				'methods'             => $this->method,
+				'callback'            => array( $this, 'run_internal' ),
+				'permission_callback' => array( $this, 'check_permission' ),
+			),
+		) );
+	}
+
+	public function run_internal( $request ) {
+		define( 'DONOTCACHEPAGE', true ); // Play nice with WP-Super-Cache
```
jeffstieler: We should check to make sure that `DONOTCACHEPAGE` isn't already defined before doing this. (See for reference: https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-cache-helper.php#L195)

DanReyLop: 🤦‍♂️  I forgot how to PHP haha

DanReyLop: Done

jeffstieler:
This establishes a new pattern of one endpoint per controller subclass.. are we sure we want that?

I don't have any particularly strong opinions about this, but would like the group to decide.

@nabsul @v18 @marcinbot @DanReyLop

DanReyLop:
Only 1 file had multiple (2) endpoints. Ideally, a "controller" should handle everything related to an "object" (get, create, update, delete), but we haven't following that, and our endpoints don't really have the shape of CRUD endpoints anyway. Maybe we shouldn't call them "controllers".

I'd rather have multiple classes per file, or as many files as needed with a sensible directory tree structure (which goes in line with @marcinbot's proposal to cleanup PHP). Adding a new endpoint shouldn't require 20 lines on the main plugin file. If the overhead is minimal, having dozens of small files is no problem.

jeffstieler:
> I'd rather have multiple classes per file, or as many files as needed with a sensible directory tree structure (which goes in line with @marcinbot's proposal to cleanup PHP). Adding a new endpoint shouldn't require 20 lines on the main plugin file. If the overhead is minimal, having dozens of small files is no problem.

Totally fair. Let's revisit if it ever becomes unwieldy.

classes/class-wc-rest-connect-packages-controller.php
```diff
@@ -8,51 +8,12 @@
...
-				'permission_callback' => array( $this, 'update_items_permissions_check' ),
-			),
-		) );
-	}
-
-	public function update_items( $request ) {
+	public function run( $request ) {
 		$packages = $request->get_json_params();
 
 		$this->service_settings_store->update_packages( $packages[ 'custom' ] );
```
jeffstieler: I definitely missed this in my review - `$this->service_settings_store` is no longer with us.

PR #928, opened 2017-03-14
Remove "Preview" from the Labels Purchase dialog and add a "Print test label" button on the help page

### LINKED ISSUE #819: Shipping labels: promote savings to merchant

DanReyLop:
Fixes #819 

In this PR:
* Refactored a bit our "React app loader" logic so that multiple "mini-apps" can be loaded in the same page.
* Removed the "Preview" from the Purchase Label dialog. Now the sidebar only has the Paper Size selector (it looks quite empty).
* Added a "Print test label" section to the self-help page. I've added it as a separate React app because it didn't make sense to use the data-driven logic of the rest of the self-help.

Screenshot:
test-print

Right now the "Print test label" can be put as the first section of the self-help, or as the last section, but not in-between. To re-arrange the sections freely, it would be needed to break down the rest of the self-help sections into independent React apps. Should be really easy.


nabsul:
Ideally, I think the test print row would go best right above the debug logs. However, at the bottom is also not terrible. At the top would be weird in my opinion.

Otherwise, the changes test well. I was able to print new labels and test print in the status page.

:shipit:

jeffstieler:
> If you are having trouble configuring your printer, use this button to print a test shipping label.

Let's keep in mind that not everyone who wants to print a test label is having trouble - perhaps they're just being thorough in the setup of their store.

Can we say this another way?

olaolusoga:
If the only two options for now are top or bottom: bottom wins. Would degrade context @ the top.

> Let's keep in mind that not everyone who wants to print a test label is having trouble - perhaps they're just being thorough in the setup of their store.

Agree. It's fine within system status as a solution for when a printing problem does arise for the entire store, but would be more useful + a better UX to add within the standard printing setup, for scenarios where a store owner wants to test the printing before an "actual print" (the thorough store owner).

We can open a new branch for that specific function though, so the thought doesn't hold this up.

DanReyLop:
> Agree. It's fine within system status as a solution for when a printing problem does arise for the entire store, but would be more useful + a better UX to add within the standard printing setup, for scenarios where a store owner wants to test the printing before an "actual print" (the thorough store owner).

Are you proposing a "Test Print" inside the Purchase Label dialog **instead** of in the System Status page? Or in both places?

Adding a "Test Print" button to the Purchase Label dialog should be fairly easy, but I agree that it shouldn't hold this PR.

What text do you propose instead of `If you are having trouble configuring your printer, use this button to print a test shipping label.`? I can't think of anything clearer than that.

olaolusoga:
>Are you proposing a "Test Print" inside the Purchase Label dialog instead of in the System Status page? Or in both places?

In both places. Two different scenarios:

1. User is having trouble configuring their printer store wide (systems status)
2. User is thorough and wants to test label printing with a real order/product (Purchase Label dialog)

>What text do you propose instead of If you are having trouble configuring your printer, use this button to print a test shipping label.? I can't think of anything clearer than that.

**How's:**
Having trouble configuring your printer?
You can run a test print for shipping labels by selecting the paper size, and clicking the print button.

Question: If they try to print and it doesn't work, do we offer suggestions on what the problem might be? Do we offer solutions? Or is it just a print button within system status?

kellychoffman:
@olaolusoga Only nitpick I have about the text is to not refer to `click`ing because of the possibility they might be on an iPad.

olaolusoga:
> Only nitpick I have about the text is to not refer to clicking because of the possibility they might be on an iPad.

Great point @kellychoffman 

**How's:**
Having trouble configuring your printer?
You can run a test print for shipping labels by selecting the paper size, then print.

kellychoffman:
That sounds good 👍

jeffstieler:
> Removed the "Preview" from the Purchase Label dialog. Now the sidebar only has the Paper Size selector (it looks quite empty).

Sorry for dragging this out, but I think we should have something to fill the space left by removing the preview before shipping this. Perhaps the label rate and applicable discount?



Though, I'm not sure there will be a discount for the merchant for all services..

.. thoughts @olaolusoga @kellychoffman ?

jeffstieler:
@kellychoffman @olaolusoga here's a screenshot of the modal without the preview:

screen shot 2017-03-21 at 8 37 40 am

DanReyLop:
I've changed the text in the self-help page.

@jeffstieler I don't think we should hold this PR until we decide what to put in the sidebar. Right now it looks a bit empty but what we had there until now was basically useless, so I don't think it's too bad to release this like it is.

> Question: If they try to print and it doesn't work, do we offer suggestions on what the problem might be? Do we offer solutions? Or is it just a print button within system status?

Sorry @olaolusoga I missed that question. It's just a print button, but if our printing code fails (and it's not a "silent error" from the browser or something undetectable) it will show a big red message with the error details. I don't know the cases on which that code could fail, but if it does, hopefully the merchant can create a ticket showing us the exact error :)

> User is thorough and wants to test label printing with a real order/product (Purchase Label dialog)

What about adding a button `"Test"` in the Purchase Label dialog sidebar, at the right of the "Paper size" dropdown? With the same UI and behaviour as in the System Status page. This would be fairly easy to implement and it doesn't add a lot of noise to the dialog.

olaolusoga:
>I think we should have something to fill the space left by removing the preview before shipping this. Perhaps the label rate and applicable discount?

@jeffstieler After the usability tests today and Maria asking where the price was, having a visual on the right for price + savings context would be great for users before they "buy & print".

And alt could be removing the right column all together and placing the "Paper Size" option below "Rates", which will also require making the 1 column full-width. 

The former feels more useful.

>What about adding a button "Test" in the Purchase Label dialog sidebar

@DanReyLop Great idea. What will the test print produce when selected; A Proof of the label, or just the default computer print?

jeffstieler:
@olaolusoga what do you think about removing the preview without having anything to replace it for the time being?

Context: I'm still working with our vendor on trying to get retail and commercial rates in our API responses (right now we just see commercial).

DanReyLop:
> What will the test print produce when selected; A Proof of the label, or just the default computer print?

@olaolusoga It will produce something similar to the "Print test label" button in the self-help page, with a few changes:
* It will print the correct number of labels, one for each package.
* When we support more carriers, each label will be of the corresponding carrier. Apart from that, they will be completely fake: fake address, fake weight, fake barcode, etc.
* It will prompt the browser `Print` dialog, just like the button in the self-help page.

Does that seem reasonable?

olaolusoga:
> what do you think about removing the preview without having anything to replace it for the time being?

@jeffstieler that should be fine for the time being

Ties closely with this: https://github.com/Automattic/woocommerce-services/issues/819#issuecomment-289914161

>It will print the correct number of labels, one for each package.
For a test print can we limit it to one label per order. Since it's a "test" print, unsure if printing all labels past the first would be useful.

>Does that seem reasonable?
2-3 sound good

PR #934, opened 2017-03-17
Persisting the state and reloading the page on nonce error

### LINKED ISSUE #905: Nonce invalidation issue on user session refresh
If a WCS form is up (any form), and a user is logged out, the following request will fail with the error: `{"code":"rest_cookie_invalid_nonce","message":"Cookie nonce is invalid","data":{"status":403}}`.

**To reproduce:**

* Open any WCS form - e.g. shipping method settings
* Verify that saving works
* Open dashboard in a new window/tab
* Log out
* Return to WCS form window/tab - attempt another save

**Initial thoughts on a solution:**

We probably need to hook into the heartbeat system and update nonce values, ideally only when absolutely necessary.

marcinbot:
Fixes #905 

It was a bit tricky to come up with a decent user experience on this, so any suggestions welcome.

To test:
* open any of the plugin admin pages, make some changes, but do not save them
* open an admin page on a separate tab, logout
* back on the first tab, attempt to save
* you should see an alert: `There was a problem saving your settings. Please try again after the page is reloaded.`
* the page should automatically refresh (you might have to log in and then be redirected back to it)
* your changes should still be there and you should be able to save now

Couple of things about the persisted state:
* the persisted state expires after 24 hours, so refreshing the page after that time will show the settings read from the DB
* the state is deleted right after it is read, so refreshing the page twice is going to show the settings read from the database


jeffstieler:
In testing this, after the refresh I'm seeing a `TypeError` in a red notice, and this in the preview pane:

```javascript
{"code":"rest_cookie_invalid_nonce","message":"Cookie nonce is invalid","data":{"status":403}}
```

> the state is deleted right after it is read, so refreshing the page twice is going to show the settings read from the database

Perhaps we can persist the state in more cases than this? What if a user accidentally refreshes during a label purchase, but is still logged in? etc etc..

Pinging @olaolusoga and @kellychoffman for UX input 👍

olaolusoga:
>the state is deleted right after it is read, so refreshing the page twice is going to show the settings read from the database
Agree with @jeffstieler it'd be great to persist the state in more cases.

>The potential problem with persisting/reloading the state is that it's going to be difficult to determine whether users want to continue working on their changes of if they want to load the settings from the database and modify them when they refresh the page.
We can lean on the assumption that most users are intentionally making changes for the better, and would want their progress "saved" incase a refresh happens. So in that case, having a persistent state would be ideal. 

If they want to go back, they would have to adjust the "updates" that have been made, compared to refreshing to return to the db info. 

>the persisted state expires after 24 hours, so refreshing the page after that time will show the settings read from the DB
We should keep this, it feels like normal UI behavior.

client/account-settings/index.js
```diff
@@ -32,6 +32,15 @@ export default ( { formData, formMeta, storeOptions } ) => ( {
 		};
 	},
 
+	getStateForPersisting( state ) {
+		delete state.notices;
```
DanReyLop: Don't mutate the `state` variable, return a new state tree instead. Since this is run in the `onbeforeunload` handler, if the user chooses to "stay", the `notices` will be gone from the Redux state without Redux even noticing to trigger a re-render.

DanReyLop: My bad, I didn't completely grok the `onbeforeunload` logic, ths is fine.

olaolusoga:
Looks fine. We can take a deeper look @ it down the road.

The system notice helps users with context, and acts as a good warning.
screen shot 2017-04-06 at 5 58 33 pm

PR #952, opened 2017-03-24
Remove "usps" hardcoding to support more carriers for label purchases

### LINKED ISSUE #891: Remove hardcoded 'usps' usage
In order to set up WCS for multiple shipping carriers, we need to remove the hardcoded `'usps'` strings from the plugin.

This issue represents the following work:
* Identify cases where removal of the hardcoded string requires significant effort or a coordinated server change
* Remove all other existing hardcoded `'usps'` instances

Related: https://github.com/Automattic/woocommerce-services/issues/525

DanReyLop:
Fixes #891 

Requires the `update/support-more-carriers` branch in the server.

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -145,17 +145,19 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 				$formatted_packages[ $package_id ] = $package;
 			}
 
-			$service_id = 'usps'; //TODO: remove hardcoding
-			$predefined_packages_schema = $this->service_schemas_store->get_predefined_packages_schema_for_service( $service_id );
-			$enabled_predefined_packages = $this->settings_store->get_predefined_packages_for_service( $service_id );
+			$predefined_packages_schema = $this->service_schemas_store->get_predefined_packages_schema();
+			$enabled_predefined_packages = $this->settings_store->get_predefined_packages();
 
-			foreach ( $predefined_packages_schema as $group ) {
-				foreach ( $group->definitions as $package ) {
-					if ( ! $package->is_flat_rate && ! in_array( $package->id, $enabled_predefined_packages ) ) {
-						continue;
+			foreach ( $predefined_packages_schema as $service_id => $service_predefined_packages_schema ) {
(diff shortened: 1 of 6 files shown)


classes/class-wc-connect-shipping-label.php
```diff
@@ -145,30 +145,33 @@ protected function get_all_packages() {
...
+
+			foreach ( $predefined_packages_schema as $service_id => $service_predefined_packages_schema ) {
+				$service_enabled_predefined_packages = isset( $enabled_predefined_packages[ $service_id ] ) ? $enabled_predefined_packages[ $service_id ] : array();
+				foreach ( $service_predefined_packages_schema as $group ) {
+					foreach ( $group->definitions as $package ) {
+						if ( ! $package->is_flat_rate && ! in_array( $package->id, $service_enabled_predefined_packages ) ) {
+							continue;
+						}
+
+						$formatted_packages[ $package->id ] = $package;
```
DanReyLop: I don't like to flatten this data structure this way (same as in the `get_package_lookup` method). This will break in subtle ways if packages from multiple carriers have the same ID. This should be more robust.

marcinbot: We spoke about it after this comment has been added, but the package IDs are created and managed by us on the server, so this should be fine. Package names, on the other hand, can be shared.

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -67,7 +64,7 @@ public function run( $request ) {
 				'tracking' => $label_data->label->tracking_id,
 				'refundable_amount' => $label_data->label->refundable_amount,
 				'created' => $label_data->label->created,
-				'carrier_id' => $settings[ 'carrier' ],
+				'carrier_id' => $label_data->label->carrier_id,
```
jeffstieler: @DanReyLop @marcinbot I'm seeing a case difference (all caps vs. lowercase) with this change that's causing the tracking link to not work before a page reload.

DanReyLop: Good catch! Fixed in https://github.com/Automattic/woocommerce-connect-server/pull/879

PR #961, opened 2017-03-28
Version 1.3.2

jeffstieler:
Update plugin version to 1.3.2 and add change log entries

* Hide destination address normalized order meta
* Log rate retrieval failures as errors with admin notice


nabsul:
:shipit:

PR #968, opened 2017-03-29
Fixed TOS button styling issues

### LINKED ISSUE #689: Odd color changing behavior when opting-in after activating connect
> There was some weird color-changing when I clicked the “Connect” button in the very first banner (went from purple to Blue, which is what color my Primary buttons normally are).

screen shot 2017-02-24 at 2 15 09 pm

marcinbot:
Fixes #689 

This is a simple change that overwrites a bunch of button pseudoclasses with WooCommerce colors.

* To show the TOS banner without nuking everything: change the value of `wc_connect_options` in `wp_options` table - replace `"tos_accepted";b:1` with `"tos_accepted";b:0`
* You might need to do `npm run dist` if the new styles don't show.

### DIFF
--- a/assets/stylesheets/banner.scss
+++ b/assets/stylesheets/banner.scss
@@ -14,16 +14,26 @@
 		}
 	}
 
-    .button-primary {
+	.button-primary {
 		background: $purple-light;
 		border-color: $purple-dark;
 		box-shadow: 0 1px 0 $purple-dark;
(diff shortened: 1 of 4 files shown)


jeffstieler:
I still saw the blue button in my tests, even after running `npm run dist`.. it looks like the primary WP styles are more specific than ours.

PR #974, opened 2017-03-31
Fixed label repurchasing without page refresh

### LINKED ISSUE #944: Label purchase modal re-uses an already used shipment ID
If a label (or group of labels) is purchased and we immediately go back to purchase the again, the old shipment ID will be re-used, causing an error on the server.

Once a label is purchased, we should invalidate the form state so that the process is restarted again and it fetches new rates and address validation data.

marcinbot:
Fixes #944 

To test:
* purchase a label
* without refreshing the page, open the label modal again
* the rates should be re-requested
* purchase the label again
* verify there are no errors in the server log (there should be an error about repurchasing a shipment id in other branches)

This also fixes the flow in some of the browsers (such as Firefox):
* purchase a label, the button should change to "Print"
* it should now be possible to close the modal, and the button should still be "Print" on reopening (before it wouldn't be possible to close the modal)
* clicking print should close the modal and open a new tab with the label ready to print
* reopening the modal should trigger a new rates request and allow to repurchase a label, as outlined in the steps above

### DIFF
--- a/client/shipping-label/state/actions.js
+++ b/client/shipping-label/state/actions.js
@@ -197,8 +197,13 @@ export const openPrintingFlow = () => (
 	dispatch( { type: OPEN_PRINTING_FLOW } );
 };
 
-export const exitPrintingFlow = ( force ) => {
-	return { type: EXIT_PRINTING_FLOW, force };
+export const exitPrintingFlow = ( force ) => ( dispatch, getState ) => {
+	const form = getState().shippingLabel.form;
+	if ( form.needsPrintConfirmation ) {
+		dispatch( clearAvailableRates() );
+	}
+
+	dispatch( { type: EXIT_PRINTING_FLOW, force } );
 };
 
 export const updateAddressValue = ( group, name, value ) => {
--- a/client/shipping-label/state/actions.js
+++ b/client/shipping-label/state/actions.js
@@ -110,6 +110,10 @@ const convertToApiPackage = ( pckg ) => {
 	return _.pick( pckg, [ 'id', 'box_id', 'service_id', 'length', 'width', 'height', 'weight' ] );
 };
 
+export const clearAvailableRates = () => {
+	return { type: CLEAR_AVAILABLE_RATES };
+};
+
 const getLabelRates = ( dispatch, getState, handleResponse, { getRatesURL, nonce } ) => {
 	const formState = getState().shippingLabel.form;
 	const {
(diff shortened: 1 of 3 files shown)


client/shipping-label/state/actions.js
```diff
@@ -487,6 +492,9 @@ export const purchaseLabel = () => ( dispatch, getState, context ) => {
 			} else if ( error ) {
 				console.error( error );
 				dispatch( NoticeActions.errorNotice( error.toString() ) );
+				//re-request the rates on failure to avoid attempting repurchase of the same shipment id
+				dispatch( clearAvailableRates() );
+				getLabelRates( dispatch, getState, _.noop, { getRatesURL, nonce } );
```
jeffstieler: Can you elaborate on why we'd want to retrieve rates under an error condition? At a glance, this seems unnecessary.

marcinbot: Yeah, we purchase the shipments from EasyPost at the end, so it seems unnecessary in case something like BillingDaddy fails or the rates are not fetched. 

However, we send separate EasyPost purchase requests per label. If any of them fails for any reason, we need to refresh the rates, otherwise it will be impossible to retry purchasing.

I ran into that while working on another server PR, so admittedly this is en edge case. I'm happy reverting this.

jeffstieler: Is there any opportunity to selectively refresh rates depending on the type of error?

In the case where our server can't successfully reach the API for rates/labels - would a refresh be necessary? Couldn't we just try the same purchase request(s) again?

marcinbot: The server doesn't return descriptive codes/messages for these errors. It's all internal server errors.

You're right - if rates cannot be reached the refresh is unnecessary. Same for BillingDaddy. But if the actual purchase fails, we might have to refresh (depending on how many labels we're trying to purchase and which ones failed)

marcinbot: I reproduced the issue I was talking about:
* attempted to purchase 2 labels
* made the second purchase request fail (by throwing an error in the code)
* received an error in the client
* attempted to repurchase without refresh
* repurchase error occurred in the backend

jeffstieler: Ok, I think the refresh is fine for now.

> The server doesn't return descriptive codes/messages for these errors. It's all internal server errors.

We need to change this.

client/shipping-label/state/actions.js
```diff
@@ -192,8 +197,13 @@ export const openPrintingFlow = () => (
 	dispatch( { type: OPEN_PRINTING_FLOW } );
 };
 
-export const exitPrintingFlow = ( force ) => {
-	return { type: EXIT_PRINTING_FLOW, force };
+export const exitPrintingFlow = ( force ) => ( dispatch, getState ) => {
+	const form = getState().shippingLabel.form;
+	if ( form.needsPrintConfirmation ) {
+		dispatch( clearAvailableRates() );
+	}
```
jeffstieler: Let's do this after dispatching the exit flow action so that we're:
1) consistent with the other calls
2) avoid the user seeing the rates being cleared (even if only for a second)

jeffstieler: I went ahead and took care of this.

PR #978, opened 2017-04-04
Added a notice with a button that will automatically install/activate Jetpack if not present

### LINKED ISSUE #915: Add Notice That JetPack Install Is Mandatory For WooCommerce Services
The extension does not work without JetPack installed. It's not mentioned in the plugin description in the WordPress.org listing, but it's mentioned in the installation instructions. It would make sense to:
1. Add a notice after the installation of the plugin to install and activate a JetPack connection (if not already present).
2. Add this in the plugin description in WordPress.org itself.

DanReyLop:
Fixes #915 

I went a bit beyond the scope of #915, but I believe this code will live one way or another in our improved future NUX flow.

I've added a `Jetpack_Installer` helper class, to handle automatically install & activate Jetpack. Its logic is very similar (someone could even said I copied it) to what we did in Jetpack: https://github.com/Automattic/jetpack/blob/master/3rd-party/woocommerce-services.php

The notice appears in every admin page, and it's not dismissable (I don't think it would make sense, but I could change that). It appears in 3 different cases:

**Jetpack is not installed:**
The button installs and activates Jetpack automatically (it takes a good while, maybe 20 seconds, and there's no other feedback than the page being reloaded). It then redirects to the Jetpack dashboard where the user will be prompted to `Connect to WordPress.com`.
install_jetpack

**Jetpack is installed, but deactivated:**
The button activates Jetpack automatically this operation is almost instant). It then redirects to the Jetpack dashboard where the user will be prompted to `Connect to WordPress.com`.
activate_jetpack

**Jetpack is installed and activated, but it's not connected to a `WordPress.com` account:**
The button starts the connection process, it would be the same as clicking one of the Jetpack's `Conenct to WordPress.com` buttons.
connect_jetpack

@olaolusoga What do you think about this? Of course, I need to know what words to put in the different variants of the notice, I'm really bad at thinking this stuff.

**Other things to consider:**
* Installing Jetpack takes time, it's a 5MB+ plugin zipped. Some feedback during install would be nice, at least changing the button to `disabled`.
* The notice will appear in every admin page, even in the ones where Jetpack is already screaming `Connect to WordPress.com!!!`, like the `Jetpack` Dashboard or the `Plugins` page. Maybe we should hide it in those?
* Should it be dismissable? In that case, what should happen when it's dismissed? Deactivate `Services`, or just stop nagging the user about that (and the plugin won't start to work until he decides to connect Jetpack)?

I can implement those things, depending on how far do we want to take this. If it clashes with our future NUX plans, it's better to leave it as a temporary patchwork, but if it could live on in the future NUX, then we should try to polish it.


olaolusoga:
This looks great. For the amount of hoops a user is required to jump through to get WCS up and running, this will help mitigate the difficulty, till we can dive deep into the NUX and outline an ideal POA.

>The notice appears in every admin page, and it's not dismissible (I don't think it would make sense, but I could change that)

Agree, make it none dismissible.

Reason: If user intentionally installed WCS, we **NEED** them to complete these steps to fully utilize the plugin. The plugin's a dud w/o it.

>Jetpack is not installed... Some feedback during install would be nice, at least changing the button to disabled.

Agree. Can we make the button a disabled color, and change copy to "Installing..."

Original Copy: "Please install Jetpack to use WooCommerce Services."
Suggestion Copy: "To get started you'll need to install Jetpack."
Button: "Install Jetpack"

>Jetpack is installed, but deactivated:

Original Copy: "Please activate Jetpack to use WooCommerce Services."
Suggestion Copy: "To get started you'll need to activate Jetpack."
Button: "Activate Jetpack"

>Jetpack is installed and activated, but it's not connected to a WordPress.com account:

Original copy: "Please connect Jetpack to your WordPress.com account to use WooCommerce Services"
Suggestion Copy: " To get started please connect Jetpack to your Worpress.com account."
Button" "Connect to WordPress.com"

>The notice will appear in every admin page, even in the ones where Jetpack is already screaming Connect to WordPress.com!!!, like the Jetpack Dashboard or the Plugins page. Maybe we should hide it in those?

We should hide in scenarios like these, so we don't duplicative CTAs.

jeffstieler:
> > Jetpack is not installed... Some feedback during install would be nice, at least changing the button to disabled.
> 
> Agree. Can we make the button a disabled color, and change copy to "Installing..."?

I learned recently about a javascript API for installing/updating plugins in WP - checkout `/wp-admin/js/updates.js`, specifically the `wp.updates.installPlugin()` method. Hit me up in Slack for more info / example code.

DanReyLop:
> Also is it possible to alternate copy if this process is consecutive?

That is more trouble than it's worth. In our experience, users simply ignore notices, no matter how annoying they are :(

I've fixed everything else:
* Changed texts
* Install & Activate through AJAX instead of PHP
* Made some adjustments to the button style when it's disabled
* Don't show the notice in pages where there's already a "Connect to WordPress.com" CTA

@jeffstieler The solution I ended up implementing it's not the most elegant but works. Creating a new app entry point (like the `admin-pointers.js` did) and passing around data from PHP to JS was more cumbersome than what is worth. So I ended up writing a small chunk of JS inside a PHP string, tell me if you're OK with that.

woocommerce-services.php
```diff
@@ -819,6 +822,74 @@ public function deactivate_notice() {
...
+			$button_onclick = '';
+			$button_url = '#';
+
+			if ( class_exists( 'Jetpack_Data' ) ) {
+				$screen = get_current_screen();
+				if ( 'jetpack' === $screen->parent_base || 'plugins' === $screen->base ) {
+					return; // Jetpack Dashboard and the plugins list screen already have big "Connect to WordPress.com" buttons
+				}
+
+				$notice_text = __( 'To get started please connect Jetpack to your Worpress.com account.', 'woocommerce-services' );
```
jeffstieler: Typo :)

Let's go with: `'To get started, please connect Jetpack to your WordPress.com account.'`

woocommerce-services.php
```diff
@@ -819,6 +822,74 @@ public function deactivate_notice() {
...
+
+					wp_enqueue_script( 'updates' );
+					$button_progress_label = esc_js( __( 'Installing...', 'woocommerce-services' ) );
+					$error_message = esc_js( __( 'There was an error installing Jetpack. Please try installing it manually.', 'woocommerce-services' ) );
+					$button_onclick = "this.onclick = null;" .
+						"this.innerHTML = '$button_progress_label';" .
+						"this.className += ' disabled';" .
+						"wp.updates.installPlugin( { slug: 'jetpack' } )" .
+							".then( () => window.location = '$activate_url' )" .
+							".fail( () => { this.parentNode.className = 'error'; this.parentNode.innerHTML = '$error_message'; } )";
```
jeffstieler: Rather than output this as an `onclick` attribute through PHP, let's put this in it's own JS file (perhaps `woocommerce-services-banner.js`).

The initially referenced code that inspired the use of `wp.updates` turned the button from "install" to "activate" after the operation was successful (without a page refresh IIRC). Is that something we can do as well?

woocommerce-services.php
DanReyLop: `capital_p_dangit`! :)

woocommerce-services.php
DanReyLop: > Rather than output this as an onclick attribute through PHP, let's put this in it's own JS file (perhaps woocommerce-services-banner.js).

Yep, this was too hacky.

> The initially referenced code that inspired the use of wp.updates turned the button from "install" to "activate" after the operation was successful (without a page refresh IIRC). Is that something we can do as well?

After the plugin is installed via AJAX, it's automatically activated, by redirecting the page to the `plugins.php?action=activate&plugin=jetpack/jetpack.php` URL. So, the experience is that you click the `Install` button, and after a while the page refreshes and you get a big Jetpack NUX splash screen.

DanReyLop: > Rather than output this as an onclick attribute through PHP, let's put this in it's own JS file (perhaps woocommerce-services-banner.js).

Done.

woocommerce-services.php
```diff
@@ -819,6 +823,72 @@ public function deactivate_notice() {
...
+					$notice_text = __( 'To get started you need to activate Jetpack.', 'woocommerce-services' );
+					$button_label = __( 'Activate Jetpack', 'woocommerce-services' );
+					$button_url = $activate_url;
+				} else {
+					$notice_text = __( 'To get started you need to install Jetpack.', 'woocommerce-services' );
+					$button_label = __( 'Install Jetpack', 'woocommerce-services' );
+					$button_url = '#';
+					$button_class .= ' wcc-install-jetpack';
+
+					wp_enqueue_script( 'updates' );
```
jeffstieler: You can define `'updates'` as a dependency for `wc_connect_banner` in the call to `wp_register_script()`.

DanReyLop: Done

woocommerce-services.php
```diff
@@ -357,14 +357,17 @@ public function load_textdomain() {
 		 * @codeCoverageIgnore
 		 */
 		public function init() {
+			$this->load_dependencies();
 			add_action( 'admin_init', array( $this, 'admin_enqueue_scripts' ) );
 
 			if ( ! WC_Connect_Options::get_option( 'tos_accepted', false ) ) {
 				add_action( 'admin_init', array( $this, 'admin_tos_notice' ) );
 				return;
+			} else if ( ! $this->check_jetpack_install() ) {
```
jeffstieler: With this being an `else if`, does a new user have to accept our TOS before seeing the Jetpack install notice?

DanReyLop: Yes. It's either that, or the other way around (install Jetpack and then accept the ToS). I don't think we should show 2 notices at the same time, but I don't really care about the order :)

jeffstieler: The problem with this approach is that the notice is never shown. I think you might be seeing it because perhaps you still have some leftover plugin option data?

With no WCS options in the `wp_options` table, and no Jetpack installed, this notice doesn't show.

See the `can_accept_tos()` function.

DanReyLop: Ouch, I missed that logic. You are totally right, I didn't erase the `wp_options` so I already had the ToS accepted.

I've changed it so the `Jetpack` notice appears first, and then simplified the `can_accept_tos()` logic since at that point Jetpack is already guaranteed to be installed.

jeffstieler:
I'm not seeing the correct styling applied - perhaps there's something wrong with my setup? (I've tried in both Chrome and Firefox..)

screen shot 2017-04-11 at 9 24 03 am

At the risk of turning this into an epic PR - what can we do to "spruce up" this notice? Compared to Jetpack and Storefront.. it looks rather bland, and doesn't contain much info. (cc: @olaolusoga)

Storefront for example:
screen shot 2017-04-11 at 9 21 43 am

We could potentially reuse the NUX code from Storefront (breaking out into a new public repo for reuse purposes if necessary): https://github.com/woocommerce/storefront/tree/master/inc/nux

DanReyLop:
> I'm not seeing the correct styling applied - perhaps there's something wrong with my setup? (I've tried in both Chrome and Firefox..)

Huh, that's weird. I've tried both using `npm start` with `define( 'WOOCOMMERCE_CONNECT_DEV_SERVER_URL', 'http://localhost:8085/' )` and using `npm run dist`, and both work for me.

Can you please try cleaning your cache? We haven't changed the plugin version in a while, so you may have an old `woocommerce_services_banner.css` file cached.

If that doesn't work, can you see the `woocommerce_services_banner.css?ver=1.3.2` file in the Chrome Devtools `Network` tab? Is it returning something other than `200 OK`? Is it empty for some reason?

olaolusoga:
> what can we do to "spruce up" this notice? Compared to Jetpack and Storefront.. it looks rather bland, and doesn't contain much info.

@jeffstieler Agree. We need to improve the visual experience. We should push this out as a baseline to start, and circle back to improve.

PR #981, opened 2017-04-13
API requests - using 2 letter language code in Accept-Language header

### LINKED ISSUE #591: Get Canadian French working
What do we need to do to support US English and Français du Canada? Most of our stuff appears in English even when the site is set for French:

This is needed for Beta because we are offering Canada Post and merchants may have their sites set for French, especially in Quebec.

screen shot 2016-09-28 at 5 57 11 pm

marcinbot:
To test:
* Set WordPress language to French or Canadian French
* Verify that the Canada Post settings present service names in French:
screen shot 2017-04-13 at 11 41 33


jeffstieler:
Note: this doesn't completely resolve #591.

This works for the strings that come from our server, but not for the strings in the plugin itself that need translation. Can you see if we have any internal resources for French Canadian translation?

PR #983, opened 2017-04-14
Changing the purchase button label depending on browser pdf support

### LINKED ISSUE #976: Labels modal - Cancel button not working between label purchase and print on some browsers
For certain browsers that don't support PDFs natively (for example Firefox) this happens:
* purchase labels
* the "Purchase" button will now be labelled "Print" (this is by design - we need an additional click from the user to open the PDF in a new tab)
* click on the "Cancel" button next to "Print"
* nothing happens

Potential fixes:
* simply make the Cancel button work
* hide the Cancel button and force the user to click on Print button
* some other solution, for example a new dialog that prompts the user to print

@jeffstieler @olaolusoga

marcinbot:
Fixes #976 

To test:
* on firefox open the label modal
* the purchase button label should be "Buy"
* purchase a label
* cancel button should disappear and the button label should change to "Print"

The resetting of available rates after closing will be handled in #974

### DIFF
--- a/client/shipping-label/views/purchase/index.js
+++ b/client/shipping-label/views/purchase/index.js
@@ -33,10 +33,10 @@ const PrintLabelDialog = ( props ) => {
 			const ratesTotal = getRatesTotal( props.form.rates );
 
 			if ( noNativePDFSupport ) {
-				return sprintf( __( 'Buy (%1$s%2$.2f)' ), currencySymbol, ratesTotal );
+				return sprintf( __( 'Buy (%(currencySymbol)s%(ratesTotal).2f)' ), { currencySymbol, ratesTotal } );
 			}
 
-			return sprintf( __( 'Buy & Print (%1$s%2$.2f)' ), currencySymbol, ratesTotal );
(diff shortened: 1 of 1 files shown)


jeffstieler:
Requesting reviews other than me, now that I've committed code. 😄

client/shipping-label/views/purchase/index.js
```diff
@@ -43,6 +53,22 @@ const PrintLabelDialog = ( props ) => {
...
+	const buttons = [
+		{
+			isDisabled: ! props.form.needsPrintConfirmation && ( ! props.canPurchase || props.form.isSubmitting ),
+			onClick: getPurchaseButtonAction(),
+			isPrimary: true,
+			label: getPurchaseButtonLabel(),
+		},
+	];
+
+	if ( ! props.form.needsPrintConfirmation ) {
```
DanReyLop: Note that, even without the `Cancel` button, the dialog can still be closed by clicking outside it or pressing `Esc`. If we want to prevent that, you'll have to add the `needsPrintConfirmation` condition here: https://github.com/Automattic/woocommerce-services/blob/master/client/shipping-label/state/reducer.js#L64

jeffstieler: It's OK if the modal gets closed - labels can still be (re)printed using the link in the metabox.

PR #985, opened 2017-04-14
Fix quotation marks serialization

### LINKED ISSUE #984: Quotation marks anywhere in package data cause disasters
Having `"` character in the package data (such as ID or name) leads to various JSON decoding issues in PHP.

This also affects server-side defined packages, such as `Window Flat Rate Envelope (12.5" x 9.5")`

Issues I noticed
* custom package with `"` as its name
  * place an order where the package has been used
  * open the order admin page
  * notice the warning next to the labels box
* flat rate using `Window Flat Rate Envelope (12.5" x 9.5")`
  * place an order that will get packed into the package above
  * open the order admin page
  * no warnings
  * purchase labels and refresh the page
  * Unexpected server error
  * label data will be missing, the stringified JSON data in the database will be `null`

This could also be the cause behind #982 

Other JSON-unsafe characters probably cause issues as well

I'm using PHP v7.0.12

@jeffstieler @DanReyLop @nabsul @v18

marcinbot:
Fixes #984

This PR replaces usages of vanilla `json_encode` with a custom function that does the following:
* uses `wp_json_encode` to ensure that the quotation marks are escaped
* escapes them again, since unescaped backslashes are stripped before writing to the database

It also adds regexes to try and handle the faulty JSONs that were created before this fix.

To test:
* place an order using the following:
  * predefined package with quotes in its name 
  * custom package with quotes in its name
  * package without quotes in its name
  * product with the quotes in its name
  * product without the quotes in its name
* open the order or orders as a merchant
* purchase labels
* refresh the page and verify that the labels are still listed in the metabox


DanReyLop:
According to the [update_post_meta](https://codex.wordpress.org/Function_Reference/update_post_meta) documentation, we should be able to just call the function with the `array`, without encoding it to JSON first. Have you tried that?

Also, it would minimize the impact in existing code if you use [a filter](https://core.trac.wordpress.org/browser/tags/4.7/src/wp-includes/meta.php#L486) instead of changing usages of the `get_post_meta` function (I'm not sure that would be more readable though).

jeffstieler:
Now that this code is causing problems, we can revisit the JSON aspect of it.

To @DanReyLop's point, the WordPress post meta functions will handle the (de)serialization of non-scalar values automatically.

Is there a clean way to move to using the built in WordPress conventions where we don't actually need JSON?

classes/class-wc-connect-shipping-label.php
```diff
@@ -90,14 +90,38 @@ protected function get_items_as_individual_packages( $order ) {
...
+
+			//attempt to recover the boxes with unescaped quotation marks in their box_id field
+			preg_match_all( '/"box_id":"(.+?)","/', $packages_json, $box_id_matches );
+			if ( 2 === count( $box_id_matches ) ) {
+				foreach ( $box_id_matches[ 0 ] as $idx => $match ) {
+					$box_id = $box_id_matches[ 1 ][ $idx ];
+					$escaped_id = preg_replace( '/(?<!\\\)"/', '\\"', $box_id );
+					$packages_json = str_replace( $match, '"box_id":"' . $escaped_id . '","', $packages_json );
+				}
+			}
```
jeffstieler: Can this "recovery" logic be broken out into a reusable function?

classes/class-wc-connect-shipping-method.php
```diff
@@ -360,7 +360,7 @@ public function calculate_shipping( $package = array() ) {
 						'label'     => self::format_rate_title( $rate->title ),
 						'cost'      => $rate->rate,
 						'meta_data' => array(
-							'wc_connect_packages' => json_encode( $rate->packages ),
+							'wc_connect_packages' => WC_Connect_Utils::json_encode_for_db( $rate->packages ),
```
jeffstieler: Can we avoid all of the JSON headaches here by using PHP's serialization methods? It's not done automatically for rate `meta_data` like it is for post meta, but perhaps it's still better than JSON.

classes/class-wc-connect-utils.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+if ( ! class_exists( 'WC_Connect_Utils' ) ) {
+	class WC_Connect_Utils {
+		/**
+		 * Encodes an object to JSON. Uses wp_json_encode to escape quotes and then double-escapes them for
+		 * saving in the database
```
jeffstieler: If this function continues to be relevant (with the JSON vs serialization discussion above), can you elaborate on the need for it? Is there a specific function that WP calls upon retrieval of DB values that requires the double escaping?

marcinbot: It's not relevant anymore. I removed it

classes/class-wc-connect-shipping-method.php
marcinbot: I was confused by different behaviours yesterday. So WC2.6 serializes the non-scalar values when storing them, but doesn't unserialize them. Serializing them into string would make the string to be serialized in turn which would complicate manual unserializing.

WC3 on the other hand, unserializes them without problems.

I added code to handle all cases on retrieval: non-scalar object, PHP serialization and JSON serialization.

tests/php/test_class-wc-connect-service-settings-store.php
```diff
@@ -2,26 +2,42 @@
 
 class WP_Test_WC_Connect_Service_Settings_Store extends WC_Unit_Test_Case {
 
-	protected $settings_store;
```
jeffstieler: For now I think it's fine to cache a copy of the `WC_Connect_Service_Settings_Store` instance for reuse. Is there a reason a new instance is used for each test?

tests/php/test_class-wc-connect-service-settings-store.php
```diff
@@ -0,0 +1,154 @@
...
+				'label_id'          => 143,
+				'tracking'          => '9405536897846106800337',
+				'refundable_amount' => 5.95,
+				'created'           => 1492165890165,
+				'carrier_id'        => 'usps',
+				'service_name'      => 'USPS - Priority Mail',
+				'package_name'      => 'box',
+				'product_names'     => array( 'product' )
+			),
+		);
```
jeffstieler: Can this value be stored in a reusable variable?

classes/class-wc-connect-service-settings-store.php
```diff
@@ -124,15 +124,114 @@ public function set_preferred_paper_size( $size ) {
...
+		 */
+		public function get_label_order_meta_data( $order_id ) {
+			$label_data = get_post_meta( ( int ) $order_id, 'wc_connect_labels', true );
+			//return an empty array if the data doesn't exist
+			if ( ! $label_data ) {
+				return array();
+			}
+
+			//labels stored as an array, return
+			if ( ! is_string( $label_data ) ) {
```
jeffstieler: The comment claims to test for an array, but the conditional just checks "not string". Why not `is_array()` here?

tests/php/test_class-wc-connect-service-settings-store.php
marcinbot: Just good practices - if we were to add tests that change the internal state of the class, we wouldn't have to write new code. I don't think caching for performance is important here, and if there are variables to be shared between tests, they shouldn't be objects under test.

classes/class-wc-connect-shipping-method.php
```diff
@@ -363,7 +363,7 @@ public function calculate_shipping( $package = array() ) {
 						'label'     => self::format_rate_title( $rate->title ),
 						'cost'      => $rate->rate,
 						'meta_data' => array(
-							'wc_connect_packages' => json_encode( $rate->packages ),
+							'wc_connect_packages' => $rate->packages,
```
nabsul: You're changing this metadata from a string to an array / object. I'm thinking of users when they upgrade, will they end up with differently encoded `wc_connect_packages` in their meta data tables?

marcinbot: Yes, but the method that reads this data handles both arrays and strings: https://github.com/Automattic/woocommerce-services/pull/985/files#diff-cc0cefd4ac82fa521d8d0ed20fc65bbbR95

nabsul:
- Unit tests run successfully (THANK YOU for adding those!)
- I was able to print labels for old and new orders
- I was able to trigger the serialization error and then observe that this PR fixes that problem

I noticed that the logic doesn't actually correct the badly serialized data. As it stands, we'll have to forever support both formats of serialization. 

What if we approach this issue as a "migration" problem? That is, upon install / upgrade of the plugin we run through all the data once and correct the entries that are bad.

This would make our code simpler in that everywhere else, we can trust that the data is fine and we don't need to keep trying both serializations forever.

nabsul:
I synced up with @marcinbot. The migration approach needs more research and discussion and we should let that block this fix.

:shipit:

Issue #123, opened 2016-03-31
Fix method_title vs title, usps_account and origin in JSX


allendav:
I realize we are going to data drive this eventually, but we should go ahead and match our settings schema now.

First, in https://github.com/Automattic/woocommerce-connect-client/blob/master/client/components/shipping-service-setup.js#L19 `method_title` should be simply `title`

Second, please use `account_id` instead of `usps_account`

Lastly, please add a `FormTextInput` for `origin` (zip code)

![screen shot 2016-03-31 at 12 41 20 pm](https://cloud.githubusercontent.com/assets/1595739/14188688/be95cd20-f73e-11e5-9c86-374017981517.png)

PR #989, opened 2017-04-18
Don't break when printing a label for a non-existent product or variation

### LINKED ISSUE #982: Error while buying label
Today I started getting Unexpected Server Error when I hit the Buy & Print button. I tried twice, and my card still got charged both times. Is there a way to check the logs and see what the error was? I installed a couple updates last night, including going from Woocommerce 3.0.1 to 3.0.2.

DanReyLop:
Fixes #982 

To test:
* As a user, buy a product, and checkout using WCS-provided rates.
* As an admin, delete said product entirely (not just "Trash").
* Go to the `Edit Order` page, try to buy a label.

You can also try purchasing / deleting a variation instead of a whole product, it's the same case.

In `master`: The item won't have a name in the `Packages` step, and after correctly purchasing a label for it, the PHP server will crash. That means that the merchant loses money and doesn't get a label.

With this PR: The item will have a (non-clickable) name in the `Packages` step, it will be `#123 - [Deleted product]`. The label purchase can get through without any problem. After that, in the `Labels` list, hovering over the label ID you will be able to see the same "name" in the package contents: `#123 - [Deleted product]`.

I'm sure there's an esoteric way of getting the product name, such as parsing the order lines, but this is a case so rare that I'm perfectly fine just preventing a crash and providing a reasonable fallback.

### DIFF
--- a/classes/class-wc-connect-compatibility-wc26.php
+++ b/classes/class-wc-connect-compatibility-wc26.php
@@ -83,12 +83,15 @@ if ( ! class_exists( 'WC_Connect_Compatibility_WC26' ) ) {
 		 * @return string The product (or variation) name, ready to print
 		 */
 		public function get_product_name_from_order( $product_id, $order ) {
-			foreach( $order->get_items() as $line_item ) {
+			foreach ( $order->get_items() as $line_item ) {
 				if ( (int) $line_item[ 'product_id' ] === $product_id || (int) $line_item[ 'variation_id' ] === $product_id ) {
-					return sprintf( '#%s - %s', $product_id, $line_item[ 'name' ] );
+					/* translators: %1$d: Product ID, %2$s: Product Name */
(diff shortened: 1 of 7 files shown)


DanReyLop:
@jeffstieler I've updated the PR. Now it tries to get the product name from the order lines: https://github.com/Automattic/woocommerce-services/pull/989/commits/ec4e72a72fde21789a5570bd23237d1ddcf09907

I'm not sure if that's too much complexity for what it's worth, I'm leaving the decision to you :)

jeffstieler:
Does the `should_show_meta_box()` logic need to be updated? I'm not seeing the labels metabox at all.

DanReyLop:
@jeffstieler My bad, I had only tried with an order with multiple items and one of them deleted. I've updated the `should_show_meta_box` to make a "best-effort" guess: If the order has packaging meta-data (i.e. it used our checkout rates), then the meta-box should show even if there are no shippable products, because the products' dimensions won't be read.

Issue #995, opened 2017-05-01
[WCS in WPCOM] Refactor the client so the React app is entirely REST-driven


DanReyLop:
Right now, when the PHP code enqueues our JS bundle for execution, it gives also a bunch of data in a global variable (`wcConnectData`) so the React code can use that to initialise itself.

When we move that code (or parts of it) to Calypso, we won't have that luxury. The React sub-applications need to be able to bootstrap themselves without relying on any data. Most of the time, that will mean creating a REST endpoint that will be called when the bundle is initialized, and that endpoint will return all the necessary information.

jeffstieler:
Can this be closed now?

DanReyLop:
Yep.

PR #996, opened 2017-05-01
[WCS in WPCOM] Refactor the client so the React app is entirely REST-driven

DanReyLop:
Fixes partially #995

For this first round, I've decided to focus on the `Packaging Manager` and the `Payment Selector` boxes, since they are the first that are going to be incorporated to Calypso.

The objective is to have those "sub-applications" be as independent of their WP-Admin context as possible. They must be able to initialize themselves completely using just REST endpoints.

This PR is big, I recommend reviewing it by commit (they are independent and self-contained). If you don't understand the "why" of some of these decision, please ask and I'll clarify (and probably add comments).

Since this is a partial refactor, there are some things I couldn't fully remove, such as `save-form.js`.

To test:
* Go to the Packaging Manager.
* Notice that it doesn't load instantly anymore, now it has a spinner that last for a second or so.
* Go to the Shipping Labels settings page (the Payment selector thingy).
* Notice that it doesn't load instantly anymore, now it has a spinner that last for a second or so.
* Check that, apart from that, everything works as usual. That includes the other parts of WCS, such as buying a label, configuring USPS, self-help, etc.

@olaolusoga You may want to take a look at this.

This is the "Label Settings" when it's loading:
progress-fetch

And this is if it couldn't load its settings (shouldn't happen almost ever):
error-fetch

The "Packaging Manager" follows the same pattern, spinner & disabled button when it's loading, generic message & disabled button on error.


client/api/request.js
```diff
@@ -0,0 +1,43 @@
+import parseJson from 'lib/utils/parse-json';
```
DanReyLop: The idea that I have floating in my mind about code reuse is something like this: The code for Calypso and for our plugin will be exactly the same, except for some files. So, for example, you will have:
* client/api/request.calypso.js
* client/api/request.js

And `Webpack` trickery will pick `request.calypso.js` instead of `request.js` if the environment is Calypso. Probably it will be the other way around (I would prefer to mess with our Webpack config instead of Calypso's), but you get the general idea.

This file (`request.js`) has the WP-Admin specific logic to make REST requests. Calypso won't use `nonce`, will use another way to construct the request themselves, and won't have the `window.persistState` logic. But the other API-related files (`api/index.js`, `api/url.js`) will be common for both the systems.

The refactor will continue in that direction: pushing the code that will differ between Calypso and WP-Admin to small, isolated files.

client/main.js
```diff
@@ -14,10 +14,16 @@ import AccountSettings from './account-settings';
 import PrintTestLabel from './print-test-label';
 import Packages from './packages';
 import _ from 'lodash';
+import { setNonce, setBaseURL } from 'api/request';
 
-( global.wcConnectData || [] ).forEach( ( wcConnectData ) => {
-	const Route = ( ( rootView ) => {
-		switch ( rootView ) {
+if ( global.wcConnectData ) {
```
DanReyLop: Note that `wcConnectData` is still used to pass down some properties to the JS bundle, but those properties (`nonce`, `baseUrl`) are specific to WP-Admin.

nabsul:
I tried out the "shipping labels page" and I get this message:

image

It also appears in master, but it goes away and shows my credit card payment methods after less than a second.

Additionally, the hot module replacement looks like it's broken. It doesn't auto refresh when I make changes to the code.

nabsul:
Instead of replacing the entire component with a spinner, I think it would look a lot better if you rendered the things that you can render, and only put spinners in place of the data that we're waiting for.

DanReyLop:
> Instead of replacing the entire component with a spinner, I think it would look a lot better if you rendered the things that you can render, and only put spinners in place of the data that we're waiting for.

Agreed. I'll see if it doesn't add much more complexity to the code.

> It also appears in master, but it goes away and shows my credit card payment methods after less than a second.

Ups, my bad, I didn't know that was the intended behaviour. Fixed.

> Additionally, the hot module replacement looks like it's broken. It doesn't auto refresh when I make changes to the code.

That has always been semi-broken in my machine. Does it happen also in `master`? I don't think I've made any changes that would affect that.

nabsul:
> That has always been semi-broken in my machine. Does it happen also in master? I don't think I've made any changes that would affect that.

@DanReyLop  It works fine in master

client/packages/views/index.js
```diff
@@ -112,40 +113,67 @@ const Packages = ( props ) => {
...
+						<div style={ { margin: 'auto' } }>
+							<Spinner size={ 24 } />
+						</div>
+					</CompactCard>
+				);
+			}
+			return (
+				<CompactCard className="settings-group-card">
+					<p className="error-message">
+						{ __( 'Unable to get your settings. Please try again.' ) }
```
marcinbot: Could this be clearer about what "trying again" means? Something like: 'Unable to get your settings. Please refresh the page to try again.'

DanReyLop:
> I don't mind the whole components being replaced by a spinner, at least in case of account settings and packages. Both are responsible for one functionality each and splitting the form data fetch into multiple requests feels forced.

I get @nabsul's point. We shouldn't make multiple requests when 1 is enough, but we could still show some parts of the UI (title, title of the sections...) instead of replacing the whole thing with an spinner.

DanReyLop:
@nabsul I've implemented what you suggested. This is what it looks now:

loading-payments
loading-pckages

Can I get a look at those changes (in https://github.com/Automattic/woocommerce-services/pull/996/commits/51f1333ebe810be33cbbdcc36e8cf648ea20af51) before 🚢 it?

DanReyLop:
We don't know beforehand the number of elements the UI will have, but at least we could replace the spinner with a generic pulsatinc placeholder, to be consistent with the Calypso guidelines.

@olaolusoga What do you think about that?

DanReyLop:
> Why make route changes? Anything other than aesthetics?

Mainly for aesthetics/consistency. I figured it was better to change them now, while we don't have any other codebase making REST queries.

> The biggest missing piece for Calypso was access to the Shipping Labels, Account, and Package settings through the API - right? What's the harm in using wp_localize_script() when we have the luxury of doing so?

I suppose we could keep passing all the data in `wcConnectData`, and have the React app fetch it from the REST API *only* if it's not present in the Redux tree (i.e. in Calypso). Give me a few minutes :)

> If the above is true, should we remove the "Unable to get your settings" cases from the components? (Assuming errors will be handled differently on the Calypso side of the house)

I think the errors in Calypso should be handled the same way. The point of this refactor is that we will probably be able to reuse all the code, so our reducer & actions & components will be ours, independently on the conventions WooCommerce will adopt in Calypso.

DanReyLop:
> I suppose we could keep passing all the data in wcConnectData, and have the React app fetch it from the REST API only if it's not present in the Redux tree (i.e. in Calypso). Give me a few minutes :)

Done. Now it's working like in `master` (no Spinner, no delay, everything loads instantly), but if the React app is initialized without `wcConnectData`, the REST call is made with the spinner and all the new behaviour. Now it's the best of both worlds :)

jeffstieler:
> I think the errors in Calypso should be handled the same way. The point of this refactor is that we will probably be able to reuse all the code, so our reducer & actions & components will be ours, independently on the conventions WooCommerce will adopt in Calypso.

It shouldn't hold up this PR, and I know it's the result of discussions we've had, but let's run all of that by both the other devs and designers on the store project. Our opportunity for reuse hinges on the acceptability of a potentially different experience in the components we own.

nabsul:
Code tests fine, everything seems to be working.

PR #997, opened 2017-05-04
Rename PurchaseLabelDialog to PrintLabelDialog

v18:
A small naming change to be consistent with the name we use to export and import the component.


nabsul:
:shipit:

PR #998, opened 2017-05-04
Update <ShippingRates />

v18:
This PR:
- removed unnecessary props from `<ShippingRates />`
- removed boolean parameter from `renderRateNotice()`
- change `renderTitle` to `getTitle`

### DIFF
--- a/client/shipping-label/views/purchase/steps/rates/list.js
+++ b/client/shipping-label/views/purchase/steps/rates/list.js
@@ -30,7 +30,7 @@ const ShippingRates = ( {
 	} ) => {
 	const packageNames = getPackageDescriptions( selectedPackages, allPackages, true );
 
-	const renderTitle = ( pckg, pckgId ) => {
+	const getTitle = ( pckg, pckgId ) => {
 		if ( 1 === Object.keys( selectedPackages ).length ) {
 			return __( 'Choose rate' );
 		}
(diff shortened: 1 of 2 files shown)


client/shipping-label/views/purchase/steps/rates/list.js
```diff
@@ -63,7 +61,7 @@ const ShippingRates = ( {
 
 	return (
 		<div>
-			{ renderRateNotice( showRateNotice ) }
+			{ shouldShowRateNotice && renderRateNotice() }
```
nabsul: Are you sure this is equivalent to the old code? How does the value `false` get rendered in this div?

nabsul: Just to make the code clearer, you could go with something like:
`shouldShowRateNotice ? renderRateNotice() : ''`

PS: the previous implementation of `renderRateNotice()` returning a null wasn't much better. It should have also returned an empty string if the boolean was false.

v18: React will ignore non-0 falsey values, so if `shouldShowRateNotice` evaluates to `false` or `null`, it won't get rendered by React ([ref](https://facebook.github.io/react/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored)). There isn't a need to return an empty string in the false case, then.

The `&&` is a common pattern for condition rendering ([ref](https://facebook.github.io/react/docs/conditional-rendering.html#inline-if-with-logical--operator)), and we also use it frequently in Calypso, so would prefer to keep that.

nabsul:
Looks good. :shipit:

PR #999, opened 2017-05-04
Add custom error type to FieldError

v18:
This change allows a user to specify an error type as a prop passed to `< FieldError />`.  The default error type is left as `input-validation`. The type is used to add the right CSS class.

### DIFF
--- a/client/components/field-error/index.js
+++ b/client/components/field-error/index.js
@@ -1,10 +1,14 @@
 import React, { PropTypes } from 'react';
 import sanitizeHTML from 'lib/utils/sanitize-html';
 import Gridicon from 'gridicons';
+import classNames from 'classnames';
 
-const FieldError = ( { text } ) => {
+const FieldError = ( { text, type = 'input-validation' } ) => {
 	return (
-		<div className="form-input-validation is-error">
+		<div className={ classNames(
+			'is-error',
+			`form-${ type }`
+		) }>
 			<Gridicon size={ 24 } icon="notice-outline" /> <span dangerouslySetInnerHTML={ sanitizeHTML( text ) } />
 		</div>
 	);
@@ -12,6 +16,7 @@ const FieldError = ( { text } ) => {
 
 FieldError.propTypes = {
 	text: PropTypes.string,
+	type: PropTypes.string,
 };
 
 export default FieldError;


nabsul:
I see that several of these PRs are being merged into branches other than master. To avoid reviewing the same changes twice, how about:

- merge this into master (and rebase other branches that need this change)
- merge this into the branch that needs it without a review (and it'll get reviewed together with the other changes)

v18:
> To avoid reviewing the same changes twice

Each PR has totally separate sets of changes, so each change is reviewed only once. The changes in an earlier branch don't show up in the PR for the next branch since the branches are stacked like this:

``` text
           * PR3
           /
          * PR2
         /
        * PR1
       /
------* master
```

> merge this into master (and rebase other branches that need this change)

Sure. My plan was to merge all the changes into master in one go once they're all reviewed, but am also open to merging them in one-by-one as they're reviewed.

nabsul:
Really? I thought that if you merge PR3 into PR2, and then PR2 into PR1, when you merge PR1 into master, the changelist is going to include everything that was in PR2 and PR3.

I could be be wrong about that. Nevermind :-)

client/components/field-error/index.js
```diff
@@ -1,17 +1,22 @@
...
 import Gridicon from 'gridicons';
+import classNames from 'classnames';
 
-const FieldError = ( { text } ) => {
+const FieldError = ( { text, type = 'input-validation' } ) => {
 	return (
-		<div className="form-input-validation is-error">
+		<div className={ classNames(
+			'is-error',
+			[ `form-${ type }` ]
```
nabsul: Why are you providing the second parameter to `classNames` as an array?

nabsul: Should we be hard-coding the `form` prefix to the CSS class? What if I want to use some other CSS class for the div?

jeffstieler: > Should we be hard-coding the form prefix to the CSS class? What if I want to use some other CSS class for the div?

What's your imagined use case?

The `Field` in the `FieldError` component references a `<form>` field.

v18: That's a mistake on my part, fixed in 2afcbfad39196d68b0d1c99d41a85d6fd577c0f7

v18: 👍 I'm using `form-` as the prefix since this is part of a form. Would prefer to leave a more generic-use-case refactoring for when it's needed.

nabsul:
:shipit:

v18:
Squashed the commits.

PR #1001, opened 2017-05-04
Show errors in rate step

v18:
# Changes in this PR
- check rates prop exists before using it in the shipping-label reducer
- add server errors to `getRatesErrors` selector
- if there's an error sent from the server for a package's rates, show a `FieldError`
- add Enzyme tests for `<ShippingRates />`

# Testing
Check out the [counterpart PR on the server](https://github.com/Automattic/woocommerce-connect-server/pull/802). Use the testing instructions there to simulate different responses to see the errors show up on the rates step. You can trigger a re-fresh of the rates by re-saving your packages on the modal :)

# What does it look like?

When there are no rates, only errors:

Where there is both a rate and an error (we don't currently support sending down both. if there are errors and rates we'll send only the rates - but there is good stylin' in just in case that situation ever happens):

# Stacked PRs!
The following PRs are related and stacked:
https://github.com/Automattic/woocommerce-services/pull/997
https://github.com/Automattic/woocommerce-services/pull/998
https://github.com/Automattic/woocommerce-services/pull/999
https://github.com/Automattic/woocommerce-services/pull/1000

### DIFF
--- a/client/shipping-label/state/selectors/errors.js
+++ b/client/shipping-label/state/selectors/errors.js
@@ -58,10 +58,10 @@ const getPackagesErrors = ( values ) => _.mapValues( values, ( pckg ) => {
 	return errors;
 } );
 
-const getRatesErrors = ( { values: selectedRates, available: allRates } ) => {
+export const getRatesErrors = ( { values: selectedRates, available: allRates } ) => {
 	return {
 		server: _.mapValues( allRates, ( rate ) => {
-			if( ! rate.errors ) {
+			if ( ! rate.errors ) {
 				return;
 			}
 
--- a/client/shipping-label/state/selectors/errors.js
+++ b/client/shipping-label/state/selectors/errors.js
@@ -58,7 +58,22 @@ const getPackagesErrors = ( values ) => _.mapValues( values, ( pckg ) => {
 	return errors;
 } );
 
-const getRatesErrors = ( values ) => _.mapValues( values, ( ( rate ) => rate ? null : __( 'Please choose a rate' ) ) );
+const getRatesErrors = ( { values: selectedRates, available: allRates } ) => {
+	return {
+		server: _.mapValues( allRates, ( rate ) => {
+			if( ! rate.errors ) {
+				return;
+			}
+
+			return rate.errors.map( ( error ) =>
+				error.userMessage ||
+				error.message ||
+				__( "We couldn't get a rate for this package, please try again." )
+			);
+		} ),
+		form: _.mapValues( selectedRates, ( ( rate ) => rate ? null : __( 'Please choose a rate' ) ) ),
+	};
+};
 
 const getSidebarErrors = ( paperSize ) => {
 	const errors = {};
@@ -80,7 +95,7 @@ export default createSelector(
 			origin: getAddressErrors( form.origin, countriesData ),
 			destination: getAddressErrors( form.destination, countriesData ),
 			packages: getPackagesErrors( form.packages.selected ),
-			rates: getRatesErrors( form.rates.values ),
+			rates: getRatesErrors( form.rates ),
 			sidebar: getSidebarErrors( paperSize ),
 		};
 	}
--- a/client/shipping-label/views/purchase/steps/rates/test/list.js
+++ b/client/shipping-label/views/purchase/steps/rates/test/list.js
@@ -1,0 +1,369 @@
+/*
+ * External dependencies
+ */
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+import { noop } from 'lodash';
+
+/*
+ * Internal Dependencies
+ */
+import ShippingRates from '../list';
+import Dropdown from 'components/dropdown';
+import FieldError from 'components/field-error';
+
+/*
+ * Useful data for testing
+ */
+const allPackages = {
+	small_flat_box: {
+		inner_dimensions: '8.63 x 5.38 x 1.63',
+		outer_dimensions: '8.63 x 5.38 x 1.63',
+		id: 'small_flat_box',
+		name: 'Small Flat Rate Box',
+		dimensions: [ {
+			inner: '8.63 x 5.38: x 1.63',
+			outer: '8.63 x 5.38 x 1.63',
+		} ],
+		box_weight: 0,
+		max_weight: 70,
+		is_letter: false,
+		is_flat_rate: true,
+		group_id: 'pri_flat_boxes',
+	},
+	medium_flat_box_top: {
+		inner_dimensions: '11 x 8.5 x 5.5',
+		outer_dimensions: '11.25 x 8.75 x 6',
+		id: 'medium_flat_box_top',
+		name: 'Medium Flat Rate Box 1, Top Loading',
+		dimensions: [ {
+			inner: '11 x 8.5 x 5.5',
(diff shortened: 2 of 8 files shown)


client/shipping-label/views/purchase/steps/rates/test/list.js
```diff
@@ -0,0 +1,384 @@
...
+						},
+					} }
+				/>
+			);
+			it( 'should not display rates dropdown', () => {
+				expect( wrapper.find( Dropdown ) ).to.have.length( 0 );
+			} );
+			it( 'should display error', () => {
+				const errorWrapper = wrapper.find( FieldError );
+				// const errorText = errorWrapper.text();
```
jeffstieler: Please remove this line.

client/shipping-label/views/purchase/steps/rates/test/list.js
```diff
@@ -0,0 +1,384 @@
...
+			);
+			it( 'should not display any dropdowns', () => {
+				expect( wrapper.find( Dropdown ) ).to.have.length( 0 );
+			} );
+			it( 'should display error for each package', () => {
+				expect( wrapper.find( FieldError ) ).to.have.length( 2 );
+			} );
+			it( 'should display extra heading for package title', () => {
+				const packageNames = wrapper.find( 'p' ).map( node => node.text() );
+				expect( wrapper.find( 'p' ) ).to.have.length( 2 );
```
jeffstieler: Can another `.find()` call be avoided here by reusing the `packageNames` variable? Or perhaps can this line be omitted in favor of the line below?

client/shipping-label/views/purchase/steps/rates/test/list.js
```diff
@@ -0,0 +1,384 @@
...
+				expect( packageNames ).to.eql( [ '#1 - Package 1 Name', '#2 - Package 2 Name' ] );
+			} );
+		} );
+
+		describe( 'Both rates and errors for all packages', () => {
```
jeffstieler: Can we add a test case for only one package (of multiple) having errors?

client/shipping-label/state/selectors/errors.js
```diff
@@ -58,7 +58,22 @@ const getPackagesErrors = ( values ) => _.mapValues( values, ( pckg ) => {
...
-const getRatesErrors = ( values ) => _.mapValues( values, ( ( rate ) => rate ? null : __( 'Please choose a rate' ) ) );
+const getRatesErrors = ( { values: selectedRates, available: allRates } ) => {
+	return {
+		server: _.mapValues( allRates, ( rate ) => {
+			if ( rate.errors ) {
+				return rate.errors.map( ( error ) => {
+					return {
+						message: ( error.userMessage ||
+							error.message ||
+							"We couldn't get a rate for this package, please try again." ),
```
jeffstieler: Missing translation.

client/shipping-label/state/selectors/errors.js
```diff
@@ -58,7 +58,22 @@ const getPackagesErrors = ( values ) => _.mapValues( values, ( pckg ) => {
 	return errors;
 } );
 
-const getRatesErrors = ( values ) => _.mapValues( values, ( ( rate ) => rate ? null : __( 'Please choose a rate' ) ) );
+const getRatesErrors = ( { values: selectedRates, available: allRates } ) => {
+	return {
+		server: _.mapValues( allRates, ( rate ) => {
+			if ( rate.errors ) {
```
v18: Yes, it's intended - the form errors return value will be the same as before, and the server return value will be either an array containing all the errors, or an empty array.

I added tests so that the behaviour is clarified in 3e0276feab8f6625fbf33e42a0dc8d7b30259ca8. Also tests 🎉

Note that I also changed the errors so that we're only getting the error message, and not the entire object (related to your next comment).

client/shipping-label/state/selectors/errors.js
v18: Fixed in 15a2063739c7f756a31f77aa3d6a13aefd3fa8d0

client/shipping-label/state/selectors/errors.js
```diff
@@ -58,7 +58,22 @@ const getPackagesErrors = ( values ) => _.mapValues( values, ( pckg ) => {
...
+				return rate.errors.map( ( error ) => {
+					return {
+						message: ( error.userMessage ||
+							error.message ||
+							"We couldn't get a rate for this package, please try again." ),
+					};
```
v18: You're right, should just be a plain error string. Fixed in 189f98ef79af32d59791a90ae8ae9b0afa6447da

client/shipping-label/state/selectors/errors.js
```diff
@@ -58,7 +58,22 @@ const getPackagesErrors = ( values ) => _.mapValues( values, ( pckg ) => {
...
+				} );
+			}
+		} ),
```
v18: Challenge accepted 😉 One indentation level was removed when using a string instead of an error object, and I removed two more in 15a2063739c7f756a31f77aa3d6a13aefd3fa8d0

client/shipping-label/views/purchase/steps/rates/test/list.js
```diff
@@ -0,0 +1,384 @@
...
+			);
+			it( 'should display rates dropdown', () => {
+				expect( wrapper.find( Dropdown ) ).to.have.length( 1 );
+			} );
+			it( 'should not display any errors', () => {
+				expect( wrapper.find( FieldError ) ).to.have.length( 0 );
+			} );
+			it( 'should not display extra heading for package title', () => {
+				// the package title is displayed using the dropdown legend
+				expect( wrapper.find( 'p' ) ).to.have.length( 0 );
```
v18: Sure - fixed in e35b2109ab925a4de55c2ae7fdd94acec3d7efac

client/shipping-label/views/purchase/steps/rates/test/list.js
v18: 👍 Removed in 08a888b935617c03f6a7f0c21e826b36921a84c3

client/shipping-label/views/purchase/steps/rates/test/list.js
v18: Yes, fixed in e35b2109ab925a4de55c2ae7fdd94acec3d7efac

client/shipping-label/views/purchase/steps/rates/test/list.js
v18: Sure, I can add it, but it seems a bit superfluous to me. We're already testing the display of these cases:
- multiple packages
- errors only for each package
- rates only for each package
- both rates and errors for a package

Since we are capable of displaying multiple packages, and for each package we are capable of displaying each type: both errors and rates, rates only, and errors only, then in terms of display, we're all set.

What do you think?

In any case, I did some of the groundwork and it's available in 8c0b08ef000df3dd28e3d97842402e87fd04307c

v18:
@jeffstieler thanks for the review. I fixed all the issues you mentioned, just had a comment on the extra test comment: https://github.com/Automattic/woocommerce-services/pull/1001#discussion_r117020597

v18:
Removed the unneeded test code, and squashed. Retested on the old and new server versions (i.e. with and without sending errors in the response). Tests pass. Merging in now.

PR #1002, opened 2017-05-04
Update NumberInput to use a JS class

v18:
# The Changes
React.createClass will be deprecated in React 16. This change updated `<NumberInput />` to use a js class instead.

# How to test

Check that `<NumberField />` and `<NumberInput /`> are working as intended in these cases:

### `<ShippingServiceEntry />` and `<SettingsItem />`
*in client/settings/views/services/entry.js* 
and *client/settings/views/wcc-settings-form/settings-item.js*
1. Go to WooCommerce (sidebar) > Settings (sidebar) > Shipping (top tab)
2. Click on a  WCS Shipping Method (link)
3. In the Rates section, open the Services to see the price adjustments
4. See the fallback rate

### `<PackageInfo />`
In *client/shipping-label/views/purchase/steps/packages/package-info.js*
1. Go to an order package and open the printing modal
2. Add a package - see the total weight input

### DIFF
--- a/client/components/number-field/number-input.js
+++ b/client/components/number-field/number-input.js
@@ -1,44 +1,38 @@
-import React, { PropTypes } from 'react';
+import React, { Component, PropTypes } from 'react';
 import FormTextInput from 'components/forms/form-text-input';
 
-export default React.createClass( {
-	displayName: 'NumberInput',
-
-	propTypes: {
+export default class NumberInput extends Component {
+	static propTypes = {
 		onChange: PropTypes.func,
-	},
+	};
 
-	getDefaultProps() {
-		return {
-			onChange: () => {},
-		};
-	},
+	static defaultProps = {
+		onChange: () => {},
+	};
 
-	getInitialState() {
-		return {
-			focused: false,
-			text: this.props.value,
-		};
-	},
+	state = {
+		focused: false,
+		text: this.props.value,
+	};
 
 	componentWillReceiveProps( nextProps ) {
 		if ( ! this.state.focused && nextProps.value !== this.props.value ) {
 			this.setState( { text: nextProps.value } );
 		}
-	},
+	}
 
-	handleChange( event ) {
+	handleChange = ( event ) => {
 		this.setState( { text: event.target.value } );
 		this.props.onChange( event );
-	},
+	}
 
-	handleBlur( event ) {
+	handleBlur = ( event ) => {
 		this.setState( {
 			focused: false,
 			text: this.props.value,
 		} );
 		this.props.onChange( event );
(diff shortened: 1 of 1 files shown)


client/components/number-field/number-input.js
```diff
@@ -1,44 +1,38 @@
...
-	getInitialState() {
-		return {
-			focused: false,
-			text: this.props.value,
-		};
-	},
+	state = {
+		focused: false,
+		text: this.props.value,
+	};
```
jeffstieler: I'm sure this is equivalent, but the [React docs](https://facebook.github.io/react/docs/state-and-lifecycle.html#adding-local-state-to-a-class) and components within Calypso set state in the constructor.

client/components/number-field/number-input.js
```diff
@@ -1,44 +1,38 @@
...
 
 	componentWillReceiveProps( nextProps ) {
 		if ( ! this.state.focused && nextProps.value !== this.props.value ) {
 			this.setState( { text: nextProps.value } );
 		}
-	},
+	}
 
-	handleChange( event ) {
+	handleChange = ( event ) => {
```
jeffstieler: Can you explain these changes?

From what I understand, arrow functions [don't bind `this`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions).

client/components/number-field/number-input.js
v18: Some of the newer syntax that's available didn't make it to the docs yet. This is the [reference](https://babeljs.io/blog/2015/06/07/react-on-es6-plus) I use when converting ES5 to ES6+.

In particular I'm using [property initializers](https://babeljs.io/blog/2015/06/07/react-on-es6-plus#property-initializers) here, which are available to WCS because we're using the [stage 1](https://github.com/Automattic/woocommerce-services/blob/master/webpack.config.js#L24) preset, which [includes](https://babeljs.io/docs/plugins/preset-stage-1/) stage 2, which currently includes [`transform-class-properties`](https://babeljs.io/docs/plugins/transform-class-properties/).

client/components/number-field/number-input.js
v18: Definitely! Arrow functions' `this` value is always set to the value of `this` of their enclosing scope. In this case, it's referring to an instance of `<NumberField />`. For this reason, we don't need to also bind handleChange to `this` in the constructor.

This syntax is widely used in Calypso, and also listed/explained in the [above reference](https://babeljs.io/blog/2015/06/07/react-on-es6-plus#arrow-functions) to React on ES6+.

PR #1012, opened 2017-05-17
Check for WooCommerce installation and activation

### LINKED ISSUE #1011: [NUX Flow] Add logic to auto-install WooCommerce and Jetpack
On startup, the plugin should check if WooCommerce and Jetpack are installed. If they're not, the admin should be able to install them with one click.

Note: I'll implement these as very simple admin notices for now, pending design by @olaolusoga 

@v18

nabsul:
Closes #1011


nabsul:
This is based on old code that no longer exists.

Issue #1014, opened 2017-05-19
Payments and Packages settings - migrate the changes from Calypso to plugin


marcinbot:
Make the UI look similar to this:
<img width="691" alt="screen shot 2017-05-19 at 16 42 37" src="https://cloud.githubusercontent.com/assets/800604/26253064/55652ea8-3cb2-11e7-8d20-7b9c156a7f4c.png">

The main change is just the paper size dropdown, but there are also some UI changes.

The idea is to try and keep those in sync between WCS and Calypso UI or ideally to reuse the components from one library in another.

CC @DanReyLop @jeffstieler @olaolusoga 

Let me know if this conflicts with the NUX work

DanReyLop:
I have some concrete details to point out (differences between the Calypso designs and our current plugin):

In the `Packages` section, what happened with the USPS predefined packages? I guess it was just an oversight when doing the Calypso design.

The `ADD ANOTHER CREDIT CARD` button, what should it do (in both cases)? Should it just open `wordpress.com/me/billing` in a new tab? In the same tab? Should it open a form in a modal to add a credit card right there?

marcinbot:
Thanks, that's some good questions that I just assumed the simplest answers to, but it's still worth asking them.

> I have some concrete details to point out (differences between the Calypso designs and our current plugin):

> In the Packages section, what happened with the USPS predefined packages? I guess it was just an oversight when doing the Calypso design.

It shouldn't be a big UI change to just add another table below the custom boxes. But we might need some designer input.

> The ADD ANOTHER CREDIT CARD button, what should it do (in both cases)? Should it just open wordpress.com/me/billing in a new tab? In the same tab? Should it open a form in a modal to add a credit card right there?

I assumed it would open the .com billing page in a new tab.

☝️ CC @kellychoffman @jameskoster

jeffstieler:
> > The ADD ANOTHER CREDIT CARD button, what should it do (in both cases)? Should it just open wordpress.com/me/billing in a new tab? In the same tab? Should it open a form in a modal to add a credit card right there?
> 
> I assumed it would open the .com billing page in a new tab.

Given that this was designed inside of Calypso, my assumption is that it's meant to open a modal (or similar) with a credit card form.

That might be a point where we have to deviate inside of our plugin, but let's let the designers weigh in first.

kellychoffman:
> Given that this was designed inside of Calypso, my assumption is that it's meant to open a modal (or similar) with a credit card form.

[...]

PR #1015, opened 2017-05-23
Add label settings UI from Calypso

marcinbot:
Adds the labels part of #1014

Makes the labels settings page look like this:
screen shot 2017-05-23 at 13 59 46

* added the paper size dropdown
* I changed the webpack config a bit, so we can use more CSS straight from Calypso and we don't have to copy the card images (or other images) to our repository


classes/class-wc-connect-service-settings-store.php
```diff
@@ -70,6 +74,12 @@ public function update_account_settings( $settings ) {
 				return false;
 			}
 
+			if ( isset( $settings[ 'paper_size' ] ) ) {
```
DanReyLop: Why wouldn't `paper_size` be set?

marcinbot: Just being safe, but I removed it after inspecting the code further

client/account-settings/index.js
```diff
@@ -44,6 +44,6 @@ export default ( { formData, formMeta } ) => ( {
 	},
 
 	View: () => (
-		<AccountSettingsRootView />
+		<AccountSettingsRootView storeOptions={ storeOptions } />
```
DanReyLop: Don't pass any props to the component, they won't be there in Calypso. Instead, store the `storeOptions` in the Redux state.

Also, this "app" must work even if it doesn't receive any arguments (`{ formData, formMeta, storeOptions }` can be `undefined`), so you'll need to:
* Add `storeOptions` property to the output of the `/wc/v1/connect/account/settings` endpoint. See the `fetchSettings` function on `client/account-settings/state/actions.js`, and the `classes\class-wc-rest-connect-account-settings-controller.php` file. You'll also need to add the `paperSize` there. Also, on the `output_account_screen` function of `classes/class-wc-connect-settings-pages.php`.
* Make the whole UI work when there's no data. See the "isLoading / renderSpinner" logic in the (now deleted) `PaymentMethodSelector` component. A good way to "simulate" it is to remove the 3rd param of the `do_action( 'enqueue_wc_connect_script' ...)` in the `output_account_screen` function of `classes/class-wc-connect-settings-pages.php`.

marcinbot: Thanks for catching that. It should be now fixed.

> You'll also need to add the paperSize there.

Paper size is fetched in `$this->service_settings_store->get_account_settings()`

DanReyLop:
Nitpick: Shouldn't the Credit Card `checkboxes` be `<input type="radio">` instead? Since only 1 option can be selected, the checkbox seems an odd element.

DanReyLop:
> That's how it was designed for Calypso. Personally I don't mind they are checkboxes, a lot of places use them as radio buttons nowadays.

I'm OK merging this as long as we have Kelly's blessing :)

PR #1017, opened 2017-05-26
Switch to using i18n-calypso

marcinbot:
Switches from our own `i18n` mixin and `i18n-calypso-stub` to `i18n-calypso`

Everything should work as before, i.e. `npm run i18n` should generate a list of strings to translate in PHP, and the pages should display correctly.

The bundle size increased from ~780KB to 1.3MB, so this still might need to be addressed (exclude moment etc?)


webpack.config.js
```diff
@@ -171,6 +168,7 @@ if ( isProd ) {
 		},
 		mangle: {
 			screw_ie8: true,
+			except: [ 'translate' ],
```
DanReyLop: It would be better to keep all the "mangling" for `production` builds, but disable mangling `translate` it for the `i18n` task. Would that be too messy?

DanReyLop:
Keep in mind that this still doesn't address the "plurals" issue. Since the strings are translated in PHP, the JS app only receives a bunch of `'string': 'translatedString` entries. We need to somehow pass a JSON file to the JS app.

> The bundle size increased from ~780KB to 1.3MB, so this still might need to be addressed (exclude moment etc?)

It will need to be addressed, but I'm ok deferring it to a different issue.

DanReyLop:
> generate .pot and .po files (they could be downloaded from .org repo by a script)

`i18n-calypso` can generate the `.po` file directly. See: https://github.com/Automattic/wp-calypso/blob/master/package.json#L210
Maybe that would simplify the workflow a bit?

> There's a lot of potential for further automation, but we can figure this out as we add more translations.

I agree with that. Akeda is working on stuff related to this process, and the full JS part of this repo will probably end up living in `wp-calypso`, so for now, the simplest the solution, the better.

> I tested by going through that process for Polish and translating an example string and the multiple plural forms worked!

Nice! (I'll have to take your word for it haha)

Even if we still need to work on the i18n process, having this use `i18n-calypso` will make things much easier in the near future. When I have a few hours I'll see what I can do about that bundle size though :)

DanReyLop:
🤔 
Yeah, `i18n-calypso` only extracts from JS files. I was picturing generating 1 POT file for PHP and another for JS. I think plugins are allowed to have multiple POT files? I honestly have no idea. But anyway, the way you've done it is fine for now, at least until we decide the future of the JS code. I'll take another quick look.

PR #1019, opened 2017-05-31
Calyspofied translation methods

marcinbot:
TBD once #1018 is merged and rebased

This adds the following:
* replaces `dangerouslySetInnerHtml` React attribute with [mixed strings and components](https://github.com/Automattic/i18n-calypso#mixing-strings-and-markup)
* replaces `stringf` calls with [i18n-calypso's string substitution](https://github.com/Automattic/i18n-calypso#string-substitution)
* uses [proper plural translate calls](https://github.com/Automattic/i18n-calypso#pluralization)
* fixes a crashing labels modal (it's not possible to pass an empty string or any falsey values to `__`)


client/shipping-label/views/purchase/steps/rates/index.js
```diff
@@ -22,20 +21,24 @@ const ratesSummary = ( selectedRates, availableRates, total, currencySymbol, pac
...
+				args: {
+					serviceName: rateInfo.title,
+					rate: rateInfo.rate,
+					currencySymbol,
+				},
 			} );
 		}
 
-		return __( '' );
+		return __( ' ' );
```
DanReyLop: Just `return '';`, an empty string works in every language :)

client/shipping-label/views/purchase/index.js
```diff
@@ -33,10 +32,10 @@ const PrintLabelDialog = ( props ) => {
 			const ratesTotal = getRatesTotal( props.form.rates );
 
 			if ( noNativePDFSupport ) {
-				return sprintf( __( 'Buy (%(currencySymbol)s%(ratesTotal).2f)' ), { currencySymbol, ratesTotal } );
+				return __( 'Buy (%(currencySymbol)s%(ratesTotal)s)', { args: { currencySymbol, ratesTotal } } );
 			}
 
-			return sprintf( __( 'Buy & Print (%(currencySymbol)s%(ratesTotal).2f)' ), { currencySymbol, ratesTotal } );
+			return __( 'Buy & Print (%(currencySymbol)s%(ratesTotal)s)', { args: { currencySymbol, ratesTotal } } );
```
jeffstieler: Why the switch here from a 2 digit precise float to a string? [Automattic/i18n-calypso](https://github.com/Automattic/i18n-calypso) still uses `sprintf` under the hood.

client/shipping-label/views/purchase/steps/rates/index.js
```diff
@@ -22,20 +21,24 @@ const ratesSummary = ( selectedRates, availableRates, total, currencySymbol, pac
...
 	}
 
 	// Otherwise, just show the total
-	return sprintf( __( 'Total rate: %(currencySymbol)s%(total).2f' ), {
-		total,
-		currencySymbol,
+	return __( 'Total rate: %(currencySymbol)s%(total)s', {
```
jeffstieler: Same comment about loss of precision.

package.json
```diff
@@ -9,6 +9,7 @@
     "test-watch": "cross-env NODE_ENV=test karma start --auto-watch --no-single-run",
     "release": "node tasks/release.js",
     "i18n": "cross-env NODE_ENV=i18n webpack && node tasks/i18n",
+    "i18njson": "node tasks/i18njson",
```
jeffstieler: Where does this fit in with the existing `i18n` task? Should there be two separate tasks?

client/shipping-label/views/purchase/index.js
marcinbot: The number gets formatted as a string earlier (in `getRatesTotal` I think). Calypso's `sprintf` errors out if string is passed as an argument to `%.2f`

client/shipping-label/views/purchase/steps/rates/index.js
marcinbot: Same answer as above :)

package.json
marcinbot: It's a completely different step of a process that I anticipate will require more work in the future (this line was actually supposed to be a part of #1017 but got lost when I was fixing automatic builds).

The process would be:
1. Extract strings to PHP using `i18n`
2. Generate .pot and .po files (this could be automated - for example Jetpack downloads them from the .org repository)
3. Run `i18njson` that transforms .pot files into JSON files that then get loaded into our app

client/shipping-label/views/purchase/index.js
jeffstieler: Ah, yep. https://github.com/Automattic/woocommerce-services/blob/d54cf492db5c928ee90ff988d00d1d59946434b7/client/shipping-label/state/selectors/rates.js#L19

PR #1023, opened 2017-06-01
Fix issues in `can_accept_tos()`

### LINKED ISSUE #1020: Development issue: Site crashes if no Jetpack but JETPACK_DEV_DEBUG is set
If the JETPACK_DEV_DEBUG PHP constant exists, the plugin will try to access the `Jetpack_Data` class before checking if it exists. So if the jetpack plugin is not installed, the result is the site fatally crashes.

### LINKED ISSUE #1021: PHP Error When no Token Created
The `can_accept_tos()` function doesn't first check that `get_access_token()` an object.

> Notice: Trying to get property of non-object in wp-content\plugins\woocommerce-services\woocommerce-services.php on line 792

This error happens when WooCommerce services is newly installed AND Jetpack is newly installed / not configured.

nabsul:
Fixes #1020 
Fixes #1021 

Adds more checks to `can_accept_tos()` that resolve several issues:

- Check that `Jetpack_Data` class exists before attempting to access it
- Account for the possibility of `Jetpack_Data::get_access_token()` returning a non-object
- Allow any admin to accept TOS if in Jetpack Dev Mode

To Test:
- Start out with a fresh unconfigured install of Jetpack and WooCommerce Services
- Deactivate Jetpack (in master this will cause a crash, in this PR it won't)
- Activate Jetpack and set it to dev mode with `define('JETPACK_DEV_DEBUG', true);`
- Observer that TOS now appears

@jeffstieler @DanReyLop @marcinbot @v18

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -788,7 +788,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		 * @return bool
 		 */
 		private function can_accept_tos() {
-			if ( defined( 'JETPACK_DEV_DEBUG' ) && JETPACK_DEV_DEBUG ) {
+			if ( defined( 'JETPACK_DEV_DEBUG' ) && JETPACK_DEV_DEBUG && current_user_can( 'manage_options' ) ) {
 				return true;
 			}
 
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -365,7 +365,9 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			if ( ! $this->check_jetpack_install() ) {
 				add_action( 'admin_init', array( $this, 'admin_jetpack_notice' ) );
 				return;
-			} else if ( ! WC_Connect_Options::get_option( 'tos_accepted', false ) ) {
+			}
+
+			if ( ! WC_Connect_Options::get_option( 'tos_accepted', false ) ) {
 				add_action( 'admin_init', array( $this, 'admin_tos_notice' ) );
 				return;
 			}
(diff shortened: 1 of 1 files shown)


woocommerce-services.php
```diff
@@ -788,8 +788,16 @@ private function dismiss_tos_notice() {
 		 * @return bool
 		 */
 		private function can_accept_tos() {
+			if ( defined( 'JETPACK_DEV_DEBUG' ) && JETPACK_DEV_DEBUG && current_user_can( 'manage_options' ) ) {
+				return true;
+			}
+
+			if ( ! class_exists( 'Jetpack_Data' ) ) {
```
DanReyLop: This check isn't needed, because `can_accept_tos` will only be called if `Jetpack` is installed and enabled (see https://github.com/Automattic/woocommerce-services/pull/1023/files#diff-57ae8afb227e356d71d37582f678a831R365 )

woocommerce-services.php
```diff
@@ -788,8 +788,16 @@ private function dismiss_tos_notice() {
...
+				return false;
+			}
+
 			$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
-			return get_current_user_id() === $user_token->external_user_id;
+			return is_object( $user_token ) && isset( $user_token->external_user_id) && get_current_user_id() === $user_token->external_user_id;
```
jeffstieler: You should be able to skip the `is_object()` check here and shorten this line a little. Also, missing a space in the `isset()` check.

woocommerce-services.php
jeffstieler: It might be overly defensive now, but isn't harming anything and allows for some flexibility for calling this function elsewhere.

DanReyLop: I'm ok with that, :shipit:

v18:
> Deactivate Jetpack (in master this will cause a crash, in this PR it won't)

I didn't see a crash on master, either. Let me know if there's a specific page I should be testing this on.

> Activate Jetpack and set it to dev mode with define('JETPACK_DEV_DEBUG', true);
> Observer that TOS now appears

This also doesn't work for me, but the notice does appear if JETPACK_DEV_DEBUG is not defined. I'm not sure if that's a big deal or not, can probably ship as-is since it's a dev-only feature.

woocommerce-services.php
nabsul: @DanReyLop The issue is that in `check_jetpack_install()` we return true if the php constant is defined, regardless of whether or not jetpack is installed.

Actually, maybe it's better to fix `check_jetpack_install()`... I'll take a shot at that.

v18:
Tried it again. I get a crash with Jetpack deactivated + tos not agreed to + JETPACK_DEV_DEBUG enabled. It's fixed in this branch.

PR #1024, opened 2017-06-01
Fix missing order items in purchase label dialog

jeffstieler:
Orders with no packaging metadata had package ID collisions, which caused order items to be missing from the purchase label dialog.

To test:
* Manually create an order ("Add order") with two different products that will require shipping
* Verify that only one product shows in the Packages step of the label flow
* Check out this branch, refresh order
* Verify that both products show as in "original packaging" in the Packages step


nabsul:
I went ahead and merged this because some of my work depends on it.

Issue #1022, opened 2017-06-01
Allow for packing all items together


jobthomas:
As per: https://wordpress.org/support/topic/add-all-items-to-package-at-once/#post-9187810

Feature request:

When USPS live services are **not** used, allow store owners to bulk add all items in 1 order to 1 package

nabsul:
Great timing! I just started working on this area.

jobthomas:
Great minds ... 📦

PR #1028, opened 2017-06-02
Improve label purchase for free shipping scenario

nabsul:
Fixes: #1022 

Changes made:

- In "saved for later": Added a button to quickly pack individually.
- For individually packed items: Removed controls related to adding and removing items. You shouldn't be able to do that.
- Merged list of individual and regular packages into one list.
- Moved the "add package" button to the top of the left panel to make it easier to find

@jeffstieler @DanReyLop @marcinbot @v18


client/shipping-label/views/purchase/steps/packages/add-item.js
```diff
@@ -27,34 +25,20 @@ const AddItemDialog = ( {
 	};
 
 	const renderRadioButton = ( pckgId, itemIdx, item ) => {
```
jeffstieler: Seems like we can rename this now that it doesn't render radio buttons. Perhaps `renderMovePackageControl` or similar?

client/shipping-label/views/purchase/steps/packages/index.js
```diff
@@ -40,6 +41,13 @@ const PackagesStep = ( {
 			};
 		}
 
+		if ( ! isValidPackageType ) {
+			return {
+				isError: true,
+				summary: __( 'Please Select a Box Type' ),
```
jeffstieler: Let's say "package type", since envelopes are options as well.

client/shipping-label/views/purchase/steps/packages/item-info.js
```diff
@@ -3,8 +3,47 @@ import { translate as __ } from 'i18n-calypso';
 import Gridicon from 'gridicons';
 import Button from 'components/button';
 
-const ItemInfo = ( { item, itemIndex, packageId, showRemove, openItemMove, removeItem } ) => {
+const ItemInfo = ( { item, itemIndex, packageId, showRemove, removeItem, showPackIndividually, openItemMove, moveItem, isIndividualPackage, addPackage } ) => {
+	const renderPackIndividually = () => {
+		if ( ! showPackIndividually ) {
```
jeffstieler: Can we save adding a prop and determine this using `isIndividualPackage`?

client/shipping-label/views/purchase/steps/packages/item-info.js
```diff
@@ -3,8 +3,47 @@ import { translate as __ } from 'i18n-calypso';
...
+
+	const renderAddToPackage = () => {
+		if ( ! showPackIndividually ) {
+			return null;
+		}
+
+		return (
+			<Button className="wcc-package-item__move" compact onClick={ () => {
+				addPackage( itemIndex );
+				//moveItem( '', itemIndex, 'new' );
```
jeffstieler: Dead line.

client/shipping-label/views/purchase/steps/packages/item-info.js
```diff
@@ -40,6 +81,9 @@ ItemInfo.propTypes = {
 	itemIndex: PropTypes.number.isRequired,
 	showRemove: PropTypes.bool,
 	packageId: PropTypes.string.isRequired,
+	showPackIndividually: PropTypes.bool,
+	openMoveItem: PropTypes.func,
+	moveItem: PropTypes.func,
```
jeffstieler: Missing `isIndividualPackage`.

client/shipping-label/views/purchase/steps/packages/list.js
```diff
@@ -64,18 +64,18 @@ const PackageList = ( { selected, all, unpacked, packageId, openPackage, addPack
...
-	if ( packed.length ) {
-		packed.unshift( renderPackageListHeader( 'boxed-header', __( 'Boxed' ) ) );
+	if ( packed.length || individual.length ) {
+		packed.unshift( renderPackageListHeader( 'boxed-header', __( 'Packages to be Shipped' ) ) );
 	}
 
 	if ( individual.length ) {
-		individual.unshift( renderPackageListHeader( 'individual-header', __( 'Original packaging' ) ) );
+		//individual.unshift( renderPackageListHeader( 'individual-header', __( 'Original packaging' ) ) );
 	}
```
jeffstieler: This block is dead now.

client/shipping-label/views/purchase/steps/packages/package-info.js
```diff
@@ -81,22 +93,26 @@ const PackageInfo = ( { packageId, selected, all, flatRateGroups, unpacked, dime
...
-					}
-
-					return <optgroup label={ group.title } key={ groupId }>
-						{ _.map( group.definitions, renderPackageOption ) }
-					</optgroup>;
-				} ) }
-			</FormSelect>
+			<div>
+				<div className="wcc-package-items-header">
+					<FormLegend className="wcc-package-item__name">{ __( 'Box Package for Multiple Items' ) }</FormLegend>
```
jeffstieler: I think "Package" or "Package type" is simpler and more accurate here. This may not be a box, and it may not contain multiple items.

classes/class-wc-connect-shipping-label.php
```diff
@@ -283,14 +283,23 @@ protected function get_form_data( WC_Order $order ) {
...
 			$destination            = $this->get_destination_address( $order );
+			$unpacked = array();
+
+			// if there's no packaging information, add everything to the unpacked list
+			if ( ! $this->get_packaging_metadata( $order ) ) {
+				foreach( $selected_packages as $package ) {
+					$unpacked[] = $package[ 'items' ][ 0 ];
+				}
+				$selected_packages = new stdClass();
+			}
```
nabsul: That's definitely the right way to do it. I cut that corner to save some time but if you think it's worth the refactor I can do that.

olaolusoga:
Added some feedback to P2, was a bit long for github.
https://heyhydra.wordpress.com/2017/06/05/label-purchase-package-settings-proposed-changes/#comment-2549

**tl;dr**
We can improve the taxonomy + UX.

olaolusoga:
Added feedback to p7bbVw-1Aa-P2

tl;dr

Still feels like the UX + taxonomy can be slightly improved.

classes/class-wc-connect-shipping-label.php
```diff
@@ -41,51 +41,60 @@ public function __construct(
 			$this->payment_methods_store = $payment_methods_store;
 		}
 
+		private function get_item_data( WC_Order $order, WC_Order_Item $item ) {
```
jeffstieler: Let's avoid adding `private` methods going forward unless we need them. It's much easier to test `public` methods.

classes/class-wc-connect-shipping-label.php
```diff
@@ -148,7 +157,25 @@ protected function get_name( WC_Product $product ) {
...
 			$packages = $this->get_packaging_metadata( $order );
 			if ( ! $packages ) {
-				return $this->get_items_as_individual_packages( $order );
+				$items = array();
+				$weight = 0;
+
+				foreach( $this->get_all_items( $order ) as $id => $item ) {
+					$items[ $id ] = $item;
+					$weight += $item[ 'weight' ];
+				}
```
jeffstieler: What do you think about storing `$this->get_all_items( $order )` in `$items`, and calculating the weight sum without a loop?

Something like.. `$weight = array_sum( wp_list_pluck( $items, 'weight' ) );`

classes/class-wc-connect-shipping-label.php
```diff
@@ -180,6 +207,26 @@ public function get_selected_packages( WC_Order $order ) {
 			return $formatted_packages;
 		}
 
+		private function get_all_items( WC_Order $order ) {
```
jeffstieler: Same comment about `private` functions.

classes/class-wc-connect-shipping-label.php
```diff
@@ -180,6 +207,26 @@ public function get_selected_packages( WC_Order $order ) {
...
+			if ( $this->get_packaging_metadata( $order ) ) {
+				return array();
+			}
+
+			$items = array();
+			foreach( $order->get_items() as $item ) {
```
jeffstieler: Missing space after `foreach`.

classes/class-wc-connect-shipping-label.php
```diff
@@ -180,6 +207,26 @@ public function get_selected_packages( WC_Order $order ) {
...
+				$item_data = $this->get_item_data( $order, $item );
+				if ( null === $item_data ) {
+					continue;
+				}
+
+				for( $i = 0; $i < $item[ 'qty' ]; $i++ ) {
```
jeffstieler: Missing space after `for`.

client/shipping-label/views/purchase/steps/packages/package-info.js
```diff
@@ -81,21 +94,40 @@ const PackageInfo = ( { packageId, selected, all, flatRateGroups, unpacked, dime
...
-					}
-
-					return <optgroup label={ group.title } key={ groupId }>
-						{ _.map( group.definitions, renderPackageOption ) }
-					</optgroup>;
-				} ) }
-			</FormSelect>
+			<div>
+				<div className="wcc-package-items-header">
+					<FormLegend className="wcc-package-item__name">{ __( 'Package of Multiple Items' ) }</FormLegend>
```
jeffstieler: This should be "Shipping Package", according to designs.

client/shipping-label/views/purchase/steps/packages/package-info.js
```diff
@@ -81,21 +94,40 @@ const PackageInfo = ( { packageId, selected, all, flatRateGroups, unpacked, dime
...
+				</div>
+				<FormSelect onChange={ packageOptionChange } value={ pckg.box_id } isError={ pckgErrors.box_id }>
+					<option value={ 'not_selected' } key={ 'not_selected' }>{ __( 'Please select a box' ) }</option> )
```
jeffstieler: This matches the design, but should probably be "package" or "package type", since envelopes are options. cc: @olaolusoga

classes/class-wc-connect-shipping-label.php
nabsul: Are there any situations where `private` is ever needed? We can definitely agree to keep everything public for the purpose of this project, but I don't think it's ideal.

For reference: Standard best practices recommend using private by default to promote well-encapsulate code. Many people would also argue that if the only reason you're making a function public is for testing, then either the class or the test is not well-designed.

v18:
One nitpick I have is that when I'm on a different package and there's an error in the other package, it's not clear why I can't click on 'Use these packages':

screen shot 2017-06-09 at 3 34 37 pm

Here's the other package:

screen shot 2017-06-09 at 3 36 26 pm


Maybe we could highlight the packages with errors somehow?

screen shot 2017-06-09 at 3 36 23 pm

v18:
Is it ok that we're not alerting the user their weight may be incorrect if we don't have the weight of all of the items in the order?

v18:
The changes are super smooth, and much better, Nabeel!! Great job 🎉

classes/class-wc-connect-shipping-label.php
```diff
@@ -41,51 +41,60 @@ public function __construct(
...
+			$width = 0;
+
+			if ( $product->has_dimensions() ) {
+				$height = $product->get_height();
+				$length = $product->get_length();
+				$width  = $product->get_width();
+			}
+
+			$product_data = array(
+				'height'     => ( float ) $height,
```
v18: no spaces are needed for cast statements

classes/class-wc-connect-shipping-label.php
```diff
@@ -41,51 +41,60 @@ public function __construct(
...
+				'length'     => ( float ) $length,
+				'quantity'   => 1,
+				'weight'     => ( float ) $weight,
+				'width'      => ( float ) $width,
+				'name'       => $this->get_name( $product ),
+				'url'        => get_edit_post_link( WC_Connect_Compatibility::instance()->get_parent_product_id( $product ), null ),
+			);
+
+			if ( $product->is_type( 'variation' ) ) {
+				$product_data[ 'attributes' ] = WC_Connect_Compatibility::instance()->get_formatted_variation( $product, true );
```
v18: no need for spaces here `[ 'attributes' ]`

classes/class-wc-connect-shipping-label.php
```diff
@@ -41,51 +41,60 @@ public function __construct(
...
-
 					$packages[ $id ] = array(
 						'id'     => $id,
 						'box_id' => 'individual',
-						'height' => ( float ) $height,
-						'length' => ( float ) $length,
-						'weight' => ( float ) $weight,
-						'width'  => ( float ) $width,
-						'items'  => array( $product_data ),
+						'height' => $item_data[ 'height' ],
```
v18: same here for spaces in array keys

classes/class-wc-connect-shipping-label.php
```diff
@@ -148,17 +157,30 @@ protected function get_name( WC_Product $product ) {
...
+				return array(
+					'default_box' => array(
+						'id'     => 'default_box',
+						'box_id' => 'not_selected',
+						'height' => 0,
+						'length' => 0,
+						'weight' => $weight,
+						'width'  => 0,
+						'items'  => $items,
+					)
```
v18: need comma here

classes/class-wc-connect-shipping-label.php
```diff
@@ -148,17 +157,30 @@ protected function get_name( WC_Product $product ) {
...
 			$formatted_packages = array();
 
-			foreach( $packages as $package_obj ) {
+			foreach ( $packages as $package_obj ) {
 				$package = ( array ) $package_obj;
 				$package_id = $package[ 'id' ];
 				$formatted_packages[ $package_id ] = $package;
 
-				foreach( $package[ 'items' ] as $item_index => $item ) {
+				foreach ( $package[ 'items' ] as $item_index => $item ) {
```
v18: extra space around array key

classes/class-wc-connect-shipping-label.php
```diff
@@ -180,12 +202,32 @@ public function get_selected_packages( WC_Order $order ) {
...
+			}
+
+			$items = array();
+			foreach ( $order->get_items() as $item ) {
+				$item_data = $this->get_item_data( $order, $item );
+				if ( null === $item_data ) {
+					continue;
+				}
+
+				for ( $i = 0; $i < $item[ 'qty' ]; $i++ ) {
```
v18: array key spacing

classes/class-wc-connect-shipping-label.php
```diff
@@ -41,51 +41,60 @@ public function __construct(
 			$this->payment_methods_store = $payment_methods_store;
 		}
 
+		public function get_item_data( WC_Order $order, $item ) {
+			$product = WC_Connect_Compatibility::instance()->get_item_product( $order, $item );
+			if ( ! $product || ! $product->needs_shipping() ) {
```
v18: under what conditions might we get a false or null product?

client/shipping-label/index.js
```diff
@@ -68,7 +67,7 @@ export default ( { formData, labelsData, paperSize, storeOptions, paymentMethod
 						retrievalInProgress: false,
 					},
 				},
-				openedPackageId: labelsData ? '' : Object.keys( formData.selected_packages )[ 0 ] || '',
+				openedPackageId: Object.keys( formData.selected_packages )[ 0 ] || '',
```
v18: why was the check for labelsData removed here?

client/shipping-label/views/purchase/steps/packages/index.js
```diff
@@ -40,6 +39,13 @@ const PackagesStep = ( {
 			};
 		}
 
+		if ( ! isValidPackageType ) {
+			return {
+				isError: true,
+				summary: __( 'Please Select a Package Type' ),
```
v18: I'm not sure we should capitalize the error here

classes/class-wc-connect-shipping-label.php
jeffstieler: > Are there any situations where private is ever needed? We can definitely agree to keep everything public for the purpose of this project, but I don't think it's ideal.

It probably makes sense at some point for us to audit the code and determine what should me public vs. private/protected. Whatever remains public should be considered code that needs to be maintained for backwards compatibility.

> Many people would also argue that if the only reason you're making a function public is for testing, then either the class or the test is not well-designed.

Sure, and my reasoning right now is really only to avoid having to use `ReflectionClass` to make individual methods testable.

classes/class-wc-connect-shipping-label.php
jeffstieler: > under what conditions might we get a false or null product?

If the product has been deleted after the order was placed.

classes/class-wc-connect-shipping-label.php
nabsul: I want eslint for PHP so badly!

client/shipping-label/index.js
nabsul: This was needed in the case that all items were in the "saved for later" list and there were no packages. But in the new paradigm, we'll always at least have one package. Speaking of which, I just realized there's a major bug in our UI.... If you want to move everything to individual packages, you won't be able to delete the remaining empty box. I'll get that fixed.

client/shipping-label/views/purchase/steps/packages/index.js
nabsul: agreed.

olaolusoga:
>Maybe we could highlight the packages with errors somehow?

Suggestion for error notice
slack-imgs

Number of items replaced by exclamation mark till issue solved.

client/shipping-label/views/purchase/steps/packages/get-package-descriptions.js
```diff
@@ -10,7 +10,7 @@ export default ( selected, all, addNames ) => {
 			return pckg.items[ 0 ].name;
 		}
 
-		let pckgType = __( 'Package %d' );
+		let pckgType = __( 'Box %d' );
```
jeffstieler: Why did this change from "package" to "box"?

nabsul: There have been so many iterations on this that I've lost track, but I THINK @olaolusoga suggested it.

olaolusoga: We should keep it "package".

PR #1031, opened 2017-06-06
Add NUX banners

v18:
# Changes in this PR
1. A new function, `set_up_nux_notices()`, to show banners is hooked into `admin_init` in `WC_Connect_Loader->init()`
2. `set_up_nux_notices()` checks for Jetpack status and user capabilities and either adds a banner using `admin_notice` before or after a Jetpack connection is established
3. Each notice may show different content based on:
- new store owner vs existing (based on # of orders)
- Jetpack status
4. All Jetpack actions are done using js - install, activate, connect

Video of flow:

https://cloudup.com/cxrZz773unw

cc @olaolusoga 

# Changes not in this PR (coming soon to a PR near you!)
- removing the older TOS and JP connect notices
- button action/link after connection is done
- removing the JP notice on the plugins page
- styles


classes/class-wc-connect-nux.php
```diff
@@ -1,5 +1,16 @@
 <?php
 
+if ( ! class_exists( 'Jetpack_Install_Status' ) ) {
+	abstract class Jetpack_Install_Status {
```
nabsul: Why abstract?

classes/class-wc-connect-nux.php
```diff
@@ -1,5 +1,16 @@
 <?php
 
+if ( ! class_exists( 'Jetpack_Install_Status' ) ) {
```
nabsul: I'd go ahead and put this class in its own file.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +94,168 @@ public function register_add_service_to_zone_pointer( $pointers ) {
 			);
 			return $pointers;
 		}
+
+		public function get_jetpack_install_status() {
```
nabsul: How about adding this function to the `Jetpack_Install_Status` class?

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +94,168 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+			if( 0 === validate_plugin( 'jetpack/jetpack.php' ) ) {
+				$status = Jetpack_Install_Status::INSTALLED;
+			}
+
+			if ( ! class_exists( 'Jetpack_Data' ) ) {
+				// installed, but not activated
+				return $status;
+			}
+
+			if ( defined( 'JETPACK_DEV_DEBUG' ) ) {
```
nabsul: So if Jetpack is not unstalled, we're going to be returning `Jetpack_Install_statis::DEV`. Is this what we want to return?

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +94,168 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				$status = Jetpack_Install_Status::DEV;
+				return $status;
+			}
+
+			if ( ! method_exists( 'Jetpack_Data', 'get_access_token' )
+				|| ! class_exists( 'Jetpack' )
+			) {
+				$status = Jetpack_Install_Status::ERROR;
```
nabsul: This could happen simply because Jetpack is not installed.

classes/class-wc-connect-nux.php
jeffstieler: These could also be class constants in the `WC_Connect_Nux` class, perhaps prefixed with `JETPACK_`.

classes/class-wc-connect-nux.php
jeffstieler: This should also check if `JETPACK_DEV_DEBUG` is `true`.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +94,168 @@ public function register_add_service_to_zone_pointer( $pointers ) {
 			);
 			return $pointers;
 		}
+
+		public function get_jetpack_install_status() {
+			$status = Jetpack_Install_Status::UNINSTALLED;
+
+			if( 0 === validate_plugin( 'jetpack/jetpack.php' ) ) {
```
jeffstieler: Missing space after `if`. _(Please check all control structure whitespace usage)_

classes/class-wc-connect-nux.php
jeffstieler: Is this a sanity check for a potentially corrupt Jetpack plugin? Like a file is missing?

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +94,168 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+
+			if ( ! method_exists( 'Jetpack_Data', 'get_access_token' )
+				|| ! class_exists( 'Jetpack' )
+			) {
+				$status = Jetpack_Install_Status::ERROR;
+				return $status;
+			}
+			$status = Jetpack_Install_Status::ACTIVATED;
+
+
```
jeffstieler: Extra newline here.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +94,168 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+			$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
+			if ( $user_token
+				&& is_object( $user_token )
+				&& isset( $user_token->external_user_id )
+			) {
+				$status = Jetpack_Install_Status::CONNECTED;
+			};
```
jeffstieler: Unneeded semicolon.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +94,168 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+			) {
+				$status = Jetpack_Install_Status::ERROR;
+				return $status;
+			}
+			$status = Jetpack_Install_Status::ACTIVATED;
+
+
+			$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
+			if ( $user_token
+				&& is_object( $user_token )
```
jeffstieler: This isn't needed - the `isset()` check is enough to determine if `$user_token` is an object.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +94,168 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+					add_action(
+						'admin_notices',
+						array( $this, 'show_banner_after_connection_see_how_it_works' )
+					);
+					break;
+			}
+		}
+
+		public function show_banner_before_connection_get_access( $content ) {
+			if( function_exists( 'get_current_screen' ) ) {
```
jeffstieler: This seems overly defensive - in what scenario would `get_current_screen()` be undefined?

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +94,168 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				'title'          => __( 'Get access to discount shipping labels by connecting to WordPress.com', 'woocommerce-services' ),
+				'description'    => __( 'WooCommerce Services is almost ready to go. Once you connect your store to WordPress.com you can begin printing labels and saving money with discounted shipping rates all from your dashboard.', 'woocommerce-services' ),
+				'url'            => $connect_url,
+				'button_text'    => __( 'Connect your store to WordPress.com', 'woocommerce-services' ),
+				'image_url'      => 'https://cldup.com/WpkrskfH_r.jpg',
+				'should_show_jp' => true,
+			) );
+		}
+
+		public function show_banner_before_connection_welcome( $content ) {
```
jeffstieler: [`admin_notices`](https://developer.wordpress.org/reference/hooks/admin_notices/) doesn't pass any arguments.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +94,168 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				&& 'post' === $screen->base
+				&& 'product' === $screen->post_type
+			) ) {
+				return;
+			}
+			if (  'add' === $screen->action ) {
+				$redirect = admin_url( 'post-new.php?post_type=product' );
+			} else {
+				global $post;
+				$redirect = get_edit_post_link( $post->ID, false );
```
jeffstieler: You can save accessing the global `$post` by passing in `null` instead.

See:
https://developer.wordpress.org/reference/functions/get_edit_post_link/
https://developer.wordpress.org/reference/functions/get_post/

classes/class-wc-connect-nux.php
v18: Ugh, apologies. Didn't run the linter before pushing. Fixed all linting issues & rebased :)

classes/class-wc-connect-nux.php
v18: @jeffstieler I encapsulated the constants in their own class since they're not directly related to NUX & the Jetpack cxn constants are likely useful outside of NUX (have not run into that yet, though, but I think the first reason stands on its own, too). To me, adding the Jetpack-related constants in this class would seem a bit out of place. Is there another reason to keep them in the NUX class that I'm not thinking of? Maybe simplicity for now? Not tied to either approach atm :)

> I'd go ahead and put this class in its own file.

Sounds good. Will do in a bit (if we keep the class).

classes/class-wc-connect-nux.php
v18: Yeh, I think that makes sense. Let me go down a bit further into this & I'll add it there.

classes/class-wc-connect-nux.php
v18: Fixed

classes/class-wc-connect-nux.php
v18: I saw that in [Codex](https://codex.wordpress.org/Function_Reference/get_current_screen) - `get_current_screen()` may be undefined in a few admin screens. They're not all listed, and I didn't dig in to see which ones and under which conditions. Can dig in to make sure it's none of the ones we care about.

classes/class-wc-connect-nux.php
v18: Thanks, fixed in all cases.

classes/class-wc-connect-nux.php
v18: Nice! Updated.

classes/class-wc-connect-nux.php
v18: Fixed

classes/class-wc-connect-nux.php
v18: So that we don't have to instantiate it, as described in this [PHP doc](http://php.net/manual/en/language.oop5.abstract.php). Since this class doesn't rely on object properties - they're all basically static properties - there's no need for an instance. Does this make sense in the general realm of PHP & WP?

classes/class-wc-connect-nux.php
v18: Nice catch. I tried to write some unit tests for this, but I didn't get very far when trying to install Jetpack.

On another note, I refactored this function quite a bit. Makes a lot more sense now :D Also fixed the `true` issue.

classes/class-wc-connect-nux.php
v18: Yeh, it was. But, it turns out it's not really needed. Jetpack itself doesn't check for this, and if that method doesn't exist, it fatals. So, we're good to rely on it given that the class `Jetpack_Data` exists. If Jetpack changes its method name, we can change our code, too.

classes/class-wc-connect-nux.php
v18: Neat! `$user_token` can be false, but that works for `isset()`, too. TIL 😄 Now we just have the one check: `$user_token->external_user_id` 👌

v18:
Note: made a bunch of changes & rebased.

classes/class-wc-connect-nux.php
jeffstieler: You don't have to instantiate any defined class, abstract or not. The difference with abstract classes is that they *can not* be instantiated - it causes a fatal error.

v18: @jeffstieler so which use makes more sense?

classes/class-wc-connect-nux.php
jeffstieler: After looking at the Codex and inspecting the hook firing order (using the [Debug Bar Actions and Filters Addon](https://wordpress.org/plugins/debug-bar-actions-and-filters-addon/)), I think we're good to remove this check.

The codex states:
> The function returns null if called from the `admin_init` hook. It should be OK to use in a later hook such as `current_screen`.

I'm seeing `admin_notices` occur after `current_screen`.

For your consideration, an alternate approach could be to hook into `current_screen` action initially, which gets the `$screen` object passed to it, determine which notices belong on that screen, and _then_ attach the appropriate display callbacks to the `admin_notices` action.

classes/class-wc-connect-nux.php
jeffstieler: @v18 personally, I'd just add the class constants to `WC_Connect_Nux`, since to me it's just a mechanism to avoid some "magic" strings. Ultimately, I'm not too picky on this, so if you want to keep the abstract class, go for it. 👍

classes/class-wc-connect-nux.php
nabsul: Even without the abstract keyword you don't have to create an instance to access the constants. They are basically static variables on the class name:

```
class test_class {	const VAL = 'hello'; }
echo test_class::VAL;
```

Abstract is used to define an "incomplete" parent class, whose abstract functions must be defined in derived classes for them to be non-abstract. It's for that reason that abstract classes are impossible to instantiate.

classes/class-wc-connect-nux.php
nabsul: Oh, I added an elaboration on `abstract` in the wrong thread :-)

nabsul: Personally I like the idea of keeping classes small, but like Jeff, no strong feelings.

classes/class-wc-connect-nux.php
v18: Thanks for enlightening me, @nabsul :D

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +91,272 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				'image_url'       => $image_url,
+				'should_show_jp'  => true,
+			);
+
+			$base_location = wc_get_base_location();
+			$country = isset( $base_location['country'] )
+				? $base_location['country']
+				: '';
+			switch ( $country ) {
+				case 'CA':
```
jeffstieler: Does this show for non-US and CA stores?

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +91,272 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+			return 0 === $order_count;
+		}
+
+		private function get_jetpack_redirect_url() {
+			$full_path = add_query_arg( array() );
+			// Remove [...]/wp-admin so we can use admin_url().
+			$new_index = strpos( '/wp-admin', $full_path ) + strlen( '/wp-admin' );
+			$path = substr( $full_path, $new_index );
+			return admin_url( $path );
+		}
```
jeffstieler: This isn't the most helpful comment, but this doesn't feel like the best way to accomplish getting the current URL. I'll try to find an alternative.

client/banner.js
```diff
@@ -1,19 +1,81 @@
...
+						btn.html( wp.updates.l10n.installing );
+						return wp.updates.installPlugin( { slug: 'jetpack' } );
+					} );
+			}
+			return $.Deferred().resolve();
+		}
+
+		function activateStep() {
+			if ( 'installed' === wcs_nux_notice.initial_install_status ||
+				'uninstalled' === wcs_nux_notice.initial_install_status
```
v18: Hmm, I think this is probably not written in the best possible way. Here we're checking if we want to go through the activation step at all, not whether it will be the first step.

We do want to go through activation step (possibly after the installation step) in these cases:
1. When Jetpack is not yet installed (condition 1)
2. After installing Jetpack from scratch (condition 2)

Maybe there's a better way to indicate that's what's happening. Maybe using a variable like `shouldDoActivationStep`?

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +91,272 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				echo 'success';
+			} else {
+				if ( is_wp_error( $result ) ) {
+					echo esc_html( $result->get_error_message() );
+				} else {
+					echo 'error';
+				}
+			}
+
+			wp_die();
```
v18: @jeffgolenski do you think I should use them here? Doesn't seem needed since it's a simple text response.

classes/class-wc-connect-nux.php
v18: Yes. We don't have specific text for them, though.

classes/class-wc-connect-nux.php
v18: Thanks! I wasn't sure about this either, but it was the best way I could find that wouldn't require hard-coding each case separately.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +91,272 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+
+		private function get_number_of_published_products() {
+			$count = wp_count_posts( 'product' );
+			return $count->publish;
+		}
+
+		private function is_new_store() {
+			$order_count = 0;
+			$posts_by_status = wp_count_posts( 'shop_order' );
+			foreach ( $posts_by_status as $post_status => $count ) {
```
v18: Not 100% sure I understand your question. I didn't see a relevant WC function that would work here.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +91,272 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+			return self::JETPACK_ACTIVATED;
+		}
+
+		private function should_display_nux_notice_on_screen( $screen ) {
+			if ( // Display if on any of these admin pages.
+				( // Products list.
+					'product' === $screen->post_type
+					&& 'edit' === $screen->base
+				)
+				|| ( // Edit product page if at least one product already published
```
v18: I spoke with @olaolusoga about this since the meeting, and we decided to add it to the products page when there's more than 1 published product already, since it does not detract from the intended WC NUX experience at that point.

At the meeting we discussed specifically that showing the banner on the 'Edit Product' page would be detrimental on the first product, but not subsequent products.

classes/class-wc-connect-nux.php
v18: To close the loop - constants were added to the main nux class instead.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +91,272 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+			// check if connected
+			$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
+			if ( isset( $user_token->external_user_id ) ) { // always an int
+				return self::JETPACK_CONNECTED;
+			}
+
+			return self::JETPACK_ACTIVATED;
+		}
+
+		private function should_display_nux_notice_on_screen( $screen ) {
```
v18: 👍 Fixed in bc19e65c10dc728a36b26f79dced7aa9ac3ae117

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +91,272 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				|| 'plugins' === $screen->base
+			) {
+				return true;
+			}
+			return false;
+		}
+
+		private function get_number_of_published_products() {
+			$count = wp_count_posts( 'product' );
+			return $count->publish;
```
v18: Fixed.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +91,272 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+					'wc-on-hold',
+					'wc-completed',
+					'wc-cancelled',
+					'wc-refunded',
+					'wc-failed',
+				), true ) ) {
+					$order_count += $count;
+				}
+			}
+			return 0 === $order_count;
```
v18: Sounds good to me, updated.

client/banner.js
```diff
@@ -1,19 +1,81 @@
...
+			) {
+				return $.when()
+				.then( function() {
+					btn.html( wcs_nux_notice.translations.activating + '...' );
```
v18: Thanks! Moved the ellipsis to the translation function.

woocommerce-services.php
```diff
@@ -361,6 +361,7 @@ public function load_textdomain() {
 		public function init() {
 			$this->load_dependencies();
 			add_action( 'admin_init', array( $this, 'admin_enqueue_scripts' ) );
+			add_action( 'admin_init', array( $this->nux, 'set_up_nux_notices' ) );
```
v18: Fixed

classes/class-wc-connect-nux.php
v18: I added it as an agenda item to discuss in today's meeting.

classes/class-wc-connect-nux.php
jeffstieler: Asked another way - why is a `foreach` working on what should be an `object`? Look at the definition of the `wp_count_posts()` function.

classes/class-wc-connect-nux.php
jeffstieler: They're a convenience. If they're not convenient, don't use them. In this case it might only save a call to `wp_die()`.

client/banner.js
jeffstieler: I seemed to have just steamrolled over the `initial_` prefix on the variable name. Sorry.

classes/class-wc-connect-nux.php
v18: Gotcha, thanks :)

client/banner.js
v18: am I right?, Yoda conditions.

Could probably still be better, but leaving for now.

v18:
Rebased on master

client/banner.js
jeffstieler: What would be best is to check the install state in the `activateStep()` function, rather than make a guess based on the initial state.

classes/class-wc-connect-nux.php
v18: Removed.

classes/class-wc-connect-nux.php
v18: Yep, I see it's returning an object. As far as I can see, I can iterate through the object properties using foreach since PHP 5, as noted in this [PHP object doc](http://php.net/manual/en/language.oop5.iterations.php). I think I'm missing something? :)

classes/class-wc-connect-nux.php
v18: @jeffstieler any news on this?

classes/class-wc-connect-nux.php
v18: Removed.

client/banner.js
v18: Opened an issue for that: #1034

v18:
This code should now be good to go ( cc @jeffstieler ). I rebased in a separate branch, so when we're ready I'll port that over to here.

cc @nabsul about the upcoming rebase since you're working off of this branch. There will be no changes between the two branches at the tip, so shouldn't be hard to `pull --rebase` as usual.

Blockers for getting this merged:
- [ ] hi-fi images and design spec from @olaolusoga
- [ ] adding stylin' & new images

woocommerce-services.php
v18: @jeffstieler - I don't think I  understand where you'd like this to go after all. Can you elaborate?

v18:
Added the dismiss button. @nabsul or @jeffstieler can I get a review on 85b1e8399ca84ebd5356025ff5dc59140ce2fdae please?

jeffstieler:
> Added the dismiss button. @nabsul or @jeffstieler can I get a review on 85b1e83 please?

jeffstieler:
@v18 we need a way to determine if there's an existing connected WCS install - I'm seeing the banner show up in the "Setup Complete!" state on an existing install.

woocommerce-services.php
```diff
@@ -361,16 +361,7 @@ public function load_textdomain() {
...
-				return;
-			}
-
-			if ( ! WC_Connect_Options::get_option( 'tos_accepted', false ) ) {
-				add_action( 'admin_init', array( $this, 'admin_tos_notice' ) );
-				return;
-			}
+			add_action( 'admin_init', array( $this->nux, 'set_up_nux_notices' ) );
 
 			$this->schedule_service_schemas_fetch();
```
nabsul: Does fetching the service schema make any sense if the user hasn't connected, accepted TOS, etc.?

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +97,288 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				return;
+			}
+
+			if ( ! $this->should_display_nux_notice_for_current_store_locale() ) {
+				return;
+			}
+
+			$jetpack_install_status = $this->get_jetpack_install_status();
+
+			$ajax_data = array(
```
nabsul: You're only using data in one part of a switch statement. Seems kind of wasteful to execute it in all situations. Moreover, it looks like this code will be executed on every page load, even after the NUX flow has been completed...

v18: Yeh, we need to do a check for tos acceptance before displaying this, which I wanted to get done in a seprate TOS-related PR.

As for the `$ajax_data`, sure I'll move it.

woocommerce-services.php
v18: No, I don't think it does. We were relying on the connection and TOS-related checks before scheduling the schema, which were removed in #1036.

classes/class-wc-connect-nux.php
v18: Moved the ajax array over.

v18:
@jeffstieler, I added a few more commits to cover these things:
- add styles
- remove old styles
- update TOS message

Can you check it please before I merge to the new `q2-nux` branch?

> we need a way to determine if there's an existing connected WCS install - I'm seeing the banner show up in the "Setup Complete!" state on an existing install.

As discussed yesterday in Slack I'll do this in a subsequent PR, which will include all new/existing and TOS + connect related flows.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +97,300 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				<div class="wcs-nux__notice-logo">
+					<img src="<?php echo esc_url( $content['image_url'] );  ?>">
+				</div>
+				<div class="wcs-nux__notice-content">
+					<h1><?php echo esc_html( $content['title'] ); ?></h1>
+					<p class="wcs-nux__notice-content-text">
+						<?php echo esc_html( $content['description'] ); ?>
+					</p>
+					<?php if ( $content['should_show_jp'] ) : ?>
+						<p><?php printf( esc_html__( 'By clicking "%1$s", you agree to the %2$sTerms of Service%3$s and understand that some of your data will be passed to external servers. You can find more information about how your data is handled %4$shere%5$s.', 'woocommerce-services' ),
```
jeffstieler: Let's add a "translators:" comment here so they know what the placeholders represent. See: https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#descriptions

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +97,300 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				: '';
+			switch ( $country ) {
+				case 'CA':
+					$localized_content = array(
+						'description'     => __( "WooCommerce Shipping is almost ready to go! Once you connect your store you'll be able to show your customers live shipping rates when they check out.", 'woocommerce-services' ),
+					);
+					break;
+			}
+
+			$this->show_nux_banner( wp_parse_args( $localized_content = null, $default_content ) );
```
jeffstieler: What do you think about moving the defaults handling into the `show_nux_banner()` method, instead of only having it for this call? It would save having to check the passed in `$content` array.

client/banner.js
```diff
@@ -10,17 +12,77 @@ import jQuery from 'jquery';
...
-			.addClass( 'disabled' );
+		function activateStep() {
+			if ( 'installed' === wcs_nux_notice.initial_install_status ||
+				'uninstalled' === wcs_nux_notice.initial_install_status
+			) {
+				return $.when()
+				.then( function() {
+					btn.html( wcs_nux_notice.translations.activating );
+					return $.post( ajaxurl, {
+						action: 'woocommerce_services_activate_jetpack',
```
jeffstieler: This doesn't have to be done in order to merge - but I just had the thought: if we're adding a custom ajax endpoint to handle activation, why not just use the single endpoint to perform both install and activation, depending on what work is required given the state of Jetpack on the store?

v18: @jeffstieler there isn't a PHP-based `install_plugin` function, only that JavaScript function. And vice versa for activation - there isn't a JS-based function, only the PHP `activate_plugin()`. That's why the install has to be done in JavaScript and the activation in PHP.

classes/class-wc-connect-nux.php
v18: I wouldn't want to do that since there are different defaults for `show_banner_after_connection()` and `show_banner_before_connection()`, but they both use `show_nux_banner()`.

classes/class-wc-connect-nux.php
v18: Updated with new syntax based on advice from @hewsut & added a translators comment in 6f44247d94b672d8100f6d5d1134b2fcd2a3e60f

client/banner.js
jeffstieler: @v18 you'd install using the `Plugin_Upgrader` class, see this example: https://github.com/Automattic/jetpack/blob/master/3rd-party/woocommerce-services.php#L101

classes/class-wc-connect-nux.php
jeffstieler: Ok, then we should either check to make sure the array keys exist that we're accessing in `show_nux_banner()`, or have an additional `wp_parse_args()` call in it with some empty strings or safe generic defaults.

client/banner.js
v18: :D thanks! adding this to my list to update

classes/class-wc-connect-nux.php
v18: Shelving this for now.

classes/class-wc-connect-nux.php
```diff
@@ -83,5 +97,307 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+			return false;
+		}
+
+		public function get_jetpack_redirect_url() {
+			$full_path = add_query_arg( array() );
+			// Remove [...]/wp-admin so we can use admin_url().
+			$new_index = strpos( $full_path, '/wp-admin' ) + strlen( '/wp-admin' );
+			$path = substr( $full_path, $new_index );
+			return admin_url( $path );
+		}
```
jeffstieler: This might be a decent alternative:

```php
site_url( add_query_arg( array() ) )
```

v18: That doesn't work for sites installed in a subfolder, it will output the subfolder twice in the URL:

`http://example.com/wp/wp/wp-admin/plugins.php`

This is because `add_query_arg( array() )` returns the queries _and_ the full path. So, if the WordPress install is in a subfolder, the result will look like this: `/wp/wp-admin/plugins.php`. 

`site_url()` uses the URL for the current site where WordPress files are installed, and appends the path passed in. Since they both contain the subfolder, the result has the subfolder twice.

I looked through the code for [`wp_safe_redirect`](https://core.trac.wordpress.org/browser/tags/4.8/src/wp-includes/pluggable.php#L1281) and related functions, but WordPress does not get the full URL to do the redirects, only the path passed in, since relative URLs will work.

We can't pass in just the path (i.e. the result of `add_query_arg( array() )`) here because Jetpack expects a full URL for the redirect. Otherwise, it will end up at `wordpress.com/$path_passed_in` after the connection is done.

classes/class-wc-connect-nux.php
v18: Closing the loop, [the conversation continues here](https://github.com/Automattic/woocommerce-services/pull/1031#discussion_r122555461).

client/banner.js
v18: Opened an issue for this: #1040

PR #1032, opened 2017-06-12
Added Calypso's eslint configuration

DanReyLop:
As part of our ongoing effort to somehow merge WCS into Calypso, I've added `wp-calypso` eslint configuration to our plugin.

See each individual commit for each eslint error that was fixed, that's the only sane way to review this PR.

The only rule I couldn't fix was the one that checks CSS `className`s for our `component-name__modifier` naming convention. Since some of our component folders are simply called `views`, the "correct" classNames would make no sense. We'll need some serious folder reshuffling to comply with that one, and I didn't want to make this PR any bigger.


DanReyLop:
Thanks @marcinbot ! I'll leave this up at least until :us: wakes up.

@nabsul you have a big PR incoming, and rebase with this will probably be a pain. Please tell me if you want to merge first.

jeffstieler:
UI seems unaffected by the changes. 👍

nabsul:
Thanks for the consideration @DanReyLop . I'd really appreciate getting my PR merged first :-)

v18:
@marcinbot - in order to use `href="#"`, we need to make sure to use `event.preventDefault()` on the onClick event (etc) to prevent unexpected page scrolling.

I would personally prefer to go with `event.preventDefault` since `javascript:void(0)` is an older workaround.

I took a look at some of the `void` changes, and it looks like in a lot of cases we could also switch to use `<button>` since no link is required in our uses cases if we don't always want to use `event.preventDefault()`. Buttons do not by default get redirected anywhere upon click.

(Mentioned ^ in our Slack channel already, re-posting here for reference.)

DanReyLop:
Changed back to `href="#"` (adding `e.preventDefault()` where it was missing) in https://github.com/Automattic/woocommerce-services/pull/1032/commits/a83a65d1799767db8280a11b4bcd65a69e4e09a4

I honestly didn't expect so much discussion over that trivial part of the PR 😂

DanReyLop:
I've squashed and rebased after @nabsul merged his PR. There were a lot of conflicts. I don't expect you folks to review this all over again, but I made a quick UI test and everything seems to work fine (it would fail pretty spectacularly if I had screwed up).

nabsul:
Since it's my fault, I'll walk around our UI and see if I find any issues :-)

nabsul:
Everything I tested is working as expected! :shipit:

PR #1036, opened 2017-06-14
Hide Jetpack Banner when Using WCS Banner

nabsul:
In this PR:

- Remove jetpack's banners from the `admin_notices` hook if we are adding the WCS connect banner
- Delete old jetpack connect and tos notice code from the plugin

To Test:
- Start with a setup where Jetpack is not connected
- Make sure you're not using `JETPACK_DEV_DEBUG` in your config
- Go the plugins page and notice that the WCS connect banner is showing without the Jetpack banner
- Disable the WCS plugin, notice that the jetpack banner now shows
- Make sure nothing in the plugin is broken

@jeffstieler @v18 @DanReyLop @marcinbot @olaolusoga


classes/class-wc-connect-nux.php
```diff
@@ -212,6 +212,7 @@ public function set_up_nux_notices() {
 						array( $this, 'ajax_get_jetpack_connect_url' )
 					);
 					add_action( 'admin_notices', array( $this, 'show_banner_before_connection' ) );
+					add_action( 'admin_notices', array( $this, 'remove_jetpack_notices' ), 1 );
```
jeffstieler: Should this get called inside `show_banner_before_connection()` instead? There is still a check inside `show_banner_before_connection()` that determines if the NUX notice should display.

PR #1038, opened 2017-06-16
Update NUX flow logic for all scenarios

v18:
These are the scenarios that need to be handled:

<table class="table table-bordered table-hover table-condensed">
<tbody><tr>
<td>Install type</td>
<td>JP cxned</td>
<td>TOS accepted</td>
<td>First banner</td>
<td>After cxn banner</td>
<td>After TOS banner</td>
</tr>
<tr>
<td>New </td>
<td>No</td>
<td>No</td>
<td>Original before</td>
<td>Original after</td>
<td>N/A</td>
</tr>
<tr>
<td>New </td>
<td>Yes</td>
<td>No</td>
<td>TOS (new)</td>
<td>N/A</td>
<td>Original after</td>
</tr>
<tr>
<td>Update</td>
<td>No</td>
<td>No</td>
<td>Original before</td>
<td>Original after</td>
<td>N/A</td>
</tr>
<tr>
<td>Update</td>
<td>Yes</td>
<td>No</td>
<td>TOS (new)</td>
<td>N/A</td>
<td>Original after</td>
</tr>
<tr>
<td>Update</td>
<td>No</td>
<td>Yes</td>
<td>Original before</td>
<td>Original after</td>
<td>N/A</td>
</tr>
<tr>
<td>Update</td>
<td>Yes</td>
<td>Yes</td>
<td>N/A</td>
<td>N/A</td>
<td>N/A</td>
</tr>
</tbody></table>

Original before = `show_banner_before_connection()`
Original after = `show_banner_after_connection`
TOS (new) = new TOS banner (does not yet exist)

# Testing instructions
1. Deactivate Jetpack, and delete all WCS options in the DB. Should see: the before connection banner before connecting to JP, the after connection banner after connecting to JP. You should be able to dismiss the after connection banner and not see it again. You should see the `tos_accepted` option saved.
2. Deactivate Jetpack. Keep `tos_activated` still saved in the DB. You should see the same as # 1 above: the before connection, and after connection banners.
3. Add logging on line [238](https://github.com/Automattic/woocommerce-services/pull/1038/files#diff-02948c6603eae6cf021f42bde08b17c8R238). Remove the `tos_accepted` option, and keep Jetpack connected. You should see the TOS-only logging you just added.
4. After you're both connected to Jetpack and have TOS enabled, and you've dismissed the after connection notice, you should see no more notices.
5. Run the php unit tests

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -256,6 +256,8 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 					add_action( 'admin_notices', array( $this, 'show_banner_after_connection' ) );
 					break;
 				case 'tos_only_banner':
+					wp_enqueue_style( 'wc_connect_banner' );
+					add_action( 'admin_notices', array( $this, 'show_tos_banner' ) );
 					break;
 			}
 		}
@@ -335,6 +337,26 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 				'button_text'    => __( 'Got it, thanks!', 'woocommerce-services' ),
 				'button_link'    => add_query_arg( array(
 					'wcs-nux-notice' => 'dismiss',
+				) ),
+				'image_url'      => plugins_url(
+					'images/nux-printer-laptop-illustration.png', dirname( __FILE__ )
+				),
+				'should_show_jp' => false,
+			) );
+		}
+
+		public function show_tos_banner() {
+			if ( isset( $_GET['wcs-nux-tos'] ) && 'accept' === $_GET['wcs-nux-tos'] ) {
+				WC_Connect_Options::update_option( 'tos_accepted', true );
+				wp_safe_redirect( remove_query_arg( 'wcs-nux-tos' ) );
+				exit;
+			}
+			$this->show_nux_banner( array(
+				'title'          => __( 'Setup complete! We need you to accept our TOS', 'woocommerce-services' ),
+				'description'    => __( 'Everything is ready to roll, we just need you to agree to our Terms of Service.', 'woocommerce-services' ),
+				'button_text'    => __( 'I accept the TOS!', 'woocommerce-services' ),
+				'button_link'    => add_query_arg( array(
+					'wcs-nux-tos' => 'accept',
 				) ),
 				'image_url'      => plugins_url(
 					'images/nux-printer-laptop-illustration.png', dirname( __FILE__ )
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -333,7 +333,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 			WC_Connect_Options::update_option( 'tos_accepted', true );
 
 			$this->show_nux_banner( array(
-				'title'          => __( 'Setup complete! You can now access discounted shipping rates and printing services' ),
+				'title'          => __( 'Setup complete! You can now access discounted shipping rates and printing services', 'woocommerce-services' ),
 				'description'    => __( 'When you’re ready, you can purchase discounted labels from USPS, and print USPS labels at home.', 'woocommerce-services' ),
 				'button_text'    => __( 'Got it, thanks!', 'woocommerce-services' ),
 				'button_link'    => add_query_arg( array(
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -299,6 +299,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 				'button_text'     => $button_text,
 				'image_url'       => $image_url,
 				'should_show_jp'  => true,
+				'should_show_terms' => true,
 			);
 
 			$base_location = wc_get_base_location();
@@ -345,6 +346,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 					'images/nux-printer-laptop-illustration.png', dirname( __FILE__ )
 				),
 				'should_show_jp' => false,
+				'should_show_terms' => false,
 			) );
 		}
 
@@ -365,6 +367,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 					'images/nux-printer-laptop-illustration.png', dirname( __FILE__ )
 				),
 				'should_show_jp' => false,
+				'should_show_terms' => false,
 			) );
 		}
 
@@ -379,7 +382,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 					<p class="wcs-nux__notice-content-text">
 						<?php echo esc_html( $content['description'] ); ?>
 					</p>
-					<?php if ( $content['should_show_jp'] ) : ?>
+					<?php if ( isset( $content['should_show_terms'] ) && $content['should_show_terms'] ) : ?>
 						<p><?php
 						/* translators: %1$s example values include "Install Jetpack and CONNECT >", "Activate Jetpack and CONNECT >", "CONNECT >" */
 						printf(
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -311,9 +311,11 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 						'description'     => __( "WooCommerce Shipping is almost ready to go! Once you connect your store you'll be able to show your customers live shipping rates when they check out.", 'woocommerce-services' ),
 					);
 					break;
+				default:
+					$localized_content = array();
 			}
 
-			$this->show_nux_banner( wp_parse_args( $localized_content, $default_content ) );
+			$this->show_nux_banner( array_merge( $default_content, $localized_content ) );
 		}
 
 		public function show_banner_after_connection() {
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -6,9 +6,9 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 		/**
 		 * Jetpack status constants.
 		 */
-		const JETPACK_UNINSTALLED = 'uninstalled';
-		const JETPACK_INSTALLED = 'installed';
-		const JETPACK_ACTIVATED = 'activated';
+		const JETPACK_NOT_INSTALLED = 'uninstalled';
+		const JETPACK_INSTALLED_NOT_ACTIVATED = 'installed';
+		const JETPACK_ACTIVATED_NOT_CONNECTED = 'activated';
 		const JETPACK_DEV = 'dev';
 		const JETPACK_CONNECTED = 'connected';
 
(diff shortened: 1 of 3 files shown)


v18:
This PR is ready for testing, I would like to make one change if time permits: add a separate function to decide which banner to display, with unit tests.

It's also missing the function to actually show the TOS-only banner, and the button to save the TOS acceptance on that banner.

nabsul:
In #1042, I've also implemented the logic checking the `tos_accepted` option and other things that you're already doing. As we discussed offline, let's just cannibalize the one or two functions of code that are not redundant in #1042 so that we only have one PR to review.

jeffstieler:
Got this warning in testing step #2:
`Notice: Undefined variable: localized_content in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/classes/class-wc-connect-nux.php on line 312`

classes/class-wc-connect-nux.php
```diff
@@ -98,6 +98,37 @@ public function register_add_service_to_zone_pointer( $pointers ) {
 			return $pointers;
 		}
 
+		public static function get_banner_type_to_display( $status = array() ) {
```
jeffstieler: When taking in associative array parameters, we should check for keys existing or use a defaults mechanism to avoid PHP warnings for missing array keys.

Also, this function is defined as `static` but not called statically.

tests/php/test_class_wc-connect-nux.php
```diff
@@ -0,0 +1,109 @@
...
+
+	public static function setupBeforeClass() {
+		require_once( dirname( __FILE__ ) . '/../../classes/class-wc-connect-nux.php' );
+	}
+
+	public function test_get_banner_type_to_display_dev_jp() {
+		$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
+			'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_DEV,
+			'tos_accepted'                           => false,
+			'should_display_after_cxn_banner' => false,
```
nabsul: I don't have a preference, but array's arrows should either all be aligned or all be singled spaced.

classes/class-wc-connect-nux.php
```diff
@@ -285,12 +319,17 @@ public function show_banner_after_connection() {
...
 			if ( isset( $_GET['wcs-nux-notice'] ) && 'dismiss' === $_GET['wcs-nux-notice'] ) {
-				WC_Connect_Options::update_option( self::SUCCESS_BANNER_IS_DISMISSED, true );
+				// No longer need to keep track of whether the before connection banner was displayed.
+				WC_Connect_Options::delete_option( self::SHOULD_SHOW_AFTER_CXN_BANNER );
 				wp_safe_redirect( remove_query_arg( 'wcs-nux-notice' ) );
+				exit;
 			}
 
+			// By going through the connection process, the user has accepted our TOS
+			WC_Connect_Options::update_option( 'tos_accepted', true );
```
jeffstieler: It feels a bit odd that we'd mark the TOS as accepted when displaying a banner "after connection/accept" - but what's the alternative? Can/should we mark the TOS accepted in one of the AJAX handlers for the actual CTA click?

classes/class-wc-connect-nux.php
```diff
@@ -303,6 +342,26 @@ public function show_banner_after_connection() {
...
+				'title'          => __( 'Setup complete! We need you to accept our TOS', 'woocommerce-services' ),
+				'description'    => __( 'Everything is ready to roll, we just need you to agree to our Terms of Service.', 'woocommerce-services' ),
+				'button_text'    => __( 'I accept the TOS!', 'woocommerce-services' ),
+				'button_link'    => add_query_arg( array(
+					'wcs-nux-tos' => 'accept',
+				) ),
+				'image_url'      => plugins_url(
+					'images/nux-printer-laptop-illustration.png', dirname( __FILE__ )
+				),
+				'should_show_jp' => false,
```
jeffstieler: Doesn't have to happen in this PR I suppose, but when we finalize the copy for this notice we should consider an additional parameter like `should_show_terms` so we can show the TOS/privacy links without the JP logo.

classes/class-wc-connect-nux.php
```diff
@@ -98,6 +98,37 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				1. show_banner_before_connection()
+				2. connect to JP
+				3. show_banner_after_connection(), which sets the TOS acceptance in options
+			- Case 2: Jetpack connected, no TOS
+				1. show_tos_only_banner(), which accepts TOS on button click
+			- Case 3: Jetpack connected, and TOS accepted
+				This is an existing user. Do nothing.
+			*/
+			switch ( $status['jetpack_connection_status'] ) {
+				case self::JETPACK_UNINSTALLED:
```
nabsul: How about we name this `JETPACK_NOT_INSTALLED`? Uninstalled makes it sound like the plugin has been actively removed.

tests/php/test_class_wc-connect-nux.php
```diff
@@ -0,0 +1,109 @@
...
+
+	public function test_get_banner_type_to_display_with_jp_cxn_with_tos_acceptance() {
+		// Jetpack is already connected, TOS is already accepted
+		// did not show before connection banner
+		$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
+			'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_CONNECTED,
+			'tos_accepted'                           => true,
+			'should_display_after_cxn_banner' => false,
+		) ), false );
+	}
```
jeffstieler: 🎉  tests!

For future reference, PHPUnit [Data Providers](https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers) can help reduce duplication for tests like this.

classes/class-wc-connect-nux.php
```diff
@@ -98,6 +98,37 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+				case self::JETPACK_INSTALLED:
+				case self::JETPACK_ACTIVATED:
```
nabsul: Similarly, I think several of these constants could be renamed to something that is clearer:
JETPACK_INSTALLED -> JETPACK_NOT_ACTIVATED
JETPACK_ACTIVATED -> JETPACK_NOT_CONNECTED

classes/class-wc-connect-nux.php
```diff
@@ -98,6 +98,37 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+						return 'tos_only_banner';
+					}
+				default:
+					return false;
+			}
+		}
+
 		public function get_jetpack_install_status() {
 			// check if Jetpack is activated
 			if ( ! class_exists( 'Jetpack_Data' ) ) {
```
nabsul: If you first check `validate_plugin()`, you can make this function a little flatter and avoid else-statements:
```
if( 0 !== validate_plugin() ) return JETAPCK_UNINSTALLED;
if ( ! class_exists() ) return JETPACK_INSTALLED;
if( defined() ) return dev
```

classes/class-wc-connect-nux.php
```diff
@@ -186,11 +217,14 @@ public function set_up_nux_notices() {
 			}
 
 			$jetpack_install_status = $this->get_jetpack_install_status();
+			$banner_to_display = $this->get_banner_type_to_display( array(
```
nabsul: What do you think about merging `get_jetpack_install_status()` into `get_banner_type_to_display()`?

I think it'll make the code easier to read:

```
if ( ! class_exists() ) return 'before_jetpack_connection';
```

nabsul: If you do decide to follow up on this route, then all my comments about renaming the constants are void :-)

v18:
@jeffstieler about the warning:

> Notice: Undefined variable: localized_content in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/classes/class-wc-connect-nux.php on line 312

This got introduced in b6c06c4b71715c768225273f7f482b5c718c315c. I think it should be handled in a separate PR - we need to have a default $localized_content array.

classes/class-wc-connect-nux.php
v18: I don't think we should do that because the JP connection can fail, and that would mean that legally they haven't accepted TOS yet, no?

classes/class-wc-connect-nux.php
v18: Yeh, we should do that.

tests/php/test_class_wc-connect-nux.php
v18: Cool! 🎉

classes/class-wc-connect-nux.php
v18: I think I tried doing that before, but some of the other logic didn't quite pan out in a neat way, so I went with this.

classes/class-wc-connect-nux.php
v18: I'd prefer not to merge them since they cover two different things:
1. The Jetpack install status
2. The banner to display based on 3 factors: the JP install status, whether we have displayed the before banner, and whether TOS was accepted.

Doing it the way it's done right splits it up by function, and also allows for neater unit and integration testing.

Happy to rename the constants, though.

classes/class-wc-connect-nux.php
v18: Sure, done. Use `JETPACK_NOT_INSTALLED`, `JETPACK_INSTALLED_NOT_ACTIVATED`, and `JETPACK_ACTIVATED_NOT_CONNECTED`.

classes/class-wc-connect-nux.php
v18: Fixed.

classes/class-wc-connect-nux.php
v18: Done in 0ac28ef8b723f747571f80bfa0b007d069e7046f

tests/php/test_class_wc-connect-nux.php
v18: Done.

classes/class-wc-connect-nux.php
v18: Sure, done.

classes/class-wc-connect-nux.php
v18: Done renaming the constants.

v18:
@jeffstieler  Fixed the `localized_content` issue by adding a new default array in 06632ff1e41f36139389756d97ead8499b16db2d.

v18:
All of the previous comments have been answered.

classes/class-wc-connect-nux.php
nabsul: The reason I suggested merging (or alternatively calling `get_jetpack_install_status()` directly from `get_banner_type_to_display()`) is that the `$status` object you're creating and passing is only used by `get_banner_type_to_display()`. 

It seems to me like you might as well fetch the jetpack state and the two options directly from `get_banner_type_to_display`.

nabsul: Ah, I do see the advantage for unit testing though. Interesting dilema.

nabsul:
To Test:

- [ ] WC & WCS installed but no jetpack
- [ ] WC + WCS + Jetpack deactivated
- [ ] WC + WCS + Jetpack activated but not connected
- [ ] WC + WCS + Jetpack connected but no TOS signed

PR #1039, opened 2017-06-16
Add order page pointer to shipping labels

nabsul:
This PR adds a pointer that highlights the shipping labels metabox. Details of features implemented:

- Added a new pointer to the labels metabox
- Added divs, css and jQuery for dimming everything but the metabox and pointer on the orders page
- Metabox now displays different text depending on if the user has any payment methods
- Added a numPaymentMethods field to redux init data for the above feature
- Implemented MySQL query to detect if the store has already created labels and therefore doesn't need to see the pointer.

Other questions:
- Did I add the CSS in the right place?
- I felt that a direct `show()` was abrupt so I faded the dimming in and out. @olaolusoga ?

To Test:
- For the user you're testing, make sure you've deleted `dismissed_wp_pointers` from usermeta
- Either have a clean site with no printed labels, or delete the `wc_connect_labels` metadata in your postmeta table.
- Open an order to see the pointer
- Note that as long as Jetpack isn't connected, the pointer should not show 
- Note that the pointer shows up after you've successfully connected jetpack
- Try reloading the page and dismissing
- The pointer on the shipping zones page should not do any dimming

### DIFF
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -467,6 +467,18 @@
 	}
 }
 
+.wcs-pointer-page-dimmer {
+	display: none;
+	position: absolute;
+	width: 100%;
+	height: 100%;
+	background-color: black;
+	top: 0;
+	left: 0;
+	z-index: 9998;
+	opacity: 0.5;
+}
+
 // Gridicons
 .gridicon {
 	fill: currentColor;
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -469,12 +469,12 @@
 
 .wcs-pointer-page-dimmer {
 	display: none;
-	position: absolute;
-	width: 100%;
-	height: 100%;
+	position: fixed;
 	background-color: black;
 	top: 0;
+	bottom: 0;
 	left: 0;
+	right: 0;
 	z-index: 9998;
 	opacity: 0.5;
 }
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -48,9 +48,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 
 		private function init_pointers() {
 			add_filter( 'wc_services_pointer_woocommerce_page_wc-settings', array( $this, 'register_add_service_to_zone_pointer' ) );
-			if ( $this->is_new_labels_user() ) {
-				add_filter( 'wc_services_pointer_post.php', array( $this, 'register_order_page_labels_pointer' ) );
-			}
+			add_filter( 'wc_services_pointer_post.php', array( $this, 'register_order_page_labels_pointer' ) );
 		}
 
 		public function show_pointers( $hook ) {
@@ -111,18 +109,21 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 		}
 
 		public function register_order_page_labels_pointer( $pointers ) {
-			$pointers[] = array(
-				'id' => 'wc_services_labels_metabox',
-				'target' => '#woocommerce-order-label',
-				'options' => array(
-					'content' => sprintf( '<h3>%s</h3><p>%s</p>',
-						__( 'Discounted Shipping Labels' ,'woocommerce-services' ),
-						__( 'When you\'re ready, purchase and print discounted labels from USPS right here.', 'woocommerce-services' )
+			if ( $this->is_new_labels_user() ) {
+				$pointers[] = array(
+					'id' => 'wc_services_labels_metabox',
+					'target' => '#woocommerce-order-label',
+					'options' => array(
+						'content' => sprintf( '<h3>%s</h3><p>%s</p>',
+							__( 'Discounted Shipping Labels' ,'woocommerce-services' ),
+							__( 'When you\'re ready, purchase and print discounted labels from USPS right here.', 'woocommerce-services' )
+						),
+						'position' => array( 'edge' => 'right', 'align' => 'left' ),
 					),
-					'position' => array( 'edge' => 'right', 'align' => 'left' ),
-				),
-				'dim' => true,
-			);
+					'dim' => true,
+				);
+			}
+
 			return $pointers;
 		}
 
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -393,10 +393,11 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 								),
 							) ),
 							esc_html( $content['button_text'] ),
-							'https://woocommerce.com/terms-conditions/',
-							'https://woocommerce.com/terms-conditions/services-privacy/'
-						);
-						?></p>
+							'<a href="https://woocommerce.com/terms-conditions/">',
+							'</a>',
+							'<a href="https://woocommerce.com/terms-conditions/services-privacy/"/>',
+							'</a>'
+						); ?></p>
 					<?php endif; ?>
 					<?php if ( isset( $content['button_link'] ) ) : ?>
 						<a
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -12,7 +12,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 		const JETPACK_DEV = 'dev';
 		const JETPACK_CONNECTED = 'connected';
 
-		const SHOW_LABEL_METABOX_POINTER_OPTION = 'show_label_metabox_pointer';
+		const TRANSIENT_IS_NEW_LABEL_USER = 'wcc_is_new_label_user';
 
 		/**
 		 * Option name for dismissing success banner
@@ -102,10 +102,17 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 		}
 
 		private function is_new_labels_user() {
-			global $wpdb;
-			$query = "SELECT meta_key FROM {$wpdb->postmeta} WHERE meta_key = 'wc_connect_labels' LIMIT 1";
-			$results = $wpdb->get_results( $query );
-			return 0 === count( $results );
+			$is_new_user = get_transient( self::TRANSIENT_IS_NEW_LABEL_USER );
+			if ( ! is_string( $is_new_user )) {
+				error_log( 'calculating if the user is new' );
+				global $wpdb;
+				$query = "SELECT meta_key FROM {$wpdb->postmeta} WHERE meta_key = 'wc_connect_labels' LIMIT 1";
+				$results = $wpdb->get_results( $query );
+				$is_new_user  = 0 === count( $results ) ? 'yes' : 'no';
+				set_transient( self::TRANSIENT_IS_NEW_LABEL_USER, $is_new_user );
+			}
+
+			return 'yes' === $is_new_user;
 		}
 
 		public function register_order_page_labels_pointer( $pointers ) {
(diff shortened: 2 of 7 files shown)


classes/class-wc-connect-nux.php
```diff
@@ -315,20 +315,13 @@ public function show_nux_banner( $content ) {
 						<?php echo esc_html( $content['description'] ); ?>
 					</p>
 					<?php if ( $content['should_show_jp'] ) : ?>
-						<p><?php
-						/* translators: %1$s example values include "Install Jetpack and CONNECT >", "Activate Jetpack and CONNECT >", "CONNECT >" */
```
v18: This code seems to be from an errant commit - cc4d442bd8b5bae9fa476a4a1167826043985a62. It doesn't look like it should be in this branch.

classes/class-wc-connect-nux.php
```diff
@@ -93,11 +102,44 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+
+		public function register_order_page_labels_pointer( $pointers ) {
+			if ( $this->is_new_labels_user() && $this->shipping_label->should_show_meta_box() ) {
+				$pointers[] = array(
+					'id' => 'wc_services_labels_metabox',
+					'target' => '#woocommerce-order-label',
+					'options' => array(
+						'content' => sprintf( '<h3>%s</h3><p>%s</p>',
+							__( 'Discounted Shipping Labels' ,'woocommerce-services' ),
+							__( 'When you\'re ready, purchase and print discounted labels from USPS right here.', 'woocommerce-services' )
```
v18: We can use double quote-marks here to [avoid escaping](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#single-and-double-quotes) using `\`.

classes/class-wc-connect-nux.php
```diff
@@ -93,11 +102,44 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
-				)
+				),
 			);
 			return $pointers;
 		}
 
+		private function is_new_labels_user() {
+			$is_new_user = get_transient( self::TRANSIENT_IS_NEW_LABEL_USER );
+			if ( ! is_string( $is_new_user )) {
+				error_log( 'calculating if the user is new' );
```
v18: I'm not sure we need this error log in production.

classes/class-wc-connect-nux.php
```diff
@@ -93,11 +102,44 @@ public function register_add_service_to_zone_pointer( $pointers ) {
 						__( 'To ship products to customers using USPS or Canada Post, you will need to add them as a shipping method to an applicable zone. If you don\'t have any zones, add one first.', 'woocommerce-services' )
 					),
 					'position' => array( 'edge' => 'right', 'align' => 'left' ),
-				)
+				),
 			);
 			return $pointers;
 		}
 
+		private function is_new_labels_user() {
```
v18: I think Jeff mentioned using public functions to facilitate unit testing.

classes/class-wc-connect-nux.php
```diff
@@ -93,11 +102,44 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
+			$is_new_user = get_transient( self::TRANSIENT_IS_NEW_LABEL_USER );
+			if ( ! is_string( $is_new_user )) {
```
v18: Missing a space between brackets here ` $is_new_user ))`

classes/class-wc-connect-nux.php
nabsul: Was just following the example in the function above. I'll fix both.

classes/class-wc-connect-nux.php
nabsul: woops, left over debugging code

classes/class-wc-connect-shipping-label.php
```diff
@@ -362,6 +364,14 @@ protected function get_states_map() {
 		}
 
 		public function should_show_meta_box() {
+			if ( null === $this->show_metabox ) {
```
v18: Why did this separate class property (`$this->show_metabox`) and new function `should_show_meta_box()` get added? From what I can see, a new instance of `WC_Connect_Shipping_Label` is created each time it's called anyway, so seems like it's not needed.

client/admin-pointers.js
```diff
@@ -22,17 +20,29 @@ jQuery( document ).ready( function( $ ) {
...
+
 				show_pointer( pointers, i + 1 );
 			},
 		} );
 
-		$( pointer.target ).pointer( options ).pointer( 'open' );
+		target.pointer( options ).pointer( 'open' );
+		if ( pointer.dim ) {
+			target.css( 'z-index', 9999 );
+			$( 'body' ).append( '<div id="wcs-pointer-page-dimmer" class="wcs-pointer-page-dimmer"></div>' );
```
jeffstieler: What do you think about always rendering the overlay div, but having it's default style be hidden? That way we don't have to manipulated the DOM to add the overlay.

Also - I noticed that the overlay still allows the page to be scrolled behind it. Can we add a class to the body for "dim" pointers that sets `overflow: hidden` to prevent scrolling?

client/admin-pointers.js
```diff
@@ -22,17 +20,29 @@ jQuery( document ).ready( function( $ ) {
...
+
 		const options = $.extend( pointer.options, {
 			close: function() {
 				$.post( ajaxurl, {
 					pointer: pointer.id,
 					action: 'dismiss-wp-pointer',
 				} );
+
+				if ( pointer.dim ) {
+					$( '#wcs-pointer-page-dimmer' ).fadeOut( 500 );
```
jeffstieler: In the effort of getting this landed sooner than later this can likely be done later, but..

Should this check to see if the next pointer is dimmed as well, and skip the fade out if so?

classes/class-wc-connect-shipping-label.php
nabsul: The idea is to avoid running all the code in `should_show_metabox()`. The value is needed twice: Once to decide if the pointer is being shown, and once to decide if the metabox is to be show.

I don't think two instances of `WC_Connect_Shipping_Label`, I'm passing the same one to the constructor of `_NUX` that's used elsewhere...

classes/class-wc-connect-nux.php
```diff
@@ -90,14 +99,46 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
 		}
 
+		public function is_new_labels_user() {
+			$is_new_user = get_transient( self::TRANSIENT_IS_NEW_LABEL_USER );
+			if ( ! is_string( $is_new_user ) ) {
+				global $wpdb;
+				$query = "SELECT meta_key FROM {$wpdb->postmeta} WHERE meta_key = 'wc_connect_labels' LIMIT 1";
+				$results = $wpdb->get_results( $query );
+				$is_new_user  = 0 === count( $results ) ? 'yes' : 'no';
+				set_transient( self::TRANSIENT_IS_NEW_LABEL_USER, $is_new_user );
```
v18: What do you think about short-circuiting the db query if we already got a value of > 1 at least once? If we get a value of > 1 (e.g. `$is_new_user` is 'no'/false etc) then we can dismiss the pointer using code, like this:

`add_user_meta( get_current_user_id(), 'dismissed_wp_pointers', 'wc_services_labels_metabox' );`

And check for the dismissal in `register_order_page_labels_pointer` using:

`get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true )`

That way, even when the transient expires we don't have to check the db again.

assets/stylesheets/style.scss
```diff
@@ -467,6 +467,18 @@
 	}
 }
 
+.wcs-pointer-page-dimmer {
+	display: none;
+	position: absolute;
```
v18: Right now, with `position: absolute;` the background doesn't extend to the bottom of the page if the user scrolls down. You can fix it by using these properties instead:

``` css
position: fixed;
top: 0;
left: 0;
bottom: 0;
right: 0;
```

Works in Chrome, and pretty sure it will work in IE, but needs to be tested there.

classes/class-wc-connect-shipping-label.php
v18: Hmm. Every time I refresh the edit order page, both functions run for me.

classes/class-wc-connect-nux.php
```diff
@@ -90,14 +99,46 @@ public function register_add_service_to_zone_pointer( $pointers ) {
...
 					'position' => array( 'edge' => 'right', 'align' => 'left' ),
-				)
+				),
 			);
 			return $pointers;
 		}
 
+		public function is_new_labels_user() {
+			$is_new_user = get_transient( self::TRANSIENT_IS_NEW_LABEL_USER );
+			if ( ! is_string( $is_new_user ) ) {
```
nabsul: No pref

classes/class-wc-connect-shipping-label.php
nabsul: are you sure `calculate_should_show_meta_box()` is run twice? What should happen is that `should_show_metabox()` and `calculate_should_show_meta_box()` are both run one time, and then the second time only `should_show_metabox` is run.

Note: If might be that the pointer code gets executed before the render of the metabox, so it's possible that our usage here DOES result in `calculate_` getting run, but then when the calculation is skipped in the meta_box render code.

assets/stylesheets/style.scss
nabsul: Works fine in Windows Chrome, Firefox and Edge.

classes/class-wc-connect-nux.php
nabsul: great idea.

client/admin-pointers.js
nabsul: It would probably be cleaner, but actually `show_pointer()` is only executed for the first valid pointer. We don't have to worry about it executing multiple times.

client/admin-pointers.js
nabsul: I added the overlay to the footer if we invoke any pointers. It feels like it executes differently, but it seems to be working. Let me know if you notice any issues with it.

Not sure I understand the thing about scrolling. I don't think disabling it is a very good idea. On a smaller screen the user might actually want to scroll down to the metabox and it's pointer?

v18:
@nabsul hey, I noticed that you marked this PR as ready to merge, but there is still that errant commit in there - 5160d72fd347869bc8b0ba0c6b4234c258da0c7a. Can you remove it? It has content in it that has been changed in the main branch `q2-nux`.

Also - did you change your mind about dismissing the pointer if we get a DB query that indicates the user is not new to labels?

classes/class-wc-connect-shipping-label.php
v18: @nabsul yep, seeing it twice using these steps:

1. Load order page at `wp-admin/post.php?post=123&action=edit`
2. See that both functions are called
3. Re-load the same page
4. See that both functions are called.

From what I can see, this happens because a new instance of `WC_Connect_Shipping_Label` is created each time.

nabsul:
Sorry @v18, I must have accidentally labelled it green instead of yellow last night, and the woke up this morning thinking someone signed off.

I'll follow up on the issues individually.

classes/class-wc-connect-shipping-label.php
v18: Was able to see this to work as expected 👍

nabsul: Oh good! I forgot to follow up on this.

PR #1042, opened 2017-06-21
Add a banner for accepting the TOS

nabsul:
If a user already has Jetpack fully configured when they install WCS, we can't just go straight to the "hurray" page. We still need the user to accept the WCS TOS.

In this PR:

- [x] Add a couple of new states to `get_jetpack_install_status()`
- [x] Add new banner for accepting the TOS
- [ ] If the user goes through the flow of connecting Jetpack we should auto-set `tos_accepted`

@jeffstieler @v18


v18:
I mentioned this on Slack, too, and noting it here for reference. This PR has conflicts with #1038, which is ready for review. This is because #1038 contains the logic for showing all the banners, including the new TOS-only banner.

nabsul:
Not needed any more.

PR #1048, opened 2017-06-22
dimmer div not rendered correctly

### LINKED ISSUE #1047: Q2 nux: fix admin_footer php warning
When I go to an edit order page, before the edit order page pointer has been dismissed I see this warning:

```
PHP Warning:  call_user_func_array() expects parameter 1 to be a valid callback, function '<div id="wcs-pointer-page-dimmer" class="wcs-pointer-page-dimmer"></div>' not found or invalid function name in /Applications/MAMP/htdocs/wp.dev/wp-includes/class-wp-hook.php on line 298
```

I think WordPress is [expecting](https://codex.wordpress.org/Plugin_API/Action_Reference/admin_footer) something like:

``` php
<?php
add_action('admin_footer', 'show_edit_order_page_pointer_dimmer_div');
function show_edit_order_page_pointer_dimmer_div() {
	echo '<div id="wcs-pointer-page-dimmer" class="wcs-pointer-page-dimmer"></div>';
}
?>
```

nabsul:
Fixes #1047 

No, you can't pass a string to a WordPress hook. It must be a function.

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -87,7 +87,6 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 				return;
 			}
 
-			add_action( 'admin_footer', array( $this, 'add_dimming_div' ) );
 			wp_enqueue_style( 'wp-pointer' );
 			wp_localize_script( 'wc_services_admin_pointers', 'wcSevicesAdminPointers', $valid_pointers );
 			wp_enqueue_script( 'wc_services_admin_pointers' );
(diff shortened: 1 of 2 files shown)


v18:
With the current implementation, the dimmer div gets added anytime there are any pointers to show on the page. That should get fixed as well if we're going this route. However...

> The idea being that it's cleaner than post-manipulating the DOM.

@nabsul @jeffstieler The pointers themselves work by manipulating the DOM after it's ready. The process is:
1) The pointers get added to the global js scope in a variable called `wcSevicesAdminPointers` using `wp_localize_script`
2) Each pointer gets shown using our `admin_pointers.js` script, which uses [WordPress's core pointer functions.](https://core.trac.wordpress.org/browser/branches/4.8/src/wp-includes/js/wp-pointer.js)

Since it's all done after the DOM is ready, I'd say it's ok to add a new div as part of this, since that div is effectively part of our 'pointer' in this case.

<i>Comment copied over from https://github.com/Automattic/woocommerce-services/issues/1047#issuecomment-310522598</i>

jeffstieler:
> All the parts related to showing the pointer are together (they're all in JS)

This is a good enough reason to keep it in JS. Sorry for the back and forth @nabsul.

PR #1049, opened 2017-06-22
Add labels metabox to the dismissed pointers list if the user is not new

### LINKED ISSUE #1045: Do not make extra DB call if previous call had >1 labels
We can dismiss the edit order page pointer if we get a DB query that indicates the user is not new to labels using ([reference](https://github.com/Automattic/woocommerce-services/pull/1039#discussion_r123359158)):

`add_user_meta( get_current_user_id(), 'dismissed_wp_pointers', 'wc_services_labels_metabox' );`

nabsul:
Fixes #1045 

This will help avoid checking the database in the future.

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -122,12 +122,24 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 		}
 
 		public function register_order_page_labels_pointer( $pointers ) {
-			$dismissed_pointers = explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) );
+			$user_id = get_current_user_id();
+			$dismissed_data = (string) get_user_meta( $user_id, 'dismissed_wp_pointers', true );
+			$dismissed_pointers = explode( ',', $dismissed_data );
 			if ( $dismissed_pointers && in_array( 'wc_services_labels_metabox', $dismissed_pointers ) ) {
 				return $pointers;
 			}
 
-			if ( $this->is_new_labels_user() && $this->shipping_label->should_show_meta_box() ) {
+			// If the user is not new to labels, we should just dismiss this pointer
+			if ( ! $this->is_new_labels_user() ) {
+				if ( strlen( $dismissed_data ) ) {
+					$dismissed_data .= ',';
+				}
+				$dismissed_data .= 'wc_services_labels_metabox';
+				update_user_meta( $user_id, 'dismissed_wp_pointers', $dismissed_data );
+				return $pointers;
+			}
+
+			if ( $this->shipping_label->should_show_meta_box() ) {
 				$pointers[] = array(
 					'id' => 'wc_services_labels_metabox',
 					'target' => '#woocommerce-order-label',
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -131,7 +131,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 
 		public function register_order_page_labels_pointer( $pointers ) {
 			$dismissed_pointers = $this->get_dismissed_pointers();
-			if ( $dismissed_pointers && in_array( 'wc_services_labels_metabox', $dismissed_pointers ) ) {
+			if ( in_array( 'wc_services_labels_metabox', $dismissed_pointers ) ) {
 				return $pointers;
 			}
 
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -70,12 +70,12 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 				return;
 			}
 
-			$dismissed_pointers = explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) );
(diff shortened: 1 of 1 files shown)


classes/class-wc-connect-nux.php
```diff
@@ -122,12 +122,24 @@ public function is_new_labels_user() {
...
 				return $pointers;
 			}
 
-			if ( $this->is_new_labels_user() && $this->shipping_label->should_show_meta_box() ) {
+			// If the user is not new to labels, we should just dismiss this pointer
+			if ( ! $this->is_new_labels_user() ) {
+				if ( strlen( $dismissed_data ) ) {
+					$dismissed_data .= ',';
+				}
+				$dismissed_data .= 'wc_services_labels_metabox';
```
v18: Getting back a comma delimited string would look a bit neater using `implode` - got that idea directly from [core](https://core.trac.wordpress.org/browser/tags/4.8/src/wp-admin/includes/ajax-actions.php#L2382). They also check if the key has already been added to the dismissed pointers option.

Not sure if it would necessarily be better, but we could also use the core function `wp_ajax_dismiss_wp_pointer` here.

classes/class-wc-connect-nux.php
```diff
@@ -122,12 +122,24 @@ public function is_new_labels_user() {
 		}
 
 		public function register_order_page_labels_pointer( $pointers ) {
-			$dismissed_pointers = explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) );
+			$user_id = get_current_user_id();
+			$dismissed_data = (string) get_user_meta( $user_id, 'dismissed_wp_pointers', true );
+			$dismissed_pointers = explode( ',', $dismissed_data );
 			if ( $dismissed_pointers && in_array( 'wc_services_labels_metabox', $dismissed_pointers ) ) {
```
v18: I think this could use a comment since it's not clear quickly that we're trying to avoid a DB call if a transient expired here.

v18: `in_array` should use a third parameter `true`, for strict comparison

nabsul: Will fix here and in `show_pointers()`

classes/class-wc-connect-nux.php
nabsul: There's a bit of an edge case around when $dismissed_data is empty. In that situation `implode()` gave me a string of `,wc_services_labels_metabox` with a leading comma in the front of the string. Let me see what I can do.

Can't use `wp_ajax_dismiss_wp_pointer` because it's hard coded to read fro `$_POST`

v18: > when $dismissed_data is empty

Was thinking we could use the `$dismissed_pointers` array without string manipulation on `$dismissed_data`. Something like:

```
if ( ! $this->is_new_labels_user() ) {
  $dismissed_pointers[] = 'wc_services_labels_metabox';
  update_user_meta( $user_id, 'dismissed_wp_pointers', implode( $dismissed_pointers ) );
}
```

> it's hard coded to read from $_POST

Was thinking of refreshing the page with a post request, but can see how that can get even more complicated since we'll need to catch that post data somewhere.

classes/class-wc-connect-nux.php
```diff
@@ -93,6 +93,14 @@ public function show_pointers( $hook ) {
 			wp_enqueue_script( 'wc_services_admin_pointers' );
 		}
 
+		public function get_dismissed_pointers() {
+			$data = get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true );
+			if ( is_string( $data ) && 0 < strlen( $data ) ) {
```
v18: We can return this here:

```
explode( ',' get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) );
```

In that case ^, I'm not sure we need a separate function.

Background: the two checks aren't required because`$data` is always a string and `explode($data)` will return an array with a single empty string if `$data` is an empty string. `implode` will return an empty string for both an empty array and an array with a item that is an empty string:

```
implode(array()); // ''
implode(array('')); // ''
```

And `in_array` will work as usual.

classes/class-wc-connect-nux.php
```diff
@@ -122,12 +130,20 @@ public function is_new_labels_user() {
 		}
 
 		public function register_order_page_labels_pointer( $pointers ) {
-			$dismissed_pointers = explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) );
-			if ( $dismissed_pointers && in_array( 'wc_services_labels_metabox', $dismissed_pointers ) ) {
+			$dismissed_pointers = $this->get_dismissed_pointers();
+			if ( in_array( 'wc_services_labels_metabox', $dismissed_pointers ) ) {
```
v18: Needs a 'true' 3rd parameter for a strict equality check.

classes/class-wc-connect-nux.php
nabsul: The problem is that when we later add a new element to the imploded array, we'll get something like this:
```
implode(array('', 'new_dismissed_pointer')); // ',new_dismissed_pointer'
```

v18: 🤦‍♂️ right!

PR #1050, opened 2017-06-22
disable plugin features if no tos

nabsul:
This makes sure we don't do things like enabling our features or talk to the server until the TOS has been accepted.

To Test:
- If you have TOS accepted already you can either disconnect jetpack or remove dev mode and plugin features like the labels metabox should disappear.

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -366,7 +366,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			// if the TOS has not bee accepted or Jetpack is not Connected do nothing else.
 			$jetpack_status = $this->nux->get_jetpack_install_status();
 			$is_jetpack_connected = WC_Connect_Nux::JETPACK_CONNECTED === $jetpack_status || WC_Connect_Nux::JETPACK_DEV === $jetpack_status;
-			if ( true !== WC_Connect_Options::get_option( 'tos_accepted' ) && $is_jetpack_connected ) {
+			if ( true !== WC_Connect_Options::get_option( 'tos_accepted' ) || ! $is_jetpack_connected ) {
 				return;
 			}
 
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -364,10 +364,12 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			add_action( 'admin_init', array( $this, 'admin_enqueue_scripts' ) );
 			add_action( 'admin_init', array( $this->nux, 'set_up_nux_notices' ) );
 
-			// if the TOS has not been accepted or Jetpack is not Connected do nothing else.
+			// Plugin should be enabled if dev mode or connected + TOS
 			$jetpack_status = $this->nux->get_jetpack_install_status();
-			$is_jetpack_connected = WC_Connect_Nux::JETPACK_CONNECTED === $jetpack_status || WC_Connect_Nux::JETPACK_DEV === $jetpack_status;
-			if ( true !== WC_Connect_Options::get_option( 'tos_accepted' ) || ! $is_jetpack_connected ) {
+			$is_jetpack_connected = WC_Connect_Nux::JETPACK_CONNECTED === $jetpack_status;
+			$is_jetpack_dev_mode = WC_Connect_Nux::JETPACK_DEV === $jetpack_status;
+			$tos_accepted = WC_Connect_Options::get_option( 'tos_accepted' );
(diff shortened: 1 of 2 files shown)


nabsul:
Good catch @v18, we also need to allow it to be enabled in DEV mode

woocommerce-services.php
```diff
@@ -363,6 +363,13 @@ public function init() {
 			add_action( 'admin_init', array( $this, 'admin_enqueue_scripts' ) );
 			add_action( 'admin_init', array( $this->nux, 'set_up_nux_notices' ) );
 
+			// if the TOS has not bee accepted or Jetpack is not Connected do nothing else.
```
v18: Super small, but there's a misspelling: 'has not bee' should be 'has not been'

v18:
I ran into an error when uninstalling Jetpack:

```
Call to undefined function validate_plugin()
```

This happens because `$this->nux->get_jetpack_install_status()` is called in `init()`, which is run on the `woocommerce_init` hook.

Instead, we can run the checks using a separate function that runs on the `admin_init` hook:

```
		public function start_services_after_connection() {
			// if the TOS has not been accepted or Jetpack is not Connected do nothing else.
			$jetpack_status = $this->nux->get_jetpack_install_status();
			$is_jetpack_connected = WC_Connect_Nux::JETPACK_CONNECTED === $jetpack_status || WC_Connect_Nux::JETPACK_DEV === $jetpack_status;
			if ( true !== WC_Connect_Options::get_option( 'tos_accepted' ) || ! $is_jetpack_connected ) {
				return;
			}

			$this->schedule_service_schemas_fetch();
			$this->service_settings_store->migrate_legacy_services();
			$this->attach_hooks();
		}
```

```
public function load_dependencies() {
...
add_action( 'admin_init', array( $this, 'start_services_after_connection' ) ); // new line in load_dependencies function
...
}

```

`load_dependencies` is also run in `init()`

@jeffstieler would love to hear your thoughts on hook order here.

jeffstieler:
> @jeffstieler would love to hear your thoughts on hook order here.

If none of the hooks added in `::attach_hooks()` fire before `admin_init`, this should be a safe change to make. 👍

woocommerce-services.php
```diff
@@ -360,10 +360,12 @@ public function load_textdomain() {
 		 */
 		public function init() {
 			$this->load_dependencies();
+
+			add_action( 'admin_init', array( $this, 'load_admin_dependencies' ) );
```
jeffstieler: For completeness, let's move this to `attach_hooks()`, and move the instantiation of `WC_Connect_Help_View` to `load_admin_dependencies()` so we aren't adding the WCS tab to the status page, etc, before TOS acceptance.

nabsul: Oh, thanks for the reminder! I noticed that as well and then forgot about it.

woocommerce-services.php
```diff
@@ -360,9 +360,17 @@ public function load_textdomain() {
 		 */
 		public function init() {
 			$this->load_dependencies();
+
 			add_action( 'admin_init', array( $this, 'admin_enqueue_scripts' ) );
 			add_action( 'admin_init', array( $this->nux, 'set_up_nux_notices' ) );
 
+			// if the TOS has not been accepted or Jetpack is not Connected do nothing else.
+			$jetpack_status = $this->nux->get_jetpack_install_status();
+			$is_jetpack_connected = WC_Connect_Nux::JETPACK_CONNECTED === $jetpack_status || WC_Connect_Nux::JETPACK_DEV === $jetpack_status;
```
v18: This could be separated on two lines for easier readability.

v18:
A question: we'll be displaying the connection banner for users in dev mode. The button on that banner will cause an error because Jetpack will not connect while in DEV mode. Is that ok? I say it's low priority & is fine to ship as-is.

woocommerce-services.php
```diff
@@ -360,9 +360,17 @@ public function load_textdomain() {
...
+			if ( true !== WC_Connect_Options::get_option( 'tos_accepted' ) || ! $is_jetpack_connected ) {
```
jeffstieler: @nabsul This check is failing for me because `tos_accepted` is coming back as a string `1`. Are we now storing the value differently? Do we need to migrate existing settings or use a non-strict check here?

nabsul:
@v18 The way we handle DEV mode should definitely be reviewed, but due to time constraints let's do that at a later stage.

woocommerce-services.php
jeffstieler: Following up.. it looks like we've been using a boolean `true` since the creation of the `tos_accepted` option - my data must be bad.

PR #1051, opened 2017-06-23
Refactor :allthethings: to comply with Calypso's CSS naming rules

DanReyLop:
This PR reshuffles pretty much all our React code to comply with Calypso's `eslint` rules for naming CSS classes. You can see the rule in question here: https://github.com/Automattic/eslint-plugin-wpcalypso/blob/master/docs/rules/jsx-classname-namespace.md

I've used this opportunity to make our components a bit more modular, and bundling the `style.scss` file on each component's folder instead of having a big `stylesheets/style.scss` file.

This PR size is insane, I don't expect for anyone to look at the diff. There has been a lot ot moved files and Git sometimes has troble showing that in a diff. Instead, I recommend checking out the branch and looking at the code structure. All the UI-related code has been affected by this change.

I've changed the "sub-applications" folders structure from this:
```
+ app-name (shipping-label, print-test-label, etc)
  - index.js ("Route" object)
  + views
    - index.js (main view)
    + some-component
      - index.js
    + some-other-component
      - index.js
```
To this:
```
+ app-name (shipping-label, print-test-label, etc)
  - index.js ("Route" object)
  - view.js (main view)
  + components
    + some-component
      - index.js
    + some-other-component
      - index.js
```
Also, every component and "sub-application" has now its own `style.scss` file.

If there is something broken in this PR, it would be visual. Like a CSS class that's mispelled, or WP-Admin styles that now have higher priority. I've been doing a visual inspection of all our UI and I couldn't find anything wrong. I suggest you do the same instead of trying to read all the modified code.

The only outstanding issue is with the `<FoldableCard>` component. We stopped using it (and instead opted for copying its markup) because it had some missing functionallity. I've done the right thing here and submitted a PR to Calypso: https://github.com/Automattic/wp-calypso/pull/15423 So until it's merged (and the dependency bumped) the Shipping Label purchase steps won't automatically expand / collapse unless expliticly clicked (but everything else works fine). It's definitely a blocker, but I don't want to delay getting this PR for review any longer.


client/account-settings/components/label-payment-method/style.scss
```diff
@@ -0,0 +1,44 @@
...
+		margin-right: 14px;
+	}
+}
+
+.label-payment-method__checkbox {
+	margin-right: 20px;
+	float: left;
+}
+
+.payment-logo {
```
v18: Do we need both this rule for `.payment-logo` and the one above it, for `.card.is-compact.label-payment-method .payment-logo`?

client/components/loading-spinner/index.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+import Spinner from 'components/spinner';
+
+const LoadingSpinner = ( { inline } ) => {
+	const spinner = <Spinner size={ 24 } />;
+	if ( inline ) {
+		return spinner;
+	}
+	return (
+		<div className="loading-spinner loading-spinner__container">
```
v18: It looks like the class `loading-spinner` is not needed?

client/components/loading-spinner/index.js
```diff
@@ -0,0 +1,27 @@
...
+ * External dependencies
+ */
+import React, { PropTypes } from 'react';
+
+/**
+ * Internal dependencies
+ */
+import Spinner from 'components/spinner';
+
+const LoadingSpinner = ( { inline } ) => {
```
v18: We can remove the inline loading spinner and use the [busy button component from Calypso](https://wpcalypso.wordpress.com/devdocs/design/button-group).

If we do that, we can remove `<LoadingSpinner />` entirely from our codebase.

v18: Calypso also has its own loading button - [`<SpinnerButton />`](https://wpcalypso.wordpress.com/devdocs/design/spinner-button) - which we can use instead, but it doesn't seem to be working atm.

DanReyLop: Nope. We are using spinners everywhere, not only in buttons. Eventually we'll need to switch to using the cool "pulsating placeholder" Calypso pattern, but that's out of the scope of this PR.

client/components/loading-spinner/index.js
DanReyLop: Correct! Fixed.

client/account-settings/components/label-payment-method/style.scss
DanReyLop: No, not at all. Good catch! Fixed.

client/components/loading-spinner/index.js
v18: I only see two inline `<LoadingSpinner />` - in `PrintLabelDialog` and `StepContainer`. Since `LoadingSpinner` was first introduced here, I thought I'd mention it.

But yeh, can definitely defer to a later PR 👍.

assets/stylesheets/_components.scss
```diff
@@ -66,7 +66,7 @@
 //@import 'components/mobile-back-to-sidebar/style';
 @import 'components/notice/style';
 //@import 'components/olark-chatbox/style';
-//@import 'components/payment-logo/style';
+@import 'components/payment-logo/style';
```
v18: Should we use this opportunity to remove the commented-out components? We can import the Calypso component style manually when we start using a new one.

DanReyLop: Fine by me :)

Done

DanReyLop:
I've merged the `FoldableCard` PR into Calypso `master` and I updated the dependencies in this PR. This should be ready to merge, pending :shipit: of course :)

v18:
I see that there are some components in `client/components` and others in `client/shipping-label/components`. The tree structure looks like this:

screen shot 2017-07-06 at 10 15 45 am

To match Calypso's style for these larger sections, we can remove the `components` folder and place each of the sub-components in the main folder. Here's an example from Calypso:

screen shot 2017-07-06 at 10 20 28 am

Also, what do you think about placing the larger components - `account-settings`, `shipping-label`, `packages`, `print-test-label`, `settings` - under one folder called `blocks` or similar to make it easy to see at a glance that `components` and `blocks` are the only places where folks can find React components.

client/account-settings/index.js
```diff
@@ -2,12 +2,12 @@
  * External dependencies
  */
 import React from 'react';
-import { combineReducers } from 'redux';
+import { combineReducers } from 'state/utils';
```
v18: We don't have a `combineReducers` in `state/utils` - did you mean to import it from Calypso (it's what they use)? I don't think we need to import from Calypso since we're not using any of the state persistence logic, which is what that function was created for.

client/shipping-label/components/label-purchase-modal/index.js
```diff
@@ -9,13 +9,14 @@ import { translate as __ } from 'i18n-calypso';
...
-import PackagesStep from './steps/packages';
-import RatesStep from './steps/rates';
+import AddressStep from './address-step';
+import PackagesStep from './packages-step';
+import RatesStep from './rates-step';
 import Sidebar from './sidebar';
 import { getRatesTotal } from 'shipping-label/state/selectors/rates';
+import FormSectionHeading from 'components/forms/form-section-heading';
 
 const PrintLabelDialog = ( props ) => {
```
v18: Let's rename `PrintLabelDialog` to `PurchaseDialog` to match the others:

```
import PrintLabelDialog from './components/label-purchase-modal';
```

vs

```
import RefundDialog from './components/label-refund-modal';
import ReprintDialog from './components/label-reprint-modal';
```

v18:
@DanReyLop can you rebase this branch on master, there are a couple of changes in there that fixed some bugs, specifically these two: https://github.com/Automattic/woocommerce-services/commit/695dd4361ad04653c1dcc3d7da058117579a03d2, https://github.com/Automattic/woocommerce-services/commit/0b0810f7d1fc5c628a527e26b09d8ae7b4a1ebef

assets/stylesheets/style.scss
```diff
@@ -23,15 +23,11 @@
...
-
 	// Main
 	@import 'main'; // global layout and responsive styles
 
 	// Components
 	@import 'components';
-	@import 'components/payment-logo/style';
 
 	// Breakpoints
 	$breakpoints: 480px, 660px, 780px, 960px, 1040px; // adds wp-admin specific breakpoints
```
v18: Calypso doens't use 780 - let's remove our uses of it. [Calypso's breakpoints](https://github.com/Automattic/wp-calypso/blob/master/assets/stylesheets/shared/mixins/_breakpoints.scss#L6) are:

`$breakpoints: 480px, 660px, 960px, 1040px, 1280px, 1600px;`

client/shipping-label/components/label-purchase-modal/style.scss
```diff
@@ -0,0 +1,94 @@
...
+		flex-basis: 30%;
+	}
+}
+
+.label-purchase-modal__content {
+	display: flex;
+	flex-direction: column;
+	height: 100%;
+
+	.gridicon.is-success {
```
v18: It looks like we don't need the extra specificity here, and can just use `is-success`. Same with `is-warning` below.

client/shipping-label/components/label-purchase-modal/style.scss
```diff
@@ -0,0 +1,94 @@
...
+	50% {
+		opacity: 1.0;
+	}
+	100% {
+		opacity: 0.5;
+	}
+}
+
+.label-purchase-modal__step-title,
+.label-purchase-modal__step-status {
```
v18: From what I can see, these classes (`label-purchase-modal__*`) are being used only in `/client/shipping-label/components/label-purchase-modal/step-container.js` (`StepContainer`). `StepContainer` is getting this CSS because it's being used by the larger LabelPurchaseModal. Can we move `StepContainer` to its own folder with its own CSS file so that it's clearer where its CSS is coming from?

client/shipping-label/components/label-purchase-modal/step-confirmation-button/index.js
```diff
@@ -0,0 +1,30 @@
...
+		</div>
+	);
+};
+
+SetpConfirmationButton.propTypes = {
+	disabled: PropTypes.bool,
+	onClick: PropTypes.func.isRequired,
+};
+
+export default SetpConfirmationButton;
```
v18: StepConfirmationButton is misspelled here

client/components/error-notice/index.js
```diff
@@ -0,0 +1,26 @@
...
+			{ children }
+		</Notice>
+	);
+};
+
+ErrorNotice.propTypes = {
+	isWarning: PropTypes.boolean,
+};
+
+export default ErrorNotice;
```
v18: I see that there are 2 uses of `<ErrorNotice isWarning />`, 1 use of `<ErrorNotice>` and 4 uses of the original `<Notice />` left. Should we either create one WCS-specific notice for all 7 cases, or use the original `<Notice />` in all cases?

client/shipping-label/components/label-purchase-modal/packages-step/style.scss
```diff
@@ -1,51 +1,78 @@
...
+	@include breakpoint( ">480px" ) {
+		top: 10%;
+		right: calc(50% - 200px);
+		bottom: 10%;
+		left: calc(50% - 200px);
+		width: 400px;
+	}
+
+	.packages-step__dialog-package-option {
+		font-weight: normal;
```
v18: Looks like the font-weight and margin changed here (from just `font-weight:bold` in `.wcc-move-item-dialog__package-name`) - what's the reason?

client/shipping-label/components/label-purchase-modal/packages-step/add-item.js
```diff
@@ -29,7 +30,7 @@ const AddItemDialog = ( {
 
 	const packageLabels = getPackageDescriptions( selected, all, true );
 	const getPackageNameElement = ( pckgId ) => {
-		return <span className="wcc-move-item-dialog__package-name">{ packageLabels[ pckgId ] }</span>;
+		return <span className="packages-step__dialog-package-option">{ packageLabels[ pckgId ] }</span>;
```
v18: This should be `packages-step__dialog-package-name`, no?

client/shipping-label/components/label-purchase-modal/packages-step/add-item.js
```diff
@@ -67,24 +68,22 @@ const AddItemDialog = ( {
...
-							}
-						} ) }
-					</p>
-					{ itemOptions }
-				</div>
-				<ActionButtons buttons={ [
-					{ label: __( 'Add' ), isPrimary: true, onClick: () => moveItem( sourcePackageId, movedItemIndex, openedPackageId ) },
-					{ label: __( 'Close' ), onClick: closeAddItem },
-				] } />
+				additionalClassNames="wcc-root packages-step__dialog" >
```
v18: I wonder why `wcc-root` doens't throw an error? Should we also rename it?

client/shipping-label/components/label-purchase-modal/packages-step/package-info.js
```diff
@@ -99,12 +99,12 @@ const PackageInfo = ( {
 
 	const renderPackageSelect = () => {
 		if ( isIndividualPackage ) {
-			return ( <div className="wcc-package__desc">
-				<div className="wcc-package-items-header">
-					<FormLegend className="wcc-package-item__name">{ __( 'Individually Shipped Item' ) }</FormLegend>
+			return ( <div>
+				<div className="packages-step__package-items-header">
+					<FormLegend className="packages-step__package-item-name">{ __( 'Individually Shipped Item' ) }</FormLegend>
```
v18: `packages-step__package-item-name` doesn't seem to have any CSS properties defined. Should we remove the class name?

client/shipping-label/view.js
```diff
@@ -32,38 +32,28 @@ class ShippingLabelRootView extends Component {
...
 
-		this.needToFetchLabelsStatus = true;
-
 		this.state = {
-			showTooltips: props.shippingLabel.labels.map( () => false ),
+			needToFetchLabelsStatus: true,
 		};
 	}
 
-	openTooltip( index ) {
```
v18: Why did we remove these tooltip-related functions?

client/shipping-label/components/label-refund-modal/index.js
```diff
@@ -20,36 +21,34 @@ const RefundDialog = ( { refundDialog, labelActions, storeOptions, created, refu
 		<Modal
 			isVisible={ refundDialog && refundDialog.labelId === label_id }
 			onClose={ labelActions.closeRefundDialog }
-			additionalClassNames="wcc-shipping-label-refund">
-			<div className="wcc-shipping-label-refund__content">
```
v18: why did we remove the class `wcc-shipping-label-refund__content`?

client/shipping-label/views/reprint/index.js
```diff
@@ -1,63 +0,0 @@
...
-import Dropdown from 'components/dropdown';
-import { getPaperSizes } from 'lib/pdf-label-utils';
-
-const ReprintDialog = ( { reprintDialog, labelActions, paperSize, storeOptions } ) => {
-	return (
-		<Modal
-			isVisible={ Boolean( reprintDialog ) }
-			onClose={ labelActions.closeReprintDialog }
-			additionalClassNames="wcc-shipping-label-reprint">
-			<div className="wcc-shipping-label-reprint__content">
```
v18: why did we remove the class `wcc-shipping-label-reprint__content`?

client/shipping-label/view.js
```diff
@@ -187,54 +177,40 @@ class ShippingLabelRootView extends Component {
...
-			<span>
-				<span className="wcc-metabox-label-item__detail"
-						onMouseEnter={ onMouseEnter }
-						onMouseLeave={ onMouseLeave }
-						ref={ 'label-details-' + index }>
-					{ __( 'Label #%(labelNum)s', { args: { labelNum } } ) }
-				</span>
-				<Tooltip
-					className="wc-connect-popover"
-					isVisible={ this.state.showTooltips[ index ] }
```
v18: Why was `isVisible={ this.state.showTooltips[ index ] }` removed? In `<InfoTooltip />`, `isVisible` is always set to `true`.

client/components/info-tooltip/style.scss
```diff
@@ -2,3 +2,33 @@
...
+
+	.info-tooltip__contents {
+		padding: 4px;
+		color: $white;
+
+		h1, h2, h3 {
+			margin: 0 0 8px 0;
+			color: $white;
+			font-weight: 600;
+			font-size: 1.3em;
```
v18: why were `font-weight: 600;` and `font-weight: 600;` added here from `info-tooltip__contents`?

assets/stylesheets/style.scss
```diff
@@ -23,15 +23,11 @@
 	//@import 'shared/welcome'; // welcome messages
 	//@import 'shared/infinite-scroll-end'; // Last page marker once infinite scroll has reached end
 
-	@import 'shipping-services'; // Shipping Services Groups, Group and Entry Component styles
-	@import 'shipping-label'; // Shipping Labels specific styles
-
```
v18: Why were these two imports removed?

client/account-settings/components/label-payment-method/style.scss
```diff
@@ -0,0 +1,38 @@
+.card.is-compact.label-payment-method {
```
v18: Why do we need both `.card` and `.is-compact` here? Previously we were only using `.card`.

client/account-settings/components/label-payment-method/style.scss
```diff
@@ -0,0 +1,38 @@
...
+	margin-bottom: 14px;
+	display: flex;
+	flex-direction: row;
+	align-items: center;
+	cursor: pointer;
+
+	.payment-logo {
+		float: left;
+		margin-top: 0;
```
v18: Why did margin top change to 0 here?

client/account-settings/view.js
```diff
@@ -37,9 +38,9 @@ const AccountSettingsRootView = ( { formData, formMeta, actions, noticeActions,
 	const renderContent = () => {
 		if ( ! formData && ! formMeta.isFetching ) {
 			return (
-				<p className="error-message">
+				<ErrorMessage>
```
v18: We only have 2 `<ErrorMessage />`s - maybe it's not worth creating a new component for it?

v18:
@DanReyLop - I finished reviewing a large chunk so far. Feel free to reply to comments, or leave it to when I'm done the whole PR. I'll continue reviewing more chunks tomorrow.

client/components/indicators/index.js
```diff
@@ -19,20 +19,20 @@ const renderLastUpdated = ( lastUpdated ) => {
 	}
 
 	return (
-		<div className="form-setting-explanation indicator__last-updated">
+		<div className="form-setting-explanation indicators__indicator-last-updated">
```
v18: Why do we need `form-setting-explanation` here? Why doesn't it show a linting error?

client/components/indicators/style.scss
```diff
@@ -16,38 +40,14 @@
 	}
 }
 
-.form-setting-explanation.indicator__last-updated {
+.form-setting-explanation.indicators__indicator-last-updated {
```
v18: Why is the extra `.form-setting-explanation` needed here?

client/components/info-tooltip/index.js
```diff
@@ -45,21 +46,23 @@ export default class InfoTooltip extends Component {
...
 			<span
 				onMouseEnter={ this.openTooltip }
 				onMouseLeave={ this.closeTooltip }
 				className={ classNames( 'info-tooltip', this.props.className ) } >
-				<Gridicon ref="icon" icon="info-outline" size={ 18 } />
+				<span ref="anchor">{ anchor }</span>
 				{ this.state.showTooltip &&
 					<Tooltip
-						className="wc-connect-popover"
+						className="info-tooltip__container wcc-root"
```
v18: why do we need `wcc-root`?

client/components/info-tooltip/style.scss
```diff
@@ -2,3 +2,33 @@
 	cursor: help;
 	color: $gray-dark;
 }
+
+&.tooltip.popover.info-tooltip__container {
+	.popover__inner {
```
v18: The css for `info-tooltip__container` seems new - did we need new CSS, or did this come from somewhere else?

client/components/info-tooltip/style.scss
```diff
@@ -2,3 +2,33 @@
...
+		background: $gray-dark;
+	}
+
+	.popover__arrow {
+		border-top-color: $gray-dark;
+	}
+
+	.info-tooltip__contents {
```
v18: do we need to nest this?

client/components/nux-notice/index.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * External dependencies
+ */
+import React, { PropTypes } from 'react';
+import Gridicon from 'gridicons';
+import { translate as __ } from 'i18n-calypso';
+
+const NuxNotice = ( { children, onDismiss } ) => {
```
v18: We use this just once. Maybe it's not worth its own component? Probably you wanted to separate the dismissal logic? Can we use a Calypso component like `<Notice />` for that?

client/components/nux-notice/index.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * External dependencies
+ */
+import React, { PropTypes } from 'react';
```
v18: Getting `PropTypes` from react directly will be deprecated in React 16. For new components, we should start using the prop-types package from npm.

client/components/nux-notice/index.js
v18: Oh, now I see the dismissal function comes from the parent component anyway. Yeh, I definitely think we should move this pack to the parent component.

client/components/settings-group-card/index.js
```diff
@@ -0,0 +1,28 @@
...
+import React, { PropTypes } from 'react';
+import classnames from 'classnames';
+
+/**
+ * Internal dependencies
+ */
+import CompactCard from 'components/card/compact';
+import FormSectionHeading from 'components/forms/form-section-heading';
+
+const SettingsGroupCard = ( { heading, children } ) => {
```
v18: I'm not sure a separate component is worth it here. It's used only 5 times, and makes it harder to understand what's going on in the components that use it.

client/packages/components/add-package/index.js
```diff
@@ -175,34 +175,32 @@ const AddPackageDialog = ( props ) => {
...
-				<div className="wcc-shipping-add-package-weight">
+			<FormFieldset className="add-package__weight-group">
+				<div className="add-package__weight">
 					<FormLabel htmlFor="box_weight">{ __( 'Package weight' ) }</FormLabel>
 					<FormTextInput
 						id="box_weight"
 						name="box_weight"
 						placeholder={ __( 'Weight of box' ) }
 						value={ box_weight || '' }
-						className={ fieldClassName }
```
v18: Why do we no longer need the class here? For reference, here's the definition:

```
const fieldClassName = is_user_defined ? null : 'flat-rate-package__inner-dimensions__read-only';
```

client/packages/components/add-package/style.scss
```diff
@@ -0,0 +1,81 @@
+&.dialog.card.add-package {
+	@include breakpoint( ">480px" ) {
```
v18: where did this CSS get copied from - `&.dialog.card.add-package`?

client/packages/components/packages-list/style.scss
```diff
@@ -0,0 +1,107 @@
...
+	display: flex;
+	height: 26px;
+	border-bottom: 1px solid lighten( $gray, 30% );
+}
+
+.packages-list__item,
+.packages-list__empty {
+	margin-top: 8px;
+	display: flex;
+	//align-items: center;
```
v18: Can we remove the comment?

client/print-test-label/index.js
```diff
@@ -25,6 +25,10 @@ export default ( { paperSize, storeOptions } ) => ( {
 		};
 	},
 
+	getStateKey() {
```
v18: What is this used for?

client/settings/components/shipping-services/group.js
```diff
@@ -56,7 +49,7 @@ const ShippingServiceGroup = ( props ) => {
 	const onChange = ( event ) => updateAll( event, updateValue, services );
 
 	const renderHeader = () => {
-		return <div className="wcc-shipping-services-group-header">
+		return <div className="shipping-services__group-header">
```
v18: Do we need `shipping-services__group-header`? - I don't see any CSS rules for it in the code.

client/settings/components/shipping-services/group.js
```diff
@@ -75,16 +68,15 @@ const ShippingServiceGroup = ( props ) => {
...
 			compact
-			actionButton={ actionButton }
-			actionButtonExpanded={ actionButton }
+			screenReaderText={ __( 'Expand Services' ) }
 			expanded={ ! _.isEmpty( errors ) }
 		>
-			<div className="wcc-shipping-service-entry multi-select-header">
-				<span className="wcc-shipping-service-header service-name">{ __( 'Service' ) }</span>
-				<span className="wcc-shipping-service-header price-adjustment">
+			<div className="shipping-services__entry multi-select-header">
```
v18: Is `multi-select-header` allowed?

client/settings/view.js
```diff
@@ -6,15 +6,15 @@ import React, { PropTypes } from 'react';
 /**
  * Internal dependencies
  */
-import WCCSettingsForm from './wcc-settings-form';
```
v18: Let's also rename `WCCSettingsForm` in `/client/settings/components/settings-form/index.js`.

client/account-settings/index.js
DanReyLop: Yes, it's imported from Calypso (I forgot to move it to `Internal dependencies`). If you call it without an `schema` argument, it's basically equivalent to the `redux` method, so there's no reason to *not* use it.
Moved those imports to `Internal imports` in https://github.com/Automattic/woocommerce-services/pull/1051/commits/fa305cab23ed07de90ba0b01eeeb90428f1b18d0

client/shipping-label/components/label-purchase-modal/index.js
DanReyLop: Fixed in https://github.com/Automattic/woocommerce-services/pull/1051/commits/9e7d3bc41612be97fdab88fab7e181de448a06bf

assets/stylesheets/style.scss
DanReyLop: Good catch! I've removed the 780px breakpoint in https://github.com/Automattic/woocommerce-services/pull/1051/commits/e028d3afc2469ef861e3cbfa4fa2253319207a3e

client/shipping-label/components/label-purchase-modal/style.scss
DanReyLop: We do. We apply the `is-whatever` class to both the `<Gridicon>` and to the step summary. We want `error` summaries to be red, but we don't want them changing color in the other cases (yellow is hard to read, and green isn't needed for the "normal" case of a step that is correct).

client/shipping-label/components/label-purchase-modal/step-confirmation-button/index.js
DanReyLop: LOL good catch, fixed in https://github.com/Automattic/woocommerce-services/pull/1051/commits/ee4d40e101ccf3030d14002090f26904d4e42288

client/components/error-notice/index.js
DanReyLop: `<ErrorNotice>` has some custom CSS, that's why I separated it into a different component.

client/shipping-label/components/label-purchase-modal/packages-step/add-item.js
DanReyLop: ESLint only checks for `className` props. This is `additionalClassNames`. I'm not gonna complain, we *need* the `wcc-root` class here, it's the root class for *all* our styles.

client/shipping-label/components/label-purchase-modal/packages-step/add-item.js
DanReyLop: Yep, fixed in https://github.com/Automattic/woocommerce-services/pull/1051/commits/c147468c7842f5ff171592b985e70e5fa674fcf7

client/shipping-label/components/label-purchase-modal/packages-step/package-info.js
DanReyLop: Removed in https://github.com/Automattic/woocommerce-services/pull/1051/commits/9bfc8324520cedbc3983ed46f7905a0cf4be379f

client/shipping-label/components/label-refund-modal/index.js
DanReyLop: The child rules it had, I put them directly under the `wcc-shipping-label-refund` rule. Same for the Reprint dialog.

client/components/info-tooltip/index.js
DanReyLop: Because the tooltip is created in a different `<div>` separated of the rest of the React UI (think of it as a mini-Modal). So we need the root rule to apply styles to it.

client/shipping-label/view.js
DanReyLop: `<Tooltip>` is a "dumber" component and it needs the `isVisible` prop, but `<InfoTooltip>` (which is in this repo) uses `<Tooltip>` and has internal state. I figured it was cleaner for each individual tooltip to have its own `visible/notVisible` state, instead of the Labels list controlling all the tooltips.

client/components/info-tooltip/style.scss
DanReyLop: `info-tooltip__container` was previously called `wc-connect-popover`, I moved the styles from [the main style file](https://github.com/Automattic/woocommerce-services/pull/1051/files#diff-71e5f6f03e3a279beddfc328664288eaL433) to this, component-specific file.

client/components/info-tooltip/style.scss
DanReyLop: Yes, for specificity. I don't remember exactly why, but if I didn't enclose every rule as a child of `&.tooltip.popover.info-tooltip__container`, some styles weren't applied.

client/components/info-tooltip/style.scss
DanReyLop: OK I remember now. Before, some `WP-Admin` rules were being accidentally applied, and once I refactored, they weren't applied anymore and messed up the tooltip content styles. I added them back explicitly to preserve the exact same look.

assets/stylesheets/style.scss
DanReyLop: Because these files were removed. Instead, I added imports for every new component and "sub-application" at the end of the file ([here](https://github.com/Automattic/woocommerce-services/pull/1051/files/995d9174509fc081ffa6a5d130b4007088fedcaf#diff-71e5f6f03e3a279beddfc328664288eaR282))

client/account-settings/components/label-payment-method/style.scss
DanReyLop: To bead specificity. The Calypso rules for `<Card>` include: `.card.is-compact {}`, so if we used `.card.label-payment-method {}` ours would have the same specificity and it would come down to which CSS file was included first. I prefer to take the more robust approach here.

client/account-settings/components/label-payment-method/style.scss
DanReyLop: It was already set `margin-top: 0;`, just in a different rule: https://github.com/Automattic/woocommerce-services/pull/1051/files/995d9174509fc081ffa6a5d130b4007088fedcaf#diff-a3c69445a234ef3dfa87f5b244ac40acL22

client/account-settings/view.js
DanReyLop: Maybe it's a bit overkill now, but there's no harm in having a component that's only been used twice. It'll probably have more uses in the future though.

client/components/indicators/index.js
DanReyLop: No idea why it doesn't throw an error. I've changed this to use the `<FormSettingsExplanation>` component instead of akwardly reusing its CSS classes. In https://github.com/Automattic/woocommerce-services/pull/1051/commits/440da372c3efc3f01b6dcca4d3327c8f833b4651

client/components/nux-notice/index.js
DanReyLop: It's only used once and it's a small component, but it has substantial custom styles (see its associated `style.scss` file). It's about separation of concerns, I don't think those styles should live in the parent component.

client/components/nux-notice/index.js
DanReyLop: We can codemod it all at once when we upgrade to 16. `wp-calypso` will face the same transition so we'll be able to borrow its migration script from them.

client/components/settings-group-card/index.js
DanReyLop: This one too, this is not about the 5-lines JS component, its about all the style associated with it (look at its `style.scss` file). Would it help if we change its name? `SettingsGroupCard` is the best I could come up with.

client/packages/components/add-package/index.js
DanReyLop: That additional class name was only added so the `:disabled` styles would be applied. I've moved those styles [here](https://github.com/Automattic/woocommerce-services/pull/1051/files/995d9174509fc081ffa6a5d130b4007088fedcaf#diff-615f3d2f3212d2af594e2cac359817f7R22) instead, and since the `<FormTextInput>` is set to `disabled=true` when the package isn't user-defined, the additional class name is no longer needed.

client/packages/components/add-package/style.scss
DanReyLop: Here: https://github.com/Automattic/woocommerce-services/pull/1051/files/995d9174509fc081ffa6a5d130b4007088fedcaf#diff-71e5f6f03e3a279beddfc328664288eaL190

client/packages/components/packages-list/style.scss
DanReyLop: Removed in https://github.com/Automattic/woocommerce-services/pull/1051/commits/6028de9a090e8076d9c18dc57db9162f48ec4b7a

client/print-test-label/index.js
DanReyLop: It's used in `client/main.js`, to persist the Redux state in storage if the user is forced to do a reload. Not really needed for something as simple as the "Print test label" form, but it's required or the app 💥

client/settings/components/shipping-services/group.js
DanReyLop: Looks like we don't :)
Removed in https://github.com/Automattic/woocommerce-services/pull/1051/commits/22d672f42b114682bf1c1ff382c0859fccd27cd4

client/settings/components/shipping-services/group.js
DanReyLop: Technically no, but ESLint overlooked it because there's a valid class. Or else, things like `something__valid is-error` wouldn't be valid. But I've changed it anyways in https://github.com/Automattic/woocommerce-services/pull/1051/commits/eaf2e2c144cb489f3495731efbe6b11f4d4072c3

client/settings/view.js
DanReyLop: Done in https://github.com/Automattic/woocommerce-services/pull/1051/commits/75d60873b8b89176e6b58157b63f315678670eea

DanReyLop:
> To match Calypso's style for these larger sections, we can remove the components folder and place each of the sub-components in the main folder. [...] Also, what do you think about placing the larger components - account-settings, shipping-label, packages, print-test-label, settings - under one folder called blocks or similar to make it easy to see at a glance that components and blocks are the only places where folks can find React components.

The thing is, our "blocks" are far more sophisticated than what the name `blocks` leds to believe. Each one is, essentially, a self-sufficient app. The fact that they are all part of the same bundle is purely technical (a single large bundle is much smaller than several, slightly smaller bundles). Each "block" has its own ser of reducers, state initializer, actions, and yes, React components. I wouldn't want to have to look into 2 completely separated branches of the directory tree to "see what this `account-settings` thing is".

It's the old debate: separate per feature, or separate per "layer"? I personally prefer it the way it is, but I won't fight you too much.

I agree on putting all the "apps" in a single folder though. How should we call it? `blocks` would be confusing because it has a different meaning in Calypso.

client/account-settings/view.js
v18: Alright :)

client/components/settings-group-card/index.js
v18: Yeh, a name change would help here, I think.

DanReyLop:
@v18 Awesome review, thanks for the work! We need to merge https://github.com/Automattic/wp-calypso/pull/16419 before this code can fully work again, after that I'll do another full pass of visual testing in case I messed up something in the merge or the PR fixes.

v18:
> I agree on putting all the "apps" in a single folder though.

That's the main thing I wanted 👍 , so let's do that. `apps` seems ok to me

v18:
> after that I'll do another full pass of visual testing in case I messed up something in the merge or the PR fixes.

Cool. I have the previous branch saved so I can do a review of all the parts that were changed.

I think it would be worthwhile to assign each person in Hydra (cc @jeffstieler) to test out different sections of our code manually, so that we catch any UI errors before merging. Can you make a list like that to use after we get ready to merge?

DanReyLop:
> That's the main thing I wanted 👍 , so let's do that. apps seems ok to me

Done in https://github.com/Automattic/woocommerce-services/pull/1051/commits/8e7fb47c1ec03fae275460884d6b00906dd6dd62

jeffstieler:
I ran through all of the UI and only found 2 issues regarding tooltips.

Both the Price adjustment info icon and the Label number throw JS errors when hovered over.

screen shot 2017-07-24 at 4 09 43 pm

screen shot 2017-07-24 at 4 09 51 pm

DanReyLop:
@jeffstieler Yeah I noticed that too... looks like a recent change in Calypso broke it. I've created https://github.com/Automattic/wp-calypso/pull/16531 to fix it.

The Calypso `<Tooltip>` component uses a `isRTL()` selector, which indirectly uses the `getCurrentUser()` selector, which breaks in `woocommerce-services` because our `state` doesn't have the same structure as Calypso's.

DanReyLop:
I've fixed the `Tooltip` issue. I've went through these parts of the UI:

Order page:
* Labels list. Tooltip works, tracking links open in new tab.
* Reprint dialog works (even changing paper size).
* Labels list UI correctly resizes when making the screen narrower (i.e. mobile styles).
* Refund dialog works.
* Purchase label flow opens and shows correctly.
* Change origin address to something invalid, click `Use this address`, the step must show an error.
* Change origin address to something that can be auto-corrected, the "pick the correct address" UI with the yellow notice must show.
* Packages step looks fine, including all the visual fixes @dechov recently pushed.
* Rates step looks fine, shows an error if there's nothing selected.
* While purchasing, you can't dismiss the dialog.
* The purchase completes as expected.

Package settings:
* Open an existing package works. Changes save correctly. The `Outer dimensions` link works.
* Add a package works.
* Selecting / unselecting predefined packages works, and so does bulk selecting/unselecting.
* Saving the changes works.

Label settings:
* Changes are saved correctly, UI looks fine.
* The `add another cc` button takes you to `wordpress.com/me/billing` in a new tab.

Self-help:
* Can enable/disable the logging.
* Everything looks fine ("Health" indicators, external links, debug log).
* Can print a test label, even changing the paper size.

USPS settings:
* If ZIP is empty an error appears on saving.
* On the services list, the `(i)` tooltip works.
* Bulk select/unselect services, edit adjustments.
* Saving works.

Canada post settings:
* Same tests as the USPS settings.

jeffstieler:
Visually, looks good. I really like the new directory structure as well. 👍 

I defer :shipit: to @v18 who has reviewed the code changes.

client/account-settings/index.js
v18: @DanReyLop I'm still not sold on using Calypso's `combineReducers` - why do we need it? Calypso has it because they persist almost all of their state, but we persist almost none iirc.

> there's no reason to not use it.

What's the reason _to_ use it, though?

client/shipping-label/views/reprint/index.js
v18: answered [here](https://github.com/Automattic/woocommerce-services/pull/1051#discussion_r128623230)

client/components/settings-group-card/index.js
v18: Not sure about a new name. Let's leave it as is, unless @jeffstieler has a good idea before this is ready to merge.

client/settings/view.js
v18: @DanReyLop btw, still seeing one use of `WCCSettingsGroup` leftover.

client/apps/account-settings/components/label-payment-method/index.js
```diff
@@ -32,18 +32,18 @@ const PaymentMethod = ( { translate, selected, type, digits, name, expiry, onSel
 	};
 
 	return (
-		<CompactCard className="shipping__card" onClick={ onSelect }>
+		<CompactCard className="label-payment-method" onClick={ onSelect }>
```
v18: should this file be renamed just `.../payment-method/index.js`? The component name is `<PaymentMethod />`.

v18:
I've gone over the comments, only a few things left:
- `combineReducers` from redux or Calypso
- one use of WCCSettingsGroup left
- label payment method vs payment method ([#](https://github.com/Automattic/woocommerce-services/pull/1051/files#r129971930))

client/account-settings/index.js
DanReyLop: > What's the reason to use it, though?

Because `Calypso` has an `eslint` rule to fobbid using the regular Redux `combineReducers`. It's either use the `Calypso` one or add a bunch of `eslint` exceptions. Not a big deal either way, I just chose to comply instead of rebel :)

client/settings/view.js
DanReyLop: Fixed in https://github.com/Automattic/woocommerce-services/pull/1051/commits/1a7ee9b4d3d4dbf29363272d85e8b4722fbc50f9

client/apps/account-settings/components/label-payment-method/index.js
DanReyLop: I'm OK with that, there's nothing specific about labels in that component. Renamed in https://github.com/Automattic/woocommerce-services/pull/1051/commits/3a9ba3e58d8b5f9b901fcbce8ffb63df960c7e70

client/account-settings/index.js
v18: Ah, I see :P. Ok, let's follow the rule.

client/components/settings-group-card/index.js
jeffstieler: `SettingsGroupCard` makes enough sense to me.

v18:
@DanReyLop wanted to mention that I've started reviewing again. A lot more changes took place than I expected, so finishing will take longer than expected. So far so good. 

After the final 👍 we should:
1. Freeze any more code changes so that there are no more conflicts between this branch and master
2. Ask for another manual review from someone else the team
3. If all is good, ship

client/apps/packages/components/packages-list/index.js
```diff
@@ -60,10 +60,14 @@ const PackagesList = ( { packages, dimensionUnit, editable, selected, serviceId,
 	return (
 		<FormFieldset className="packages-list">
 			<div className="packages-list__header">
-				<FormLegend className="packages-list__column-actions" />
+				{ editable ? null : <FormLegend className="packages-list__column-actions" /> }
 				<FormLegend className="packages-list__column-type">{ __( 'Type' ) }</FormLegend>
 				<FormLegend className="packages-list__column-name">{ __( 'Name' ) }</FormLegend>
-				<FormLegend className="packages-list__column-dimensions">{ __( 'Dimensions (L x W x H)' ) }</FormLegend>
+				<FormLegend className="packages-list__column-dimensions">
+					{ __( 'Dimensions' ) }{ ' ' }
```
v18: From what I can see, this doesn't correspond to the most recent version:

https://github.com/Automattic/woocommerce-services/blame/master/client/packages/views/packages-list.js#L71

For example `{ __( 'Dimensions' ) }` was in its own span.

client/apps/packages/components/packages-list/index.js
```diff
@@ -60,10 +60,14 @@ const PackagesList = ( { packages, dimensionUnit, editable, selected, serviceId,
...
+					<span className="packages-list__column-dimensions-format">{ __( '(L x W x H)' ) }</span>
+				</FormLegend>
+				{ editable ? <FormLegend className="packages-list__column-actions" /> : null }
```
v18: Maybe better as `editable && <FormLegend className="packages-list__column-actions" /`? Don't need `null` here.

client/apps/shipping-label/view.js
```diff
@@ -239,8 +242,9 @@ class ShippingLabelRootView extends Component {
 	}
 
 	renderLabels() {
-		if ( ! _.every( this.props.shippingLabel.labels, 'statusUpdated' ) ) {
-			return <LoadingSpinner />;
+		if ( this.needToFetchLabelsStatus ) {
```
v18: This is a pretty curious change. From what I can see this if statement was introduced ~8 months ago in c21471a2b76dc200d0e947a15356cab9a57c5727 - do you know why it ended up in a merge conflict here?

(No need to solve this before merging, but if you know I'm curious :) )

v18:
Reviewed all the new changes. Only a things left to ship:
1. https://github.com/Automattic/woocommerce-services/pull/1051#pullrequestreview-54172238
2. https://github.com/Automattic/woocommerce-services/pull/1051#pullrequestreview-54172501
3. Merge master in again to remove conflicts - I can review that
4. Manual tests

client/apps/shipping-label/view.js
DanReyLop: The change in `master` was removing the `<Spinner>`. The correct action here is removing this conditional, since fetching the labels status is handled in `componentWillMount`. Fixed in https://github.com/Automattic/woocommerce-services/pull/1051/commits/8877cf3ceb332e56afcbdd6258d7b8f7a3b61d4f

client/apps/packages/components/packages-list/index.js
DanReyLop: Changed in https://github.com/Automattic/woocommerce-services/pull/1051/commits/ba85a116933085715395a16cd8618cc7d715c7b4

client/apps/packages/components/packages-list/index.js
DanReyLop: It was inside its own `<span>` but that `<span>` had no associated CSS, so it was pointless.

v18:
🚢!!

Let me know if you want to do manual tests before or after you merge. Given that both you and @jeffstieler manually tested this branch, I think we can do a pass before cutting a release. Happy to do it either way, though.

PR #1061, opened 2017-06-27
Fix tracking of TOS acceptance

### LINKED ISSUE #1057: Add tracks event back in after TOS acceptance
This was removed in #1036 in https://github.com/Automattic/woocommerce-services/pull/1036/files#diff-57ae8afb227e356d71d37582f678a831L769

jeffstieler:
Fixes #1057.

This PR seeks to reestablish tracking of TOS acceptance. In addition to that goal, it:
* Adds an optional "source" parameter to the `opted_in()` helper method
* Attaches shipping method related tracks hooks explicitly, rather than on instantiation
* Tracks TOS accept from "TOS only" banner and "connect" banner separately

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -21,12 +21,19 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 		const SHOULD_SHOW_AFTER_CXN_BANNER = 'should_display_nux_after_jp_cxn_banner';
 
 		/**
+		 * @var WC_Connect_Tracks
+		 */
+		protected $tracks;
+
+		/**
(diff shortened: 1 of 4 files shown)


v18:
Can we also remove `load_tracks_for_activation_hooks()` with these changes in https://github.com/Automattic/woocommerce-services/blob/59de8c638d1a9b258afbc65a8dc33174e50afed5/woocommerce-services.php#L161 and https://github.com/Automattic/woocommerce-services/blob/59de8c638d1a9b258afbc65a8dc33174e50afed5/woocommerce-services.php#L172 ?

v18:
Looks good to go except for the above comment.

PR #1062, opened 2017-06-27
Restrict TOS acceptance to Jetpack master user

### LINKED ISSUE #1058: Check for JP master user to allow user to connect
Looks like we need to check the owner of the Jetpack connection before letting the user accept the TOS, as per https://github.com/Automattic/woocommerce-services/pull/1036/files#diff-57ae8afb227e356d71d37582f678a831L793

jeffstieler:
Fixes #1058.

To test:
* Make sure TOS isn't accepted
* Log in with Jetpack master user
* Verify TOS banner displays
* Log in with different user (with `manage_woocommerce`, `install_plugins`, `activate_plugins` caps)
* Verify no TOS banner
* `define( 'JETPACK_DEV_DEBUG', true );`
* Verify TOS banner displays

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -207,6 +207,11 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 					return 'before_jetpack_connection';
 				case self::JETPACK_CONNECTED:
 				case self::JETPACK_DEV:
+					// Has the user just gone through our NUX connection flow?
+					if ( isset( $status['should_display_after_cxn_banner'] ) && $status['should_display_after_cxn_banner'] ) {
+						return 'after_jetpack_connection';
+					}
+
 					// Has the user already accepted our TOS? Then do nothing.
 					// Note: TOS is accepted during the after_connection banner
 					if (
@@ -216,11 +221,6 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 						&& $status['can_accept_tos']
 					) {
 						return 'tos_only_banner';
-					}
-
-					// Has the user just gone through our NUX connection flow?
-					if ( isset( $status['should_display_after_cxn_banner'] ) && $status['should_display_after_cxn_banner'] ) {
-						return 'after_jetpack_connection';
 					}
 
 					return false;
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -156,7 +156,28 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 			return $pointers;
 		}
 
-		public static function get_banner_type_to_display( $status = array() ) {
+		/**
+		 * Check that the current user is the owner of the Jetpack connection
+		 * - Only that person can accept the TOS
+		 *
+		 * @return bool
+		 */
+		public function can_accept_tos() {
+			// Developer case
+			if ( defined( 'JETPACK_DEV_DEBUG' ) && JETPACK_DEV_DEBUG ) {
+				return true;
+			}
+
+			$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
+			$can_accept = (
+				isset( $user_token->external_user_id ) &&
+				get_current_user_id() === $user_token->external_user_id
+			);
+
+			return $can_accept;
+		}
+
+		public function get_banner_type_to_display( $status = array() ) {
 			if ( ! isset( $status['jetpack_connection_status'] ) ) {
 				return false;
 			}
@@ -177,13 +198,19 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 				case self::JETPACK_ACTIVATED_NOT_CONNECTED:
 					return 'before_jetpack_connection';
 				case self::JETPACK_CONNECTED:
+				case self::JETPACK_DEV:
 					// Has the user just gone through our NUX connection flow?
 					if ( isset( $status['should_display_after_cxn_banner'] ) && $status['should_display_after_cxn_banner'] ) {
 						return 'after_jetpack_connection';
 					}
+
 					// Has the user already accepted our TOS? Then do nothing.
 					// Note: TOS is accepted during the after_connection banner
-					if ( isset( $status['tos_accepted'] ) && ! $status['tos_accepted'] ) {
+					if (
+						isset( $status['tos_accepted'] )
+						&& ! $status['tos_accepted']
+						&& $this->can_accept_tos()
+					) {
 						return 'tos_only_banner';
 					}
 					return false;
@@ -286,7 +313,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 			}
 
 			$jetpack_install_status = $this->get_jetpack_install_status();
-			$banner_to_display = self::get_banner_type_to_display( array(
+			$banner_to_display = $this->get_banner_type_to_display( array(
 				'jetpack_connection_status'       => $jetpack_install_status,
 				'tos_accepted'                    => WC_Connect_Options::get_option( 'tos_accepted' ),
 				'should_display_after_cxn_banner' => WC_Connect_Options::get_option( self::SHOULD_SHOW_AFTER_CXN_BANNER ),
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -165,9 +165,15 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 		 * @return bool
 		 */
 		public function can_accept_tos() {
+			$jetpack_status = $this->get_jetpack_install_status();
+
 			// Developer case
-			if ( self::JETPACK_DEV === $this->get_jetpack_install_status() ) {
+			if ( self::JETPACK_DEV === $jetpack_status ) {
 				return true;
+			}
+
+			if ( self::JETPACK_INSTALLED_NOT_ACTIVATED === $jetpack_status ) {
+				return false;
 			}
 
 			$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -160,11 +160,13 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 		 * Check that the current user is the owner of the Jetpack connection
 		 * - Only that person can accept the TOS
 		 *
+		 * @uses self::get_jetpack_install_status()
+		 *
 		 * @return bool
 		 */
 		public function can_accept_tos() {
 			// Developer case
-			if ( defined( 'JETPACK_DEV_DEBUG' ) && JETPACK_DEV_DEBUG ) {
+			if ( self::JETPACK_DEV === $this->get_jetpack_install_status() ) {
 				return true;
 			}
 
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -199,11 +199,6 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 					return 'before_jetpack_connection';
 				case self::JETPACK_CONNECTED:
 				case self::JETPACK_DEV:
-					// Has the user just gone through our NUX connection flow?
-					if ( isset( $status['should_display_after_cxn_banner'] ) && $status['should_display_after_cxn_banner'] ) {
-						return 'after_jetpack_connection';
-					}
-
 					// Has the user already accepted our TOS? Then do nothing.
 					// Note: TOS is accepted during the after_connection banner
 					if (
@@ -213,6 +208,12 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 					) {
 						return 'tos_only_banner';
 					}
+
+					// Has the user just gone through our NUX connection flow?
+					if ( isset( $status['should_display_after_cxn_banner'] ) && $status['should_display_after_cxn_banner'] ) {
+						return 'after_jetpack_connection';
+					}
+
 					return false;
 				default:
 					return false;
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -177,7 +177,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 			return $can_accept;
 		}
 
-		public function get_banner_type_to_display( $status = array() ) {
+		public static function get_banner_type_to_display( $status = array() ) {
 			if ( ! isset( $status['jetpack_connection_status'] ) ) {
 				return false;
 			}
@@ -204,7 +204,8 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 					if (
 						isset( $status['tos_accepted'] )
 						&& ! $status['tos_accepted']
-						&& $this->can_accept_tos()
+						&& isset( $status['can_accept_tos'] )
+						&& $status['can_accept_tos']
 					) {
 						return 'tos_only_banner';
 					}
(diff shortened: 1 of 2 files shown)


classes/class-wc-connect-nux.php
```diff
@@ -156,7 +156,28 @@ public function register_order_page_labels_pointer( $pointers ) {
...
-		public static function get_banner_type_to_display( $status = array() ) {
+		/**
+		 * Check that the current user is the owner of the Jetpack connection
+		 * - Only that person can accept the TOS
+		 *
+		 * @return bool
+		 */
+		public function can_accept_tos() {
+			// Developer case
+			if ( defined( 'JETPACK_DEV_DEBUG' ) && JETPACK_DEV_DEBUG ) {
```
v18: Could we use `self::JETPACK_DEV` here?

classes/class-wc-connect-nux.php
```diff
@@ -177,15 +198,22 @@ public static function get_banner_type_to_display( $status = array() ) {
...
-						return 'after_jetpack_connection';
-					}
+				case self::JETPACK_DEV:
 					// Has the user already accepted our TOS? Then do nothing.
 					// Note: TOS is accepted during the after_connection banner
-					if ( isset( $status['tos_accepted'] ) && ! $status['tos_accepted'] ) {
+					if (
+						isset( $status['tos_accepted'] )
+						&& ! $status['tos_accepted']
+						&& $this->can_accept_tos()
```
v18: Not sure that we should make this function non-pure. Can we pass `$this->can_accept_tos()` in as a parameter in the array & add unit tests for it?

classes/class-wc-connect-nux.php
v18: We used to also check whether a user is an admin by using `current_user_can( 'manage_options' )`. Should we do that in this PR by adding it to `set_up_nux_notices()` on line 304 https://github.com/Automattic/woocommerce-services/blob/c01a35f7c1ec3a245a8dd86bf8beea1e366b91fc/classes/class-wc-connect-nux.php#L304 ?

jeffstieler: I excluded the `manage_options` check mainly because we're already requiring that the user be able to install/activate plugins *and* manage the WC store. That seems like enough to me.

v18: 👍 sounds good.

classes/class-wc-connect-nux.php
```diff
@@ -156,7 +156,30 @@ public function register_order_page_labels_pointer( $pointers ) {
...
+			$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
+			$can_accept = (
+				isset( $user_token->external_user_id ) &&
+				get_current_user_id() === $user_token->external_user_id
+			);
+
+			return $can_accept;
+		}
+
+		public function get_banner_type_to_display( $status = array() ) {
```
v18: We can keep this a static function if we're keeping it pure, which would eliminate the need for the `::` to `->` changes :)

jeffstieler: Fixed in https://github.com/Automattic/woocommerce-services/pull/1062/commits/0e627381b6a8631509c0c6c785fc78dad2bb863c

classes/class-wc-connect-nux.php
```diff
@@ -156,6 +156,29 @@ public function register_order_page_labels_pointer( $pointers ) {
...
+		 *
+		 * @return bool
+		 */
+		public function can_accept_tos() {
+			// Developer case
+			if ( self::JETPACK_DEV === $this->get_jetpack_install_status() ) {
+				return true;
+			}
+
+			$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
```
nabsul: This causes the site to crash if you deactivate or delete Jetpack. You'll need to check if the class exists.

classes/class-wc-connect-nux.php
```diff
@@ -177,15 +200,23 @@ public static function get_banner_type_to_display( $status = array() ) {
...
+						isset( $status['tos_accepted'] )
+						&& ! $status['tos_accepted']
+						&& isset( $status['can_accept_tos'] )
+						&& $status['can_accept_tos']
+					) {
 						return 'tos_only_banner';
 					}
+
+					// Has the user just gone through our NUX connection flow?
+					if ( isset( $status['should_display_after_cxn_banner'] ) && $status['should_display_after_cxn_banner'] ) {
```
nabsul: I realize you probably swapped these two if-statements because of the issue I reported yesterday. Unfortunately though this won't work. The problem is that although connecting Jetpack is considered an acceptance of the TOS, we don't actually set the `tos_accepted` option when that happens.

So, if jetpack isn't already connected, you'll see the "accept tos" screen instead of the final welcome screen.

Ideally I think we should set the `tos_accepted` option when the user installs/activates/connects Jetpack, but since we're not doing that your best bet is to swap these back to the order in which they were.

classes/class-wc-connect-nux.php
jeffstieler: Fixed in https://github.com/Automattic/woocommerce-services/pull/1062/commits/d7b41835a0108fe57426d75f2cf092473f73876c.

classes/class-wc-connect-nux.php
jeffstieler: Swapped the order back and added a regression test in https://github.com/Automattic/woocommerce-services/pull/1062/commits/d8883662742f5292fa54552467bcb1f73db1d3ef

tests/php/test_class_wc-connect-nux.php
```diff
@@ -7,56 +7,91 @@ public static function setupBeforeClass() {
...
+
+		$this->assertEquals(
+			WC_Connect_Nux::get_banner_type_to_display(
+				array(
+					'jetpack_connection_status'       => WC_Connect_Nux::JETPACK_DEV,
+					'tos_accepted'                    => false,
+					'can_accept_tos'                  => false,
+					'should_display_after_cxn_banner' => true,
+				)
+			), 'after_jetpack_connection'
```
v18: Should we be displaying the 'after_jetpack_connection' banner if the user can't accept TOS?

tests/php/test_class_wc-connect-nux.php
```diff
@@ -7,56 +7,91 @@ public static function setupBeforeClass() {
...
+
+		$this->assertEquals(
+			WC_Connect_Nux::get_banner_type_to_display(
+				array(
+					'jetpack_connection_status'       => WC_Connect_Nux::JETPACK_DEV,
+					'tos_accepted'                    => true,
+					'can_accept_tos'                  => false,
+					'should_display_after_cxn_banner' => true,
+				)
+			), 'after_jetpack_connection'
```
v18: Should we display anything if `can_accept_tos` is false? Looks like the result here should equal false.

tests/php/test_class_wc-connect-nux.php
```diff
@@ -7,56 +7,91 @@ public static function setupBeforeClass() {
...
+		$this->assertEquals(
+			WC_Connect_Nux::get_banner_type_to_display(
+				array(
+					'jetpack_connection_status'       => WC_Connect_Nux::JETPACK_DEV,
+					'tos_accepted'                    => false,
+					'can_accept_tos'                  => true,
+					'should_display_after_cxn_banner' => false,
+				)
+			), 'tos_only_banner'
+		);
```
v18: We can add in the remaining 3 cases as well (all passing as-is):

``` php
$this->assertEquals(
  WC_Connect_Nux::get_banner_type_to_display(
    array(
      'jetpack_connection_status'       => WC_Connect_Nux::JETPACK_DEV,
      'tos_accepted'                    => true,
      'can_accept_tos'                  => true,
      'should_display_after_cxn_banner' => false,
    )
  ), false // passes as-is
);

$this->assertEquals(
  WC_Connect_Nux::get_banner_type_to_display(
    array(
      'jetpack_connection_status'       => WC_Connect_Nux::JETPACK_DEV,
      'tos_accepted'                    => false,
      'can_accept_tos'                  => true,
      'should_display_after_cxn_banner' => true,
    )
  ), 'after_jetpack_connection' // passes as-is
);

$this->assertEquals(
  WC_Connect_Nux::get_banner_type_to_display(
    array(
      'jetpack_connection_status'       => WC_Connect_Nux::JETPACK_DEV,
      'tos_accepted'                    => true,
      'can_accept_tos'                  => true,
      'should_display_after_cxn_banner' => true,
    )
  ), 'after_jetpack_connection' // passes as-is
);
```

tests/php/test_class_wc-connect-nux.php
```diff
@@ -7,56 +7,91 @@ public static function setupBeforeClass() {
...
 	}
 
 	public function test_get_banner_type_to_display_no_jp_cxn_without_tos_acceptance() {
 		// before going through connection
 		$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
 			'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_NOT_INSTALLED,
 			'tos_accepted'                           => false,
-			'should_display_after_cxn_banner'        =>  false,
+			'can_accept_tos'                         => false,
```
v18: In all of these cases, if `can_accept_tos` is false, shouldn't we not display any banner?

tests/php/test_class_wc-connect-nux.php
```diff
@@ -7,56 +7,91 @@ public static function setupBeforeClass() {
...
 
 		// after going through connection, TOS was never accepted
 		$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
 			'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_CONNECTED,
 			'tos_accepted'                           => false,
-			'should_display_after_cxn_banner'        =>  true,
+			'can_accept_tos'                         => false,
+			'should_display_after_cxn_banner'        => true,
 		) ), 'after_jetpack_connection' );
 	}
```
v18: In addition to the `'can_accept_tos' => false` cases, we can add the true cases:

``` php
// before going through connection, user can accept tos
$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_NOT_INSTALLED,
  'tos_accepted'                           => false,
  'can_accept_tos'                         => true,
  'should_display_after_cxn_banner'        => false,
) ), 'before_jetpack_connection' ); // new test, passes as-is

$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_INSTALLED_NOT_ACTIVATED,
  'tos_accepted'                           => false,
  'can_accept_tos'                         => true,
  'should_display_after_cxn_banner'        => false,
) ), 'before_jetpack_connection' ); // new test, passes as-is

$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_ACTIVATED_NOT_CONNECTED,
  'tos_accepted'                           => false,
  'can_accept_tos'                         => true,
  'should_display_after_cxn_banner'        => false,
) ), 'before_jetpack_connection' ); // new test, passes as-is

// after going through connection, TOS was not already accepted before connection
$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_CONNECTED,
  'tos_accepted'                           => false,
  'can_accept_tos'                         => true,
  'should_display_after_cxn_banner'        => true,
) ), 'after_jetpack_connection' );  // new test, passes as-is

// after going through connection, TOS was already accepted before connection
$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_CONNECTED,
  'tos_accepted'                           => true,
  'can_accept_tos'                         => true,
  'should_display_after_cxn_banner'        => true,
) ), 'after_jetpack_connection' ); // new test, passes as-is

// make sure we get the right banner even in 'impossible' situations
$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_CONNECTED,
  'tos_accepted'                           => false,
  'can_accept_tos'                         => false,
  'should_display_after_cxn_banner'        => true,
) ), false ); // new test, does not pass
$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_CONNECTED,
  'tos_accepted'                           => true,
  'can_accept_tos'                         => false,
  'should_display_after_cxn_banner'        => true,
) ), false ); // new test, does not pass
```

tests/php/test_class_wc-connect-nux.php
```diff
@@ -65,26 +100,30 @@ public function test_get_banner_type_to_display_no_jp_cxn_with_tos_acceptance()
...
+			'should_display_after_cxn_banner'        => false,
 		) ), 'before_jetpack_connection' );
 
 		// after going through connection, TOS was already previously accepted
 		$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
 			'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_CONNECTED,
 			'tos_accepted'                           => true,
-			'should_display_after_cxn_banner'        =>  true,
+			'can_accept_tos'                         => true,
+			'should_display_after_cxn_banner'        => true,
```
v18: We can add the `'can_accept_tos' =>  false` cases (they do not pass right now):

``` php
$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_NOT_INSTALLED,
  'tos_accepted'                           => true,
  'can_accept_tos'                         => false,
  'should_display_after_cxn_banner'        => false,
) ), false ); // new test, does not pass

$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_INSTALLED_NOT_ACTIVATED,
  'tos_accepted'                           => true,
  'can_accept_tos'                         => false,
  'should_display_after_cxn_banner'        => false,
) ), false ); // new test, does not pass

$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_ACTIVATED_NOT_CONNECTED,
  'tos_accepted'                           => true,
  'can_accept_tos'                         => false,
  'should_display_after_cxn_banner'        => false,
) ), false ); // new test, does not pass

$this->assertEquals( WC_Connect_Nux::get_banner_type_to_display( array(
  'jetpack_connection_status'              => WC_Connect_Nux::JETPACK_ACTIVATED_NOT_CONNECTED,
  'tos_accepted'                           => true,
  'can_accept_tos'                         => false,
  'should_display_after_cxn_banner'        => false,
) ), false ); // new test, does not pass
```

v18:
The testing steps work as expected:
> Make sure TOS isn't accepted
> Log in with Jetpack master user
> Verify TOS banner displays
> Log in with different user (with manage_woocommerce, install_plugins, activate_plugins caps)
> Verify no TOS banner
> define( 'JETPACK_DEV_DEBUG', true );
> Verify TOS banner displays

One question - if the site has been actually connected to Jetpack (not a case we currently differentiate) then do we want for any admin to be able to accept the TOS? What if later the DEV constant is removed and a non-master-JP-user had accepted the TOS?

tests/php/test_class_wc-connect-nux.php
jeffstieler: As discussed in Slack, the value of `can_accept_tos` is reliant on Jetpack. I've added some comments to the tests that hopefully clear things up.

tests/php/test_class_wc-connect-nux.php
jeffstieler: Per our Slack discussion, we'll circle back to refactor tests post-launch.

jeffstieler:
> One question - if the site has been actually connected to Jetpack (not a case we currently differentiate) then do we want for any admin to be able to accept the TOS? What if later the DEV constant is removed and a non-master-JP-user had accepted the TOS?

Let's create an issue to track this to work on post-merge/launch. It might make more sense to scrap the DEV handling.

PR #1083, opened 2017-07-07
Handle corrupted or empty packaging metadata.

### LINKED ISSUE #1075: Unable to open Create New Label dialog
I have one order that won't let me open the Create New Label dialog, the page fades out but there's no window. Here's the content of the Chrome debug console:

image

And there's nothing in the debug log.

jeffstieler:
Fixes #1075.

This is specifically meant to handle a WC 3.1.0 bug, but adds general resilience to shipping labels.

To test:
* Check out with a WCS-provided live rate using WC 3.1.0
- OR -
* Edit an existing order item metadata `wc_connect_packages` to be `a:1:{i:0;N;}`
* Go to order details page
* Verify that no PHP warnings are displayed / in error log
* Verify that label dialog can be opened

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -142,8 +142,14 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 
 		protected function get_packaging_metadata( WC_Order $order ) {
 			$shipping_methods = $order->get_shipping_methods();
-			$shipping_method = reset( $shipping_methods );
-			return $this->get_packaging_from_shipping_method( $shipping_method );
+			$shipping_method  = reset( $shipping_methods );
+			$packaging        = $this->get_packaging_from_shipping_method( $shipping_method );
+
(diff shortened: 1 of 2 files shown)


v18:
Curious, do we have steps to repro the data corruption?

jeffstieler:
> Curious, do we have steps to repro the data corruption?

@v18 yep - "Check out with a WCS-provided live rate using WC 3.1.0"

v18:
Oh, ouch. @jeffstieler should we also fix that bug? Sorry if there's already a reference to that, I couldn't find it.

jeffstieler:
@v18 the bug has already been fixed - you won't see it on `master`.

The fix isn't super obvious in the new WC_Data layers - I believe it's one or both of these PRs:

* https://github.com/woocommerce/woocommerce/pull/15879
* https://github.com/woocommerce/woocommerce/pull/15963

v18:
> Verify that no PHP warnings are displayed / in error log
> Verify that label dialog can be opened

Both work as expected. Code looks good, too. 🚢

Issue #1071, opened 2017-06-28
When switching package type, package weight should be remembered


v18:
When switching package type, it goes back to 0 for items without a weight:

![package-weight-goes-back-to-0](https://user-images.githubusercontent.com/11487924/27649621-496b2824-5c00-11e7-84a2-50013c097d0c.gif)

jeffstieler:
@v18 I'm having a hard time reproducing this. Does the product you're testing with have a weight defined?

v18:
@jeffstieler yeh, no weight defined for the product. otherwise it would go back to the total weight (I'm guessing).

dechov:
Can confirm it always returns to the original total weight.

PR #1087, opened 2017-07-13
Take into account current package weight when setting package type

dechov:
Fixes https://github.com/Automattic/woocommerce-services/issues/1071. If you modify the weight and then change the package type, it'll keep (or adjust, depending on box weights) the weight you entered.

fix-1071

### DIFF
--- a/client/apps/shipping-label/state/reducer.js
+++ b/client/apps/shipping-label/state/reducer.js
@@ -208,6 +208,7 @@ reducers[ UPDATE_PACKAGE_WEIGHT ] = ( state, { packageId, value } ) => {
 	newPackages[ packageId ] = {
 		...newPackages[ packageId ],
 		weight: parseFloat( value ),
+		is_user_specified_weight: true,
 	};
 
 	return { ...state,
@@ -445,11 +446,10 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
 	const newPackages = { ...state.form.packages.selected };
 	const oldPackage = newPackages[ packageId ];
 
-	const oldBox = state.form.packages.all[ oldPackage.box_id ];
-	const newBox = state.form.packages.all[ boxTypeId ];
+	const box = state.form.packages.all[ boxTypeId ];
 	const weight = round(
-		( newBox ? newBox.box_weight : 0 ) - ( oldBox ? oldBox.box_weight : 0 ) +
-			( oldPackage.weight != null ? oldPackage.weight : _.sumBy( oldPackage.items, 'weight' ) )
+		oldPackage.is_user_specified_weight ? oldPackage.weight
+			: ( box ? box.box_weight : 0 ) + _.sumBy( oldPackage.items, 'weight' )
 	);
 
 	if ( 'not_selected' === boxTypeId ) {
@@ -463,7 +463,7 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
 			box_id: boxTypeId,
 		};
 	} else {
-		const { length, width, height } = getBoxDimensions( newBox );
+		const { length, width, height } = getBoxDimensions( box );
 		newPackages[ packageId ] = {
 			..._.omit( oldPackage, 'service_id' ),
 			height,
--- a/client/apps/shipping-label/state/reducer.js
+++ b/client/apps/shipping-label/state/reducer.js
@@ -445,21 +445,25 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
 	const newPackages = { ...state.form.packages.selected };
 	const oldPackage = newPackages[ packageId ];
 
+	const oldBox = state.form.packages.all[ oldPackage.box_id ];
+	const newBox = state.form.packages.all[ boxTypeId ];
+	const weight = round(
+		(newBox != null ? newBox.box_weight : 0) - ( oldBox != null ? oldBox.box_weight : 0 ) +
+			( oldPackage.weight != null ? oldPackage.weight : _.sumBy( oldPackage.items, 'weight' ) )
+	);
+
 	if ( 'not_selected' === boxTypeId ) {
 		// This is when no box is selected
 		newPackages[ packageId ] = {
 			..._.omit( oldPackage, 'service_id' ),
-			height: 0, length: 0, width: 0,
-			weight: 0 + _.sumBy( oldPackage.items, 'weight' ),
+			height: 0, length: 0, width: 0, weight,
 			box_id: boxTypeId,
 		};
 	} else {
-		const box = state.form.packages.all[ boxTypeId ];
-		const { length, width, height } = getBoxDimensions( box );
+		const { length, width, height } = getBoxDimensions( newBox );
 		newPackages[ packageId ] = {
 			..._.omit( oldPackage, 'service_id' ),
-			height, length, width,
-			weight: round( box.box_weight + _.sumBy( oldPackage.items, 'weight' ) ),
+			height, length, width, weight,
 			box_id: boxTypeId,
 		};
 	}
--- a/client/apps/shipping-label/state/reducer.js
+++ b/client/apps/shipping-label/state/reducer.js
@@ -208,7 +208,7 @@ reducers[ UPDATE_PACKAGE_WEIGHT ] = ( state, { packageId, value } ) => {
 	newPackages[ packageId ] = {
 		...newPackages[ packageId ],
 		weight: parseFloat( value ),
-		is_user_specified_weight: true,
+		isUserSpecifiedWeight: true,
 	};
 
 	return { ...state,
@@ -448,7 +448,7 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
 
 	const box = state.form.packages.all[ boxTypeId ];
 	const weight = round(
-		oldPackage.is_user_specified_weight ? oldPackage.weight
+		oldPackage.isUserSpecifiedWeight ? oldPackage.weight
 			: ( box ? box.box_weight : 0 ) + _.sumBy( oldPackage.items, 'weight' )
 	);
 
--- a/client/apps/shipping-label/state/reducer.js
+++ b/client/apps/shipping-label/state/reducer.js
@@ -448,7 +448,7 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
 	const oldBox = state.form.packages.all[ oldPackage.box_id ];
 	const newBox = state.form.packages.all[ boxTypeId ];
 	const weight = round(
-		( newBox != null ? newBox.box_weight : 0 ) - ( oldBox != null ? oldBox.box_weight : 0 ) +
+		( newBox ? newBox.box_weight : 0 ) - ( oldBox ? oldBox.box_weight : 0 ) +
 			( oldPackage.weight != null ? oldPackage.weight : _.sumBy( oldPackage.items, 'weight' ) )
 	);
 
@@ -456,14 +456,20 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
 		// This is when no box is selected
 		newPackages[ packageId ] = {
 			..._.omit( oldPackage, 'service_id' ),
-			height: 0, length: 0, width: 0, weight,
+			height: 0,
+			length: 0,
+			width: 0,
+			weight,
 			box_id: boxTypeId,
 		};
 	} else {
 		const { length, width, height } = getBoxDimensions( newBox );
 		newPackages[ packageId ] = {
 			..._.omit( oldPackage, 'service_id' ),
-			height, length, width, weight,
+			height,
+			length,
+			width,
+			weight,
 			box_id: boxTypeId,
 		};
 	}
--- a/client/apps/shipping-label/state/reducer.js
+++ b/client/apps/shipping-label/state/reducer.js
@@ -448,7 +448,7 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
 	const oldBox = state.form.packages.all[ oldPackage.box_id ];
 	const newBox = state.form.packages.all[ boxTypeId ];
 	const weight = round(
-		(newBox != null ? newBox.box_weight : 0) - ( oldBox != null ? oldBox.box_weight : 0 ) +
+		( newBox != null ? newBox.box_weight : 0 ) - ( oldBox != null ? oldBox.box_weight : 0 ) +
 			( oldPackage.weight != null ? oldPackage.weight : _.sumBy( oldPackage.items, 'weight' ) )
 	);
 
--- a/client/apps/shipping-label/state/test/reducer.js
+++ b/client/apps/shipping-label/state/test/reducer.js
@@ -16,6 +16,7 @@ import {
 	addPackage,
 	removePackage,
 	setPackageType,
+	updatePackageWeight,
 	savePackages,
 	removeIgnoreValidation,
 	updateAddressValue,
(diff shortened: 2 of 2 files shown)


v18:
@dechov looks like the CI errors are just due to some linting errors in `reducer.js`:

```
428:3   error  There must be a space inside this paren  space-in-parens
428:42  error  There must be a space inside this paren  space-in-parens
429:93  error  Missing semicolon                        semi
```

client/shipping-label/state/reducer.js
```diff
@@ -422,21 +422,24 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
 	const newPackages = { ...state.form.packages.selected };
 	const oldPackage = newPackages[ packageId ];
 
+	const oldBox = state.form.packages.all[ oldPackage.box_id ];
+	const newBox = state.form.packages.all[ boxTypeId ];
+	const weight =
+		( newBox != null ? newBox.box_weight : 0 ) - ( oldBox != null ? oldBox.box_weight : 0 ) +
```
v18: Here we're checking to see if `newBox` is not `undefined` right? If so, we can rely on the falsy if check like so:

`( newBox ? newBox.box_weight : 0 )`

client/shipping-label/state/reducer.js
```diff
@@ -422,21 +422,24 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
 	const newPackages = { ...state.form.packages.selected };
 	const oldPackage = newPackages[ packageId ];
 
+	const oldBox = state.form.packages.all[ oldPackage.box_id ];
+	const newBox = state.form.packages.all[ boxTypeId ];
+	const weight =
```
v18: Do we know how many decimal places we're using for weight? We should make sure that we are not running into the JS addition bug for small numbers:

https://stackoverflow.com/a/5037927

client/shipping-label/state/reducer.js
```diff
@@ -422,21 +422,24 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
 	const newPackages = { ...state.form.packages.selected };
```
v18: I don't know if it makes sense to do in this PR or even soon in a new related PR, but `newPackages` should be an array and not an object of objects. Why? For easier lookup, and manipulation. Adding this to the list of things we should review: p2-p7bbVw-1Bj#comment-2697 

#1088

client/shipping-label/state/reducer.js
```diff
@@ -422,21 +422,24 @@ reducers[ SET_PACKAGE_TYPE ] = ( state, { packageId, boxTypeId } ) => {
...
 		};
 	} else {
-		const box = state.form.packages.all[ boxTypeId ];
-		const { length, width, height } = getBoxDimensions( box );
+		const { length, width, height } = getBoxDimensions( newBox );
 		newPackages[ packageId ] = {
 			..._.omit( oldPackage, 'service_id' ),
-			height, length, width,
-			weight: box.box_weight + _.sumBy( oldPackage.items, 'weight' ),
+			height, length, width, weight,
```
v18: Can we put each new object property on a new line ([eslint doc](http://eslint.org/docs/rules/object-property-newline))? I'm surprised we don't have this in our linting rules.

client/shipping-label/state/reducer.js
dmsnell: Fun fact! This isn't a JS bug. As funny as it is to see `0.1 + 0.2 !== 0.3` this is part of the IEEE standard for the number format. Check out much more detail at this post: https://medium.com/the-node-js-collection/javascripts-number-type-8d59199db1b6

dmsnell: also, I'm guessing that the kinds of errors we will get from floating-point oddities will be far lower than what would be significant with the weight of any shippable thing. if there is extreme concern then we can safely store no decimal places, instead using an integer and choosing appropriate units

client/shipping-label/state/reducer.js
dmsnell: @v18 _why_ should they not be an object? It's a valid point to raise _if there's a reason_, but did you find that something about this was not working properly on account of the object? it looks like we _are_ using the object for its indexing abilities

v18: @dmsnell no solid reason yet - it fits in with some other patterns I've noted before. It may be the case that a different structure isn't needed after all, but it's worth a look.

client/shipping-label/state/reducer.js
v18: @dmsnell It may not be actual bug (i.e. following the standards), but it's certainly not expected.

> also, I'm guessing that the kinds of errors we will get from floating-point oddities will be far lower than what would be significant with the weight of any shippable thing

Yeh, that's the assumption we need to double-check here.

client/shipping-label/state/reducer.js
dmsnell: > but it's certainly not expected.

well, and I don't want to sound offensive here, but it _is_ expected because that's what a bunch of people very publicly agreed to and standardized :wink:

it's more that we're often unfamiliar with it because we don't often run into it being a problem. in general fixed-point arithmetic has for most of programming history been a thing all to itself due to the limits of the floating-point representation. it's _acknowledged_ that all floating-point arithmetic is an approximation and that it also breaks mathematical laws.

yes! floating-point addition is not associative and commutative even though arithmetic addition is. some languages have a specific fixed-point data type (SQL does, for example) and often one is required to specify how many places will be used beyond the decimal. in some more esoteric places we end up storing actual decimal data in BCD (binary-coded decimal) which treats each digit as a nibble or half-word of four bits.

since most of our floating-point match _does_ deal with approximations this normally is safe but it always reminds us that it is not exact, rounding errors propagate, and if we _need_ fixed-point math we need to use integer data types.

client/shipping-label/state/reducer.js
dmsnell: > but it's certainly not expected.

well, and I don't want to sound offensive here, but it _is_ expected because that's what a bunch of people very publicly agreed to and standardized :wink:

it's more that we're often unfamiliar with it because we don't often run into it being a problem. in general fixed-point arithmetic has for most of programming history been a thing all to itself due to the limits of the floating-point representation. it's _acknowledged_ that all floating-point arithmetic is an approximation and that it also breaks mathematical laws.

yes! floating-point addition is not associative and commutative even though arithmetic addition is. some languages have a specific fixed-point data type (SQL does, for example) and often one is required to specify how many places will be used beyond the decimal. in some more esoteric places we end up storing actual decimal data in BCD (binary-coded decimal) which treats each digit as a nibble or half-word of four bits.

since most of our floating-point math _does_ deal with approximations this normally is safe but it always reminds us that it is not exact, rounding errors propagate, and if we _need_ fixed-point math we need to use integer data types.

dmsnell: …or we use a novel but nonstandard floating point representation, such as the [Unum](https://en.wikipedia.org/wiki/Unum_(number_format))

dmsnell: …also because this is such a fun topic, there are two zeros: a negative zero and a positive zero :smile:

v18:
Great job, this works well. 

I took this out for a test ride. There are a few flow-type issues I see:

**Flow issue 1: User can enter a weight that's less than the package weight**
Steps:
1. Create a custom package of weight 10
2. Go to label purchase flow > packages
3. Select custom package
4. Enter weight less than 10, see that it works

I realize that this bug existed before this branch, but...

When this happens, if you select another package, the weight will be a negative number:

screen shot 2017-07-13 at 7 57 01 am

**Flow issue #2: When selecting a different package, as a user I may not realize the previous weight is still saved.**

This is exactly the problem we were trying to solve... but maybe it's not the right answer.

I should have done this when opening the issue initially, and now I wonder if before we solve the problem of 'not saving the weight' (i.e. #1071) we think through why a user would enter a package weight.

When someone enters a weight for a package with products with no weight - what are they trying to accomplish?

I think the underlying issue is letting users enter a weight for a package with products with no weight, instead of entering the weight for the products themselves.

Another assumption we can challenge: what's the weight for? For flat rate packages, the weight 

@olaolusoga - I'd love to hear your your opinion about the best way to solve #1071

client/shipping-label/state/reducer.js
v18: @dechov I wanted to circle back to this issue. We do expect to have weights that are less than 1 [unit] of measurement, e.g. 0.2lbs, etc.

The questions for our code here are:
1. How many decimal places do we need for the weight measurement?
2. Will we need to use a different method for adding the weights together, rather than the currently-used floating-point addition option?

dechov:
It might alleviate "Flow issue 1" to put a minimum of 0 on the weight value, but I'll await some more consideration of how the weight input is or would be used.

jeffstieler:
> I should have done this when opening the issue initially, and now I wonder if before we solve the problem of 'not saving the weight' (i.e. #1071) we think through why a user would enter a package weight.

It mainly offers flexibility - perhaps some items require more packing material than others, resulting in varying weights?

> Another assumption we can challenge: what's the weight for? For flat rate packages, the weight

This is a good point - perhaps for flat rate packaging, we should just display the weight limit, and omit the user-editable "total weight" field.

jeffstieler:
@dechov - sorry, I didn't realize I froze progress here.

> > Another assumption we can challenge: what's the weight for? For flat rate packages, the weight
> 
> This is a good point - perhaps for flat rate packaging, we should just display the weight limit, and omit the user-editable "total weight" field.

@olaolusoga what do you think about this ☝️  ?

client/shipping-label/state/reducer.js
dechov: @v18 No idea what minimum precision we need but let's say we can agree on some limit `digits` that's less than 10. Will `+arithmeticResult.toFixed( digits )` be good enough?

Test case demonstrating the issue independent of this PR (i.e. in `master`): put a weight 3 item and a weight 0.3 item in the same package, and then take out the weight 3 item — the Total Weight winds up as 0.2999999999999998

v18: @dechov in that case, it would mean that we're showing the user a total weight of 0.299 when they change back to the previous package? That doesn't sound exactly right to me, but if it's a widespread issue I can see the case for solving it in a different PR.

dechov: I believe it would show 0.3 after they move the heavier item away: `+( 3.3 - 3 ).toFixed( n )` for 0 < n < 16. I can see why a separate PR might be appropriate, since it exists independently of this thread's design discussion.

dechov: @dmsnell Does truncating with `.toFixed` sound like an acceptable alternative to using integer values and micro units?

dmsnell: @dechov yes! I think it's not only an _acceptible_ solution but probably the standard and preferred solution :wink:

technically it's not truncating either…it's rounding, which is probably even more what we want here.

dechov: https://github.com/Automattic/woocommerce-services/pull/1096!

dechov:
Thinking about it, I suspect the chance to input the package weight at this point is valuable for two reasons:
- The need for precision (and even for entering a weight at all) is rather more obvious when shipping than when product and package details are declared previously. This is *not dependent* on package type and would benefit from the change in this PR if precise values were predefined for the package weights — which would be more often the case than for product weights, I imagine.
- Using packing material would require manually adjusting the weight (good point @jeffstieler). This is likely *dependent* on package type in many cases, so when you change the package, the previous weight you entered (and this PR) would only apply if you're just directly transferring the same packing material.

> For flat rate packages, the weight doesn't really make a difference unless it's over 70lbs. So some more custom logic would be great to add there to make it easier for the user to grok what's going on.

Agreed, and if it's obvious it's nowhere near 70 lb. then we shouldn't present a "required"-looking weight input to the user. But otherwise, both of the above reasons matter for flat-rate packages as well — hence I think we should keep the input — if only just to tell you if it's above the package's maximum weight or the USPS weight restriction, which it doesn't currently do but probably should? (Also, if it's over 70 lb., the rate just doesn't compute and it doesn't explain why, which may also be what happens if the [length + girth restriction](https://www.usps.com/ship/priority-mail.htm) is exceeded.)

olaolusoga:
Sorry for delay @jeffstieler @v18 @dechov 

> Another assumption we can challenge: what's the weight for? For flat rate packages, the weight...

>This is a good point - perhaps for flat rate packaging, we should just display the weight limit, and omit the user-editable "total weight" field.

>When someone enters a weight for a package with products with no weight - what are they trying to accomplish?

They are trying to get a label price estimate. The weight is just a step/hurdle on the path for them.

It's worth leaving weight field in for the reason @jeffstieler mentioned (packing material). Also agree with your thoughts @v18 
>I think the underlying issue is letting users enter a weight for a package with products with no weight, instead of entering the weight for the products themselves.

Focusing on getting a person to enter the product weights when creating a product would help solve the issue of no weight, but we would still have to account for the people who will choose not to input their product weight during product creation.

What could help close the gap/missing component will be figuring out how to merge rate estimates into the packaging step. By doing this, no matter the weight on a flat rate package the user will have a visual indicator (the rate estimate) showing the unchanging price.

In the interim we can introduce notifications that help guide the user. If flat rate is selected, we can say "$XX Flat rate price for packages under 70lb". For other packages we can simply say "To get label price estimates, enter the total weight of your package."

Also for **flow issue 1**,  can we retain the weight the store owner enters regardless of "package browsing". If not, @dechov suggestion to **put a minimum of 0 on the weight value** would be ideal.

jeffstieler:
@olaolusoga I love the ideas - let's put those into our Q3 onboarding work. 👍 

> Also for flow issue 1, can we retain the weight the store owner enters regardless of "package browsing". If not, @dechov suggestion to put a minimum of 0 on the weight value would be ideal.

@dechov let's keep the scope of this PR to the above ☝️ .

dechov:
To summarize the existing alternatives: currently (in master), the weight is always replaced upon selecting a new package type, with a new sum of all items plus the package's weight. With the change in this PR, the user-modified weight is kept but adjusted by the difference in old/new package weights (currently 0 for the predefined packages).

> Also for flow issue 1, can we retain the weight the store owner enters regardless of "package browsing".

@olaolusoga Yup, we can do this by flagging the weight for that package as *user-modified*, after which a newly-selected package type would no longer figure into the weight — is this preferable?

And also, should this state be communicated as a deliberately entered amount that the user is responsible for and may want to modify after changing the package type / adding packing materials? This may also be a consideration in the other scenarios, re: packing materials. Is it obvious that they may want to weigh it themselves in that case despite a precise weight being automatically filled for them?

I'm fine with deferring that latter question to the rest of Q3, and going ahead with either a) leaving the weight that the user enters, or b) leaving the behavior as it is but clamping the adjusted weight to 0.

jeffstieler:
> Yup, we can do this by flagging the weight for that package as user-modified, after which a newly-selected package type would no longer figure into the weight — is this preferable?

I think it is preferable - it's not always intuitive why inputs change after you've interacted with them. (I'm also building on the assumption that merchant should be weighing their packages before shipping)

> And also, should this state be communicated as a deliberately entered amount that the user is responsible for and may want to modify after changing the package type / adding packing materials?
> 
> I'm fine with deferring that latter question to the rest of Q3, and going ahead with either a) leaving the weight that the user enters ...

👍 option `a` please 😄

dechov:
Now simply marks once the user has modified the field, and that package's weight doesn't respond to package type from that point forward.

client/apps/shipping-label/state/reducer.js
```diff
@@ -208,6 +208,7 @@ reducers[ UPDATE_PACKAGE_WEIGHT ] = ( state, { packageId, value } ) => {
 	newPackages[ packageId ] = {
 		...newPackages[ packageId ],
 		weight: parseFloat( value ),
+		is_user_specified_weight: true,
```
jeffstieler: We should be using camelCase in JavaScript.

dechov: I don't disagree but is this a documented standard for keys? I see quite a few in snake case, but then again maybe there's an overriding reason for those like conforming to DB schema.

jeffstieler: AFAIK the ones we have in snake case are initialized from data that originates on the PHP side of the house. (Where we aren't doing anything to data we're getting directly from WC or the WP database before passing it to `wp_localize_script()`.

Issue #1016, opened 2017-05-24
On Label purchase, make sure user can't accidentally click and buy a second label


nabsul:
https://wordpress.org/support/topic/two-labels-purchased/#post-9161430

Based on the conversation above, it seems that it might be possible to click "BUY" in that time between successfully making a first purchase and closing the modal.

jeffstieler:
I can confirm that this is possible - if your connection is too fast, you can use Chrome's developer tools to throttle the speed (under Network tab).

dechov:
`state.form.isSubmitting` turns false when the [label purchase response comes in](https://github.com/Automattic/woocommerce-services/blob/a059559475dc129aeba36f48e112c9efc31f32a9/client/shipping-label/state/actions.js#L472), again inviting the user to "Buy & Print" until the document to print is loaded, at which point the [printing flow is exited](https://github.com/Automattic/woocommerce-services/blob/a059559475dc129aeba36f48e112c9efc31f32a9/client/shipping-label/state/actions.js#L508).

Is there some way to preserve the error handling while bringing those actions closer together? Otherwise, do we want another disabled state after "Purchasing..." such as "Loading..." / "Downloading..."?

jeffstieler:
I don't think we sacrifice error handling by decreasing the time between those action dispatches.

The delay looks to be caused by the `printDocument()` call. I'm thinking that we should look into either not clearing `form.isSubmitting` on successful purchase responses, or moving the success notice and flow exit actions to before `printDocument()`, or both.

PR #1091, opened 2017-07-14
Persist “purchasing…” button until label modal is dismissed.

jeffstieler:
Fixes: #1016.

This PR seeks to avoid accidental clicks of the "buy" button after a label has successfully been purchased, but before the modal dismisses.

It also adds some tests for the `PURCHASE_LABEL_RESPONSE` reducer.

To test:
* Throttle network connection (see issue comments)
* Buy label in native PDF supporting browser (Safari or Chrome)
* Verify that there is no opportunity to click the "buy" button again
* Verify that non-native PDF supporting browsers still work for purchasing label (Firefox)
* Verify that tests pass

### DIFF
--- a/client/shipping-label/state/actions.js
+++ b/client/shipping-label/state/actions.js
@@ -505,12 +505,14 @@ export const purchaseLabel = () => ( dispatch, getState, context ) => {
 								}
 							);
 							dispatch( NoticeActions.successNotice( noticeText ) );
-							dispatch( exitPrintingFlow( true ) );
-							dispatch( clearAvailableRates() );
 						} )
 						.catch( ( err ) => {
 							console.error( err );
 							dispatch( NoticeActions.errorNotice( err.toString() ) );
+						} )
+						.then( () => {
+							dispatch( exitPrintingFlow( true ) );
+							dispatch( clearAvailableRates() );
 						} );
 				}
 			}
--- a/client/shipping-label/state/reducer.js
+++ b/client/shipping-label/state/reducer.js
@@ -510,9 +510,6 @@ reducers[ PURCHASE_LABEL_RESPONSE ] = ( state, { response, error } ) => {
 	}
 
 	return { ...state,
-		form: { ...state.form,
-			isSubmitting: false,
-		},
 		labels: [
 			...response.map( ( label ) => ( { ...label,
 				statusUpdated: true,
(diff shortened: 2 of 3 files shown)


dechov:
Cool— nice and simple solution. Just one error case I can see it affecting: if `printDocument( printUrl )` rejects, the button would stay disabled ("Purchasing...") instead of allowing you to "Buy & Print" again, which actually seems to me like a mild improvement but I'm not sure what exactly would trigger this case.

jeffstieler:
> Just one error case I can see it affecting: if `printDocument( printUrl )` rejects, the button would stay disabled ("Purchasing...") instead of allowing you to "Buy & Print" again, which actually seems to me like a mild improvement but I'm not sure what exactly would trigger this case.

I believe a rejection would be caused by something like the PDF url timing out or similar (in the hidden iframe we use). I also think this is a mild improvement, but perhaps in the `printDocument()` error case we should also dismiss the modal? We only call `printDocument()` after a successful purchase request, so we definitely shouldn't show "buy" again.

dechov:
I agree it should dismiss the modal: after successful purchase, the only purpose it appears to serve is communicating that something's still happening, which after rejection isn't true. In this case, should the error notification also a) let you know that the purchase succeeded, and b) direct you to try and Reprint maybe? Currently it displays the raw DOM error only.

jeffstieler:
> I agree it should dismiss the modal: after successful purchase, the only purpose it appears to serve is communicating that something's still happening, which after rejection isn't true. In this case, should the error notification also a) let you know that the purchase succeeded, and b) direct you to try and Reprint maybe? Currently it displays the raw DOM error only.

Can you detail your testing steps? I'd like to be able to recreate the DOM error and see if this PR negatively effects that flow.

If that PDF/DOM error flow isn't changed by this PR, let's capture any improvements in a separate issue as to not hold this up.

dechov:
I'd inelegantly added `http_response_code(404);` to `class-wc-rest-connect-shipping-label-print-controller.php`, which may or may not approximate a real error case but did cause the promise to reject. I don't believe this PR affects the flow in this regard at all, however, so LGTM! :shipit:

jeffstieler:
The purchase dialog will now close if the `printDocument()` promise rejects.

dechov:
Keeps the button disabled until the labels are ready, and then closes the dialog immediately even if there was an error in printing the labels. Looks good to me :shipit:

Issue #1033, opened 2017-06-13
In "Add Items": Replace radio buttons with check boxes


nabsul:
Related: https://github.com/Automattic/woocommerce-services/pull/1028#pullrequestreview-43581864

PR #1092, opened 2017-07-14
Enhance "Add Item" dialog by allowing multiple items to be checked

dechov:
Fixes: https://github.com/Automattic/woocommerce-services/issues/1033

Allows adding one _or more_ items to a package at once by using checkbox controls instead of radio buttons.

Before:
add-items-before

After:
add-items

Also disables the primary action button (for this dialog and the "Move" dialog) if no selection is made, indicating that no consequential action has been selected.

### DIFF
--- a/client/shipping-label/state/reducer.js
+++ b/client/shipping-label/state/reducer.js
@@ -341,14 +341,17 @@ reducers[ CLOSE_ADD_ITEM ] = ( state ) => {
 };
 
 reducers[ SET_ADDED_ITEM ] = ( state, { sourcePackageId, movedItemIndex, added } ) => {
-	const itemIndices = state.addedItems[ sourcePackageId ] || []
+	let newItemIndices;
+	if ( added ) {
+		const itemIndices = state.addedItems[ sourcePackageId ] || [];
+		newItemIndices = _.includes( itemIndices, movedItemIndex ) ? itemIndices : [ ...itemIndices, movedItemIndex ];
+	} else {
+		newItemIndices = _.without( state.addedItems[ sourcePackageId ], movedItemIndex );
+	}
+
 	return {
 		...state,
-		addedItems: { ...state.addedItems,
-			[ sourcePackageId ]: ! added
-				? _.without( itemIndices, movedItemIndex )
-				: _.includes( itemIndices, movedItemIndex ) ? itemIndices : [ ...itemIndices, movedItemIndex ]
-		},
+		addedItems: { ...state.addedItems, [ sourcePackageId ]: newItemIndices },
 	};
 };
 
--- a/client/shipping-label/state/reducer.js
+++ b/client/shipping-label/state/reducer.js
@@ -334,16 +334,19 @@ reducers[ OPEN_ADD_ITEM ] = ( state ) => {
 reducers[ CLOSE_ADD_ITEM ] = ( state ) => {
 	return {
 		...state,
-		movedItemIndex: -1,
+		addedItems: null,
 		showAddItemDialog: false,
 	};
 };
 
-reducers[ SET_ADDED_ITEM ] = ( state, { sourcePackageId, movedItemIndex } ) => {
+reducers[ SET_ADDED_ITEM ] = ( state, { sourcePackageId, movedItemIndex, added } ) => {
 	return {
 		...state,
-		sourcePackageId,
-		movedItemIndex,
+		addedItems: { ...state.addedItems,
+			[ sourcePackageId ]: { ...( state.addedItems || {} )[ sourcePackageId ],
+				[ movedItemIndex ]: added,
+			}
+		},
 	};
 };
 
--- a/client/shipping-label/state/reducer.js
+++ b/client/shipping-label/state/reducer.js
@@ -341,21 +341,23 @@ reducers[ CLOSE_ADD_ITEM ] = ( state ) => {
 };
 
 reducers[ SET_ADDED_ITEM ] = ( state, { sourcePackageId, movedItemIndex, added } ) => {
+	const itemIndices = state.addedItems[ sourcePackageId ] || []
 	return {
 		...state,
 		addedItems: { ...state.addedItems,
-			[ sourcePackageId ]: added
-				? { ...state.addedItems[ sourcePackageId ], [ movedItemIndex ]: true }
-				: _.omit( state.addedItems[ sourcePackageId ], movedItemIndex )
+			[ sourcePackageId ]: ! added
+				? _.without( itemIndices, movedItemIndex )
+				: _.includes( itemIndices, movedItemIndex ) ? itemIndices : [ ...itemIndices, movedItemIndex ]
 		},
 	};
 };
 
 reducers[ ADD_ITEMS ] = ( state, { targetPackageId } ) => {
 	// For each origin package
-	_.each( state.addedItems, ( inPackage, originPackageId ) => {
-		// Move items in reverse order of index, to maintain validity as items are removed
-		_.sortBy( Object.keys( inPackage ), ( d ) => -d ).forEach( ( movedItemIndex ) => {
+	_.each( state.addedItems, ( itemIndices, originPackageId ) => {
+		// Move items in reverse order of index, to maintain validity as items are removed.
+		// e.g. when index 0 is removed from the package, index 1 would become index 0
+		_.sortBy( itemIndices, ( i ) => -i ).forEach( ( movedItemIndex ) => {
 			state = reducers[ MOVE_ITEM ]( state, { originPackageId, movedItemIndex, targetPackageId } );
 		} );
 	} );
--- a/client/shipping-label/state/reducer.js
+++ b/client/shipping-label/state/reducer.js
@@ -344,24 +344,22 @@ reducers[ SET_ADDED_ITEM ] = ( state, { sourcePackageId, movedItemIndex, added }
 	return {
 		...state,
 		addedItems: { ...state.addedItems,
-			[ sourcePackageId ]: { ...state.addedItems[ sourcePackageId ],
-				[ movedItemIndex ]: added,
-			}
+			[ sourcePackageId ]: added
+				? { ...state.addedItems[ sourcePackageId ], [ movedItemIndex ]: true }
+				: _.omit( state.addedItems[ sourcePackageId ], movedItemIndex )
 		},
 	};
 };
 
 reducers[ ADD_ITEMS ] = ( state, { targetPackageId } ) => {
-	return _.reduce( state.addedItems, ( prevState, inPackage, originPackageId ) => {
-		const movedItemIndices = _.toPairs( inPackage )
-			.filter( ( [ , added ] ) => added )
-			.map( ( [ movedItemIndex ] ) => +movedItemIndex );
-		return _.sortBy( movedItemIndices ).reverse().reduce( ( prevPrevState, movedItemIndex ) => {
-			return reducers[ MOVE_ITEM ]( prevPrevState, { originPackageId, movedItemIndex, targetPackageId } );
-		}, prevState );
-	}, { ...state,
-		showAddItemDialog: false,
+	// For each origin package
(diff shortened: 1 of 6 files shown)


client/shipping-label/views/purchase/steps/packages/add-item.js
```diff
@@ -81,7 +80,16 @@ const AddItemDialog = ( {
 					{ itemOptions }
 				</div>
 				<ActionButtons buttons={ [
-					{ label: __( 'Add' ), isPrimary: true, onClick: () => moveItem( sourcePackageId, movedItemIndex, openedPackageId ) },
+					{ label: __( 'Add' ), isPrimary: true, onClick: () => {
+						Object.keys( addedItems || {} ).forEach( ( sourcePackageId ) => {
+							Object.keys( addedItems[ sourcePackageId ] || {} ).forEach( ( movedItemIndex ) => {
+								if ( addedItems[ sourcePackageId ][ movedItemIndex ] ) {
+									moveItem( sourcePackageId, movedItemIndex, openedPackageId );
```
jeffstieler: What do you think about refactoring `moveItem()` to become `moveItems()` so we only dispatch one action?

client/shipping-label/views/purchase/steps/packages/add-item.js
```diff
@@ -32,17 +31,17 @@ const AddItemDialog = ( {
...
 			: item;
 
-		const onChange = () => setAddedItem( pckgId, itemIdx );
+		const onChange = ( event ) => setAddedItem( pckgId, itemIdx, event.target.checked );
 		return (
 			<FormLabel
 				key={ `${ pckgId }-${ itemIdx }` }
 				className="wcc-move-item-dialog__package-option">
-				<FormRadio checked={ pckgId === sourcePackageId && itemIdx === movedItemIndex }
+				<FormCheckbox checked={ ! ! ( ( addedItems || {} )[ pckgId ] || {} )[ itemIdx ] }
```
jeffstieler: Lodash's [`_.get()`](https://lodash.com/docs#get) can make this a lot more readable and still safely handle missing object properties. Something like `_.get( addedItems, [ pckgId, itemIdx ], false )`.

client/shipping-label/state/reducer.js
```diff
@@ -334,16 +334,19 @@ reducers[ OPEN_ADD_ITEM ] = ( state ) => {
...
 };
 
-reducers[ SET_ADDED_ITEM ] = ( state, { sourcePackageId, movedItemIndex } ) => {
+reducers[ SET_ADDED_ITEM ] = ( state, { sourcePackageId, movedItemIndex, added } ) => {
 	return {
 		...state,
-		sourcePackageId,
-		movedItemIndex,
+		addedItems: { ...state.addedItems,
+			[ sourcePackageId ]: { ...( state.addedItems || {} )[ sourcePackageId ],
```
jeffstieler: Shouldn't this be checking `state.addedItems[ sourcePackageId ]` instead?

`addedItems` should be added to the initial state here: https://github.com/Automattic/woocommerce-services/blob/748e310eba1fa1c0a804c535b213955c56cd9d11/client/shipping-label/index.js#L31

dechov: Good point. Trying a different initialization approach — setting `addedItems` to `{}` when opening the dialog — let me know what you think.

client/shipping-label/views/purchase/steps/packages/add-item.js
dechov: Ah, yes, much better. Or the [modern version](https://twitter.com/paul_irish/status/885975704055062528) 😉

client/shipping-label/views/purchase/steps/packages/add-item.js
dechov: Why I must say, I think it would be a swell thing to refactor as it looks like the only way to fix my apparently broken original implementation 😄 that is, if multiple items were being moved from a single package, the indexing would be thrown off when acting in sequence.

dechov:
Added a test for the logic in the `ADD_ITEMS` reducer.

client/shipping-label/state/reducer.js
```diff
@@ -328,25 +329,41 @@ reducers[ OPEN_ADD_ITEM ] = ( state ) => {
...
+		const movedItemIndices = _.toPairs( inPackage )
+			.filter( ( [ , added ] ) => added )
+			.map( ( [ movedItemIndex ] ) => +movedItemIndex );
+		return _.sortBy( movedItemIndices ).reverse().reduce( ( prevPrevState, movedItemIndex ) => {
+			return reducers[ MOVE_ITEM ]( prevPrevState, { originPackageId, movedItemIndex, targetPackageId } );
+		}, prevState );
+	}, { ...state,
+		showAddItemDialog: false,
+	} );
+};
```
jeffstieler: As discussed in Slack, it's not immediately clear what's going on in this reducer.

Adding some comments to explain lodash function usage would help explain things. For instance, what's the purpose of using `_.sortBy()` with no second argument, and why does the result need `reverse()`ing?

dechov:
The reducer I added was overly complex, as pointed out to me by @jeffstieler — should be much clearer now. `state.addedItems` only stores `true` leaf values now.

client/shipping-label/state/reducer.js
```diff
@@ -328,25 +329,39 @@ reducers[ OPEN_ADD_ITEM ] = ( state ) => {
...
-reducers[ SET_ADDED_ITEM ] = ( state, { sourcePackageId, movedItemIndex } ) => {
+reducers[ SET_ADDED_ITEM ] = ( state, { sourcePackageId, movedItemIndex, added } ) => {
 	return {
 		...state,
-		sourcePackageId,
-		movedItemIndex,
+		addedItems: { ...state.addedItems,
+			[ sourcePackageId ]: added
+				? { ...state.addedItems[ sourcePackageId ], [ movedItemIndex ]: true }
+				: _.omit( state.addedItems[ sourcePackageId ], movedItemIndex )
```
jeffstieler: How much cleaner would this be if we weren’t using the item index as an object property?
 
`{ addedItems: { 'package_1': [ 0, 3, 4 ] } }` array instead?

client/shipping-label/state/reducer.js
```diff
@@ -328,25 +329,39 @@ reducers[ OPEN_ADD_ITEM ] = ( state ) => {
...
+		},
 	};
 };
 
+reducers[ ADD_ITEMS ] = ( state, { targetPackageId } ) => {
+	// For each origin package
+	_.each( state.addedItems, ( inPackage, originPackageId ) => {
+		// Move items in reverse order of index, to maintain validity as items are removed
```
jeffstieler: > to maintain validity as items are removed

What would be invalid about moving items in the opposite order?

dechov: This was the bug with the original implementation in this PR. Adding example in expanded comment, but let me know if you find it to be a bad practice / code smell.

client/shipping-label/state/reducer.js
dechov: I don't think the code is much cleaner (though I'm open to suggestions), but I made the change for the sake of having the state be a bit cleaner.

dechov:
Changed to storing the index of the item to add in an array instead of an object, the main benefit of which is to store it as a `Number` — more appropriate than `String`.

client/shipping-label/state/test/reducer.js
```diff
@@ -123,6 +126,41 @@ describe( 'Label purchase form reducer', () => {
...
+		existingState.form.packages.selected.weight_1_custom1.items[ 1 ] = {
+			product_id: 457,
+			weight: 3.4,
+		};
+
+		existingState = [
+			setAddedItem( 'weight_0_custom1', 0, true ),
+			setAddedItem( 'weight_1_custom1', 0, true ),
+			setAddedItem( 'weight_1_custom1', 1, true ),
+		].reduce( reducer, existingState );
```
jeffstieler: Why modify initial state for package data, but then use the reducer for selecting items to add?

My gut says that we should either add some expectations around the result of that reducer, or use an initial state that already contains the necessary value for `addedItems`.

client/shipping-label/views/purchase/steps/packages/add-item.js
```diff
@@ -72,7 +72,7 @@ const AddItemDialog = ( {
 				<h1 className="form-section-heading">{ __( 'Add item' ) }</h1>
 				<div className="wcc-label-packages-dialog__body">
 					<p>
-						{ __( 'Which item would you like to add to {{pckg/}}?', {
+						{ __( 'Which items would you like to add to {{pckg/}}?', {
```
jeffstieler: This should be a single/plural case based on the number of items available to move.

client/shipping-label/views/purchase/steps/packages/move-item.js
```diff
@@ -105,7 +105,12 @@ const MoveItemDialog = ( {
 					{ renderIndividualOption() }
 				</div>
 				<ActionButtons buttons={ [
-					{ label: __( 'Move' ), isPrimary: true, onClick: () => moveItem( openedPackageId, movedItemIndex, targetPackageId ) },
+					{
+						label: __( 'Move' ),
+						isPrimary: true,
+						isDisabled: targetPackageId === openedPackageId,
```
jeffstieler: When would this be encountered in the move dialog?

client/shipping-label/state/test/reducer.js
dechov: I agree about adding expectations of the intermediate result. I wanted to test `setAddedItem` as well, since there's a tiny bit of complexity there too. And `openAddItem` now that I think about it, to test that initialization as well.

client/shipping-label/views/purchase/steps/packages/move-item.js
dechov: When you first open it — I'll leave a comment explaining that.

client/shipping-label/views/purchase/steps/packages/add-item.js
dechov: I'm not sure but I feel like if we're explicitly acknowledging that we know there's only one item, it doesn't make sense to proceed to ask which one, and in that case we should go further — like reformulating the prompt as a confirmation?

jeffstieler: Good point. I'm not too crazy about the whole string, so let's just leave it "items".

dechov:
Expanded the test to cover the whole flow of adding items to a package.

DanReyLop:
@dechov Something happened with your commit history, looks like you rebased and brought some extra commits to the branch. Could you please fix that or squash your changes? I'm trying to see if this PR will conflict with https://github.com/Automattic/woocommerce-services/pull/1051, the last of us to merge will probably need to fix some nasty conflicts (I'll help you if I merge first).

dechov:
@DanReyLop I tried removing those extra commits — as a rule should I not be rebasing on master? Also, any idea why the Travis build is failing?

DanReyLop:
> I tried removing those extra commits — as a rule should I not be rebasing on master?

Yes, you should rebase on `master`, but here clearly something went wrong. If you create `branch-1` off `master`, and then you create `branch-2` off `branch-1`, when you rebase `branch-2` on `master` you'll drag with it all the `branch-1` commits, so maybe that's what it happened here. Anyway, now you fixed it :)

I see now that there are probably going to be some conflicts, but nothing too crazy.

> any idea why the Travis build is failing?

Looking at the build result: https://travis-ci.org/Automattic/woocommerce-services/builds/257797571
Only 1 instance of the tests is failing. Normally that means it's just caused by some instability in the tests, and if you retry the build it should pass. It would be OK if you merged with Travis in that state.
If only some of the test instances are failing, as a rule of thumb just check if there's something common about them. For example, once I commited something that was only valid in PHP 5.3+, so the instances that tested in PHP 5.2 failed spectacularly, that was a clear giveaway :)

PR #1093, opened 2017-07-18
Removed sprintf-js from the list of dependencies

marcinbot:
As a part of bringing WCS in line with Calypso we want to revise the non-dev dependencies we use. This will allow to bring our components over to Calypso more easily.

`sprintf` was a low hanging fruit and can be replaced either by built-in JS methods or `i18n-calypso` methods.

To test:
* verify that the labels load
* tracking links should be of a correct format
* label refund can be requested
* package names are shown correctly in the label modal (package list on the left in the packages step, and package names above the price dropdown in the rates step)

### DIFF
--- a/client/shipping-label/state/actions.js
+++ b/client/shipping-label/state/actions.js
@@ -2,7 +2,6 @@
 /**
  * External dependencies
  */
-import { sprintf } from 'sprintf-js';
 import { translate as __ } from 'i18n-calypso';
 import _ from 'lodash';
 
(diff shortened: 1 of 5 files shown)


DanReyLop:
You need to generate it with `npm shrinkwrap --dev`, it's important since at this time the `devDependencies` and the `dependencies` are badly classified so we can't have a `shkinrwrap` covering only arbitrary parts of our dependency tree.

Issue #936, opened 2017-03-21
Fix table header overflow on mobile


RiaanKnoetze:
![http://cld.wthms.co/mkO6](http://cld.wthms.co/mkO6+)
Full Image: http://cld.wthms.co/mkO6+

When creating a new package on a mobile device, there are some slight CSS overflow issues. This could potentially be exacerbated with localised text.

PR #1094, opened 2017-07-19
Appearance and alignment tweaks in WC Settings

dechov:
- Slightly increase maximum width of `.wcc-root` to match Calypso form width
- Change 'Shipping Labels' to 'Shipping labels' for consistency
- Increase margin above packages and labels settings forms
- Add margin to form section header in Packages settings
- Improve alignment and wrapping of 'Custom packages' table headers (should fix https://github.com/Automattic/woocommerce-services/issues/936)

Before | After
-- | --
screen shot 2017-07-18 at 9 25 51 pm | screen shot 2017-07-18 at 9 26 28 pm

- Improve alignment of checkboxes to text
- Avoid 'Edit package' dialog flickering to 'Add a package' dialog upon closing

Before:
packages-before

After:
packages-after

### DIFF
--- a/client/packages/views/packages-list.js
+++ b/client/packages/views/packages-list.js
@@ -68,7 +68,9 @@ const PackagesList = ( { packages, dimensionUnit, editable, selected, serviceId,
 				<FormLegend className="package-type">{ __( 'Type' ) }</FormLegend>
 				<FormLegend className="package-name">{ __( 'Name' ) }</FormLegend>
 				<FormLegend className="package-dimensions">
-					{ __( 'Dimensions (L x W x H)' ).replace( /\(.+\)/g, ( d ) => d.replace( / /g, '\u00a0' ) ) }
+					<span className="package-dimensions-title">{ __( 'Dimensions' ) }</span>
+					{ ' ' }
+					<span className="package-dimensions-format">{ __( '(L x W x H)' ) }</span>
 				</FormLegend>
 				{ editable ? <FormLegend className="package-actions" /> : null }
 			</div>
--- a/client/packages/views/packages-list.js
+++ b/client/packages/views/packages-list.js
@@ -64,10 +64,13 @@ const PackagesList = ( { packages, dimensionUnit, editable, selected, serviceId,
 	return (
 		<FormFieldset className="wcc-shipping-packages-list">
 			<div className="wcc-shipping-packages-list-header">
-				<FormLegend className="package-actions" />
+				{ editable ? null : <FormLegend className="package-actions" /> }
 				<FormLegend className="package-type">{ __( 'Type' ) }</FormLegend>
 				<FormLegend className="package-name">{ __( 'Name' ) }</FormLegend>
-				<FormLegend className="package-dimensions">{ __( 'Dimensions (L x W x H)' ) }</FormLegend>
+				<FormLegend className="package-dimensions">
+					{ __( 'Dimensions (L x W x H)' ).replace( /\(.+\)/g, ( d ) => d.replace( / /g, '\u00a0' ) ) }
+				</FormLegend>
+				{ editable ? <FormLegend className="package-actions" /> : null }
 			</div>
 			{ renderList() }
 		</FormFieldset>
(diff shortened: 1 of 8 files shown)


dechov:
Things to review:
- [Custom package dialog state management no longer cleans up after itself](https://github.com/Automattic/woocommerce-services/pull/1094/files#diff-4cd96525cace674b25800165b5d660c9)
- [Swapping space character to avoid breaking inside parentheses](https://github.com/Automattic/woocommerce-services/pull/1094/files#diff-685792d574c1f48f24907f9442907011R71)
- Anything else...

dechov:
Thanks @jeffstieler — fixed this by resetting when you reopen the form, if it the package was added. Also found and fixed a bug in which, while the previous "type" value was preserved, the dropdown wasn't reflecting it. (Test: select "Envelope", close the dialog, and open it again.)

client/packages/views/packages-list.js
```diff
@@ -64,10 +64,13 @@ const PackagesList = ( { packages, dimensionUnit, editable, selected, serviceId,
 	return (
 		<FormFieldset className="wcc-shipping-packages-list">
 			<div className="wcc-shipping-packages-list-header">
-				<FormLegend className="package-actions" />
+				{ editable ? null : <FormLegend className="package-actions" /> }
 				<FormLegend className="package-type">{ __( 'Type' ) }</FormLegend>
 				<FormLegend className="package-name">{ __( 'Name' ) }</FormLegend>
-				<FormLegend className="package-dimensions">{ __( 'Dimensions (L x W x H)' ) }</FormLegend>
+				<FormLegend className="package-dimensions">
+					{ __( 'Dimensions (L x W x H)' ).replace( /\(.+\)/g, ( d ) => d.replace( / /g, '\u00a0' ) ) }
```
jeffstieler: Can we accomplish this using CSS instead? https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

dechov: Ah, didn't realize before but `white-space: nowrap` would work. Would have to split "Dimensions" and the rest into two separate spans, assuming we want to keep the string intact— is that preferable to what I did? If so, I can make that change.

jeffstieler: I think spans are preferable to string replacement. Let's give each span a distinct class so we can handle RTL languages in the future.

dechov: Sounds good. Think it's valid to assume we should do that dynamically to preserve the translation `__( 'Dimensions (L x W x H)' )`?

jeffstieler: No, we should probably just break up the translated strings and put them in separate spans.

dechov: Thanks— made that change.

client/packages/views/packages-list.js
```diff
@@ -64,10 +64,15 @@ const PackagesList = ( { packages, dimensionUnit, editable, selected, serviceId,
...
 		<FormFieldset className="wcc-shipping-packages-list">
 			<div className="wcc-shipping-packages-list-header">
-				<FormLegend className="package-actions" />
+				{ editable ? null : <FormLegend className="package-actions" /> }
 				<FormLegend className="package-type">{ __( 'Type' ) }</FormLegend>
 				<FormLegend className="package-name">{ __( 'Name' ) }</FormLegend>
-				<FormLegend className="package-dimensions">{ __( 'Dimensions (L x W x H)' ) }</FormLegend>
+				<FormLegend className="package-dimensions">
+					<span className="package-dimensions-title">{ __( 'Dimensions' ) }</span>
+					{ ' ' }
```
jeffstieler: Won't this space break RTL? Can we _(CSS newbie warning)_ add a `1 em` margin to achieve the same effect?

dechov: It's still just inline text, and it doesn't seem so _(RTL rookie alert)_: https://codepen.io/ahmadalfy/pen/arqLy?editors=1000
(Adapted from https://alfy.me/2014/07/26/lets-talk-about-rtl.html)
Would a margin be safer or otherwise better?

jeffstieler: Thanks for making me challenge an (incorrect) assumption I held. 👍 Looks like we can leave the space.

PR #1096, opened 2017-07-20
Maintain fixed precision of weight values when moving items among packages

dechov:
Avoids floating point arithmetic artifacts (such as `9.879999999999999`) being stored and displayed in the Total Weight value when moving items around in the shipping label dialog.

Before:
precision-before

After:
precision-after

### DIFF
--- a/client/shipping-label/state/reducer.js
+++ b/client/shipping-label/state/reducer.js
@@ -60,7 +60,8 @@ const generateUniqueBoxId = ( keyBase, boxIds ) => {
 	}
 };
 
-const toFixed = ( n ) => +n.toFixed( 6 );
+/* Irons out floating point arithmetic artifacts */
+const round = ( n ) => _.round( n, 8 );
 
 const reducers = {};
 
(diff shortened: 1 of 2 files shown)


client/shipping-label/state/reducer.js
```diff
@@ -60,6 +60,8 @@ const generateUniqueBoxId = ( keyBase, boxIds ) => {
 	}
 };
 
+const toFixed = ( n ) => +n.toFixed( 6 );
```
marcinbot: We could probably use lodash's [`_.round(<number>,2)`](https://lodash.com/docs/4.17.4#round) here, but I'm ok with this implementation.

dechov: Ah, yeah, I'd forgotten about that. I'm OK with both too, but also prefer `_.round`

dechov:
Looks like supporting 6 decimal places would probably be borderline acceptable, since that's the limit of precision I'm seeing for consumer [high precision scales](https://www.amazon.com/s/ref=nb_sb_noss_1?url=search-alias%3Dgarden&field-keywords=high+precision+scale), but I raised it up to 8.

PR #1104, opened 2017-07-26
Add support for asynchronous requests

DanReyLop:
To test this, you'll need to checkout the `add/async-requests` branch in the server.

In this PR:
* Refactored the `Purchase Label` PHP endpoint to make the transition a bit clearer.
* Refactored the `Purchase Label` JS Redux action so it uses the new `api` module instead of the hacky `saveForm`.
* Add support in PHP to interpret when a REST response is `async`, implementing the whole flow described in the server-side PR.
* Add polling behaviour to the `api` module so the result is transparent to the rest of the JS code.

To test, just buy a label with the Chrome `Network` tab open. The HTTP requests & responses that happen during the `Purchase` action will speak from themselves (check out how the `async` flow works in the server-side PR first, to understand it more easily). The UI should behave exactly as before.

### DIFF
--- a/classes/class-wc-rest-connect-base-controller.php
+++ b/classes/class-wc-rest-connect-base-controller.php
@@ -8,6 +8,14 @@ if ( class_exists( 'WC_REST_Connect_Base_Controller' ) ) {
 	return;
 }
 
+class WC_Connect_Api_Exception extends Exception {
+	public $response;
+	public function __construct( $response ) {
+		$this->response = $response;
+		parent::__construct();
+	}
+}
+
 abstract class WC_REST_Connect_Base_Controller extends WP_REST_Controller {
 
 	/**
@@ -63,14 +71,44 @@ abstract class WC_REST_Connect_Base_Controller extends WP_REST_Controller {
 		if ( ! defined( 'DONOTCACHEPAGE' ) ) {
 			define( 'DONOTCACHEPAGE', true ); // Play nice with WP-Super-Cache
 		}
-		return $this->get( $request );
+		try {
+			return $this->get( $request );
+		} catch ( WCC_Api_Exception $error ) {
+			return $error->response;
+		}
 	}
 
 	public function post_internal( $request ) {
 		if ( ! defined( 'DONOTCACHEPAGE' ) ) {
 			define( 'DONOTCACHEPAGE', true ); // Play nice with WP-Super-Cache
 		}
-		return $this->post( $request );
+		try {
+			return $this->post( $request );
+		} catch ( WC_Connect_Api_Exception $ex ) {
+			return $ex->response;
+		}
+	}
+
+	protected function api_request( $method /*, ...$args */ ) {
+		$args = array_slice( func_get_args(), 1 );
+		$response = call_user_func_array( array( $this->api_client, $method ), $args );
+
+		if ( is_wp_error( $response ) ) {
+			$error = new WP_Error(
+				$response->get_error_code(),
+				$response->get_error_message(),
+				array( 'message' => $response->get_error_message() )
+			);
+			$this->logger->debug( $error, get_class( $this ) );
+			throw new WC_Connect_Api_Exception( $error );
+		}
+
+		if ( isset( $response->async_token ) || isset( $response->async_status ) ) {
+			// Not an error per-se, but we want to interrupt further processing of the response at this point
+			throw new WC_Connect_Api_Exception( $response );
+		}
+
+		return $response;
 	}
 
 	/**
--- a/classes/class-wc-rest-connect-base-controller.php
+++ b/classes/class-wc-rest-connect-base-controller.php
@@ -73,7 +73,7 @@ abstract class WC_REST_Connect_Base_Controller extends WP_REST_Controller {
 		}
 		try {
 			return $this->get( $request );
-		} catch ( WCC_Api_Exception $error ) {
+		} catch ( WC_Connect_Api_Exception $error ) {
 			return $error->response;
 		}
 	}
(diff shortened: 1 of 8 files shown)


classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -19,28 +19,43 @@ public function post( $request ) {
...
 			);
-			$this->logger->debug( $error, __CLASS__ );
-			return $error;
 		}
 
-		$label_ids = array();
-		$purchased_labels_meta = array();
-		$package_lookup = $this->settings_store->get_package_lookup();
+		$settings[ 'async' ] = true; // TODO: only make it async if the request comes from the Jetpack proxy?
+		$response = $this->api_request( 'send_shipping_label_request', $settings );
```
DanReyLop: The only real changes needed for an endpoint to start accepting `async requests` are these:
* Replace `$response = $this->api_client->method_name( $arg1, $arg2 );` for `$response = $this->api_request( 'method_name', $arg1, $arg2 );`
* Make sure that no server state is changed **before** calling `$this->api_request()`.
* Make sure that the `async_token` parameter sent by the client will end up in the `$body` of the HTTP request.

Handling program flow using exceptions is a bit hacky, and I don't feel super proud of it, but it allows the individual endpoint handlers to remain virtually unchanged, hiding all the async-related complexity in the parent class.

client/api/request.js
```diff
@@ -27,22 +29,39 @@ const _request = ( url, data, method ) => {
...
-			}
+				if ( json.success ) {
+					return json;
+				}
 
-			throw json;
+				if ( 'rest_cookie_invalid_nonce' === json.code ) {
+					window.persistState = true;
+					alert( __( 'There was a problem saving your settings. Please try again after the page is reloaded.' ) );
+					location.reload();
```
DanReyLop: Hum... edge case indeed, if there is polling going on every few seconds, the `nonce` should never expire, right?

Resuming polling after refresh would be tricky because right now all the polling logic is tied to this file, it isn't added to the Redux state or persisted anywhere.

client/api/request.js
```diff
@@ -27,22 +29,39 @@ const _request = ( url, data, method ) => {
...
+			return parseJson( response ).then( ( json ) => {
+				if ( json.async_token ) {
+					request.body = JSON.stringify( { ...( data || {} ), async_token: json.async_token } );
+					return poll( performRequest );
+				}
+				switch ( json.async_status ) {
+					case 'IN_PROGRESS':
+						return poll( performRequest );
+					case 'NOT_FOUND':
+						throw { code: 'async_token_not_found', message: __( 'The request expired' ) }; // TODO: How to phrase this error?
```
DanReyLop: This is equivalent to a "503" error or something like that. It will happen if the server had the async request `IN_PROGRESS` for a few minutes and then it was purged from memory, so it can only be caused if the purchase failed in some fatal way, or for some reason the client hasn't polled for a few minutes.

I guess we could go with a generic `An error occured, please try again` (no need to refresh), what do you think?

classes/class-wc-rest-connect-base-controller.php
```diff
@@ -63,14 +71,44 @@ public function get_internal( $request ) {
...
 		}
-		return $this->post( $request );
+		try {
+			return $this->post( $request );
+		} catch ( WC_Connect_Api_Exception $ex ) {
+			return $ex->response;
+		}
+	}
+
+	protected function api_request( $method /*, ...$args */ ) {
```
DanReyLop: Oh, that's to indicate that the function receives additional arguments (see the next line). I've always found quite confusing if the function for example is declared as receiving 1 argument but then with `func_get_args` it turns out that it receives more.

client/api/request.js
marcinbot: Yep, sounds good.

classes/class-wc-rest-connect-base-controller.php
marcinbot: Ok, I'm fine with leaving this in.

classes/class-wc-rest-connect-base-controller.php
```diff
@@ -63,14 +71,44 @@ public function get_internal( $request ) {
...
+				$response->get_error_code(),
+				$response->get_error_message(),
+				array( 'message' => $response->get_error_message() )
+			);
+			$this->logger->debug( $error, get_class( $this ) );
+			throw new WC_Connect_Api_Exception( $error );
+		}
+
+		if ( isset( $response->async_token ) || isset( $response->async_status ) ) {
+			// Not an error per-se, but we want to interrupt further processing of the response at this point
```
DanReyLop: There's no more processing *in this function*, but there is in its caller. The endpoint calls `$this->api_request()`, and after that it processes the response (which would be the list of purchased labels). So throwing an exception here ensures that the rest of the endpoint code won't be run. That's why I said the this is "using exceptions as flow control", which is not very good practice. It allows the endpoint to work without any changes though.

jeffstieler: Gotcha.

What do you think about refactoring other REST controllers to use `api_request()` so they can benefit from the built in `WP_Error` handling? It looks like a good opportunity to remove some redundant code.

DanReyLop: > What do you think about refactoring other REST controllers to use api_request() so they can benefit from the built in WP_Error handling? It looks like a good opportunity to remove some redundant code.

That was the intention, but it would add more noise to this PR.

> We can leave the refactor for later work, which gives us time to rethink the exception approach as well.

👍

client/api/request.js
DanReyLop: What do you think about this solution: https://github.com/Automattic/woocommerce-services/pull/1104/commits/c52ba797b6d8cb6a2f1efd16f21a2afd1b891628

If the user refreshes the page mid-polling, the response will be lost in limbo, but at least now the user will have a confirmation prompt before he navigates away. We can try to make it more robust down the road, but I believe this is enough for now. Thoughts?

DanReyLop:
The whole `async` approach was flawed because we can't keep global state in the server. Closing this, altough some of the refactors done in this PR will be used.

PR #1108, opened 2017-07-31
Added REST endpoint for accepting TOS

marcinbot:
Related to #1107

This adds a REST endpoint that accepts the TOS when called. Calypso will call it on initial store setup.

I tested it with Postman.

### DIFF
--- a/classes/class-wc-rest-connect-tos-controller.php
+++ b/classes/class-wc-rest-connect-tos-controller.php
@@ -1,0 +1,48 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( class_exists( 'WC_REST_Connect_Tos_Controller' ) ) {
+	return;
+}
+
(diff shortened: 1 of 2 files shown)


classes/class-wc-rest-connect-tos-controller.php
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	public function post( $request ) {
+		$settings = $request->get_json_params();
+
+		if ( ! $settings || ! isset( $settings[ 'accepted' ] ) || ! $settings[ 'accepted' ] ) {
+			return new WP_REST_Response( array(
+				'success'  => false,
+				'accepted' => WC_Connect_Options::get_option( 'tos_accepted' ),
+			), 200 );
```
DanReyLop: Can this be a `WP_Error` instead? I think it should be something like a `400 Bad Request`, since the client should never send a request without `accepted: true`.

PR #1112, opened 2017-08-04
PDF printing - Fixed a conflict with plugins that try to pass arrays to kses

### LINKED ISSUE #1111: Label print dialog fails when certain plugins and Jetpack modules are activated
To repro:
* have the `shortcodes` jetpack module activated
* activate a plugin that always calls `kses` on all `get` parameters
* try to print a label
* weep

screen shot 2017-08-04 at 17 54 38

@v18 @dechov

marcinbot:
Fixes #1111 

This was caused by the label IDs and their captions being passed as arrays in GET string. Some plugins try to call `kses` on all query params. With Jetpack, this could produce a warning if one of the params has been already converted to an array.

To test:
* check that a single label can be printed
* check that multiple labels can be printed
* optional
  * Activate the `shortcode embeds` module at `/wp-admin/admin.php?page=jetpack_modules`
  * Install an offending plugin (see slack channel for details p1501775680572853-slack-hydra )
  * verify that everything still prints

@melindahelt @v18

### DIFF
--- a/classes/class-wc-rest-connect-shipping-label-print-controller.php
+++ b/classes/class-wc-rest-connect-shipping-label-print-controller.php
@@ -18,9 +18,9 @@ class WC_REST_Connect_Shipping_Label_Print_Controller extends WC_REST_Connect_Ba
 		$params[ 'paper_size' ] = $raw_params[ 'paper_size' ];
 		$this->settings_store->set_preferred_paper_size( $params[ 'paper_size' ] );
 
-		$label_ids = isset( $raw_params[ 'label_id_csv' ] ) && $raw_params[ 'label_id_csv' ] ? explode( ',', $raw_params[ 'label_id_csv' ] ) : array();
+		$label_ids = ! empty( $raw_params[ 'label_id_csv' ] ) ? explode( ',', $raw_params[ 'label_id_csv' ] ) : array();
 		$n_label_ids = count( $label_ids );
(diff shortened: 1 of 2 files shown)


classes/class-wc-rest-connect-shipping-label-print-controller.php
```diff
@@ -18,8 +18,10 @@ public function get( $request ) {
 		$params[ 'paper_size' ] = $raw_params[ 'paper_size' ];
 		$this->settings_store->set_preferred_paper_size( $params[ 'paper_size' ] );
 
-		$n_label_ids = isset( $raw_params[ 'label_ids' ] ) ? count( $raw_params[ 'label_ids' ] ) : 0;
-		$n_captions = isset( $raw_params[ 'captions' ] ) ? count( $raw_params[ 'captions' ] ) : 0;
+		$label_ids = isset( $raw_params[ 'label_id_csv' ] ) && $raw_params[ 'label_id_csv' ] ? explode( ',', $raw_params[ 'label_id_csv' ] ) : array();
```
dechov: Could `! empty` be used instead here, or is there a reason this is preferable?

client/lib/pdf-label-utils/index.js
```diff
@@ -42,8 +42,8 @@ const _getPDFURL = ( paperSize, labels, baseURL, nonce ) => {
 	const params = {
```
dechov: Worth leaving a comment explaining why it's not just using an array, and warning against using QS arrays in the future?

jeffstieler:
I'm wondering why this isn't a fix in Jetpack: https://github.com/Automattic/woocommerce-services/issues/1111#issuecomment-322069730

PR #1119, opened 2017-08-08
Checking the Jetpack master user in account settings REST endpoint

marcinbot:
This is part of the work on automattic/wp-calypso#16958

I made the following changes in this PR:
* moved all Jetpack utils to a single static class `WC_Connect_Jetpack`
* in `class-wc-rest-connect-account-settings-controller.php` I perform the same checks as when the account settings page is rendered to verify if the current user is a master user and if they can manage payments

To test:
* everything should work as before, especially on the accounts page
* in dev mode, Jetpack will not return a master user
* I tested the API changes by uploading the updated version of the plugin to my AT site. I got the following in the response when logged in as non-master: `"can_manage_payments":false,"master_user_name":"botmarcin","master_user_login":"botmarcin"`, which is as expected

### DIFF
--- a/classes/class-wc-rest-connect-account-settings-controller.php
+++ b/classes/class-wc-rest-connect-account-settings-controller.php
@@ -71,6 +71,6 @@ class WC_REST_Connect_Account_Settings_Controller extends WC_REST_Connect_Base_C
 		global $current_user;
 		$master_user = WC_Connect_Jetpack::get_master_user();
 		return WC_Connect_Jetpack::is_development_mode() ||
-			( is_a( $master_user, 'WP_User' ) && $current_user->ID == $master_user->ID );
+			( is_a( $master_user, 'WP_User' ) && $current_user->ID === $master_user->ID );
 	}
 }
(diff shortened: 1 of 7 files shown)


classes/class-wc-rest-connect-account-settings-controller.php
```diff
@@ -58,4 +67,10 @@ public function post( $request ) {
 		return new WP_REST_Response( array( 'success' => true ), 200 );
 	}
 
+	private function can_user_manage_payment_methods() {
+		global $current_user;
+		$master_user = WC_Connect_Jetpack::get_master_user();
+		return WC_Connect_Jetpack::is_development_mode() ||
+			( is_a( $master_user, 'WP_User' ) && $current_user->ID == $master_user->ID );
```
v18: @nabsul @marcinbot I'm not sure that the changes discussed here ever made it through in the merged PR? Maybe we should open a new PR to update?

marcinbot: Do you mean the equality operator? I can see it being updated in `master`: https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-rest-connect-account-settings-controller.php#L74

PR #1121, opened 2017-08-09
Label Modal - using the api module instead of saveForm

### LINKED ISSUE #1103: Label Modal - use the api module instead of saveForm
* Instead of using `saveForm`, use the methods exposed by `api`, similar to Packages and Account Settings
* Allow the page to dynamically fetch all the required data when not available

CC @DanReyLop

marcinbot:
Fixes #1103

This is a prerequisite to start moving the label modal component to Calypso. I also cleaned up some of the code in the last two commits.

To test:
* everything should work correctly:
  * address normalization
  * label rates request
  * label purchase and print
  * label refund
* to test the dynamic fetch, replace `$payload = $this->get_label_payload( $order );` with `$payload = array()` in `class-wc-connect-shipping-label.php`
  * a spinner should appear while the data is being requested
  * everything should work once the data has been loaded

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -473,15 +473,10 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		public function meta_box( $post ) {
 			$order = wc_get_order( $post );
 
-			$order_id = WC_Connect_Compatibility::instance()->get_order_id( $order );
-
 			$payload = $this->get_label_payload( $order );
 			if ( ! $payload ) {
 				return;
 			}
-
-			//pass order ID as the only constant in the context, everything else can be fetched dynamically
-			$payload[ 'orderId' ] = $order_id;
 
 			do_action( 'enqueue_wc_connect_script', 'wc-connect-create-shipping-label', $payload );
 		}
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -452,12 +452,6 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 
 			$order_id = WC_Connect_Compatibility::instance()->get_order_id( $order );
 			$payload = array(
-				'purchaseURL'             => get_rest_url( null, '/wc/v1/connect/label/' . $order_id ),
-				'addressNormalizationURL' => get_rest_url( null, '/wc/v1/connect/normalize-address' ),
-				'getRatesURL'             => get_rest_url( null, '/wc/v1/connect/label/' . $order_id . '/rates' ),
-				'labelStatusURL'          => get_rest_url( null, '/wc/v1/connect/label/' . $order_id . '/%d' ),
-				'labelRefundURL'          => get_rest_url( null, '/wc/v1/connect/label/' . $order_id . '/%d/refund' ),
-				'labelsPrintURL'          => get_rest_url( null, '/wc/v1/connect/label/print' ),
 				'orderId'                 => $order_id,
 				'paperSize'               => $this->settings_store->get_preferred_paper_size(),
 				'formData'                => $this->get_form_data( $order ),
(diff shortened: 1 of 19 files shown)


classes/class-wc-connect-shipping-label.php
```diff
@@ -447,19 +447,13 @@ public function get_selected_payment_method() {
 			return null;
 		}
 
-		public function meta_box( $post ) {
+		public function get_label_payload( $post ) {
 			$order = wc_get_order( $post );
```
marcinbot: This also gets called with `$order_id` from the REST controller. `wc_get_order` takes either a post, order or id. I updated the argument name for clarity.

classes/class-wc-connect-shipping-label.php
```diff
@@ -470,6 +464,18 @@ public function meta_box( $post ) {
...
+		}
+
+		public function meta_box( $post ) {
+			$order = wc_get_order( $post );
+
+			$order_id = WC_Connect_Compatibility::instance()->get_order_id( $order );
+
+			$payload = $this->get_label_payload( $order );
+			//pass order ID as the only constant in the context, everything else can be fetched dynamically
+			$payload[ 'orderId' ] = $order_id;
```
marcinbot: The only argument that the AJAX requests will absolutely need to know about is the orderId. Some requests will try to get the payload (not when we're preloading the data, but it still makes the switch easier), so the id can't be located inside the payload.

In Calypso it will be possible to get the order id in a different way.

client/apps/shipping-label/state/actions.js
```diff
@@ -463,15 +493,13 @@ export const updatePaperSize = ( value ) => {
 };
 
 export const purchaseLabel = () => ( dispatch, getState, context ) => {
-	const { purchaseURL, getRatesURL, addressNormalizationURL, nonce } = context;
+	const { orderId } = context;
```
marcinbot: > Alternatively, why are we keeping it an object if all that it contains is the order id?

This is how it gets initialized for all apps in `main.js`. Looks like as we move away from using `saveForm` the object will become obsolete but removing it is out of the scope of this PR.

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -24,6 +24,9 @@ public function __construct( WC_Connect_API_Client $api_client, WC_Connect_Servi
 	public function get( $request ) {
 		$order_id = $request[ 'order_id' ];
 		$payload = $this->shipping_label->get_label_payload( $order_id );
+		if ( ! $payload ) {
+			return new WP_Error( 404 );
```
jeffstieler: I don't believe this is the correct form for a 404. See the example here: https://developer.wordpress.org/rest-api/extending-the-rest-api/adding-custom-endpoints/#return-value

classes/class-wc-connect-shipping-label.php
jeffstieler: I'm not sure I understand. It looks like `orderId` is always in the returned payload from `get_label_payload()`, so what's the point in overwriting it here?

classes/class-wc-rest-connect-shipping-label-controller.php
marcinbot: I updated it and also updated other places where I used the same approach.

classes/class-wc-connect-shipping-label.php
marcinbot: Ok, I removed it. 

Just some background info: The main reason why I had it there was to make a switch between preloaded data and dynamic fetch easier. In other parts of the plugin (packages, accounts) all it takes to switch to the dynamic fetch is removing an argument from the function call. Here we always need to pass orderId

classes/class-wc-rest-connect-tos-controller.php
```diff
@@ -23,7 +23,7 @@ public function post( $request ) {
 		$settings = $request->get_json_params();
 
 		if ( ! $settings || ! isset( $settings[ 'accepted' ] ) || ! $settings[ 'accepted' ] ) {
-			return new WP_Error( 400 );
+			return new WP_Error( 'bad_request', __( 'Bad request' ), array( 'status' => 400 ) );
```
jeffstieler: Are these strings found in core WP, or are they missing the `woocommerce-services` domain?

marcinbot: My bad, I added the domain

classes/class-wc-connect-shipping-label.php
```diff
@@ -470,6 +467,17 @@ public function meta_box( $post ) {
...
 			$payload['storeOptions'] = $store_options;
 
+			return $payload;
+		}
+
+		public function meta_box( $post ) {
+			$order = wc_get_order( $post );
+
+			$payload = $this->get_label_payload( $order );
+			if ( ! $payload ) {
```
jeffstieler: Should this be testing for `false`? How would one test the async load now?

classes/class-wc-connect-shipping-label.php
```diff
@@ -473,16 +473,11 @@ public function get_label_payload( $post_order_or_id ) {
...
 
-			$order_id = WC_Connect_Compatibility::instance()->get_order_id( $order );
-
 			$payload = $this->get_label_payload( $order );
 			if ( ! $payload ) {
 				return;
 			}
 
-			//pass order ID as the only constant in the context, everything else can be fetched dynamically
-			$payload[ 'orderId' ] = $order_id;
```
jeffstieler: It looks like this change broke the async load of label data - can the order id be retrieved elsewhere in the client code?

classes/class-wc-connect-shipping-label.php
marcinbot: It should never be hit, since we initialize the script only on the existing order pages.

To test the async load replace `$payload` in `do_action` call below with `array( 'orderId' => WC_Connect_Compatibility::instance()->get_order_id( $order ) )`

classes/class-wc-connect-shipping-label.php
marcinbot: Not in a "proper" way. But I described how to restore the async load above. We do need to provide the script with `orderId` in some way.

PR #1132, opened 2017-08-22
Support for async label purchases

### LINKED ISSUE #1120: Labels - add asynchronous purchases
As discussed in p7bbVw-1WI-p2, when a label is purchased, let's follow these steps:
* send an initial request to the server
* create a database entry for a label and mark it as "in progress"
* return the in-progress label to the client and continue with the purchase
* the client polls for the label status
* once the label status is updated, update the label in the client

Also handle the legacy synchronous requests.

marcinbot:
Fixes #1120 

To test:
* Checkout the server PR: Automattic/woocommerce-connect-server#959
* Purchase a single label
* Purchase multiple labels
* Hack the server by throwing an error at any step in `lib/shipping/handler.js` and verify that the errors are handled properly in the client
* ~~If the page is refreshed during the purchase or after an error, this should show in the label history~~ (will be handled in #1135)

### DIFF
--- a/client/apps/shipping-label/state/actions.js
+++ b/client/apps/shipping-label/state/actions.js
@@ -496,15 +496,12 @@ const purchaseLabelResponse = ( response, error ) => {
 	return { type: PURCHASE_LABEL_RESPONSE, response, error };
 };
 
-const handleLabelPurchaseError = ( dispatch, getState, error, orderId, labelCount ) => {
+const handleLabelPurchaseError = ( dispatch, getState, error, orderId ) => {
 	dispatch( purchaseLabelResponse( null, true ) );
 	if ( 'rest_cookie_invalid_nonce' === error ) {
 		dispatch( exitPrintingFlow( true ) );
 	} else {
-		dispatch( NoticeActions.errorNotice( __(
-			'Shipping label didn\'t purchase. Please try again.',
-			'Some labels for this order didn\'t purchase. Please try again.',
-			{ count: labelCount } ) ) );
+		dispatch( NoticeActions.errorNotice( error ) );
 		//re-request the rates on failure to avoid attempting repurchase of the same shipment id
 		dispatch( clearAvailableRates() );
 		getLabelRates( dispatch, getState, _.noop, { orderId } );
@@ -514,7 +511,7 @@ const handleLabelPurchaseError = ( dispatch, getState, error, orderId, labelCoun
 const pollForLabelsPurchase = ( dispatch, getState, orderId, labels ) => {
 	const errorLabel = _.find( labels, { status: 'PURCHASE_ERROR' } );
 	if ( errorLabel ) {
-		handleLabelPurchaseError( dispatch, getState, 'purchase_error', orderId, labels.length );
+		handleLabelPurchaseError( dispatch, getState, errorLabel.error , orderId );
 		return;
 	}
 
@@ -527,7 +524,7 @@ const pollForLabelsPurchase = ( dispatch, getState, orderId, labels ) => {
 
 			Promise.all( statusTasks )
 				.then( ( pollResponse ) => pollForLabelsPurchase( dispatch, getState, orderId, pollResponse ) )
-				.catch( ( pollError ) => handleLabelPurchaseError( dispatch, getState, pollError, orderId, labels.length ) );
+				.catch( ( pollError ) => handleLabelPurchaseError( dispatch, getState, pollError, orderId ) );
 		}, 1000 );
 		return;
 	}
@@ -574,7 +571,6 @@ const pollForLabelsPurchase = ( dispatch, getState, orderId, labels ) => {
 export const purchaseLabel = () => ( dispatch, getState, { orderId } ) => {
 	let error = null;
 	let labels = null;
-	let labelCount = 1;
 
 	const setError = ( err ) => error = err;
 	const setSuccess = ( json ) => {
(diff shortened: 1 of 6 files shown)


jeffstieler:
The label purchase error flow is a bit off to me right now. I introduced an error just before label purchase on the server as low level as I could, in `purchaseLabelAsync()` (`/lib/shipping/usps/index.js`), rejecting the promise with:

```javascript
{
	code: 'purchase_failed',
	message: 'There was an error purchasing the label',
}
```

While this error message is never surfaced, the merchant now sees a human readable error stating there was a problem purchasing labels, and that refunds are being issued.

purchase-error-flow

Upon refreshing the page - I see that Label #1 was "purchased" but then have a tooltip repeating the refund-centric error message.

Let's bring @olaolusoga in here to evaluate flow. When label purchases failed previously we weren't surfacing the refunds to the merchant, but this PR changes that. I'm not sure we want to show these failed purchases in the order details screen or even mention refunds.

olaolusoga:
>The label purchase error flow is a bit off to me right now

Most likely due to the messaging. The error message "Some label purchased for this order failed, issuing refund" is stating a problem, but doesn't offer a solution, or second attempt, leaving money on the table & the user unhappy. If we change the copy to "Some label didn't go through, please try again" with "try again" underlined, that helps address a solution, and keeps the user on track to purchase the labels.

We could take it a step further and introduce more context while in the modal. At the moment, we're purchasing labels for the store owner, but the last time they saw what they were purchasing was in the "packaging" step. If we create a tally under "paper size" breaking down the labels and product within each, that should also help with this scenario. Where if it fails, we can show them which ones, and ask for them to "retry" while in modal, removing the need to add it to the order details. (I can mock up visual, if we want to do this)

Let's say a user closes the modal without retrying, it makes sense to have it in the order details, but instead of "purchase error", and the repeated message in tool tip. We should give context to which items where part of the label order, and have a "Try Again" (replacing purchase error) link with the tooltip giving context to what they're about to try again "You order didn't complete. Please try purchasing this label again."

They succeed by purchasing all labels to fulfill their customer order, so we shouldn't be refunding. We should be giving them a way to get their label order through (Retry all label purchases till it completes).

jeffstieler:
screen shot 2017-08-24 at 3 02 35 pm

The refundable amount isn't coming through anymore.

client/apps/shipping-label/state/actions.js
```diff
@@ -492,6 +492,91 @@ export const updatePaperSize = ( value ) => {
...
+		}
+		return;
+	}
+
+	dispatch( NoticeActions.successNotice( noticeText ) );
+};
+
 export const purchaseLabel = () => ( dispatch, getState, { orderId } ) => {
 	let error = null;
 	let response = null;
```
jeffstieler: Can this be renamed to `labels` now? It's not actually used for a response in `purchaseLabel()`, right?

client/apps/shipping-label/state/actions.js
```diff
@@ -492,6 +492,91 @@ export const updatePaperSize = ( value ) => {
 	};
 };
 
+const purchaseLabelResponse = ( response, error ) => {
+	return { type: PURCHASE_LABEL_RESPONSE, response, error };
+};
+
+const pollForLabelsPurchase = ( dispatch, getState, orderId, labels, error, showPrintDialog ) => {
```
jeffstieler: What do you think about changing this to expect an object instead of multiple arguments? I think it could make the calls easier to understand, especially around `labels`, `error`, and `showPrintDialog`. (like in `purchaseLabel()` and recursive calls inside the method.

client/apps/shipping-label/view.js
```diff
@@ -6,6 +6,7 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import Gridicon from 'gridicons';
 import { translate as __ } from 'i18n-calypso';
+import { filter } from 'lodash';
```
jeffstieler: I haven't caught this in previous reviews (sorry!) but I thought we determined that we could load all of lodash without increasing the bundle size?

marcinbot: We use `lodash` Babel plugin which allows us to that, but Calypso doesn't. I wanted to reduce the amount of refactoring once this is copied and also prepare for the future when we do actual code sharing (we might have to replace our `_` usages with manual cherrypicking or push for Calypso to start using the Babel plugin, both of which might prolong the wpcom Store development)

I can change it back here and then perform the necessary changes in Calypso. Let me know which one is preferred.

client/apps/shipping-label/view.js
```diff
@@ -230,7 +231,10 @@ class ShippingLabelRootView extends Component {
 	};
 
 	renderLabels = () => {
-		return this.props.shippingLabel.labels.map( this.renderLabel );
+		const labelsToRender = filter( this.props.shippingLabel.labels,
+			( label ) => 'PURCHASE_IN_PROGRESS' !== label.status && 'PURCHASE_ERROR' !== label.status );
```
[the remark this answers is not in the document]
marcinbot: No, by using the blacklist of states this can handle older labels with no `status` field.

jeffstieler:
Per our Slack convo: p1503939927000045-slack-hydra

Let's roll this back to displaying the error that comes from the server. We'll move the pluralized label purchase error string generation to the server side as well.

client/apps/shipping-label/state/actions.js
```diff
@@ -514,7 +511,7 @@ const handleLabelPurchaseError = ( dispatch, getState, error, orderId, labelCoun
 const pollForLabelsPurchase = ( dispatch, getState, orderId, labels ) => {
 	const errorLabel = _.find( labels, { status: 'PURCHASE_ERROR' } );
 	if ( errorLabel ) {
-		handleLabelPurchaseError( dispatch, getState, 'purchase_error', orderId, labels.length );
+		handleLabelPurchaseError( dispatch, getState, errorLabel.error , orderId );
```
dechov: Space before comma causing lint error here

PR #1136, opened 2017-08-24
Removed usages of saveForm from shipping service settings

### LINKED ISSUE #1102: Service Settings - use the api module instead of saveForm
* Instead of using `saveForm`, use the methods exposed by `api`, similar to Packages and Account Settings
* Allow the page to dynamically fetch all the required data when not available
* (optional) consider reducing data-drivenness to just the service list, and share the rest of the controls between services (could be a lot of work)

CC @DanReyLop

marcinbot:
Fixes #1102 

Replaces calls to `saveForm` with `api` on the shipping service settings page. With that, `saveForm` is no longer used in our codebase.

To test:
* the settings should work as before
  * to test the NUX notice on the shipping settings delete `wc_connect_nux_notices` row from `wp_usermeta` in the database
* to test this without data preloading replace `apply_filters( 'wc_connect_shipping_service_settings', array(), $method_id, $instance_id )` with `array( 'methodId' => $method_id, 'instanceId' => $instance_id )` in `woocommerce-services.php`


client/apps/settings/state/selectors/index.js
```diff
@@ -0,0 +1,47 @@
...
+	return form.storeOptions;
+};
+
+export const getFormLayout = ( state ) => {
+	const form = getShippingSettingsForm( state );
+	if ( ! form || ! form.loaded ) {
+		return null;
+	}
+	return form.layout;
+};
```
jeffstieler: What do you think about using `_.get()` for all of these selectors, or perhaps removing the use of it?

woocommerce-services.php
```diff
@@ -538,7 +539,7 @@ public function rest_api_init() {
 			$rest_account_settings_controller->register_routes();
 
 			require_once( plugin_basename( 'classes/class-wc-rest-connect-services-controller.php' ) );
-			$rest_services_controller = new WC_REST_Connect_Services_Controller( $this->api_client, $settings_store, $logger, $schemas_store );
+			$rest_services_controller = new WC_REST_Connect_Services_Controller( $this->api_client, $settings_store, $logger, $schemas_store, $this->nux );
```
jeffstieler: It looks like the addition of `WC_Connect_Nux` was missed in the `WC_REST_Connect_Services_Controller` class.

woocommerce-services.php
```diff
@@ -617,41 +618,53 @@ public function log_rest_api_errors( $response, $handler, $request ) {
...
+		 * as the form container is emitted into the body's HTML
+		 */
+		public function localize_and_enqueue_service_script( $method_id, $instance_id = false ) {
+			if ( ! function_exists( 'get_rest_url' ) ) {
+				return;
+			}
+
+			do_action( 'enqueue_wc_connect_script',
+				'wc-connect-service-settings',
+				apply_filters( 'wc_connect_shipping_service_settings', array(), $method_id, $instance_id ) );
```
jeffstieler: Same question about filter usage.

classes/class-wc-rest-connect-services-controller.php
```diff
@@ -16,11 +16,35 @@ class WC_REST_Connect_Services_Controller extends WC_REST_Connect_Base_Controlle
...
+
+		$service_schema = $this->service_schemas_store->get_service_schema_by_id_or_instance_id( $instance_id
+			? $instance_id
+			: $method_id );
+
+		if ( ! $service_schema ) {
+			return new WP_Error( 'schemas_not_found', __( 'Service schemas were not loaded', 'woocommerce-services' ), array( 'status' => 500 ) );
+		}
+
+		$payload = apply_filters( 'wc_connect_shipping_service_settings', array(
```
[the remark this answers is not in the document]
marcinbot: It's mostly about avoiding code duplication and being able to reference the objects initialized in the main class.

I think the reason why we have to use `wc_connect_service_admin_options` action in `WC_Connect_Shipping_Method` is that we can't pass objects to that class easily. So the simplest way to move the payload initialization out of the action hook and share it with the REST controller seemed to be the filter. 

Other options included creating a new class just for the payload init or having the init in the REST controller and then reference the controller method in `localize_and_enqueue_service_script()`.

I don't feel strongly about any of these options, so I'm open to suggestions.

PR #1137, opened 2017-08-24
stripe account setup (deferred + oauth)

nabsul:
This PR includes some basic UI to test the creation of Stripe deferred accounts and connecting a Stripe account via OAuth.

To Test:
- Checkout the stripe PR on the server side https://github.com/Automattic/woocommerce-connect-server/pull/960
- Install the Stripe gateway WooCommerce plugin
- Go to any admin page and there should be an admin notice to create/connect


classes/class-wc-connect-stripe.php
```diff
@@ -0,0 +1,87 @@
...
+		public function create_account( $email, $country ) {
+			$response = $this->api->create_stripe_account( $email, $country );
+			if ( is_wp_error( $response ) ) {
+				return $response;
+			}
+			return $this->save_stripe_keys( $response );
+		}
+
+		public function connect_oauth( $state, $code ) {
+			if ( $state !== get_transient( 'wcs_stripe_state' ) ) {
```
jeffstieler: Can you explain the reasoning behind this `state` verification? Does storing the state locally avoid having to implement a verification endpoint in our service layer?

If this should remain on the client, let's use the options API instead of transients since transients are never guaranteed to persist.

classes/class-wc-connect-stripe.php
```diff
@@ -0,0 +1,87 @@
...
+
+			if ( is_wp_error( $response ) ) {
+				return $response;
+			}
+
+			return $this->save_stripe_keys( $response );
+		}
+
+		private function save_stripe_keys( $result ) {
+			if ( ! isset( $result->accountId, $result->publishableKey, $result->secretKey ) ) {
```
jeffstieler: TIL `isset()` takes an arbitrary number of arguments!

classes/class-wc-rest-connect-stripe-oauth-controller.php
```diff
@@ -0,0 +1,41 @@
...
+		$this->stripe = $stripe;
+	}
+
+	public function post( $request ) {
+		$data = $request->get_json_params();
+
+		$response = $this->stripe->connect_oauth( $data['state'], $data['code'] );
+
+		if ( is_wp_error( $response ) ) {
+			error_log(json_encode($response));
```
jeffstieler: Please remove this `error_log()`.

classes/class-wc-connect-stripe.php
```diff
@@ -0,0 +1,87 @@
...
+		 */
+		private $logger;
+
+		public function __construct( WC_Connect_API_Client $client, WC_Connect_Options $options, WC_Connect_Logger $logger ) {
+			$this->api = $client;
+			$this->options = $options;
+			$this->logger = $logger;
+		}
+
+		public function get_settings( $return_url ) {
```
jeffstieler: Without a local UI, should this just be an OAuth init URL retrieval method to support API consumers?

classes/class-wc-rest-connect-stripe-account-controller.php
```diff
@@ -0,0 +1,46 @@
...
+	protected $rest_base = 'connect/stripe/account';
+	private $stripe;
+
+	public function __construct( WC_Connect_Stripe $stripe, WC_Connect_API_Client $api_client, WC_Connect_Service_Settings_Store $settings_store, WC_Connect_Logger $logger ) {
+		parent::__construct( $api_client, $settings_store, $logger );
+		$this->stripe = $stripe;
+	}
+
+	public function get( $request ) {
+		$payload = $this->stripe->get_settings();
```
jeffstieler: What's an example use of this endpoint? Doesn't `WC_Connect_Stripe::get_settings()` require a URL parameter?

classes/class-wc-rest-connect-stripe-settings-controller.php
```diff
@@ -0,0 +1,41 @@
...
+	private $stripe;
+
+	public function __construct( WC_Connect_Stripe $stripe, WC_Connect_API_Client $api_client, WC_Connect_Service_Settings_Store $settings_store, WC_Connect_Logger $logger ) {
+		parent::__construct( $api_client, $settings_store, $logger );
+		$this->stripe = $stripe;
+	}
+
+	public function post( $request ) {
+		$data = $request->get_json_params();
+		$settings = $this->stripe->get_settings( $data['returnUrl'] );
```
jeffstieler: Same question about this just providing an OAuth init URL for API consumers.

classes/class-wc-connect-stripe.php
```diff
@@ -0,0 +1,87 @@
...
+		private function save_stripe_keys( $result ) {
+			if ( ! isset( $result->accountId, $result->publishableKey, $result->secretKey ) ) {
+				return new WP_Error( 'Invalid credentials received from server' );
+			}
+
+			$is_test = false !== strpos( $result->publishableKey, '_test_' );
+			$prefix = $is_test ? 'test_' : '';
+
+			$option_name = 'woocommerce_stripe_settings';
+			$options = get_option( $option_name );
```
jeffstieler: Let's also enable the Stripe gateway here.

classes/class-wc-connect-stripe.php
nabsul: True, that ca be deleted. If we need it back, it should anyways be implemented in a different function.

classes/class-wc-connect-stripe.php
nabsul: We definitely want to store this state value temporarily. A transient isn't the right place for this since the state should be tied to the user. I'm thinking this is better placed in a session variable.

classes/class-wc-rest-connect-stripe-account-controller.php
nabsul: True. This was never used in my test UI because I was fetching everything I needed in the call for the oauth URL. 

This GET method should return the default email and country values that the UI could use for auto-filling the fields. Since we're not using those at the moment, I'll just delete this GET method for now.

classes/class-wc-rest-connect-stripe-settings-controller.php
```diff
@@ -0,0 +1,39 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( class_exists( 'WC_REST_Connect_Stripe_Settings_Controller' ) ) {
+	return;
+}
+
+class WC_REST_Connect_Stripe_Settings_Controller extends WC_REST_Connect_Base_Controller {
+	protected $rest_base = 'connect/stripe/settings';
```
jeffstieler: Let's change this to `connect/stripe/oauth-url` or similar and rename the controller class to suit.

nabsul:
@jeffstieler I haven't abandoned this PR :-). Thanks for the url fix. I just want to fix the other comment on enabling stripe, and also only working when the stripe plugin is installed and enabled.

I should find some time for it in the next couple of days.

jeffstieler:
> I haven't abandoned this PR :-).

🎉 @nabsul 

> I just want to fix the other comment on enabling stripe, and also only working when the stripe plugin is installed and enabled.

One thing I discovered (but not sure how to solve correctly) is that we're potentially squashing the Stripe extension's ability to set its own defaults thanks to [`WC_Settings_API ::init_settings()`](https://github.com/woocommerce/woocommerce/blob/master/includes/abstracts/abstract-wc-settings-api.php#L230) .. heads up :)

nabsul:
Alrighty, I think I resolved the outstanding issues, but I'm going to hand it over for anything else that needs to be done. Gotta start focusing on data! :-D 💾

classes/class-wc-rest-connect-stripe-oauth-connect-controller.php
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	public function post( $request ) {
+		$data = $request->get_json_params();
+
+		$response = $this->stripe->connect_oauth( $data['state'], $data['code'] );
+
+		if ( is_wp_error( $response ) ) {
+			$this->logger->debug( $response, __CLASS__ );
+			return new WP_REST_Response( array(
```
DanReyLop: We usually handle errors like this:
```php
			$response->add_data( array(
				'message' => $response->get_error_message(),
			), $response->get_error_code() );
			return $response;
```
See https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-rest-connect-shipping-label-refund-controller.php#L24-L31 for example.

Same for the other controllers.

classes/class-wc-rest-connect-stripe-oauth-connect-controller.php
```diff
@@ -0,0 +1,40 @@
...
+				'success'   => false,
+				'data'      => array(
+					'message' => $response->get_error_message(),
+				),
+			), 400 );
+		}
+
+		return new WP_REST_Response( array(
```
DanReyLop: It's enough to `return` the `array`, it's wrapped into a `WP_REST_Response(200)` by default.

classes/class-wc-connect-stripe.php
```diff
@@ -0,0 +1,106 @@
+<?php
```
DanReyLop: The `! defined( 'ABSPATH' )` check is missing.

PR #1138, opened 2017-08-25
Initialize REST API and check TOS acceptance per request

dechov:
Possibly temporary measure (pending **decision on whether we still need WCS TOS**) to allow the REST API to be accessed with the TOS accepted on a per-request basis. **Can hold off on reviewing and merging if we expect that decision soon.** This approach partially reverts https://github.com/Automattic/woocommerce-services/pull/1108, and is an alternative to https://github.com/Automattic/wp-calypso/issues/16954, which would necessitate an additional network call and related complexity.

Changes involve initializing the REST API and fetching the service schemas as long as Jetpack is connected, and adding a TOS condition (be it the usual TOS acceptance, or a `_terms_of_service ` parameter with an "accept" value) to the REST permissions check. Seeking feedback on which [other hooks](https://github.com/Automattic/woocommerce-services/blob/update/per-request-tos-acceptance/woocommerce-services.php#L460-L493) are necessary in order to support proper operation of plugin functionality via the REST API.

cc: @nabsul, author of https://github.com/Automattic/woocommerce-services/pull/1050


woocommerce-services.php
```diff
@@ -380,7 +371,8 @@ public function pre_wc_init() {
 				return;
 			}
 
-			add_action( 'rest_api_init', array( $this, 'tos_rest_init' ) );
+			add_action( 'rest_api_init', array( $this, 'rest_api_init' ) );
+			add_action( 'woocommerce_init', array( $this, 'schedule_service_schemas_fetch' ) );
```
jeffstieler: Just so I don't forget amidst discussions about TOS - we can't fetch schemas without some sort of consent first. I know we're still working that out, but as things currently sit I don't think this is OK.

Issue #1116, opened 2017-08-07
Add explanatory text to 'Shipping Labels' page


v18:
On this page:

<img width="966" alt="screen shot 2017-08-07 at 1 55 42 pm" src="https://user-images.githubusercontent.com/11487924/29039150-2052b960-7b78-11e7-8a68-1bcaedfbd2ac.png">

We should have text explaining what the credit card will be used for:

"Enter a credit card so that you can buy postage from USPS"

v18:
So, looks like that text already exists but users are not reading it :). Currently it says:

> Use your credit card on file to pay for the labels you print or add a new one.

<img width="966" alt="29039150-2052b960-7b78-11e7-8a68-1bcaedfbd2ac" src="https://user-images.githubusercontent.com/11487924/29819386-3a03e11a-8cc9-11e7-94aa-353482bfef99.png">

@olaolusoga do you think we should close this issue until it comes up again, or come up with a better location for that text, or some other solution for this problem?

olaolusoga:
> do you think we should close this issue until it comes up again, or come up with a better location for that text, or some other solution for this problem?

I say we close since it really low hanging fruit, and the text is already there. We should keep a mental note of it though, since we'll need to touch it down the road per convo here #1115

dechov:
We could maybe reword it to front-load the emphasis: "To purchase labels, use your credit card on file or add a new one."

v18:
@dechov -  I like it. Would you be able to open a PR with that change?

PR #1141, opened 2017-08-30
Re-word description for credit card settings to emphasize purpose

dechov:
Fixes https://github.com/Automattic/woocommerce-services/issues/1116 by adjusting existing description for credit card label settings so that the reason for it is harder to miss.


v18:
@marcinbot thanks for bringing that up. Does this mean it might be a good time to think about our approach to synching Calypso & the plugin? Should we start on one of the proposed ideas, or still hold off until there's more consensus around it in general (not just WCS)?

jameskoster:
Looks good to me @marcinbot

dechov:
@marcinbot Done https://github.com/Automattic/wp-calypso/pull/17759 — should that additional tweak be moved back here now?

Issue #1142, opened 2017-08-30
Labels modal - respect the enabled flag


marcinbot:
#1099 added `enabled` flag to the label settings. It's always set to `true` by default, but Calypso allows to toggle it.

At the very least, that flag should be added to the labels GET response at `WC_REST_Connect_Shipping_Label_Controller` and preloaded data at `WC_Connect_Shipping_Label` in `get_label_payload`

This will allow Calypso to decide whether to render the label controls on the order page

CC @jeffstieler @dechov @DanReyLop - it's a small task that can be picked up right now

PR #1143, opened 2017-08-31
Adjust the label payload to respect whether shipping labels are enabled

dechov:
Fixes https://github.com/Automattic/woocommerce-services/issues/1142 by adding the `enabled` flag to the shipping label payload. Also returns no other data if not `enabled`.


dechov:
Realized not returning the rest of the data would break the wp-admin experience if the flag was toggled from Calypso, and simplified it to just include the `enabled` flag.

PR #1144, opened 2017-09-01
Refactor shipping label code to use child containers

dechov:
Encapsulates `shipping-label/view` code that pertains to a single label item into a `LabelItem` component, and connects subcomponents to the state tree instead of passing everything as props from the top-level container.

cc @v18

### DIFF
--- a/client/apps/shipping-label/components/label-item/index.js
+++ b/client/apps/shipping-label/components/label-item/index.js
@@ -1,0 +1,158 @@
+/**
+ * External dependencies
+ */
+import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import Gridicon from 'gridicons';
+import { translate as __ } from 'i18n-calypso';
+
+/**
+ * Internal dependencies
+ */
+import RefundDialog from '../label-refund-modal';
+import ReprintDialog from '../label-reprint-modal';
+import TrackingLink from '../tracking-link';
+import InfoTooltip from 'components/info-tooltip';
+import formatDate from 'lib/utils/format-date';
+import timeAgo from 'lib/utils/time-ago';
+import { openRefundDialog, openReprintDialog } from '../../state/actions';
+
+class LabelItem extends Component {
+	renderRefundLink = ( label ) => {
+		const today = new Date();
+		const thirtyDaysAgo = new Date().setDate( today.getDate() - 30 );
+		if ( ( label.used_date && label.used_date < today.getTime() ) || ( label.created_date && label.created_date < thirtyDaysAgo ) ) {
+			return null;
+		}
+
+		const openDialog = ( e ) => {
+			e.preventDefault();
+			this.props.labelActions.openRefundDialog( label.label_id );
+		};
+
+		return (
+			<span>
+				<RefundDialog { ...label } />
+				<a href="#" onClick={ openDialog } >
+					<Gridicon icon="refund" size={ 12 } />{ __( 'Request refund' ) }
+				</a>
+			</span>
+		);
+	};
+
+	renderRefund = ( label ) => {
+		if ( ! label.refund ) {
+			return this.renderRefundLink( label );
+		}
+
+		let text = '';
+		let className = '';
+		switch ( label.refund.status ) {
+			case 'pending':
+				if ( label.statusUpdated ) {
+					className = 'is-refund-pending';
+					text = __( 'Refund pending' );
+				} else {
+					className = 'is-refund-checking';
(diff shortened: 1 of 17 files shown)


client/apps/shipping-label/components/label-item/index.js
```diff
@@ -0,0 +1,158 @@
...
+	}
+}
+
+LabelItem.propTypes = {
+	label: PropTypes.object.isRequired,
+};
+
+function mapDispatchToProps( dispatch ) {
+	return {
+		labelActions: bindActionCreators( { openRefundDialog, openReprintDialog }, dispatch ),
```
marcinbot: Let's remove `{labelActions}` and just return a result of `bindActionCreators` here instead. Then we can reference the actions via `this.props.<actionName>` instead of `this.props. labelActions.<actionName>` This has been a change often requested in Calypso PR reviews (not a pattern that is used there)

client/apps/shipping-label/components/label-purchase-modal/index.js
```diff
@@ -115,4 +120,23 @@ const PurchaseDialog = ( props ) => {
...
+		...shippingLabel,
+		storeOptions,
+		errors: loaded && getFormErrors( state, storeOptions ),
+		canPurchase: loaded && canPurchase( state, storeOptions ),
+	};
+}
+
+function mapDispatchToProps( dispatch ) {
+	return {
+		labelActions: bindActionCreators( { confirmPrintLabel, purchaseLabel, exitPrintingFlow }, dispatch ),
```
marcinbot: Same comment about `labelActions`

client/apps/shipping-label/components/label-refund-modal/index.js
```diff
@@ -62,4 +65,19 @@ RefundDialog.propTypes = {
...
+	const loaded = shippingLabel.loaded;
+	return {
+		refundDialog: loaded ? shippingLabel.refundDialog : {},
+		storeOptions: loaded ? shippingLabel.storeOptions : {},
+	};
+}
+
+function mapDispatchToProps( dispatch ) {
+	return {
+		labelActions: bindActionCreators( { closeRefundDialog, confirmRefund }, dispatch ),
```
marcinbot: Same comment about `labelActions`

client/apps/shipping-label/components/label-reprint-modal/index.js
```diff
@@ -58,4 +61,20 @@ ReprintDialog.propTypes = {
...
+	return {
+		reprintDialog: loaded ? shippingLabel.reprintDialog : {},
+		paperSize: shippingLabel.paperSize,
+		storeOptions: loaded ? shippingLabel.storeOptions : {},
+	};
+}
+
+function mapDispatchToProps( dispatch ) {
+	return {
+		labelActions: bindActionCreators( { closeReprintDialog, confirmReprint, updatePaperSize }, dispatch ),
```
marcinbot: Same comment about `labelActions`

client/apps/shipping-label/view.js
```diff
@@ -264,32 +128,24 @@ class ShippingLabelRootView extends Component {
...
 		needToFetchLabelStatus: loaded && ! shippingLabel.refreshedLabelStatus,
-		errors: loaded && getFormErrors( state, storeOptions ),
-		canPurchase: loaded && canPurchase( state, storeOptions ),
 	};
 }
 
 function mapDispatchToProps( dispatch ) {
 	return {
-		labelActions: bindActionCreators( ShippingLabelActions, dispatch ),
+		labelActions: bindActionCreators( { fetchLabelsStatus, openPrintingFlow }, dispatch ),
```
marcinbot: Same comment about `labelActions`

dechov:
@marcinbot Thanks — should have caught that. Working on connecting label purchase subcomponents. Should we flatten out the `components` directory, since they are not just presentational components anymore? Or keep `components/tracking-link` and move the rest up? cc @DanReyLop

DanReyLop:
@dechov A "connected" component is still a component. I'm not sure what distinction are we making between `view` and `component`, but I've seen modules inside `views/` that are presentational components. So, just use your best judgement (or whatever option involves moving around **less** files), we'll need to do a sweep anyway once the code stabilizes.

dechov:
@DanReyLop Ah, my previous understanding was that, by Calypso's conventions, components inside `components` should be purely [presentational components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0), but I don't think that's true. I won't move anything for now.

dechov:
Not sure why the build is failing — any ideas?

PR #1147, opened 2017-09-04
Fixed React warnings on the label print page

marcinbot:
Fixed some problems that were missed in #1144 

To test:
* the order page should show no react warnings on load
* the label modal should show no react warnings
* everything in the modal should render and work properly (esp. package weight change)

### DIFF
--- a/client/apps/shipping-label/components/label-purchase-modal/packages-step/package-info.js
+++ b/client/apps/shipping-label/components/label-purchase-modal/packages-step/package-info.js
@@ -25,9 +25,13 @@ import {
 } from '../../../state/actions';
 
 const renderPackageDimensions = ( dimensions, dimensionUnit ) => {
-	return `${ dimensions.length } ${ dimensionUnit } x
-			${ dimensions.width } ${ dimensionUnit } x
-			${ dimensions.height } ${ dimensionUnit }`;
+	return [
+		dimensions.length,
+		dimensions.width,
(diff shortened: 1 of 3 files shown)


client/apps/shipping-label/components/label-purchase-modal/packages-step/package-info.js
```diff
@@ -18,15 +18,15 @@ import Button from 'components/button';
...
+	updatePackageWeight,
 	removePackage,
 	setPackageType,
 	openAddItem,
 } from '../../../state/actions';
 
 const renderPackageDimensions = ( dimensions, dimensionUnit ) => {
-	return `${ dimensions.length } ${ dimensionUnit } x 
-			${ dimensions.width } ${ dimensionUnit } x 
+	return `${ dimensions.length } ${ dimensionUnit } x
```
dechov: Is there any case where this trailing space would help ensure the space after the `x`?

marcinbot: I don't think there is. This was cleaned up automatically, but [according to the template literals documentation](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals) all the whitespaces, including line breaks, are preserved. You can see the tabs and line breaks in dev tools, even if the HTML renders them as a single space.

I fixed it.

Issue #846, opened 2017-01-25
Add support for domestic shipping that WooCommerce core treats as international


allendav:
Sorry for another drive-by issue. I just ran into this on a different shipping extension of ours. WooCommerce core (as of 2.6?) treats these domestic addresses as international, so it would be good 1) to be aware that this means that connect 1.0 doesn't support these in case support questions come up as they will and 2) to be sure that these destinations don't require customs forms if they do not need to when they are implemented in connect.

Domestic, no Customs Forms
PR	Puerto Rico
VI	United States (US) Virgin Islands

Domestic, but requires Custom Forms
AS	American Samoa
GU	Guam
MP	Northern Mariana Islands
UM	United States (US) Minor Outlying Islands

"Compact of Free Association" countries (treated as domestic destinations, but it is not clear about customs forms)
FM	Federated States of Micronesia
MH	Marshall Islands
PW	Palau

WPprodigy:
Two of them are mentioned here: https://github.com/woocommerce/woocommerce/blob/906b7dbf10eba9da49c1abe24bf5668bdc99f030/i18n/countries.php

Here is the error report from a customer that is pretty detailed: 

> When trying to calculate shipping for the Marshall Islands, the error below is logged:

> 07-11-2017 @ 18:21:14 - Error. Unable to get shipping rate(s) for wc_services_usps instance id 6. (calculate_shipping) 
[...]

PR #1148, opened 2017-09-05
Added support for labels to PR and VI

### LINKED ISSUE #1129: Puerto Rico as a State
#640568-zen

My understanding is that recently Puerto Rico became a state in US, therefore the shipping now should include Puerto Rico as a state when selecting United States as a country, instead of having Puerto Rico as a separate country.

marcinbot:
Fixes #1129

According to #846 no customs forms are needed for these regions, so we should be able to print labels for them as usual.

To test:
* place an order with an address from Puerto Rico or Virgin Islands (US)
* on the order page, the label metabox should appear
* you should be able to purchase and print a label just like for mainland states

Note:
* the address validation can be a bit picky when it comes to the addresses from there but it does work

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -25,9 +25,16 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		protected $payment_methods_store;
 
 		/**
-		 * @var array array of currently unsupported US states
+		 * @var array array of currently supported countries
 		 */
-		private $unsupported_states = array( 'AA', 'AE', 'AP' );
+		private $supported_countries = array( 'US', 'PR' );
+
+		/**
+		 * @var array array of currently unsupported states, by country
+		 */
+		private $unsupported_states = array(
+			'US' => array( 'AA', 'AE', 'AP' ),
+		);
 
 		private $show_metabox = null;
 
@@ -307,19 +314,44 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 			return $form_data;
 		}
 
+		private function is_supported_state( $country_code, $state_code ) {
+			if ( ! $country_code || ! $state_code ) {
+				return true;
+			}
+
+			if ( ! array_key_exists( $country_code, $this->unsupported_states ) ) {
+				return true;
+			}
+
+			return ! in_array( $state_code, $this->unsupported_states[ $country_code ] );
+		}
+
+		private function is_supported_address( $address ) {
+			$country_code = $address['country'];
+			if ( ! $country_code ) {
+				return true;
+			}
+
+			if ( ! in_array( $country_code, $this->supported_countries ) ) {
+				return false;
+			}
+
+			$state_code = $address['state'];
+			return $this->is_supported_state( $country_code, $state_code );
+		}
+
 		protected function get_states_map() {
 			$result = array();
 			$all_countries = WC()->countries->get_countries();
-			$allowed_countries = array( 'US', 'PR' );
 
-			foreach ( $allowed_countries as $country_code ) {
+			foreach ( $this->supported_countries as $country_code ) {
 				$country_data = array( 'name' => html_entity_decode( $all_countries[ $country_code ] ) );
 				$states = WC()->countries->get_states( $country_code );
 
 				if ( $states ) {
 					$country_data['states'] = array();
 					foreach ( $states as $state_code => $name ) {
-						if ( 'US' === $country_code && in_array( $state_code, $this->unsupported_states ) ) {
+						if ( ! $this->is_supported_state( $country_code, $state_code ) ) {
 						  continue;
 						}
 						$country_data['states'][ $state_code ] = html_entity_decode( $name );
@@ -359,10 +391,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 			}
 
 			$dest_address = $order->get_address( 'shipping' );
-			if ( ( $dest_address['country'] && 'US' !== $dest_address['country']
-				//special case - include PR since it's treated as domestic
-				&& 'PR' !== $dest_address['country'] )
-				|| in_array( $dest_address['state'], $this->unsupported_states ) ) {
+			if ( ! $this->is_supported_address( $dest_address ) ) {
 				return false;
 			}
 
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -310,24 +310,23 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		protected function get_states_map() {
 			$result = array();
 			$all_countries = WC()->countries->get_countries();
-			$all_states = WC()->countries->get_states();
 			$allowed_countries = array( 'US', 'PR' );
 
-			foreach ( $allowed_countries as $code ) {
-				$country_data = array( 'name' => html_entity_decode( $all_countries[ $code ] ) );
+			foreach ( $allowed_countries as $country_code ) {
+				$country_data = array( 'name' => html_entity_decode( $all_countries[ $country_code ] ) );
+				$states = WC()->countries->get_states( $country_code );
 
-				if ( array_key_exists( $code, $all_states ) ) {
-					$states = $all_states[ $code ];
+				if ( $states ) {
 					$country_data['states'] = array();
 					foreach ( $states as $state_code => $name ) {
-						if ( 'US' === $code && in_array( $state_code, $this->unsupported_states ) ) {
+						if ( 'US' === $country_code && in_array( $state_code, $this->unsupported_states ) ) {
 						  continue;
 						}
 						$country_data['states'][ $state_code ] = html_entity_decode( $name );
 					}
 				}
 
-				$result[ $code ] = $country_data;
+				$result[ $country_code ] = $country_data;
 			}
 
 			return $result;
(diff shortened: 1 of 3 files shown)


jeffstieler:
Another use case: https://wordpress.org/support/topic/usps-not-working-with-puerto-rico/

dechov:
Looks good and tests well for destination address.

What if you were shipping from PR or VI though? On USPS's site, I believe you can select them as a "state" within the United States — should we allow this as well? Or perhaps enable switching countries, with only US, PR, and VI as options?

v18:
## Shipping to the Virgin Islands

Looks like shipping from VI to other places in the US still requires a clearance form, according to USPS (https://about.usps.com/publications/pub14/pub14_ch7_004.htm):

> The U.S. Virgin Islands lie outside the Customs Territory of the United States (CTUS) - that is, the 50 States, the District of Columbia, and the Commonwealth of Puerto Rico. Accordingly, even though domestic postage rates apply, mail originating in the U.S. Virgin Islands and addressed for delivery inside the CTUS is still subject to the same customs clearance procedures required for foreign mail.

@marcinbot do we still want to allow printing custom labels from addresses in VI? It looks like our users would be able to create their own customs forms at https://www.usps.com/international/customs-forms.htm

However, we don't alert users about this right now.

## Shipping to the Virgin Islands

It looks like there are no special custom forms need to ship to the Virgin Islands. Flat rate will be the same as domestic rate in USPS: http://about.usps.com/postal-bulletin/2009/pb22265/html/info1_004.htm

> Delivery to Alaska, Hawaii, Puerto Rico, Guam, or the U.S. Virgin Islands at domestic prices.

dechov:
Given the asymmetry pointed out by @v18, is it preferable to keep origin and destination consistent by removing VI altogether, or still allow VI (and GU?) as a destination country? I'd lean towards the latter on the grounds that friction when basing a store in VI shouldn't necessarily also mean friction when ordering things.

client/lib/initialize-labels-state/index.js
```diff
@@ -47,7 +47,7 @@ export default ( formData, labelsData, paperSize, storeOptions, paymentMethod, n
 				ignoreValidation: hasOriginAddress ? null : mapValues( formData.origin, () => true ),
 				selectNormalized: true,
 				normalizationInProgress: false,
-				allowChangeCountry: false,
+				allowedCountries: [ 'US', 'PR' ],
```
jeffstieler: What do you think about defining this on the PHP side? My thought is that if we wanted to change the countries based on other store settings (or account level in the future), its easier to do so on the WP side of the house.

jeffstieler:
It appears we might be able to support USVI as an origin if the destination is also USVI. Using https://postcalc.usps.com/CustomsForm - it shows "No form needed" for two USVI postcodes (I tested with `00801` and `00803`).

Given that this complicates the logic (making allowed destination countries dependent on origin) - we can save for another issue.

v18:
To be able to check out with a destination address to PR, I did the following:

1. Add a Puerto Rico zone
2. Add the USPS WCS shipping method to it, with a mainland US zipcode in the settings
3. Use https://github.com/Automattic/woocommerce-connect-server/pull/968 to update my local server
4. Enter 'Puerto Rico' as the country and the state in the cart

After this, I was able to successfully purchase a shipping label when the origin is in the US and the destination in PR.

These scenarios all work for me:
- ✓ origin: mainland US, destination: PR
- ✓ PR rate requests on checkout when the store origin is mainland US

These scenarios don't work:
- ✗checkout: origin: PR, destination: mainland US
- ✗ checkout: origin: PR, destination: PR

This is because our server expects state to not be empty, but when Puerto Rico is used as the store address, requesting rates (`/shipping/rates`) on check out does not work:

```
fails because [child \"base_state\" fails because [\"base_state\" is not allowed to be empty
```

client/apps/shipping-label/components/label-purchase-modal/address-step/fields.js
```diff
@@ -57,7 +57,12 @@ const AddressFields = ( props ) => {
 
 	const fieldErrors = _.isObject( errors ) ? errors : {};
 	const getId = ( fieldName ) => group + '_' + fieldName;
-	const getValue = ( fieldName ) => values[ fieldName ] || '';
+	const getValue = ( fieldName ) => {
+		if ( 'country' === fieldName && ! values[ fieldName ] ) {
+			return 'US';
+		}
```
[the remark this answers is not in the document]
marcinbot: I think I wanted to default to US if the country field is empty, but it should never be empty so I removed this in 9f9ba82

client/lib/initialize-labels-state/index.js
marcinbot: This and the above are addressed in fd1e904

classes/class-wc-connect-shipping-label.php
```diff
@@ -353,19 +353,27 @@ protected function get_form_data( WC_Order $order ) {
...
-			foreach ( WC()->countries->get_states() as $country => $states ) {
-				$result[ $country ]['states'] = array();
-				foreach ( $states as $code => $name ) {
-					if ( 'US' === $country && in_array( $code, $this->unsupported_states ) ) {
-						continue;
+			$all_countries = WC()->countries->get_countries();
+			$all_states = WC()->countries->get_states();
+			$allowed_countries = array( 'US', 'PR' );
+
+			foreach ( $allowed_countries as $code ) {
```
v18: What do you think about using `$country_code` here instead of `$code`? A bit easier to understand what's going on with `$country_code`.

classes/class-wc-connect-shipping-label.php
```diff
@@ -353,19 +353,27 @@ protected function get_form_data( WC_Order $order ) {
...
-			foreach ( WC()->countries->get_countries() as $code => $name ) {
-				$result[ $code ] = array( 'name' => html_entity_decode( $name ) );
-			}
-			foreach ( WC()->countries->get_states() as $country => $states ) {
-				$result[ $country ]['states'] = array();
-				foreach ( $states as $code => $name ) {
-					if ( 'US' === $country && in_array( $code, $this->unsupported_states ) ) {
-						continue;
+			$all_countries = WC()->countries->get_countries();
+			$all_states = WC()->countries->get_states();
```
v18: Instead of getting the global states array here, what do you think about using `WC()->countries->get_states( $country_code )` instead of `$states = $all_states[ $code ];` below to get the list of states for each allowed country?

v18:
I left a couple of comments on the code. The changes work for me:
- calculating shipping rates
- getting label rates (didn't test actually purchasing rn since that's not working for me locally atm)
- all scenarios: to/from PR, to/from mainland address

classes/class-wc-connect-shipping-label.php
```diff
@@ -353,19 +353,26 @@ protected function get_form_data( WC_Order $order ) {
...
-			foreach ( WC()->countries->get_countries() as $code => $name ) {
-				$result[ $code ] = array( 'name' => html_entity_decode( $name ) );
-			}
-			foreach ( WC()->countries->get_states() as $country => $states ) {
-				$result[ $country ]['states'] = array();
-				foreach ( $states as $code => $name ) {
-					if ( 'US' === $country && in_array( $code, $this->unsupported_states ) ) {
-						continue;
+			$all_countries = WC()->countries->get_countries();
+			$allowed_countries = array( 'US', 'PR' );
```
jeffstieler: What do you think about moving this to a class property so it can be reused in `calculate_should_show_meta_box()`?

classes/class-wc-connect-shipping-label.php
```diff
@@ -353,19 +353,26 @@ protected function get_form_data( WC_Order $order ) {
...
+
+			foreach ( $allowed_countries as $country_code ) {
+				$country_data = array( 'name' => html_entity_decode( $all_countries[ $country_code ] ) );
+				$states = WC()->countries->get_states( $country_code );
+
+				if ( $states ) {
+					$country_data['states'] = array();
+					foreach ( $states as $state_code => $name ) {
+						if ( 'US' === $country_code && in_array( $state_code, $this->unsupported_states ) ) {
```
jeffstieler: What do you think about using `$this->unsupported_states` for all countries and nesting what's currently there under a `US` key? It might be unlikely that we restrict states from other countries, but it removes some US-specific hardcoding.

classes/class-wc-connect-shipping-label.php
```diff
@@ -396,7 +403,9 @@ private function calculate_should_show_meta_box() {
 			}
 
 			$dest_address = $order->get_address( 'shipping' );
-			if ( ( $dest_address['country'] && 'US' !== $dest_address['country'] )
+			if ( ( $dest_address['country'] && 'US' !== $dest_address['country']
+				//special case - include PR since it's treated as domestic
+				&& 'PR' !== $dest_address['country'] )
```
jeffstieler: What do you think about using a class property for this? (to represent supported countries)

jeffstieler:
@v18 - just noticed you had some unanswered questions/comments - blocking merge or no?

classes/class-wc-connect-shipping-label.php
```diff
@@ -25,22 +25,27 @@ class WC_Connect_Shipping_Label {
 		protected $payment_methods_store;
 
 		/**
-		 * @var array array of currently unsupported US states
+		 * @var array array of currently supported countries
 		 */
-		private $unsupported_states = array( 'AA', 'AE', 'AP' );
+		private $supported_countries = array( 'US', 'PR' );
```
DanReyLop: You need to add `'VI'` (US Virgin Islands) too, since it doesn't require a customs form either..

marcinbot: There was some discussion around it and it looks like VI still requires it, see [Val's comment above](https://github.com/Automattic/woocommerce-services/pull/1148#issuecomment-339026116). That's why it was removed from the initial implementation.

jeffstieler:
Works as expected (now with new Calypso shiny). :shipit:

Issue #1156, opened 2017-09-14
Integrate with WooCommerce Shipment Tracking


JamesJosephFinn:
Tracking info can be exposed to the customer in `My Account dashboard >> Orders >> View order`, as well as sent to the customer in the core `order complete` email, by using the [WooCommerce Shipment Tracking](https://woocommerce.com/products/shipment-tracking/) (paid) plugin developed by the core team, *buuut....*

### Human input is an opportunity for human error

**Unfortunately**, the store operator must manually paste the tracking number into the *Woocommerce Shipment Tracking* widget on the order details page in order for these good things to happen.

I know that *WooCommerce Services* has [plans to add the automated sending of tracking info to the customer](https://docs.woocommerce.com/document/woocommerce-services-faq/#section-18), but *WooCommerce Shipment Tracking* is useful in other ways (e.g. the aforementioned `View order` page in the customer's My Account dashboard), so it would be nice if *Woocommerce Shipment Tracking* was fed this data automatically.

### Closing argument

*WooCommerce Services* and *Woocommerce Shipment Tracking* are core WooCommerce products, and it would be great if they were integrated.

The shipment tracking number *Woocommerce Services* generates as part of the label creation process should automatically be fed into *Woocommerce Shipment Tracking* rather than having to be pasted in manually.

ValeDeOro:
683743-zen

JamesJosephFinn:
@jeffstieler why has this been removed from 1.9.0? I believe I made a good argument when I created this ticket. Having to manually paste in critical data to a paid plugin is inefficient at best. Please re-consider.

jeffstieler:
@JamesJosephFinn we're going to halt releases during heightened sales periods like Black Friday and the Christmas holidays, and we're aiming to have a 1.9.0 release early this week. Since we haven't actually worked on this, it won't make it into the release.

It will be prioritized for a release after 1.9.0.

marcinbot:
A fix has been added in #1325 and will be included with the next release.

PR #1158, opened 2017-09-18
Option to feed tracking info to Shipment Tracking plugin

dechov:
Fixes https://github.com/Automattic/woocommerce-services/issues/1156

The design of this integration could perhaps use some more thought, but currently,  if the [Shipment Tracking](https://woocommerce.com/products/shipment-tracking/) plugin is active, it adds an action to the right of the tracking link:

screen shot 2017-09-18 at 6 02 22 pm

...which takes you to the filled-in Shipment Tracking form:

screen shot 2017-09-18 at 6 02 54 pm

(It also could just submit the form for you, except that you'd lose the chance to set the ship date.)

### DIFF
--- a/client/apps/shipping-label/components/label-item/index.js
+++ b/client/apps/shipping-label/components/label-item/index.js
@@ -19,6 +19,13 @@ import timeAgo from 'lib/utils/time-ago';
 import { openRefundDialog, openReprintDialog } from '../../state/actions';
 
 class LabelItem extends Component {
+	constructor( props ) {
+		super( props );
+
+		this.shipmentTracking = document.querySelector( '#woocommerce-shipment-tracking' );
+		this.feedToShipmentTracking = this.feedToShipmentTracking.bind( this );
+	}
+
 	renderRefundLink = ( label ) => {
 		const today = new Date();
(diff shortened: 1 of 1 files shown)


JamesJosephFinn:
@dechov thanks for this...

As regards whether or not the Shipment Tracking widget form should be automatically submitted, keep in mind, the data in this form gets sent to the customer with the "Order Completed" core WooCommerce email, which is sent as soon as an order is marked `completed`. In our workflow, we do not mark an order as `completed` until it has shipped, and we keep the default subject header for this email as-is, i.e. "Your order has shipped!"

For this reason, perhaps it is best for the ship date to automatically match the date on which the order is marked `completed`.

Thanks again!

jeffstieler:
For transparency - I labeled this "parked" because we're in a crunch for some other priorities and I want to evaluate if a client-side-script-only approach is appropriate for this feature.

jeffstieler:
If we're going to continue this after https://github.com/Automattic/woocommerce-services/pull/1173, we should look at using [`wc_st_add_tracking_number `](https://docs.woocommerce.com/document/shipment-tracking/#section-3)

PR #1155, opened 2017-09-08
add current proper link

Umangvaghela:
add a proper link in plugin main file

PR #1161, opened 2017-09-22
Update translation files and non-SSL urls.

jeffstieler:
In reviewing https://github.com/Automattic/woocommerce-services/pull/1155, I discovered that our translation files were very out-of-date. In the process of updating them, I noticed the addition of several Calypso strings that weren't used in our project.

This PR seeks to incorporate the changes from @Umangvaghela, use `babel-plugin-transform-imports` from the Calypso config to avoid bringing in all state selectors, and update our translation files to use the latest strings.


jeffstieler:
There are more strings to translate than the last time the files were generated, so the size increase is expected.

PR #1165, opened 2017-09-26
Setup shipping zones with live rate methods based on Setup Wizard choices

jeffstieler:
To test:

* Check out https://github.com/Automattic/woocommerce-connect-server/tree/add/shipping-zone-setup-from-3-2-wizard
* Set one or both of the following options to `1`:
  * `woocommerce_setup_domestic_live_rates_zone`
  * `woocommerce_setup_intl_live_rates_zone`
* Verify that the corresponding zone has USPS or Canada Post (based on store country) method setup up with defaults


woocommerce-services.php
```diff
@@ -390,6 +390,66 @@ public function pre_wc_init() {
...
+			if ( 'US' === $store_country ) {
+				$country_method = 'usps';
+			} elseif ( 'CA' === $store_country ) {
+				$country_method = 'canada_post';
+			} else {
+				return; // Only set up live rates for US and CA
+			}
+			
+			$zones_to_setup = array(
+				'domestic' => 1, // TODO: this might be too fragile/assumptive
```
DanReyLop: If you want to be extra-safe, you can check that the only location on the zone is of type `country` and value `$store_country`, and bail otherwise.

woocommerce-services.php
```diff
@@ -390,6 +390,66 @@ public function pre_wc_init() {
...
+				'intl'     => 0,
+			);
+			
+			foreach ( $zones_to_setup as $option => $zone_id ) {
+				$option_name = "woocommerce_setup_{$option}_live_rates_zone";
+
+				if ( get_option( $option_name ) ) {
+					$method      = $this->get_service_schemas_store()->get_service_schema_by_id( $country_method );
+					$zone        = WC_Shipping_Zones::get_zone( $zone_id );
```
DanReyLop: In the same line, to make this more robust, you should check that `$zone !== false` and bail.

DanReyLop:
I've made some changes in https://github.com/Automattic/woocommerce-services/pull/1165/commits/f93997a26e5a9df50a63767ef373d9077eaeb98b to address the "robustness" of finding the domestic shipping zones and add Live Rates to it. Everything else works fine, so if you're OK with my change, you can :shipit: 

Just one question, how did you decide which methods to enable? For USPS, it's just `Priority Mail` with boxes, is that intended?

PR #1167, opened 2017-09-27
Remove TOS flag: state, condition, and UI

dechov:
Note: **Not to merge**/release until WCS TOS are consolidated into Jetpack TOS (discussion in p7bbVw-1Rq-p2 and p7bbVw-1Ur-p2)

Removes all TOS functionality that is local to the plugin and independent from Jetpack. As of the WP.com/Jetpack Terms of Service updates accompanying the WooCommerce 3.2 release, there is no need for any additional link or acceptance step.

To test:
- Connect Jetpack (or make sure it's connected)
- Install & activate WCS plugin (or set `tos_accepted` to 0 in `wc_connect_options`)
- Verify that no TOS banner appears, and that plugin is fully initialized & functional
  - e.g. "WooCommerce Services" tab appears on WooCommerce Status page

In case any other implementation complexity comes to mind that was only there to support the TOS check, it may also be worth removing in this PR.

Closes https://github.com/Automattic/woocommerce-services/pull/1138 and https://github.com/Automattic/wp-calypso/issues/16954


client/apps/shipping-label/components/label-reprint-modal/index.js
```diff
@@ -59,6 +59,7 @@ ReprintDialog.propTypes = {
 	reprintDialog: PropTypes.object,
 	paperSize: PropTypes.string.isRequired,
 	storeOptions: PropTypes.object.isRequired,
+	labelId: PropTypes.number,
```
jeffstieler: Is this meant to be a part of these changes?

dechov: Whoops — thanks!

classes/class-wc-connect-nux.php
```diff
@@ -467,38 +413,6 @@ public function show_banner_after_connection() {
...
-				WC_Connect_Options::update_option( 'tos_accepted', true );
-
-				$this->tracks->opted_in( 'tos_banner' );
-
-				wp_safe_redirect( remove_query_arg( 'wcs-nux-tos' ) );
-				exit;
-			}
-
-			$this->show_nux_banner( array(
-				'title'          => __( 'Almost ready to enjoy discounted shipping rates', 'woocommerce-services' ),
```
DanReyLop: We need to remember to regenerate the translation files before every release. Don't you have a release checklist somewhere, @jeffstieler ? We should put the checklist into a Wiki page.

jeffstieler:
After another round of discussion, it appears that this has to stay (and get reworked a bit).

classes/class-wc-connect-nux.php
dechov: @jeffstieler shared p7bbVw-16e-p2 as a reference, with just 1-2 steps out of date IIRC.

PR #1171, opened 2017-10-02
Fix debug toggle on Status page.

jeffstieler:
Refactor shipping settings form URL generation to be a generic settings form URL method.

This allows for separate handling of other forms that use the SettingsForm components, like the Status page debug toggle.

To test:
* Verify that shipping method form saving works
* Verify that toggling debug on Status page works

### DIFF
--- a/classes/class-wc-connect-help-view.php
+++ b/classes/class-wc-connect-help-view.php
@@ -692,10 +692,8 @@ if ( ! class_exists( 'WC_Connect_Help_View' ) ) {
 				'formSchema'         => $this->get_form_schema(),
 				'formLayout'         => $this->get_form_layout(),
 				'formData'           => $this->get_form_data(),
-				'methodId'           => 'self_help',
-				'instanceId'         => 'self_help',
+				'methodId'           => 'self-help',
 				'nonce'              => wp_create_nonce( 'wp_rest' ),
-				'callbackURL'        => get_rest_url( null, '/wc/v1/connect/self-help' ),
 				'noticeDismissed'    => true,
 			) );
 
--- a/classes/class-wc-connect-help-view.php
+++ b/classes/class-wc-connect-help-view.php
@@ -692,16 +692,13 @@ if ( ! class_exists( 'WC_Connect_Help_View' ) ) {
 				'formSchema'         => $this->get_form_schema(),
 				'formLayout'         => $this->get_form_layout(),
 				'formData'           => $this->get_form_data(),
-				'methodId'           => 'self-help',
-				'nonce'              => wp_create_nonce( 'wp_rest' ),
+				'formType'           => 'self-help',
 				'noticeDismissed'    => true,
 			) );
 
 			do_action( 'enqueue_wc_connect_script', 'wc-connect-admin-test-print', array(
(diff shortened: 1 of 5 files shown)


classes/class-wc-connect-help-view.php
```diff
@@ -692,10 +692,8 @@ public function page() {
 				'formSchema'         => $this->get_form_schema(),
 				'formLayout'         => $this->get_form_layout(),
 				'formData'           => $this->get_form_data(),
-				'methodId'           => 'self_help',
-				'instanceId'         => 'self_help',
+				'methodId'           => 'self-help',
 				'nonce'              => wp_create_nonce( 'wp_rest' ),
```
marcinbot: You can remove the `nonce`

`nonce` and `labelsPreviewURL` can also be removed from the `wc-connect-admin-test-print` call below these lines

classes/class-wc-connect-help-view.php
```diff
@@ -692,10 +692,8 @@ public function page() {
 				'formSchema'         => $this->get_form_schema(),
 				'formLayout'         => $this->get_form_layout(),
 				'formData'           => $this->get_form_data(),
-				'methodId'           => 'self_help',
-				'instanceId'         => 'self_help',
+				'methodId'           => 'self-help',
```
marcinbot: It feels like this should be `formType` and then the signature and usages of `url.settingsForm()` would need to be changed to reflect that. But I'm fine with it as it is.

client/api/url.js
```diff
@@ -18,8 +23,10 @@ export const labelTestPrint = () => `${ namespace }label/preview`;
 
 export const addressNormalization = () => `${ namespace }normalize-address`;
 
-export const shippingServiceSettings = ( methodId, instanceId = false ) => instanceId
-	? `${ namespace }services/${ methodId }/${ instanceId }`
-	: `${ namespace }services/${ methodId }`;
+export const settingsForm = ( methodId, instanceId = false, formType = false ) => {
```
DanReyLop: Having some arguments have default values it's confusing, since this function is always called with the 3 arguments. Also, I think `formType` should be the first, since it'll end up being the first chunk of the URL (will make the calling code a bit clearer maybe).

jeffstieler:
Thanks for the feedback @marcinbot @DanReyLop - it's been addressed. Can you take another look?

client/api/url.js
jeffstieler: Fixed in https://github.com/Automattic/woocommerce-services/pull/1171/commits/7248ece5d72e1b0228398e34a213f569d5e483e9.

classes/class-wc-connect-help-view.php
jeffstieler: Fixed in https://github.com/Automattic/woocommerce-services/pull/1171/commits/1a732c05525072c9fb7a9bf8d0242e1725472a9f.

classes/class-wc-connect-help-view.php
jeffstieler: Fixed in https://github.com/Automattic/woocommerce-services/pull/1171/commits/1a732c05525072c9fb7a9bf8d0242e1725472a9f.

PR #1172, opened 2017-10-03
Using WC store address as default when printing labels

### LINKED ISSUE #1079: Default shipping origin address to new store address
WooCommerce 3.2 is introducing "Store Address" to general settings.

screen shot 2017-07-05 at 3 55 05 pm

We should default the shipping label origin address to these values. (See `Edit order > Create new label`)

marcinbot:
Fixes #1079

Fixes an issue raised by @kellychoffman in https://github.com/Automattic/wp-calypso/pull/18281#issuecomment-332969186

When no labels have been printed before and WC version is 3.1.1 or newer, the address supplied in the WC settings should be used.

To test:
* Delete `wc_connect_origin_address` option from the `wp_options` table in your DB
* Ensure that you're using WC v>3.1.1 and that an address was supplied
* Load an order page and verify that the labels origin address is the same as the address in WC settings

### DIFF
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -103,6 +103,7 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 			$wc_address_fields = array();
 			$wc_address_fields['company'] = get_bloginfo( 'name' );
 			$wc_address_fields['name'] = wp_get_current_user()->display_name;
+			$wc_address_fields['phone'] = '';
 
 			$wc_countries = WC()->countries;
 			// WC 3.2 introduces ability to configure a full address in the settings
@@ -123,7 +124,6 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 				$wc_address_fields['city'] = '';
 				$wc_address_fields['postcode'] = '';
 			}
-			$wc_address_fields['phone'] = '';
 
 			$stored_address_fields = WC_Connect_Options::get_option( 'origin_address', array() );
 			return array_merge( $wc_address_fields, $stored_address_fields );
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -103,13 +103,24 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 			$wc_address_fields = array();
 			$wc_address_fields[ 'company' ] = get_bloginfo( 'name' );
 			$wc_address_fields[ 'name' ] = wp_get_current_user()->display_name;
-			$base_location = wc_get_base_location();
-			$wc_address_fields[ 'country' ] = $base_location[ 'country' ];
-			$wc_address_fields[ 'state' ] = $base_location[ 'state' ];
-			$wc_address_fields[ 'address' ] = '';
-			$wc_address_fields[ 'address_2' ] = '';
-			$wc_address_fields[ 'city' ] = '';
-			$wc_address_fields[ 'postcode' ] = '';
+
+			$wc_countries = WC()->countries;
+			if ( method_exists( $wc_countries, 'get_base_address' ) ) {
+				$wc_address_fields[ 'country' ] = $wc_countries->get_base_country();
+				$wc_address_fields[ 'state' ] = $wc_countries->get_base_state();
+				$wc_address_fields[ 'address' ] = $wc_countries->get_base_address();
+				$wc_address_fields[ 'address_2' ] = $wc_countries->get_base_address_2();
+				$wc_address_fields[ 'city' ] = $wc_countries->get_base_city();
+				$wc_address_fields[ 'postcode' ] = $wc_countries->get_base_postcode();
+			} else {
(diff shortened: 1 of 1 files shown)


dechov:
Fixes https://github.com/Automattic/woocommerce-services/issues/1079, is that right?

jeffstieler:
There will likely be a design component to this as well with changes to the labels modal. Track in #1079.

classes/class-wc-connect-service-settings-store.php
```diff
@@ -103,13 +103,24 @@ public function get_origin_address() {
...
-			$base_location = wc_get_base_location();
-			$wc_address_fields[ 'country' ] = $base_location[ 'country' ];
-			$wc_address_fields[ 'state' ] = $base_location[ 'state' ];
-			$wc_address_fields[ 'address' ] = '';
-			$wc_address_fields[ 'address_2' ] = '';
-			$wc_address_fields[ 'city' ] = '';
-			$wc_address_fields[ 'postcode' ] = '';
+
+			$wc_countries = WC()->countries;
+			if ( method_exists( $wc_countries, 'get_base_address' ) ) {
```
v18: Do you think this need a comment to explain the change is due to a new feature in WC 3.2?

v18: 🤔 Do you think this would work better or worse in compatibility class, like `classes/class-wc-connect-compatibility.php` and `classes/class-wc-connect-compatibility-wc30.php`?

classes/class-wc-connect-service-settings-store.php
```diff
@@ -103,13 +103,24 @@ public function get_origin_address() {
...
+				$wc_address_fields[ 'country' ] = $wc_countries->get_base_country();
```
v18: Small linting comment: array keys don't need the extra spaces when they only contain a string, so it should be `$wc_address_fields['country']` etc.

v18:
These cases work for me:
1. WC 3.0, with or without a saved WCS address
2. WC 3.2, with or without a saved WCS address

Left a few comments on the code.

v18:
Looks like this PR fully completes @kellychoffman's original request in https://github.com/Automattic/wp-calypso/pull/18281#issuecomment-332969186:

> The Origin address should be pre-filled in with the address seen on all Settings pages

It also solves completely the original description of #1079, so I'd say we should ship it when the comments are answered.

#1079 also discusses moving the address form to another section. We can either amend #1079 to include that or create a new issue, so it's tracked after this PR is merged.

classes/class-wc-connect-service-settings-store.php
marcinbot: Fixed in d602e0d

classes/class-wc-connect-service-settings-store.php
marcinbot: I considered that, but to be done ultra-properly, we would need to add and maintain something like `classes/class-wc-connect-compatibility-wc32.php` just to handle this. But we only need to retrieve the store address here, and other places in the code can use this function, so it would be creating a wrapper of a wrapper.

An argument could also be made that this is not really a compatibility issue. The old method is still there and works in the same way as before. This is just using the additional resources, if available.

v18: Yeh, I do agree with the argument that it's not really a compatibility issue. I also think it's a bit convoluted to put in a separate class. Happy to leave it as is :)

classes/class-wc-connect-service-settings-store.php
```diff
@@ -85,32 +85,45 @@ public function update_account_settings( $settings ) {
...
+			} else {
+				$base_location = wc_get_base_location();
+				$wc_address_fields['country'] = $base_location['country'];
+				$wc_address_fields['state'] = $base_location['state'];
+				$wc_address_fields['address'] = '';
+				$wc_address_fields['address_2'] = '';
+				$wc_address_fields['city'] = '';
+				$wc_address_fields['postcode'] = '';
+			}
+			$wc_address_fields['phone'] = '';
```
v18: How about putting this line at the top before the if statement, with the rest of the array keys that are the same in both cases? Right below
`$wc_address_fields['name'] = wp_get_current_user()->display_name;`.

classes/class-wc-connect-service-settings-store.php
```diff
@@ -124,7 +137,7 @@ public function get_preferred_paper_size() {
 			// According to https://en.wikipedia.org/wiki/Letter_(paper_size) US, Mexico, Canada and Dominican Republic
 			// use "Letter" size, and pretty much all the rest of the world use A4, so those are sensible defaults
 			$base_location = wc_get_base_location();
-			if ( in_array( $base_location[ 'country' ], array( 'US', 'CA', 'MX', 'DO' ) ) ) {
+			if ( in_array( $base_location['country'], array( 'US', 'CA', 'MX', 'DO' ) ) ) {
```
v18: I know this line change is unrelated to this PR since it's for a lint fix, but if we're changing it anyway, we could also fix the non-strict in_array comparison here by adding `true` as the third argument. Happy to merge without this fix - up to you.

v18:
Awesome, thanks for opening that new issue @marcinbot! I left only one minor comment, feel free to ship at any time :shipit:

Issue #1139, opened 2017-08-29
Add Tracking Number shortcode to be used in emails provided by WC Services 


juliaamosova:
Will be great to have the functionality where WC Services would provide a  shortcode for the tracking number once the shipping label is created. 

This shortcode can then be added to the email template to be sent to the customer of the site letting them know that the product was shipped and provide a tracking number.

JamesJosephFinn:
I also created #1156 which is related to this

jeffstieler:
> This shortcode can then be added to the email template to be sent to the customer of the site letting them know that the product was shipped and provide a tracking number.

@juliaamosova can you elaborate on your imagined usage of this shortcode? How would a store owner put the shortcode into their email template? etc..

juliaamosova:
@jeffstieler I am sorry for the delayed response! 

Now that I think about it further, the shortcode idea may not necessarily work since there isn't an _easy_ way to modify the Completed Order Email Template and include shortcode (if any). What I think will be great is to have a Shipment Tracking Number automatically included in the Completed Order Email that is sent to the user upon order is marked as _Completed_. Much like it currently works with Shipment Tracking Number widget (extension). If there is a tracking number entered within the widget upon the order being marked as Completed, the number will be included in the email:

![](https://cloudup.com/cJxE-AJpU3h+)
Image Link: https://cloudup.com/cJxE-AJpU3h

![](https://cloudup.com/ciUH2-aRcbn+)
Image Link: https://cloudup.com/ciUH2-aRcbn

JamesJosephFinn:
@juliaamosova this is my thinking exactly. PR #1158 is working on it.

juliaamosova:
@JamesJosephFinn great, thanks for pointing out the PR!

joshuaja:
@JamesJosephFinn - that PR is specific to integration with the Shipment Tracking plugin - I think isssue #391 is a better reference.

joshuaja:
@juliaamosova - I'm about to submit a Pull Request for this - please see #391 for a screenshot.

jeffstieler:
Closed in https://github.com/Automattic/woocommerce-services/pull/1173 - tracking numbers are now included in the Completed Order emails.

Issue #391, opened 2016-06-08
[Labels] Expose tracking information to customer for their order


allendav:
When available, expose at least the tracking number to the customer in the order details view, for orders where the merchant has used WCC to purchase a shipping label.

See also https://github.com/Automattic/woocommerce-connect-client/issues/390

kellychoffman:
Would it be weird if it was added to the notes section? Weird because it wouldn't be an actual note that you added, but it feels like that makes the most contextual sense:

<img width="295" alt="screen shot 2016-06-24 at 12 28 53 pm" src="https://cloud.githubusercontent.com/assets/5835847/16343885/ac6b9588-3a07-11e6-9185-2f1da90dcc17.png">

kellychoffman:
Another option would be to add it below the Order status because it would also make sense in that context:

<img width="787" alt="screen shot 2016-06-24 at 12 48 47 pm" src="https://cloud.githubusercontent.com/assets/5835847/16344313/0bad895a-3a0a-11e6-8f77-3d7a4bb9dba0.png">

kellychoffman:
What do ya'll think?

DanReyLop:
I'm ok having it as an Order Note, it seems the most straight-forward option to implement. The tracking number is not so important for the merchant anyway, it's more important how we present it to the customer.

kellychoffman:
👍  Order Note sounds good to me. 

>  more important how we present it to the customer.

Agree

kellychoffman:
Expose order/tracking ID in notes section too?

allendav:
@kellychoffman @DanReyLop - all this is fine by me, but this issue is for the customer side (e.g. my-account) - this feedback is more appropriate in #482 , no?

kellychoffman:
>  this issue is for the customer side

So it is. Will move design there and add a new design to this issue.

allendav:
@DanReyLop @kellychoffman - let's put this on hold for the moment until we hear from marketing on whether this feature should be included in the MVP

I'm moving it from Sprint 4 to Sprint 6

cc @warrendholmes @avivapinchas

kellychoffman:
We should add the ability for the merchant to send an email upon printing/purchasing a label, perhaps under the paper size picker:

<img width="301" alt="screen shot 2017-01-22 at 1 01 56 pm" src="https://cloud.githubusercontent.com/assets/5835847/22184183/04a9879a-e0a3-11e6-98c1-78c928cedbe6.png">

[...]

PR #1173, opened 2017-10-04
✨ Add tracking info to emails [#1139]

joshuaja:
This addresses #1139 and the email portion of #391.

I'm a JavaScript engineer so I welcome any feedback for the PHP code, especially the templating.

I'm not a fan of inline styles but the `<table>` element matches how the [Order details](https://github.com/woocommerce/woocommerce/blob/master/templates/emails/email-order-details.php#L29) table is set up in Woocommerce core.


woocommerce-services.php
```diff
@@ -667,6 +668,58 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+            echo '<thead>';
+            echo '<tr>';
+            echo '<th class="td" scope="col">' . __( 'Provider', 'woocommerce-services' ) . '</th>';
+            echo '<th class="td" scope="col">' . __( 'Tracking number', 'woocommerce-services' ) . '</th>';
+            echo '</tr>';
+            echo '</thead>';
+            echo '<tbody>';
+            echo '<tr>';
+
+            foreach ( $labels as &$label ) {
```
jeffstieler: Please remove the value-by-reference (`&`) usage here, there isn't a need to modify the `$label` array.

woocommerce-services.php
```diff
@@ -667,6 +668,58 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+                $label = $label[0];
```
jeffstieler: Hmm.. each item in the `$labels` array shouldn't be nested - I think this line becomes unnecessary if you use the order labels helper method mentioned above.

woocommerce-services.php
```diff
@@ -667,6 +668,58 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+                $carrier = $label['carrier_id'];
+                $tracking = $label['tracking'];
```
jeffstieler: Each `$label` can be refunded - please check for a `refund` array key before adding tracking information to the order email.

In `var_dump()` format, it looks like:
```php
   ["refund"]=>
    object(stdClass)#1 (3) {
      ["request_date"]=>
      int(1506647216000)
      ["status"]=>
      string(7) "pending"
      ["refund_date"]=>
      int(1)
    }
```

woocommerce-services.php
```diff
@@ -667,6 +668,58 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+
+                echo '<td class="td" scope="col">' . strtoupper($carrier) . '</td>';
```
jeffstieler: It's best to sanitize any values we're writing to the page. This is a prime candidate for `esc_html()`.

See: https://developer.wordpress.org/themes/theme-security/data-sanitization-escaping/#escaping-securing-output

woocommerce-services.php
```diff
@@ -667,6 +668,58 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+                switch ( $carrier ) {
+                    case "fedex":
+                        $tracking_url = 'https://www.fedex.com/apps/fedextrack/?action=track&tracknumbers=' . $tracking;
+                        break;
+                    case "usps":
+                        $tracking_url = 'https://tools.usps.com/go/TrackConfirmAction.action?tLabels=' . $tracking;
+                        break;
+                }
+
+                echo '<td class="td" scope="col"><a href="' . $tracking_url . '">' . $tracking . '</a></td>';
```
jeffstieler: Please use `esc_url()` on the tracking url, and `esc_html()` on the tracking number here.

See: https://developer.wordpress.org/themes/theme-security/data-sanitization-escaping/#escaping-securing-output

woocommerce-services.php
```diff
@@ -667,6 +668,58 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+
+            echo '<table class="td" cellspacing="0" cellpadding="6" style="margin-top: 10px; width: 100%; font-family: \'Helvetica Neue\', Helvetica, Roboto, Arial, sans-serif;">';
+            echo '<thead>';
+            echo '<tr>';
+            echo '<th class="td" scope="col">' . __( 'Provider', 'woocommerce-services' ) . '</th>';
+            echo '<th class="td" scope="col">' . __( 'Tracking number', 'woocommerce-services' ) . '</th>';
+            echo '</tr>';
+            echo '</thead>';
+            echo '<tbody>';
+            echo '<tr>';
```
jeffstieler: Should the `<tr>` be inside the loop to handle multiple labels?

joshuaja:
@jeffstieler - thanks so much for reviewing. I implemented your review suggestions last night. I'll do some testing today and update the Pull Request.

woocommerce-services.php
```diff
@@ -494,6 +494,7 @@ public function attach_hooks() {
 			add_action( 'enqueue_wc_connect_script', array( $this, 'enqueue_wc_connect_script' ), 10, 2 );
 			add_action( 'admin_init', array( $this, 'load_admin_dependencies' ) );
 			add_filter( 'wc_connect_shipping_service_settings', array( $this, 'shipping_service_settings' ), 10, 3 );
+			add_action( 'woocommerce_email_after_order_table', array( $this, 'add_tracking_info_to_emails' ), 10, 1 );
```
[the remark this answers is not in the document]
joshuaja: I changed it to
 ```php
add_action( 'woocommerce_email_after_order_table', array( $this, 'add_tracking_info_to_emails' ), 10, 4 );
``` 
to match the 4 arguments passed in [email-order-details.php](https://github.com/woocommerce/woocommerce/blob/4c5bb1f0ddf2e81da418ff4c7084d8319f6b4619/templates/emails/email-order-details.php#L25); however, my `add_tracking_info_to_emails` function only accepts/needs 1 argument. Should I add the other three hook arguments to that function as well?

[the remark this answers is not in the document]
Comment (jeffstieler) [reply woocommerce-services.php]: Please do - the idea here is that if `$plain_text` is specified, we should add the tracking information sans HTML.

**Comment by joshuaja** [comment]
@jeffstieler - I pushed a follow-up commit with your review suggestions. Let me know what you think.

File: woocommerce-services.php [hunk 8]
```diff
@@ -668,56 +668,74 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+			echo '<h2>' . __( 'Tracking', 'woocommerce-services' ) . '</h2>';
+			echo '<table class="td" cellspacing="0" cellpadding="6" style="margin-top: 10px; width: 100%; font-family: \'Helvetica Neue\', Helvetica, Roboto, Arial, sans-serif;">';
+			echo '<thead>';
+			echo '<tr>';
+			echo '<th class="td" scope="col">' . __( 'Provider', 'woocommerce-services' ) . '</th>';
+			echo '<th class="td" scope="col">' . __( 'Tracking number', 'woocommerce-services' ) . '</th>';
+			echo '</tr>';
+			echo '</thead>';
+			echo '<tbody>' . $markup . '</tbody>';
+			echo '</table>';
```
jeffstieler: This is optional, but recommended - we mix PHP and HTML to avoid a lot of unnecessary `echo`s. Here's an example: https://github.com/Automattic/woocommerce-services/blob/master/woocommerce-services.php#L1023-L1029

woocommerce-services.php
```diff
@@ -494,7 +494,7 @@ public function attach_hooks() {
 			add_action( 'enqueue_wc_connect_script', array( $this, 'enqueue_wc_connect_script' ), 10, 2 );
 			add_action( 'admin_init', array( $this, 'load_admin_dependencies' ) );
 			add_filter( 'wc_connect_shipping_service_settings', array( $this, 'shipping_service_settings' ), 10, 3 );
-			add_action( 'woocommerce_email_after_order_table', array( $this, 'add_tracking_info_to_emails' ), 10, 1 );
+			add_action( 'woocommerce_email_after_order_table', array( $this, 'add_tracking_info_to_emails' ), 10, 4 );
```
jeffstieler: Let's reduce this to `3` arguments since we're not planning to use `$email`, as far as I know.

joshuaja:
@jeffstieler - I pushed a commit incorporating your suggested changes. Let me know what you think. Here's a screenshot of a plain text completed email containing tracking info.

image

woocommerce-services.php
```diff
@@ -694,28 +693,35 @@ public function add_tracking_info_to_emails( $order, $sent_to_admin, $plain_text
...
-					case "usps":
-						$tracking_url = 'https://tools.usps.com/go/TrackConfirmAction.action?tLabels=' . $tracking;
+					case 'usps':
+						$tracking_url = 'https://tools.usps.com/go/TrackConfirmAction.action?tLabels=' . esc_html( $tracking );
 						break;
 				}
 
 				$markup .= '<td class="td" scope="col">';
-				$markup .= '<a href="' . esc_html( $tracking_url ) . '">' . $tracking . '</a>';
+				$markup .= '<a href="' . esc_html( $tracking_url ) . '">' . esc_html( $tracking ) . '</a>';
```
jeffstieler: Please use `esc_url()` on the tracking url instead of `esc_html()`.

woocommerce-services.php
```diff
@@ -694,28 +693,35 @@ public function add_tracking_info_to_emails( $order, $sent_to_admin, $plain_text
...
 				switch ( $carrier ) {
-					case "fedex":
-						$tracking_url = 'https://www.fedex.com/apps/fedextrack/?action=track&tracknumbers=' . $tracking;
+					case 'fedex':
+						$tracking_url = 'https://www.fedex.com/apps/fedextrack/?action=track&tracknumbers=' . esc_html( $tracking );
 						break;
-					case "usps":
-						$tracking_url = 'https://tools.usps.com/go/TrackConfirmAction.action?tLabels=' . $tracking;
+					case 'usps':
+						$tracking_url = 'https://tools.usps.com/go/TrackConfirmAction.action?tLabels=' . esc_html( $tracking );
```
jeffstieler: No need to `esc_html()` the tracking number here (or with fedex), as the whole url will be passed to `esc_url()`.  _(see comment below)_

woocommerce-services.php
```diff
@@ -694,28 +693,35 @@ public function add_tracking_info_to_emails( $order, $sent_to_admin, $plain_text
...
 				$refunded = array_key_exists( 'refund', $label );
 
 				if ( $refunded ) {
 					continue;
 				}
 
+				if ( $plain_text ) {
+					// Should look like '- USPS: 9405536897846173912345' in plain text mode
+					$bullet = '- ';
+					$markup .= $bullet . $carrier . ': ' . $tracking . "\n";
```
jeffstieler: Let's just use the bullet inline here. Like `'- ' . $carrier` ...

woocommerce-services.php
```diff
@@ -725,16 +731,27 @@ public function add_tracking_info_to_emails( $order, $sent_to_admin, $plain_text
...
-			echo '<tr>';
-			echo '<th class="td" scope="col">' . __( 'Provider', 'woocommerce-services' ) . '</th>';
-			echo '<th class="td" scope="col">' . __( 'Tracking number', 'woocommerce-services' ) . '</th>';
-			echo '</tr>';
-			echo '</thead>';
-			echo '<tbody>' . $markup . '</tbody>';
-			echo '</table>';
+			if ( $plain_text ) {
+				echo "\n=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=\n\n";
+				echo strtoupper( __( 'Tracking', 'woocommerce-services' ) ) . "\n\n";
```
jeffstieler: I'm not sure `strtoupper()` is "safe" for translated strings. (See: https://secure.php.net/manual/en/function.strtoupper.php#112383)

It might be better to just have the string be `'TRACKING'`, or perhaps make use of `mb_strtoupper()` with the `UTF-8` argument.

woocommerce-services.php
```diff
@@ -725,16 +731,27 @@ public function add_tracking_info_to_emails( $order, $sent_to_admin, $plain_text
...
+				echo $markup;
+				return;
+			}
+
+			?>
+				<h2><?php echo __( 'Tracking', 'woocommerce-services' ) ?></h2>
```
jeffstieler: Please add a semicolon to this line (and all others like it).

woocommerce-services.php
```diff
@@ -725,16 +731,27 @@ public function add_tracking_info_to_emails( $order, $sent_to_admin, $plain_text
...
+				<table class="td" cellspacing="0" cellpadding="6" style="margin-top: 10px; width: 100%; font-family: \'Helvetica Neue\', Helvetica, Roboto, Arial, sans-serif;">
```
jeffstieler: Do we still need to escape these single quotes?

woocommerce-services.php
```diff
@@ -694,28 +693,35 @@ public function add_tracking_info_to_emails( $order, $sent_to_admin, $plain_text
 
 			// Generate a table row for each label
 			foreach ( $labels as $label ) {
-				$carrier = esc_html( $label['carrier_id'] );
-				$tracking = esc_html( $label['tracking'] );
+				$carrier = strtoupper( $label['carrier_id'] );
```
jeffstieler: Moving the `strtoupper()` call here broke the `switch()` case below. Let's either change the `case`s to be uppercase strings or call `strtoupper()` as late as possible.

joshuaja:
@jeffstieler I was out of office this week but will make the changes this weekend. Thanks!

joshuaja:
@jeffstieler - I completed the last round of changes. Please note I added a wrapper `div` to match changes made to the Woocommerce email-customer-details.php template in this [commit](https://github.com/woocommerce/woocommerce/commit/94b32e6#diff-4f465e1a9e4ce62c5827b3f4e5d69fcbR27).

joshuaja:
@jeffstieler - I added a commit addressing your requested changes. Will you have time to review this week?

Thanks much!

jeffstieler:
> Will you have time to review this week?

@joshuaja I was out of the office for a bit there. I hope to review sometime this week. 👍

woocommerce-services.php
```diff
@@ -667,6 +668,95 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+		 *
+		 * @param $order
+		 * @param $sent_to_admin
+		 * @param $plain_text
+		 */
+		public function add_tracking_info_to_emails( $order, $sent_to_admin, $plain_text ) {
+			$id = WC_Connect_Compatibility::instance()->get_order_id( $order );
+
+			// Abort if no id was passed or if the order is not marked as 'completed'
+			if ( ! $id || ! $order->has_status( 'completed' )) {
```
jeffstieler: Nitpick: the last parenthesis needs a space.

woocommerce-services.php
```diff
@@ -667,6 +668,95 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+			if ( empty( $labels ) ) {
+				return;
+			}
+
+			$markup = '';
+
+			// Generate a table row for each label
+			foreach ( $labels as $label ) {
+				$carrier = $label['carrier_id'];
+				$carrierLabel = strtoupper( $carrier );
```
jeffstieler: Nitpick: use `snake_case` for PHP variable names.

woocommerce-services.php
```diff
@@ -667,6 +668,95 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+					case 'fedex':
+						$tracking_url = 'https://www.fedex.com/apps/fedextrack/?action=track&tracknumbers=' . $tracking;
+						break;
+					case 'usps':
+						$tracking_url = 'https://tools.usps.com/go/TrackConfirmAction.action?tLabels=' . $tracking;
+						break;
+				}
+
+				$markup .= '<td class="td" scope="col">';
+				$markup .= '<a href="' . esc_url( $tracking_url ) . '">' . esc_html( $tracking ) . '</a>';
```
jeffstieler: In my testing, this link ended up being white against a white background. The culprit was the colors selected in my email settings and the interplay with `wc_light_or_dark()` used to choose a color for header text and also links in emails: https://github.com/woocommerce/woocommerce/blob/master/templates/emails/email-styles.php#L188

Let's force the tracking link text to use the color in the `woocommerce_email_text_color` option.

joshuaja:
@jeffstieler - I'm hoping to have some time to work on this today. Thanks!

joshuaja:
@jeffstieler - actually this took less time than I expected. I've implemented the latest round of review suggestions if you could review when you have a moment.

woocommerce-services.php
```diff
@@ -667,6 +668,96 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
+
+			$markup = '';
+			$link_color = get_option( 'woocommerce_email_text_color' );
+
+			// Generate a table row for each label
+			foreach ( $labels as $label ) {
+				$carrier = $label['carrier_id'];
+				$carrier_label = strtoupper( $carrier );
+				$tracking = $label['tracking'];
+				$refunded = array_key_exists( 'refund', $label );
```
marcinbot: We need to check for an `error` field in a similar way. Otherwise this results in an error in the email:
screen shot 2017-11-16 at 14 18 29

joshuaja: @marcinbot - just to confirm, you're suggesting to check for an `error` key in the $label array like this:

```php
$error = array_key_exists( 'error', $label );
$refunded = array_key_exists( 'refund', $label );

// If the label has an error, move to the next label
if ( $error || $refunded ) {
    continue;
}
```

marcinbot: Yep, looks good

joshuaja:
@marcinbot, @jeffstieler - I added a check for an `error` key in addition to a `refund`key within the label. It moves onto the next label if it finds either key.

jeffstieler:
@joshuaja thanks for your contribution! If you'd like to be listed under "Contributors" on WordPress.org, please let me know what your WordPress.org username is!

joshuaja:
@jeffstieler - thanks! My Wordpress user account is dappermountain

jeffstieler:
@joshuaja - I added `dappermountain` as a contributor on WordPress.org, but it seems that user doesn't exist. ☹️

joshuaja:
@jeffstieler - sorry, that is my username on WordPress.com. I was thinking it shared the same username. I'll figure out how to sign up on WordPress.org under the same username and let you know when it's up there.

Thanks much!

joshuaja:
@jeffstieler - it should work now. I just registered.

Thanks!

PR #1177, opened 2017-10-07
Rework TOS banner to include non-shipping functionality

### LINKED ISSUE #1175: Rework TOS banner to include non-shipping functionality
screen shot 2017-10-06 at 10 38 35 am

Now that WCS include automated taxes, stripe account creation, and shipping - we need to rework this banner to be relevant for all users.

jeffstieler:
Fixes #1175.

The best way to test this is to manipulate the value of `$banner_to_display` in `WC_Connect_Nux::set_up_nux_notices()`. (Make sure that the TOS accepted flag is set to `false`)

In addition to rewording the banner, it removes the "TOS only" banner in favor of reusing the "pre connection" banner for all cases where we want to show TOS.


classes/class-wc-connect-nux.php
```diff
@@ -342,6 +342,7 @@ public function set_up_nux_notices() {
 
 			switch ( $banner_to_display ) {
 				case 'before_jetpack_connection':
+				case 'tos_only_banner':
```
dechov: Old case can be removed: https://github.com/Automattic/woocommerce-services/blob/85dbbc637e4ba68da659bce11999a125f3807e0b/classes/class-wc-connect-nux.php#L372-L375

jeffstieler: Yep, I missed that. I'm going to rewrite this commit to include the changes. Force pull after.

jeffstieler: Fixed.

classes/class-wc-connect-nux.php
```diff
@@ -452,50 +450,26 @@ public function show_banner_after_connection() {
...
-				'description'    => sprintf( wp_kses( __( 'Everything is ready to roll, we just need you to agree to our <a href="%1$s">Terms of Service</a>.', 'woocommerce-services' ),
-					array(
-						'a' => array(
-							'href' => array(),
-						),
-					) ),
-					'https://woocommerce.com/terms-conditions/'
-				),
-				'button_text'    => __( 'I accept', 'woocommerce-services' ),
+				'title'          => __( 'Setup complete.', 'woocommerce-services' ),
```
dechov: This title has no punctuation in other past or current cases — may want to omit?

classes/class-wc-connect-nux.php
```diff
@@ -395,36 +392,37 @@ public function show_banner_before_connection() {
 
 			$jetpack_status = $this->get_jetpack_install_status();
 
-			$button_text = __( 'Connect', 'woocommerce-services' );
+			$button_text  = __( 'Connect', 'woocommerce-services' );
+			$banner_title = __( 'Connect your store to activate WooCommerce Services', 'woocommerce-services' );
```
dechov: This text and title only shows up in the case where Jetpack is active but not connected. Why "Connect Jetpack" for the other cases, but "Connect your store" with no mention of Jetpack in this one? If we're clarifying what it is we're connecting in the other two cases, why not in this one?

jeffstieler: Good catch - this also highlighted some deficiencies in our copy regarding features - working on that now.

jeffstieler:
@dechov this is ready for (yet another) review - thanks for your patience!

classes/class-wc-connect-nux.php
dechov: Ah, as the cleaned up diff makes clear, I was incorrect about punctuation not being used at the end of titles.

v18:
Looks like this PR breaks this user flow:

1. Have Jetpack already connected
2. Go through wizard & select a WCS-related service
3. See the banner, which has a JP connection button instead of a TOS-only button.

Here's an issue with more detailed repro steps:

https://github.com/Automattic/woocommerce-services/issues/1188

I think we should solve it by bringing back all `tos_only_banner` logic, and the `show_tos_banner()` function, and change the wording in the tos banner as needed. We should have two separate functions instead of relying on one banner to do the work of two (e.g. `before_jetpack_connection` and `tos_only_banner` both being displayed using `show_banner_before_connection`).

PR #1181, opened 2017-10-10
Add TOS bypass for Atomic sites

dechov:
Because the "Ecommerce" section of the [WP.com TOS](https://en.wordpress.com/tos/) covers the functionality of this plugin, we can check if the site is an Atomic site as a condition for auto-accepting the TOS and proceeding.

To test, a) install/activate on Atomic site (use `npm run release` and upload ZIP) and verify plugin is functional without any TOS prompt, and b) activate on non-atomic site and verify that TOS prompt still shows up once Jetpack is activated (outside of WC wizard), and plugin doesn't work until TOS accepted, as before.

Note: `jetpack_is_atomic_site` appears to be somewhat of an undocumented function intended for internal use — it was actually [renamed](https://github.com/Automattic/jetpack/pull/7729) from `jetpack_is_automated_transfer_site` in Jetpack 5.4 — is there some risk in relying on it?

Fixes https://github.com/Automattic/wp-calypso/issues/16954
Closes https://github.com/Automattic/woocommerce-services/pull/1138

### DIFF
--- a/classes/class-wc-connect-jetpack.php
+++ b/classes/class-wc-connect-jetpack.php
@@ -39,6 +39,18 @@ if ( ! class_exists( 'WC_Connect_Jetpack' ) ) {
 		}
 
 		/**
+		 * Helper method to get whether the current site is an Atomic site
+		 * @return bool
+		 */
+		public static function is_atomic_site() {
+			if ( function_exists( 'jetpack_is_atomic_site' ) ) {
+				return jetpack_is_atomic_site();
+			}
+
+			return false;
+		}
+
+		/**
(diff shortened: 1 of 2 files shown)


jeffstieler:
> Note: `jetpack_is_atomic_site` appears to be somewhat of an undocumented function intended for internal use — it was actually renamed from `jetpack_is_automated_transfer_site` in Jetpack 5.4 — is there some risk in relying on it?

Let's just handle both by adding an `elseif` that uses the old function here: https://github.com/Automattic/woocommerce-services/pull/1181/files#diff-3851bfa97209bb11391036b0d6ae2dd1R48

PR #1182, opened 2017-10-10
Return the PDF in JSON format (base64-encoded) if the "json" parameter is set

DanReyLop:
If the `conect/label/print` request has a POST parameter `"json": true`, instead of spitting the raw PDF, the endpoint will return a regular json, like this:
```js
{
  "mimeType": "application/pdf",
  "b64Content": "...THE....BINARY....PDF....CONTENT.....ENCODED...IN...BASE64...."
}
```

Printing from `WP-Admin` shouldn't be affected by this, this change is only for requests coming from `wp-calypso` (which will have the `json` flag set to `true`).


jeffstieler:
Please hold off on merging this until I release and tag `1.8.0`

Issue #1166, opened 2017-09-26
[Labels] Can't print labels on iOS


joshuaja:
When clicking on the "reprint" link, choosing the paper size, and clicking "Print" it does nothing on iOS. This is a non-starter for our business because the wifey prints her labels from her iPad while crafting her creations.

I'm assuming this is due to the call to `print()` within the React client component [here](https://github.com/Automattic/woocommerce-services/blob/b98ab1c14a2e93505f9448189107526551c21521/client/lib/utils/print-document.js#L46-L63)

Can we just load the PDF in a new tab instead of doing the whole iFrame thing and calling print()?

From a UX perspective it's an extra click but I think most people are used to previewing PDFs in a new tab before printing nowadays.

Help me understand the back story for the cases in the above client code and I can submit a Pull Request with a change if needed.

DanReyLop:
> I'm assuming this is due to the call to print() within the React client component here

Yeah, you're totally right, that's causing it. If you want to look at more scary cross-browser logic, [check out this](https://github.com/Automattic/woocommerce-services/blob/b98ab1c14a2e93505f9448189107526551c21521/client/lib/utils/pdf-support.js). That's the "backstory" you were missing :)

> From a UX perspective it's an extra click but I think most people are used to previewing PDFs in a new tab before printing nowadays.

It's an extra click, as opposed to not being able to print the PDF at all. I think the decision is clear :)

We'll just add `iOS` as a special case (like Firefox is now) so it opens the PDF in a new tab (with an extra click). May be a good time to try it in an Android phone too.

joshuaja:
@DanReyLop - Just a follow-up I tested this in Chrome on an Android phone (Samsung S6). It downloads the PDF to the device (I'm assuming it hits the `default` case).

joshuaja:
@DanReyLop - Fixed and tested in PR #1185

johndcoy:
Looks like this issue has resurfaced. Clicking Print on Safari using an iPad 7 does nothing other than close the reprint window.

I replicated this issue in browserstack but also on a physical iPad using two different sites.

harriswong:
I gave a few iPad a test on browserstack. It seems like a few of them stopped including `iPad` in their `navigator.userAgent`. 
![image](https://user-images.githubusercontent.com/572862/96931832-2b9a5680-147b-11eb-9bab-5ef3a8c87fdc.png)

This causes the condition to fall back to `default` and use `print()` as it described originally in this issue.

PR #1185, opened 2017-10-12
🐛 Fix label printing on iOS [#1166]

joshuaja:
Mobile Safari and Chrome are smart enough to determine that `window.open` originated from a user click and is opening the pdf in a new browser tab.

Tested and working on:

- iPhone 6S (iOS 9.0, Mobile Safari)
- iPad Air 2 (iOS 8.0, Mobile Safari)
- iPad Air 2 (iOS 11, Mobile Safari)
- iPad Air 2 (iOS 11, Chrome)

### DIFF
--- a/client/lib/utils/pdf-support.js
+++ b/client/lib/utils/pdf-support.js
@@ -13,6 +13,9 @@ import _ from 'lodash';
  */
 export default _.memoize( () => {
 	if ( /iPad|iPhone|iPod/.test( navigator.userAgent ) && ! window.MSStream ) {
+		// iOS doesn't support triggering a print dialog, so we should load the pdf in a new tab
+		// instead. Windows Phones are filtered out with `! window.MSStream` since the user agent
+		// string can contain the false positive 'like iPhone'
 		return 'addon';
 	}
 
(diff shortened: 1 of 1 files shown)


client/lib/utils/pdf-support.js
```diff
@@ -12,6 +12,10 @@ import _ from 'lodash';
  * PDFs at all.
  */
 export default _.memoize( () => {
+	if ( /iPad|iPhone|iPod/.test( navigator.userAgent ) && ! window.MSStream ) {
```
DanReyLop: Damn those silly user-agents! I've just learnt that IE Mobile uses `like iPhone` on its user-agent, thats low :(

Could you please add a comment about why this is needed? This file is full of seemingly arbitrary rules that should be explained. Then this is good to go :)

joshuaja: @DanReyLop - I added a commit containing documentation. Thanks for the review!

DanReyLop:
Thanks for your PR! We'll do a new release with this as soon as possible (not at the end of the week, that's just asking for trouble).

PR #1190, opened 2017-10-13
Fix TOS-only banner - remove connection flow

### LINKED ISSUE #1188: Show TOS-only banner for users who are already connected to Jetpack
Steps to repro:

1. Ensure Jetpack is installed + connected
2. Uninstall all WooCommerce and WooCommerce Services things. Remove all DB options, etc
3. Go through WC wizard using a US locale, and make sure to enable Stripe + enter your email to create an account automatically
4. After you finish the wizard, go to the plugins page
5. See this banner:

screen shot 2017-10-13 at 8 28 18 am

This banner directs users to connect to Jetpack. It's incorrect because in this case the user is already connected to Jetpack.

Instead, we should display this banner:

tos-only-banner

This banner should save the TOS_accepted option.

jeffstieler:
Fixes #1188.

This PR seeks to restore the TOS-only banner (with new verbiage) that doesn't redirect to WPCOM.

It also adds the "should display" logic to the TOS-only banner, so it doesn't display on every admin screen. (The idea being that will help with "nag" complaints)

To test:
* Ensure a connected Jetpack
* Either remove the WCS plugin (and options), or set `wc_connect_options.tos_accepted` to `0`
* Verify that on NUX banner pages (plugins, WC pages) the TOS-only banner shows
* Verify that clicking "connect" dismisses the banner and sets `wc_connect_options.tos_accepted` to `1` and doesn't send you to WPCOM

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -474,26 +474,21 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 
 			$jetpack_status = $this->get_jetpack_install_status();
 			$button_text    = __( 'Connect', 'woocommerce-services' );
-			$banner_title   = __( 'Connect your store to activate WooCommerce Services', 'woocommerce-services' );
+			$banner_title   = __( 'Connect Jetpack to activate WooCommerce Services', 'woocommerce-services' );
 			$image_url      = plugins_url( 'images/wcs-notice.png', dirname( __FILE__ ) );
 
 			switch ( $jetpack_status ) {
 				case self::JETPACK_NOT_INSTALLED:
 					$button_text  = __( 'Install Jetpack and connect', 'woocommerce-services' );
-					$banner_title = __( 'Connect Jetpack to activate WooCommerce Services', 'woocommerce-services' );
 					break;
 				case self::JETPACK_INSTALLED_NOT_ACTIVATED:
 					$button_text  = __( 'Activate Jetpack and connect', 'woocommerce-services' );
-					$banner_title = __( 'Connect Jetpack to activate WooCommerce Services', 'woocommerce-services' );
-					break;
-				case self::JETPACK_ACTIVATED_NOT_CONNECTED:
-					$banner_title = __( 'Connect Jetpack to activate WooCommerce Services', 'woocommerce-services' );
 					break;
 			}
 
 			$country = WC()->countries->get_base_country();
 			/* translators: %s: list of features, potentially comma separated */
-			$description_base = __( "WooCommerce Services is almost ready to go! Once you connect your store you'll have access to %s.", 'woocommerce-services' );
+			$description_base = __( "WooCommerce Services is almost ready to go! Once you connect Jetpack you'll have access to %s.", 'woocommerce-services' );
 			$feature_list     = $this->get_feature_list_for_country( $country );
 			$banner_content   = array(
 				'title'             => $banner_title,
@@ -580,7 +575,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 					'wcs-nux-tos' => 'accept',
 				) ),
 				'image_url'      => plugins_url(
-					'assets/images/wcs-notice.png', dirname( __FILE__ )
+					'images/wcs-notice.png', dirname( __FILE__ )
 				),
 				'should_show_jp'    => true,
 				'should_show_terms' => true,
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -417,7 +417,6 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 
 			switch ( $banner_to_display ) {
 				case 'before_jetpack_connection':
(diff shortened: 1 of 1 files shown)


classes/class-wc-connect-nux.php
```diff
@@ -546,6 +545,44 @@ public function show_banner_after_connection() {
...
+
+			$country = WC()->countries->get_base_country();
+			/* translators: %s: list of features, potentially comma separated */
+			$description_base = __( "WooCommerce Services is almost ready to go! Once you connect your store you'll have access to %s.", 'woocommerce-services' );
+			$feature_list     = $this->get_feature_list_for_country( $country );
+
+			$this->show_nux_banner( array(
+				'title'          => __( 'Connect your store to activate WooCommerce Services', 'woocommerce-services' ),
+				'description'    => esc_html( sprintf( $description_base, $feature_list ) ),
+				'button_text'    => __( 'Connect', 'woocommerce-services' ),
```
dechov: The verbiage here in title, description, and button_text still evokes something more involved than flipping the TOS bit — as in, if they've already connected their store to Jetpack, so what are they connecting it to now?

jeffstieler: This is the same verbiage as introduced in https://github.com/Automattic/woocommerce-services/pull/1177.

In this scenario, the user is in effect connecting their store to WooCommerce Services (SaaS layer). Since they're now opting into WCS, more of the Jetpack TOS apply to them, so we redisplay those terms.

dechov: As https://github.com/Automattic/woocommerce-services/issues/1188 (being fixed) says,
> This banner directs users to connect to Jetpack.

Even if it doesn't say "Connect to Jetpack" in this case, why overload the term? I understand that we want to redisplay the terms, but in the event we do want the term "Connect" for accepting WCS terms, we should at least clarify that they're connecting to not-Jetpack in my opinion. cc @v18

jeffstieler: @dechov @v18 can we agree to discuss the wording separate from reviewing this bug fix PR?

dechov: Yes! Sorry, I had missed the context of there being an actual bug separate from the verbal communication. Disregard ^^

classes/class-wc-connect-nux.php
```diff
@@ -471,26 +474,22 @@ public function show_banner_before_connection() {
...
 			$button_text    = __( 'Connect', 'woocommerce-services' );
-			$banner_title   = __( 'Connect your store to activate WooCommerce Services', 'woocommerce-services' );
+			$banner_title   = __( 'Connect Jetpack to activate WooCommerce Services', 'woocommerce-services' );
 			$image_url      = plugins_url( 'images/wcs-notice.png', dirname( __FILE__ ) );
 
 			switch ( $jetpack_status ) {
 				case self::JETPACK_NOT_INSTALLED:
 					$button_text  = __( 'Install Jetpack and connect', 'woocommerce-services' );
-					$banner_title = __( 'Connect Jetpack to activate WooCommerce Services', 'woocommerce-services' );
+
```
v18: Accidental extra line?

jeffstieler: Yep - going to rewrite that commit to remove it.

jeffstieler: Done in d5617a2.

classes/class-wc-connect-nux.php
```diff
@@ -546,6 +545,44 @@ public function show_banner_after_connection() {
...
+				'title'          => __( 'Connect your store to activate WooCommerce Services', 'woocommerce-services' ),
+				'description'    => esc_html( sprintf( $description_base, $feature_list ) ),
+				'button_text'    => __( 'Connect', 'woocommerce-services' ),
+				'button_link'    => add_query_arg( array(
+					'wcs-nux-tos' => 'accept',
+				) ),
+				'image_url'      => plugins_url(
+					'images/wcs-notice.png', dirname( __FILE__ )
+				),
+				'should_show_jp'    => true,
```
v18: Question: do we want to display the JP logo in this banner? We didn't used to display it here, and I'm not really sure it's needed.

classes/class-wc-connect-nux.php
v18: Agree with verbiage questions - opened separate issue for it here: https://github.com/Automattic/woocommerce-services/issues/1192#issue-265428329

dechov: Thanks — and 👍  @jeffstieler for distinguishing urgency & prioritization of different matters

classes/class-wc-connect-nux.php
jeffstieler: Whoops! Good catch - that was a copy-paste error I made.

jeffstieler: Fixed in https://github.com/Automattic/woocommerce-services/pull/1190/commits/60954b9281d0d80700250276a076bd42663cf7ef.

v18:
Looks like the bug still exists. Re-testing to see what's going wrong.

PR #1191, opened 2017-10-13
Include additional metadata in request settings

dechov:
It would be nice to know if errors reported by the production server are resolved by plugin changes we release — I see WP, WC, and JP versions reported in the error but not WCS:
```
{
   [...]
   "request": {
       "settings": {
           [...]
           "jetpack_version": "5.4",
           "wc_version": "3.2.0",
           [...]
           "wp_version": "4.8.2",
           [...]
       },
       [...]
   }
}
```

Is there already some other way to detect this? Otherwise, is there a more elegant way to implement?


DanReyLop:
There's no other way to detect it currently.

This is a good idea, the only thing I would change is to get the WCS version from the plugin header at runtime, instead of having yet another manual and error-prone step to do in each release.

jeffstieler:
This will need a corresponding server PR (validation) before we can release.

dechov:
Do we want this in Tracks too? https://github.com/Automattic/woocommerce-services/blob/d4c24e9db5916c50acff0a4ce2694f14d69b43aa/classes/class-wc-connect-tracks.php#L103-L105

If so (or even if not), where should the `get_wcs_version` function live?

v18:
Just noting that in our meeting today we also said that want to send whether a site is atomic or not.

DanReyLop:
> Do we want this in Tracks too?

I think so, we're including the WC and WP version too.

> If so (or even if not), where should the get_wcs_version function live?

As an static function in the `woocommerce-services.php` main plugin class? It's parsin the plugin header, so it kind-of makes sense maybe?

woocommerce-services.php
```diff
@@ -185,6 +185,17 @@ static function plugin_uninstall() {
 			WC_Connect_Options::delete_all_options();
 		}
 
+		/**
+		 * Get WCS plugin version
+		 *
+		 * @return string
+		 */
+		static function get_wcs_version() {
+			$file_path = plugin_dir_path( __FILE__ ) . 'woocommerce-services.php';
```
DanReyLop: **This** is the `woocommerce-services.php` file, you can just use `get_file_data(__FILE__, array(...))`

dechov: Ah, right 😴

Issue #1193, opened 2017-10-13
Shipping method init crashes when country's default service isn't supported by currency


dechov:
Services are limited by currency as of https://github.com/Automattic/woocommerce-connect-server/pull/984, which may be causing an error. Testing with country=USA and currency=GBP, this is the error:

> Notice: Trying to get property of non-object in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/woocommerce-services.php on line 443
> 
> Fatal error: Uncaught Error: Call to a member function get_service_schema() on boolean in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/woocommerce-services.php:448 Stack trace: #0 /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/woocommerce-services.php(466): WC_Connect_Loader->add_method_to_shipping_zone(0, 'usps') #1 /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/woocommerce-services.php(605): WC_Connect_Loader->init_core_wizard_shipping_config() #2 /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/woocommerce-services.php(515): WC_Connect_Loader->attach_hooks() #3 /srv/www/wordpress-default/public_html/wp-includes/class-wp-hook.php(298): WC_Connect_Loader->after_wc_init('') #4 /srv/www/wordpress-default/public_html/wp-includes/class-wp-hook.php(323): WP_Hook->apply_filters('', Array) #5 /srv/www/wordpress-default/public_html/wp-includes/plugin.php(453): WP_Hook->do_action(Array) #6 /srv/www/wordp in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/woocommerce-services.php on line 448

Setting high priority since GBP is the default and doesn't change when selecting country.

jeffstieler:
This will also need to have a corresponding change in the core setup wizard. We can be defensive and check for the symptom here, but the root cause is that we're allowing `USA`+`GBP` stores to configure live rates.

jeffstieler:
We should make sure `$instance` isn't `false` here: https://github.com/Automattic/woocommerce-services/blob/master/woocommerce-services.php?utf8=%E2%9C%93#L448

We should also align the check here (https://github.com/Automattic/woocommerce-services/blob/master/woocommerce-services.php?utf8=%E2%9C%93#L457) with the service side, which checks currency instead of country.

PR #1195, opened 2017-10-16
Map currency instead of country to shipping method in post-Wizard config init

dechov:
Fixes: https://github.com/Automattic/woocommerce-services/issues/1193

Avoids fatal error when shipping method is unavailable due to mismatch between country and currency, by adapting to the server's recently introduced mapping by _currency_ (rather than _country_ as initially established in https://github.com/Automattic/woocommerce-services/pull/1165.)

Also hardens `add_method_to_shipping_zone` by making sure the method and method instance exist before proceeding to configure them.

To test, set country to US or CA but currency to something other than USD or CAD — confirm that the wizard is completed without WCS-derived fatal error.

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -438,12 +438,12 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		}
 
 		protected function add_method_to_shipping_zone( $zone_id, $method_id ) {
-			$method      = $this->get_service_schemas_store()->get_service_schema_by_id( $method_id );
+			$method = $this->get_service_schemas_store()->get_service_schema_by_id( $method_id );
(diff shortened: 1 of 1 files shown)


woocommerce-services.php
```diff
@@ -473,9 +479,9 @@ public function init_core_wizard_shipping_config() {
 					if (
 						1 === count( $zone['zone_locations'] ) &&
 						'country' === $zone['zone_locations'][0]->type &&
-						$store_country === $zone['zone_locations'][0]->code
+						WC()->countries->get_base_country() === $zone['zone_locations'][0]->code
```
jeffstieler: Let's get the base country outside of the loop so we're not retrieving it more than necessary.

PR #1198, opened 2017-10-19
Stripe Connect: Add endpoint that fetches connected account details from server

### LINKED ISSUE #1125: Stripe: Add endpoint(s) that returns Stripe Connect User Account details, if any
At a minimum, the endpoint should, for users with manage_woocommerce capability, query the connect server for the connected stripe user (if any) and return it to the requestor (e.g. Calypso ala Jetpack).

At a minimum the endpoint should return the Stripe connected account user id (e.g. acct_14qyt6Alijdnw0EA ) and its activated state, but the endpoint (or another endpoint) should also relay:

- display_name
- email
- business_logo
- legal_entity.first_name
- legal_entity.last_name

See https://stripe.com/docs/api#account or api.stripe.com/v1/accounts/ for field definitions

If the store has no Stripe Connect User, the endpoint should probably still 200 but return false or something for the fields

TBD on whether we need to retrieve activated state stored from a webhook or if we can get activated state from api.stripe.com/v1/accounts/

allendav:
Fixes #1125 

<del>Note: Testing requires server PR 987 to be merged to 975 first.</del>

To test:
- See "old test instructions" in comment on 987

### DIFF
--- a/classes/class-wc-connect-stripe.php
+++ b/classes/class-wc-connect-stripe.php
@@ -55,6 +55,14 @@ if ( ! class_exists( 'WC_Connect_Stripe' ) ) {
 			return $this->save_stripe_keys( $response );
 		}
 
+		public function get_account_details() {
+			$response = $this->api->get_stripe_account_details();
+			if ( is_wp_error( $response ) ) {
+				return $response;
+			}
+			return $response;
+		}
+
 		public function connect_oauth( $state, $code ) {
 			if ( $state !== $this->options->get_option( self::STATE_VAR_NAME, false ) ) {
 				return new WP_Error( 'Invalid stripe state' );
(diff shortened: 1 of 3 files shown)


allendav:
@jeffstieler @DanReyLop - with 987 merged to 975 on the server, this is ready for testing - I'm working on updating calypso to talk with this now

classes/class-wc-connect-stripe.php
```diff
@@ -55,6 +55,14 @@ public function create_account( $email, $country ) {
 			return $this->save_stripe_keys( $response );
 		}
 
+		public function get_account_details() {
+			$response = $this->api->get_stripe_account_details();
+			if ( is_wp_error( $response ) ) {
```
marcinbot: This condition is unnecessary since `$response` is always returned.

allendav: Fixed in https://github.com/Automattic/woocommerce-services/pull/1198/commits/ccd27adfa92f322f384041fe9695c0b1704ca7b0

marcinbot: Thanks for fixing, this is ready to merge

PR #1199, opened 2017-10-19
Add metadata to distinguish Stripe referral vs. connect requests

dechov:
~Adds [metadata](https://stripe.com/docs/api#metadata) (key: `connect`, value: `1`) to~ Removes User-Agent from Stripe requests from accounts created via WCS after WooCommerce setup wizard opt-in (discussion in p7bbVw-1Rh-p2)

Requires WooCommerce Stripe Payment Gateway 4.0.0 ([latest](https://github.com/woocommerce/woocommerce-gateway-stripe/tree/4.0.0)) but shouldn't break without it — tested by logging request data in the Stripe plugin's [API class](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/e0ac1c3c28e8718810e5c2ffd0ae84d411e415a4/includes/class-wc-stripe-api.php#L101).

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -500,10 +500,6 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 				if ( is_wp_error( $response ) ) {
 					// TODO handle case of existing account
 					$this->logger->debug( $response, __CLASS__ );
-				} else {
-					$stripe_settings = get_option( 'woocommerce_stripe_settings', array() );
-					$stripe_settings['connect'] = 'yes';
-					update_option( 'woocommerce_stripe_settings', $stripe_settings );
 				}
 			}
 		}
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -501,8 +501,8 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 					// TODO handle case of existing account
 					$this->logger->debug( $response, __CLASS__ );
 				} else {
-					$stripe_settings = get_option( 'woocommerce_stripe_settings', false );
-					$stripe_settings[ 'connect' ] = 'yes';
+					$stripe_settings = get_option( 'woocommerce_stripe_settings', array() );
+					$stripe_settings['connect'] = 'yes';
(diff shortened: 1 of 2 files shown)


dechov:
Pending https://github.com/woocommerce/woocommerce-gateway-stripe/pull/389, I believe the filter can be changed to return `null` on tag `woocommerce_stripe_request_user_agent`.

The placement of the `add_filter` call can be reviewed already.

woocommerce-services.php
```diff
@@ -500,6 +500,10 @@ public function init_core_wizard_payments_config() {
 				if ( is_wp_error( $response ) ) {
 					// TODO handle case of existing account
 					$this->logger->debug( $response, __CLASS__ );
+				} else {
+					$stripe_settings = get_option( 'woocommerce_stripe_settings', array() );
+					$stripe_settings['connect'] = 'yes';
+					update_option( 'woocommerce_stripe_settings', $stripe_settings );
```
jeffstieler: What do you think about doing this in [`WC_Connect_Stripe::save_stripe_keys()`](https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-stripe.php#L84) instead?

dechov:
Looks like we should try and release this by Nov 24 (p6q7sZ-3Mf-p2)

woocommerce-services.php
```diff
@@ -897,6 +899,14 @@ public function woocommerce_payment_gateways( $payment_gateways ) {
...
 		}
 
+		public function woocommerce_stripe_request_headers( $headers ) {
+			$stripe_settings = get_option( 'woocommerce_stripe_settings', array() );
+			if ( isset( $stripe_settings['connect'] ) && 'yes' === $stripe_settings['connect'] ) {
+				unset( $headers['User-Agent'] );
+			}
+			return $headers;
+		}
+
```
jeffstieler: Sorry for coming late with this opinion, but let's keep the Stripe functionality in the `WC_Connect_Stripe` class.

jeffstieler:
Please squash when you merge. 👍

PR #1200, opened 2017-10-24
Make paymentMethod not required for labels view

v18:
The payment method is null when it hasn't been entered yet, so it should not be a required prop.

Fixes this error:

screen shot 2017-10-16 at 9 32 10 pm


v18:
Thanks for the review @marcinbot :)

PR #1204, opened 2017-10-28
Do not allow default 500 to be returned by stripe rest api endpoints

### LINKED ISSUE #1203: Stripe: Do not allow endpoints to return WP_Errors without an explicit status
On certain hosts, like Pressable for example, if a REST API endpoint returns a WP_Error without an explicit status, the resulting 500 is intercepted and instead of the expected JSON, an HTML page is returned to the endpoint caller.

See https://developer.wordpress.org/rest-api/extending-the-rest-api/adding-custom-endpoints/#return-value

This HTML response disrupts callers that are expecting JSON

Props @seanosh for putting me on the right track

cc @justinshreve @timmyc @budzanowski @ryelle @coderkevin @seanosh

allendav:
Fixes #1203 

To Test:
* Install this branch as a plugin on a pressable site or other site that returns HTML for 500s
* Install the stripe connect test bench
* Navigate to https://{DOMAIN}/wp-admin/admin.php?page=wc-status&tab=sctbs
* Click each of the buttons and ensure JSON is returned even for errors (i.e. not HTML), e.g.

```
{"code":"wcc_server_error_response","message":"Error: The WooCommerce Services server returned: Bad Request Account already exists for the provided email. ( 400 )","data":{"status":400}}
```

### DIFF
--- a/classes/class-wc-rest-connect-stripe-account-controller.php
+++ b/classes/class-wc-rest-connect-stripe-account-controller.php
@@ -21,12 +21,15 @@ class WC_REST_Connect_Stripe_Account_Controller extends WC_REST_Connect_Base_Con
 		$response = $this->stripe->get_account_details();
 
 		if ( is_wp_error( $response ) ) {
-			$response->add_data( array(
-				'message' => $response->get_error_message(),
-			), $response->get_error_code() );
-
 			$this->logger->debug( $response, __CLASS__ );
-			return $response;
+
+			return new WP_Error(
+				$response->get_error_code(),
+				$response->get_error_message(),
+				array(
+					'status' => 400
+				)
+			);
 		}
 
 		return array(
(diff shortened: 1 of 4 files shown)


classes/class-wc-rest-connect-stripe-account-controller.php
```diff
@@ -21,12 +21,15 @@ public function get( $request ) {
...
-			), $response->get_error_code() );
-
 			$this->logger->debug( $response, __CLASS__ );
-			return $response;
+
+			return new WP_Error(
+				$response->get_error_code(),
+				$response->get_error_message(),
+				array(
+					'status' => 400
```
DanReyLop: Can you use the original HTTP code of the `$response`? AFAIK, the server won't ever return a `500` error unless something truly blew up.

How does Pressable behave with an `503` error? That's the closest code (semantically) to the "500". I'm thinking we should always use the status from the `$response` (except if it's 500, then replace it with something less offensive). Not only for the Stripe endpoints, but for everyone.

allendav: Not without changes to the plugin's API layer. That $response is a WP_Error and doesn't include the HTTP response code. https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-api-client.php#L404

DanReyLop: Ouch. Something to look into when we improve user-facing errors in Q4. Go for it then, :shipit:

PR #1205, opened 2017-10-30
Avoid coercing null to object in coerceFormValues

v18:
Steps to repro:

Go to WooCommerce > Settings > Shipping zones > [one of your shipping zones] > Edit when there have been no USPS services selected for the zone.

You'll see this:

screen shot 2017-10-30 at 11 11 13 am

This is the error:

> TypeError: Cannot convert undefined or null to object
> Object.keys( value )

It happens in https://github.com/Automattic/woocommerce-services/blob/master/client/lib/utils/coerce-values.js#L65 when the `value` argument of `coerceValue()` is equal to null.

I added a test to explain what I mean & to prevent regression.

### DIFF
--- a/client/lib/utils/coerce-values.js
+++ b/client/lib/utils/coerce-values.js
@@ -24,9 +24,13 @@ const getFieldSchema = ( fieldSchema, definitions ) => {
  * @param {Object} definitions - Schema definitions.
  * @returns {*} - Coerced value.
  */
-const coerceValue = ( schema, value, definitions ) => {
+export const coerceValue = ( schema, value, definitions ) => {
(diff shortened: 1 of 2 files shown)


jeffstieler:
> Steps to repro:
> 
> Go to WooCommerce > Settings > Shipping zones > [one of your shipping zones] > Edit when there have been no USPS services selected for the zone.

I wasn't able to reproduce this. The zone has services selected by default, and the server validates that at least one service is selected upon saving.

v18:
@jeffstieler I don't think we need to wait to reproduce this issue on a real site before merging. I included a test so that it can verify that the fix works, and the fix doesn't break behaviour.

PR #1216, opened 2017-11-06
Import package manager, label settings and label purchase as Calypso components

marcinbot:
Replaces views/components written for WCS with components imported from Calypso for package manager, label settings and order pages.


DanReyLop:
@kellychoffman @jameskoster We've been backporting pretty much everything from Calypso back to WP-Admin. One of those things was the `Order Activity Log`. Since the WP-Admin order page already has an "Order Notes" section we decided to include the exact same component as in Calypso, but passing it always an empty array of `order notes` instead of fetching them. This is the result:

sidebar-activity-log

I **think** there are some problems there. The sidebar is obviously narrower than the Calypso card. Should we try to work around that? Try to make the CSS more responsive to make it fit? Move the "Order Activity Log" into the main section of the Order screen instead of the sidebar? Forget about it entirely and just keep using the old component?

dechov:
Here it is with the tracking number, and after removing the double padding:

kellychoffman:
To make the activity design work in the context of the metabox, I would: 

- decrease the icon size
- decrease the horizontal padding on either side of the icon
- decrease the font size of the time stamp

DanReyLop:
@kellychoffman This is definitely better:

aaa


Is it ok to ship? I tried decreasing the icon size and font size of the timestamp but it didn't make a difference. There's no way to make the `Tracking #: 123456787654321` fit in a single line unless you remove *all* the paddings and make everything super small.

jameskoster:
We should decrease the top/bottom padding of the action box to match the left/right.

> There's no way to make the Tracking #: 123456787654321 fit in a single line unless you remove all the paddings

I think it looks fine as you have it.

kellychoffman:
> Is it ok to ship?

I'd just make sure the padding is even in each main event area (where it says Label #4 printed, etc). Right now it looks like the left padding is less that the rest.

jeffstieler:
It looks like the text is smaller in the labels modal than on Calypso:

screen shot 2017-11-20 at 5 15 16 pm

vs.


screen shot 2017-11-20 at 6 36 42 pm

kellychoffman:
That’s because the base text sixes are different on wp-admin and Calypso.

kellychoffman:
No, I don't think it needs to be addressed.

jeffstieler:
> > Does it need to be addressed? I thought we were going with Calypso styles, but making the font slightly larger shouldn't be a big change
> 
> No, I don't think it needs to be addressed.

If it's not a big change, I say let's do it - the text looks a little small to me.

kellychoffman:
Cool. Looks like just the header icons need to be vertically aligned now.

PR #1248, opened 2017-11-27
Allow changing label settings (except payment) to non-Master users

DanReyLop:
Related PR: https://github.com/Automattic/wp-calypso/pull/19514

Changed the `label settings` endpoint logic so any admin user is allowed to update the settings.

If the user is not the `Jetpack master user`, then the payment setting won't be updated, but the rest of the settings will.

It also adds a flag to the `GET` endpoint (`can_edit_settings`) which will always be `true`. This is done so the Calypso UI can differentiate between versions of the endpoints:
* For previous versions, `can_edit_settings` will be `false`-y (`undefined`), so only the `Master` user will be able to update the settings.
* For versions that include the `can_edit_settings` flag in the `GET` response, it's safe to present the user with the `Settings` form, and only hide the "Payment" part of it if he's not the `Master` user.

### DIFF
--- a/classes/class-wc-rest-connect-account-settings-controller.php
+++ b/classes/class-wc-rest-connect-account-settings-controller.php
@@ -34,6 +34,7 @@ class WC_REST_Connect_Account_Settings_Controller extends WC_REST_Connect_Base_C
 			'formData' => $this->settings_store->get_account_settings(),
 			'formMeta' => array(
 				'can_manage_payments' => $this->can_user_manage_payment_methods(),
+				'can_edit_settings' => true,
 				'master_user_name' => is_a( $master_user, 'WP_User' ) ? $master_user->display_name : '',
 				'master_user_login' => is_a( $master_user, 'WP_User' ) ? $master_user->user_login : '',
  				'payment_methods' => $this->payment_methods_store->get_payment_methods(),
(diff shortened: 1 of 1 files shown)


classes/class-wc-rest-connect-account-settings-controller.php
```diff
@@ -34,6 +34,7 @@ public function get() {
 			'formData' => $this->settings_store->get_account_settings(),
 			'formMeta' => array(
 				'can_manage_payments' => $this->can_user_manage_payment_methods(),
+				'can_edit_settings' => true,
```
marcinbot: Is this field required if it's always true? `can_manage_payments` can be used to prevent users from modifying the payment-related settings as in the code below.

DanReyLop: We need the `can_edit_settings` to maintain backwards-compatibility in Calypso. So:
* `can_manage_payments: true`: Show the whole UI, the user can do whatever.
* `can_manage_payments: false`: Disable the whole UI, the user can't edit payments or anything else, because this is an old version of the plugin that will fail if he tries to save *anything*.
* `can_manage_payments: false, can_edit_settings: true`: Disable the payment UI but keep the rest of the UI, since this is a new version of the plugin that will admit changes in the non-payment label settings.

PR #1254, opened 2017-12-06
Filter PPEC request endpoint to redirect requests via WCS server

dechov:
Adds a filter to proxy all of the PPEC plugin's requests via WCS server. The server will forward the requests and authenticate them by injecting credentials — requires https://github.com/Automattic/woocommerce-connect-server/pull/1012. (Context in p7bbVw-2kH-p2.)

Inert unless a PPEC version with https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/347 is active.

Note that the PPEC plugin settings support choosing between two different sets of credentials (Sandbox and Live), which affects the URL that requests are sent to. <s>However, unless anyone would argue otherwise, I assume that proxying both Sandbox and Live requests is unnecessary, and that this could just depend on the server environment: for the production server, only the Live API Subject will be filled in for them, and for the staging server, you'll have to set Environment to Sandbox and fill in the Sandbox API Subject manually.</s>

Left to do (some things may wait for future PRs):
- filter PPEC plugin settings
  - [x] disable (at least) Payment Action, for which the non-default option won't work
  - [x] perhaps alter the copy to make the user aware that request authentication is being handled for them
  - [x] possibly add a setting for `reroute_requests`, which will be opt-in from the WC wizard (see https://github.com/woocommerce/woocommerce/pull/18039)
- [x] probably want to filter what PPEC payment gateway "supports" to disable refunds and subscriptions, which won't work
- [x] show API Subject setting if it's filled, even if other credentials are not
- [x] admin banner/notice after an order is received, inviting user to link PayPal account
- [x] hide PPEC "prompt_to_connect" when proxying requests
- [x] adjust WCS NUX banner to generalize `$supports_stripe`

Testing instructions: TBD

P.S. In my case (VVV), testing with a local server at 10.0.2.2:5000 required working around the [safe URL validation](https://developer.wordpress.org/reference/functions/wp_http_validate_url/) triggered [from the PPEC plugin](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/abb51b0325587b036a5ea75092225fb1b414892f/includes/class-wc-gateway-ppec-client.php#L133).


jeffstieler:
> I assume that proxying both Sandbox and Live requests is unnecessary, and that this could just depend on the server environment

It's still useful to allow the client to determine live/sandbox. If something isn't working as expected, or as a test before "going live" merchants can use the sandbox.

dechov:
Thanks — this seems like something well worth thinking through a bit more. As it stands, if the merchant knows and cares to do so, they're still welcome to switch their "Environment" to "Sandbox" and create and link a sandbox account for testing. In this case, the username will be filled in and so the endpoint won't be modified. When they switch back to "Live", that username will be empty and so the requests will be proxied again.

If you mean the merchant should have a chance to test out the email-only flow, that's fair. Two concerns with that though:
- the WC wizard would fill in the `api_subject` setting but not `sandbox_api_subject`, so either the merchant would have to manually enter it, or we should be copying it over for them. (This is already the situation for staging/QA purposes — which I'll describe in the test instructions — but not part of the merchant experience.)
- the sandbox accounts don't send an email to the payment recipient (i.e. API Subject) asking them to create an account / claim a payment. I had to test with live accounts & payments in order to verify that such emails are sent.

woocommerce-services.php
```diff
@@ -528,6 +528,175 @@ public function init_core_wizard_payments_config() {
...
+					// Reroute requests from the PPEC extension via WCS to pick up API credentials
+					add_filter( 'woocommerce_paypal_express_checkout_request_endpoint', array( $this, 'paypal_ec_endpoint' ) );
+
+					add_filter( 'woocommerce_payment_gateway_supports', array( $this, 'paypal_ec_supports' ), 10, 3 );
+					add_filter( 'option_wc_gateway_ppce_prompt_to_connect', array( $this, 'paypal_ec_prompt_to_connect' ) );
+					add_filter( 'option_woocommerce_ppec_paypal_settings', array( $this, 'paypal_ec_settings' ) );
+				}
+			}
+		}
+
```
jeffstieler: Let's move the PPEC code to its own file.

woocommerce-services.php
```diff
@@ -528,6 +528,175 @@ public function init_core_wizard_payments_config() {
...
+					update_option( 'woocommerce_ppec_paypal_settings', $ppec_settings );
+				}
+
+				$username_key = 'sandbox' === $ppec_settings[ 'environment' ] ? 'sandbox_api_username' : 'api_username';
+				if ( empty( $ppec_settings[ $username_key ] ) ) {
+					// Reroute requests from the PPEC extension via WCS to pick up API credentials
+					add_filter( 'woocommerce_paypal_express_checkout_request_endpoint', array( $this, 'paypal_ec_endpoint' ) );
+
+					add_filter( 'woocommerce_payment_gateway_supports', array( $this, 'paypal_ec_supports' ), 10, 3 );
+					add_filter( 'option_wc_gateway_ppce_prompt_to_connect', array( $this, 'paypal_ec_prompt_to_connect' ) );
```
jeffstieler: I suspect this will change given our discussion on using a new custom banner, but:

Using `pre_option_{option}` here would be better since you don't need the existing option value.

woocommerce-services.php
```diff
@@ -528,6 +528,175 @@ public function init_core_wizard_payments_config() {
...
+
+		public function paypal_ec_settings( $settings ) {
+			$settings['paymentaction'] = 'sale';
+
+			if ( empty( $settings['api_username'] ) ) {
+				$settings['api_username'] = '';
+			}
+			if ( empty( $settings['sandbox_api_username'] ) ) {
+				$settings['sandbox_api_username'] = '';
+			}
```
jeffstieler: Can you explain these `empty()` checks and empty strings?

woocommerce-services.php
```diff
@@ -528,6 +528,175 @@ public function init_core_wizard_payments_config() {
...
+		}
+
+		/**
+		 * Once a payment is received, show prompt to connect a PayPal account on certain screens
+		 */
+		public function paypal_ec_prompt_to_connect() {
+			function uses_ppec_payment_method( $order ) {
+				return $order->get_payment_method() === 'ppec_paypal';  // TODO WC <3.0 compatibility
+			}
+			if ( ! count( array_filter( wc_get_orders( array() ), 'uses_ppec_payment_method' ) ) ) {
```
jeffstieler: This is an expensive check to run on every `admin_notices` action ([#](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-plugin.php#L175)).

I know this logic is likely to change significantly - since we want to display the banner after an order comes in, perhaps it makes more sense to hook into order-based actions instead of querying all orders.

woocommerce-services.php
dechov: That was working from me but I shied away after spotting this in [documentation](https://developer.wordpress.org/reference/hooks/pre_option_option/):
> Passing a truthy value to the filter will short-circuit retrieving the option value, returning the passed value instead.

In this case, I want a falsy value in the case of suppressing the notice. Is `pre_option_{option}` safe to use?

woocommerce-services.php
dechov: The PPEC plugin shows the "prompt to connect" if the option set to `''` but not if it's unset, so I was trying to fill in the default here without having to save. I've since realized we need to do this for other options too to make it work out of the box, so I'm looking for a different way to just initialize the settings to their defaults.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -20,6 +20,11 @@ public function __construct( WC_Connect_Nux $nux ) {
 		public function init() {
 			if ( function_exists( 'wc_gateway_ppec' ) ) {
 				add_action( 'wp_loaded', array( $this, 'setup' ) );
+
+				if ( isset( $_GET['api_username'] ) ) {
```
jeffstieler: Is the intention here to schedule this order updating event when the API username is saved?

Does the `api_username` come in `GET` from the OAuth connection redirect?

classes/class-wc-connect-paypal-ec.php
```diff
@@ -221,5 +226,30 @@ public function maybe_set_reroute_requests() {
 
 			wp_safe_redirect( wc_gateway_ppec()->get_admin_setting_link() );
 		}
-    }
+
+		public function update_order_status() {
+			$orders = wc_get_orders( array( 'status' => 'on-hold' ) );
```
jeffstieler: What happens if there are more "on hold" orders than are included in one call to `wc_get_orders()` with default params?

classes/class-wc-connect-paypal-ec.php
```diff
@@ -221,5 +226,30 @@ public function maybe_set_reroute_requests() {
...
+			error_log( count( $orders ) );
```
jeffstieler: Oops. 😛

classes/class-wc-connect-paypal-ec.php
dechov: Yes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-ips-handler.php#L163 and I also intended this to happen when saving `api_username` manually, though I haven't tested that yet.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -221,5 +226,30 @@ public function maybe_set_reroute_requests() {
...
+
+			foreach ( $orders as $order ) {
+				$old_wc         = version_compare( WC_VERSION, '3.0', '<' );
+				$order_id       = $old_wc ? $order->id : $order->get_id();
+				$payment_method = $old_wc ? $order->payment_method : $order->get_payment_method();
```
jeffstieler: We have a compatibility layer to encapsulate WC version differences here: https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-compatibility.php. Here's an example use: https://github.com/Automattic/woocommerce-services/blob/212442410e1e9fd493c28473346efcab2cf2ddb0/classes/class-wc-connect-shipping-label.php#L52

classes/class-wc-connect-paypal-ec.php
```diff
@@ -221,5 +226,30 @@ public function maybe_set_reroute_requests() {
...
+
+				if ( 'ppec_paypal' === $payment_method ) {
```
jeffstieler: You should be able to pass `'payment_method' => 'ppec_paypal'` to `wc_get_orders()`.

classes/class-wc-connect-paypal-ec.php
dechov: A bug 👎. `'limit' => -1` should help

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,250 @@
...
+
+				if ( empty( $username ) && ! empty( $subject ) ) {
+					// Reroute Express Checkout requests from the PPEC extension via WCS to pick up API credentials
+					add_filter( 'woocommerce_paypal_express_checkout_request_body', array( $this, 'request_body' ) );
+
+					add_filter( 'option_woocommerce_ppec_paypal_settings', array( $this, 'paypal_ec_settings' ) );
+					add_filter( 'woocommerce_payment_gateway_supports', array( $this, 'supports' ), 10, 3 );
+
+					add_filter( 'pre_option_wc_gateway_ppce_prompt_to_connect', '__return_empty_string' );
+					if ( 'sandbox' !== $settings->environment ) {
```
jeffstieler: Why not test the positive case here?

`if ( 'live' === $settings->environment ) {`

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,250 @@
...
+				}
+			}
+		}
+
+		public function request_body( $body ) {
+			$methods_to_proxy = array( 'SetExpressCheckout', 'GetExpressCheckoutDetails', 'DoExpressCheckoutPayment' );
+			if ( in_array( $body['METHOD'], $methods_to_proxy ) ) {
+				add_filter( 'woocommerce_paypal_express_checkout_request_endpoint', array( $this, 'request_endpoint' ), 10, 2 );
+			} else {
+				remove_filter( 'woocommerce_paypal_express_checkout_request_endpoint', array( $this, 'request_endpoint' ), 10, 2 );
```
jeffstieler: An alternative to removing the filter here is to do it in the `request_endpoint()` method so it's only ever executed once per attachment.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,250 @@
...
+		 */
+		public function maybe_init_banner() {
+			if (
+				'yes' !== get_option( 'wc_connect_banner_ppec', null ) ||
+				'yes' === get_option( 'wc_connect_dismiss_banner_ppec', null )
+			) {
+				return;
+			}
+
+			$screen = get_current_screen();
```
jeffstieler: Now that this function is attached to `current_screen`, you can just access the screen object passed in.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,250 @@
...
+				'image_url'      => plugins_url( 'images/cashier.svg', dirname( __FILE__ ) ),
+				'should_show_jp' => false,
+				'dismiss_option' => 'ppec',
+			) );
+		}
+
+		public function initialize_settings() {
+			$settings = get_option( 'woocommerce_ppec_paypal_settings', array() );
+
+			if ( ! isset( $settings['button_size'] ) ) {
```
jeffstieler: Please add a comment explaining why this is a good setting to use for this purpose.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,250 @@
...
+				$gateway = new WC_Gateway_PPEC_With_PayPal();
+				$settings_meta = $gateway->form_fields;
+				foreach ( $settings_meta as $key => $setting_meta ) {
+					if ( ! isset( $settings[ $key ] ) && isset( $setting_meta['default'] ) ) {
+						$settings[ $key ] = $setting_meta['default'];
+					}
+				}
```
jeffstieler: Is this loop necessary? Could you use [`WC_Settings_API::process_admin_options()`](https://github.com/woocommerce/woocommerce/blob/de71c92fa0bf11de78e1a09b5a2fe1d53dfcf983/includes/abstracts/abstract-wc-settings-api.php#L174) instead?

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,250 @@
...
+				if ( ! isset( $settings['reroute_requests'] ) ) {
+					$settings['reroute_requests'] = 'no';
+				}
+				update_option( 'woocommerce_ppec_paypal_settings', $settings );
+				wc_gateway_ppec()->settings->load( true );
+			}
+		}
+
+		public function paypal_ec_settings( $settings ) {
```
jeffstieler: An explanation for this would be great.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,250 @@
...
+					unset( $settings_meta['sandbox_api_username'], $settings_meta['sandbox_api_password'], $settings_meta['sandbox_api_signature'], $settings_meta['sandbox_api_certificate'] );
+				}
+
+				$api_subject_title = __( 'Payment Email', 'woocommerce-services' );
+				$api_subject_description = __( 'Enter your email address at which to accept payments. You\'ll need to link your own account in order to perform anything other than "sale" transactions.', 'woocommerce-services' );
+				$api_subject_placeholder = __( 'Required', 'woocommerce-services' );
+
+				$settings_meta['api_subject']['title'] = $settings_meta['sandbox_api_subject']['title'] = $api_subject_title;
+				$settings_meta['api_subject']['description'] = $settings_meta['sandbox_api_subject']['description'] = $api_subject_description;
+				$settings_meta['api_subject']['placeholder'] = $settings_meta['sandbox_api_subject']['placeholder'] = $api_subject_placeholder;
```
jeffstieler: This is a little hard to read, please don't have multiple assignments per line.

classes/class-wc-connect-paypal-ec.php
dechov: Ah, hopefully — will look into it.

classes/class-wc-connect-paypal-ec.php
dechov: 👍

classes/class-wc-connect-paypal-ec.php
dechov: Actually, we can't rely on this (and in fact it doesn't work) because the endpoint is retrieved multiple times per request: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-client.php#L131-L133

Open to other suggestions for improving this though.

classes/class-wc-connect-nux.php
```diff
@@ -635,6 +657,21 @@ class="woocommerce-services__connect-jetpack wcs-nux__notice-content-button butt
 				<?php endif; ?>
 			</div>
 			<?php
+			if ( isset( $content['dismiss_option'] ) ) :
+			?>
+				<script>
+				( function( $ ) {
+					$( '.wcs-nux__notice' ).on( 'click', '.notice-dismiss', function() {
```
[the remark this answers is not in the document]
jeffstieler: Let's figure out where the `.notice-dismiss` functionality (where the notice is actually dismissed) lives since we're relying on it.

classes/class-wc-connect-nux.php
```diff
@@ -448,6 +458,18 @@ public function set_up_nux_notices() {
...
+			add_action( 'wp_ajax_wc_connect_dismiss_banner', array( $this, 'ajax_dismiss_banner' ) );
+		}
+
+		public function ajax_dismiss_banner() {
+			if ( empty( $_POST['dismiss_option'] ) ) {
+				return;
+			}
+
+			check_ajax_referer( 'wc_connect_dismiss_banner', 'nonce' );
+			update_option( 'wc_connect_dismiss_banner_' . $_POST['dismiss_option'], 'yes' );
```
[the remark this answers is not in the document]
jeffstieler: > I thought it would be helpful to have a generalized way to make any banner dismissible

Good thinking. How about in `show_nux_banner()` we check for the option that dismisses the banner if it has `dismiss_option` defined. Let's put those options in the compact options using `WC_Connect_Options`.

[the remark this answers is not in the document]
dechov: Checking for the option in `show_nux_banner` now, but I've since realized that there already appears to be code for this, e.g. https://github.com/Automattic/woocommerce-services/blob/960c12b97878a3344f43e38a293c61417c838ca6/classes/class-wc-connect-nux.php#L40-L60
...though it's apparently only used for service settings notices. Later, I can look into using it rather than duplicating.

dechov:
It was broken when using the remote server at https://api-staging.woocommerce.com due to:
- missing auth header (fixed in https://github.com/Automattic/woocommerce-services/pull/1254/commits/fa4fe3438948dc4b3b5f4034ab8f99fe59a5bf49)
- cURL certificate configuration (PR: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/364)

dechov:
Narrowed the set of screens where the banner appears down to: orders screen, order screen if PPEC payment method was used, WC checkout settings, and WC addons payment gateway screen.

Rebased on master.

dechov:
New banner text:

screen shot 2018-01-11 at 11 55 12 pm

classes/class-wc-connect-paypal-ec.php
```diff
@@ -128,19 +128,16 @@ public function maybe_init_banner( $screen ) {
 				|| ( // Edit order page.
 					'shop_order' === $screen->post_type
 					&& 'post' === $screen->base
+					&& 'ppec_paypal' === WC_Connect_Compatibility::instance()->get_payment_method( new WC_Order( $_GET['post'] ) )
```
jeffstieler: You can use `wc_get_order()` here with no parameters to get the current order being edited.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -68,6 +69,40 @@ public function init() {
...
+		public function register_refund_pointer( $pointers ) {
+			$target = '.refund-actions > button:first-child';
+			$link_url = wc_gateway_ppec()->ips->get_signup_url( wc_gateway_ppec()->settings->environment );
+			$pointers[] = array(
+				'id' => 'wc_services_refund_via_ppec',
+				'target' => $target,
+				'options' => array(
+					'content' => sprintf( '<h3>%s</h3><p>%s</p>',
+						__( 'Link a PayPal account' ,'woocommerce-services' ),
+						sprintf( __( "To issue refunds via PayPal Express Checkout, you will need to <a href=\"%s\">link a PayPal account</a> with the email address that received this payment.", 'woocommerce-services' ), $link_url )
```
jeffstieler: Use single quotes for this string, and pass the result through `wp_kses()` for safety. Example: https://github.com/Automattic/woocommerce-services/blob/367a47accc80b22d28e87df1cd87df7c106a236a/classes/class-wc-connect-help-view.php#L266-L272

classes/class-wc-connect-paypal-ec.php
```diff
@@ -68,6 +69,40 @@ public function init() {
...
+					'position' => array( 'edge' => 'bottom', 'align' => 'top' ),
+				),
+			);
+
+			wp_add_inline_script( 'wc_services_admin_pointers', sprintf( "
+				jQuery( document ).ready( function( $ ) {
+					$( '#woocommerce-order-items' )
+						.one( 'click', 'button.refund-items', function() {
+							setTimeout( function() {
+								$( 'div.wc-order-refund-items' ).promise().then( function() {
```
jeffstieler: Why is a promise needed here?

classes/class-wc-connect-paypal-ec.php
```diff
@@ -67,18 +73,20 @@ public function init() {
...
 		}
 
-		/**
-		 * Get WCS PayPal proxy endpoint
-		 */
-		public function request_endpoint( $endpoint, $environment ) {
-			return trailingslashit( WOOCOMMERCE_CONNECT_SERVER_URL ) . 'paypal/nvp/' . $environment;
+		public function proxy_request( $preempt, $r, $url ) {
+			if ( ! preg_match( '/paypal.com\/nvp$/', $url ) ) {
+				return false;
```
jeffstieler: If we aren't going to proxy, we should pass through the value of `$preempt`.

dechov: Oh, right!

classes/class-wc-connect-paypal-ec.php
dechov: The intent is to wait for the `slideDown` animation to finish, at which point the promise ([defaulting to `fx` type](https://api.jquery.com/promise/#promise-type-target)) will resolve, and the pointer will show in the right place. It still seems to show up with a slightly incorrect vertical position on occasion — not sure why.

classes/class-wc-connect-nux.php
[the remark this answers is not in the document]
dechov: 👍 It gets appended here: https://core.trac.wordpress.org/browser/trunk/src/wp-admin/js/common.js#L752

classes/class-wc-connect-nux.php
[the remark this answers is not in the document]
dechov: Updated to use those existing functions for setting and retrieving dismissal state.

classes/class-wc-connect-paypal-ec.php
dechov: I believe it does what we need, but I seem to have run into trouble constructing a new payment gateway instance at this point in order to run that method — I can look further into how to make that work another time.

dechov:
Cleaned up the code — ready for further or final reviews.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -71,14 +71,19 @@ public function init() {
...
 			$pointers[] = array(
 				'id' => 'wc_services_refund_via_ppec',
 				'target' => $target,
 				'options' => array(
 					'content' => sprintf( '<h3>%s</h3><p>%s</p>',
 						__( 'Link a PayPal account' ,'woocommerce-services' ),
-						sprintf( __( "To issue refunds via PayPal Express Checkout, you will need to <a href=\"%s\">link a PayPal account</a> with the email address that received this payment.", 'woocommerce-services' ), $link_url )
+						sprintf(
+							wp_kses(
+								__( 'To issue refunds via PayPal Express Checkout, you will need to <a href=\"%s\">link a PayPal account</a> with the email address that received this payment.', 'woocommerce-services' ),
```
jeffstieler: You don't need to escape the double quotes now that the string is single quoted.

client/admin-pointers.js
```diff
@@ -37,11 +35,35 @@ jQuery( document ).ready( function( $ ) {
 			},
 		} );
 
-		target.pointer( options ).pointer( 'open' );
-		if ( pointer.dim ) {
-			target.css( 'z-index', 9999 );
-			$( 'body' ).append( '<div id="wcs-pointer-page-dimmer" class="wcs-pointer-page-dimmer"></div>' );
-			$( '#wcs-pointer-page-dimmer' ).fadeIn( 500 );
+		function open() {
+			const target = $( pointer.target );
```
jeffstieler: Why did this need to move?

If it were to return to the outer scope, would the later `$( pointer.target )` call be necessary?

client/admin-pointers.js
```diff
@@ -37,11 +35,35 @@ jQuery( document ).ready( function( $ ) {
...
+			target.pointer( options ).pointer( 'open' );
+			if ( pointer.dim ) {
+				target.css( 'z-index', 9999 );
+				$( 'body' ).append( '<div id="wcs-pointer-page-dimmer" class="wcs-pointer-page-dimmer"></div>' );
+				$( '#wcs-pointer-page-dimmer' ).fadeIn( 500 );
+			}
+		}
+
+		// Selectors for delayed opening. Only show_button is required.
```
jeffstieler: Can you expand this comment to describe the uses of all of the `delayed_opening` parameters? _(e.g. static vs. dynamic containers)_

classes/class-wc-connect-paypal-ec.php
```diff
@@ -21,9 +21,16 @@ class WC_Connect_PayPal_EC {
...
+		/**
+		 * Express Checkout API methods to proxy.
+		 */
+		private $methods_to_proxy;
+
 		public function __construct( WC_Connect_API_Client $api_client, WC_Connect_Nux $nux ) {
 			$this->api_client = $api_client;
 			$this->nux = $nux;
+
+			$this->methods_to_proxy = array( 'SetExpressCheckout', 'GetExpressCheckoutDetails', 'DoExpressCheckoutPayment' );
```
jeffstieler: FYI: you can just define this on line 27.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,320 @@
...
+					'animating_container' => '.wc-order-refund-items',
+					'delegation_container' => '#woocommerce-order-items',
+				),
+			);
+
+			return $pointers;
+		}
+
+		/**
+		 * If PPEC order comes in, activate prompt to connect a PayPal account
```
jeffstieler: Consider moving this comment to where the action is attached, and rewriting this to be more specific to the body of the function here.

_E.g. this function isn't tied to an order coming in_

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,320 @@
...
+			$order          = wc_get_order( $order_id );
+			$payment_method = WC_Connect_Compatibility::instance()->get_payment_method( $order );
+
+			if ( 'ppec_paypal' === $payment_method ) {
+				WC_Connect_Options::update_option( 'banner_ppec', 'yes' );
+			}
+		}
+
+		/**
+		 * Once a payment is received, show prompt to connect a PayPal account on certain screens
```
jeffstieler: Consider moving this comment to where the action is attached, and rewriting this to be more specific to the body of the function here.

_E.g. this function isn't tied to payment receipt_

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,320 @@
...
+			return $settings;
+		}
+
+		/**
+		 * Modify PPEC settings form
+		 */
+		public function adjust_form_fields( $form_fields ) {
+			$settings = wc_gateway_ppec()->settings;
+
+			if ( 'yes' === $settings->reroute_requests ) {
```
jeffstieler: Please comment why `reroute_requests` is used here.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,320 @@
...
+
+				$username = $settings->get_active_api_credentials()->get_username();
+				$subject  = $settings->get_active_api_credentials()->get_subject();
+
+				if ( empty( $username ) && ! empty( $subject ) ) {
+					add_filter( 'woocommerce_paypal_express_checkout_request_body', array( $this, 'request_body' ) );
+
+					add_filter( 'option_woocommerce_ppec_paypal_settings', array( $this, 'adjust_settings' ) );
+					add_filter( 'woocommerce_payment_gateway_supports', array( $this, 'ppec_supports' ), 10, 3 );
+					add_filter( 'wc_services_pointer_post.php', array( $this, 'register_refund_pointer' ) );
```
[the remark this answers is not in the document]
dechov: I don't think so — it serves the same purpose and works the same for `sandbox` environment, as far as I can tell. On the other hand, if you dismiss it in `sandbox`, you won't see it again in `live` where it really matters, which is possibly not ideal. What do you think?

classes/class-wc-connect-nux.php
```diff
@@ -364,17 +374,27 @@ public function get_feature_list_for_country( $country ) {
...
 			$supports_labels = ( 'US' === $country );
 
-			if ( $supports_stripe && $supports_taxes && $supports_rates && $supports_labels ) {
+			$is_stripe_active = is_plugin_active( 'woocommerce-gateway-stripe/woocommerce-gateway-stripe.php' );
+			$stripe_settings  = get_option( 'woocommerce_stripe_settings', array() );
+			$is_stripe_ready  = $is_stripe_active && isset( $stripe_settings['enabled'] ) && 'yes' === $stripe_settings['enabled'];
+
+			$is_ppec_active = is_plugin_active( 'woocommerce-gateway-paypal-express-checkout/woocommerce-gateway-paypal-express-checkout.php' );
+			$ppec_settings  = get_option( 'woocommerce_ppec_paypal_settings', array() );
+			$is_ppec_ready  = $is_ppec_active && isset( $ppec_settings['enabled'] ) && 'yes' === $ppec_settings['enabled'];
```
dechov: @jeffstieler I noticed from your test today that the WCS banner is missing a mention of payments in the case that `'enabled'` isn't set, even though WCS itself will make sure it is initialized to `'yes'`. Will modify this to treat the unset case as if it were enabled.

classes/class-wc-connect-paypal-ec.php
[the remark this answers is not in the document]
jeffstieler: > On the other hand, if you dismiss it in sandbox, you won't see it again in live where it really matters, which is possibly not ideal. What do you think?

This is what I'm trying to protect against - though it is an unlikely scenario. I think the pointer has more value in `live` mode.

classes/class-wc-connect-paypal-ec.php
```diff
@@ -0,0 +1,320 @@
...
+
+					if ( 'live' === $settings->environment ) {
+						add_action( 'woocommerce_order_status_on-hold', array( $this, 'maybe_set_banner' ) );
+						add_action( 'woocommerce_payment_complete', array( $this, 'maybe_set_banner' ) );
+						add_action( 'admin_enqueue_scripts', array( $this, 'maybe_init_banner' ) );
```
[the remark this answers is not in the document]
dechov: Sounds good, though I think maybe_set_banner => maybe_trigger_banner is even clearer.

PR #1262, opened 2017-12-12
Updated NodeJS to 8.9.3

DanReyLop:
Node v8 is now the LTS version, and Calypso has started using it a few weeks ago.

This also changes `npm` to version 5, which is much faster and reliable. `npm v5` comes bundled with `nodejs v8`, but if for some reason it isn't working for you, just type `npm i -g npm@5` to install it.

This PR completely regenerates the `shrinkwrap` file, so it's not risk-free. I didn't find anything obviously broken though.


package.json
```diff
@@ -10,7 +10,8 @@
     "release": "node tasks/release.js",
     "i18n": "cross-env NODE_ENV=i18n webpack && node tasks/i18n",
     "i18njson": "node tasks/i18njson",
-    "update-deps": "rm npm-shrinkwrap.json && rm -rf node_modules && npm install --no-optional && npm install --no-optional && npm prune && npm shrinkwrap --dev"
+    "postshrinkwrap": "node node_modules/wp-calypso/bin/trim-shrinkwrap.js",
+    "update-deps": "rm npm-shrinkwrap.json && rm -rf node_modules && npm install && npm shrinkwrap"
```
marcinbot: Actually, I noticed that running `update-deps` hasn't updated the shrinkwrap file. Looks like the `npm prune` step is still necessary.

Other than that, this is still ok.

DanReyLop: > Would also be good if the postshrinkwrap step was a part of the update-deps script

It runs automatically after any operation that updates the `shrinkwrap` file (such as running `npm shrinkwrap` or installing a package).

marcinbot: Discussed in Slack, looks like there was an issue with my `npm` installation (was using v3, and had to run `npm i -g npm@5`)

PR #1264, opened 2017-12-13
Added headers to indicate support for WooCommerce up to 3.3

DanReyLop:
See the new headers explained here: https://woocommerce.wordpress.com/2017/08/28/new-version-check-in-woocommerce-3-2/

WooCommerce `3.3` isn't out yet, but I've been using WooCommerce `master` since a few months ago and I haven't noticed any breakage. Should we declare support for `3.3`, or do you guys think we should wait until `3.3` is closer to be released?

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -9,7 +9,7 @@
  * Domain Path: /i18n/languages/
  * Version: 1.9.1
  * WC requires at least: 2.6.0
- * WC tested up to: 3.3.0
+ * WC tested up to: 3.2.6
  *
  * Copyright (c) 2017 Automattic
  *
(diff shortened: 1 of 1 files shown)


michaeltieso:
I was surprised to see required at least 2.6. Has this been tested on WooCommerce 2.6? IMO we should encourage 3.0 usage as much as possible.

DanReyLop:
This plugin has been in development since before the `3.0` days, and we've maintained backwards compatibility with `2.6`. Since many merchants were probably reluctant to update to 3.0, 2.6 seemed a reasonable target.

Maybe it's time to revisit that decision? Drop compatibility completely with `2.6`? That would allow us to delete a bunch of "compatibility" code. cc/ @jeffstieler

jeffstieler:
@DanReyLop let's note tested up to `3.2.6`, and we'll update to `3.3` with the release of `3.3`

DanReyLop:
> let's note tested up to 3.2.6, and we'll update to 3.3 with the release of 3.3

Done.

PR #1268, opened 2017-12-19
Updated the Calypso imports

marcinbot:
Ran `npm update-deps` and added more stubs for missing files.

When building you might see the following warning:
```
WARNING in ./~/wp-calypso/client/lib/wp/offline-library/index.js
Module not found: Error: Can't resolve '../../../../cached-requests.json' ...
```

As far as I know this doesn't break anything, but I fixed it in https://github.com/Automattic/wp-calypso/pull/20948

To test:
* all JS areas should still work


jeffstieler:
Using node 8.9.1 and npm 5.6.0 I'm getting this error in browser: `TypeError: state.plugins is undefined`

(on edit order screen)

client/calypso-stubs/state/selectors/is-rtl.js
```diff
@@ -0,0 +1 @@
+export default () => false;
```
DanReyLop: If we're going to simulate functionality, let's do it right :)

```js
export default () => document.body.classList.contains( 'rtl' );
```

client/apps/account-settings/view-wrapper.js
```diff
@@ -36,7 +37,7 @@ const LabelSettingsWrapper = ( props ) => {
 	return (
 		<div>
 			<GlobalNotices id="notices" notices={ notices.list } />
-			<LabelSettings />
+			<LabelSettings onChange={ noop } />
```
jeffstieler: Rather than use a noop, can we use the `ProtectFormGuard` component?

client/apps/packages/view-wrapper.js
```diff
@@ -34,7 +35,7 @@ const PackagesWrapper = ( props ) => {
 	return (
 		<div>
 			<GlobalNotices notices={ notices.list } />
-			<Packages />
+			<Packages onChange={ noop } />
```
jeffstieler: Same question about `ProtectFormGuard`.

PR #1271, opened 2017-12-29
Fix missing 'plugins' state now required by wp-calypso

dechov:
Provides the `state.plugins` Redux sub-tree, now required by `shipping-label` state imported from Calypso.

Alternatively, we can look into where it's being used and whether we can avoid this call altogether.


jeffstieler:
I'm not seeing any console errors on `master` related to this when opening the labels modal - is that where it was manifesting @dechov ?

dechov:
It was breaking the edit order screen on load when using the updated wp-calypso. Not able to reproduce that issue. It certainly works now — closing.

PR #1283, opened 2018-01-19
Improved the logic for NUX banners on multisites with Jetpack installed

### LINKED ISSUE #1282: Multisite users can't accept the ToS
Even when Jetpack is installed, activated and connected, the multisite admin cannot accept the ToS and enable the plugin features.

To reproduce:
* Create a multisite network
* Create a new site in the network with a new user as an admin
* Make sure Jetpack, WooCommerce and WCS are installed and activated on the new site
* The new admin will be able to connect their site to Jetpack, but they won't see any NUX banners and thus won't be able to accept the ToS

marcinbot:
Fixes #1282 

To test:
* create a multisite
* add a new site with a new user as admin
* as a superadmin install and activate WooCommerce, Jetpack and WCS (do not connect Jetpack)
* view the site as admin and verify that the NUX banner is visible and can be connected
* the rest of the plugin should work for the admin

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -400,14 +400,16 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 		}
 
 		public function set_up_nux_notices() {
-			if ( ! current_user_can( 'manage_woocommerce' )
-				|| ! current_user_can( 'install_plugins' )
-				|| ! current_user_can( 'activate_plugins' )
-			) {
+			if ( ! current_user_can( 'manage_woocommerce' ) ) {
 				return;
 			}
 
 			$jetpack_install_status = $this->get_jetpack_install_status();
+			if ( ( self::JETPACK_NOT_INSTALLED === $jetpack_install_status && ! current_user_can( 'install_plugins' ) )
+				|| ( self::JETPACK_INSTALLED_NOT_ACTIVATED === $jetpack_install_status && ! current_user_can( 'install_plugins' ) ) ) {
+				return;
+			}
+
 			$banner_to_display = self::get_banner_type_to_display( array(
 				'jetpack_connection_status'       => $jetpack_install_status,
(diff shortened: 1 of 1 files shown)


classes/class-wc-connect-nux.php
```diff
@@ -400,14 +400,16 @@ public function get_jetpack_redirect_url() {
...
-				|| ! current_user_can( 'install_plugins' )
-				|| ! current_user_can( 'activate_plugins' )
-			) {
+			if ( ! current_user_can( 'manage_woocommerce' ) ) {
 				return;
 			}
 
 			$jetpack_install_status = $this->get_jetpack_install_status();
+			if ( ( self::JETPACK_NOT_INSTALLED === $jetpack_install_status && ! current_user_can( 'install_plugins' ) )
+				|| ( self::JETPACK_INSTALLED_NOT_ACTIVATED === $jetpack_install_status && ! current_user_can( 'install_plugins' ) ) ) {
```
jeffstieler: Should this be checking `activate_plugins` instead of `install_plugins`?

classes/class-wc-connect-nux.php
```diff
@@ -400,14 +400,16 @@ public function get_jetpack_redirect_url() {
 		}
 
 		public function set_up_nux_notices() {
-			if ( ! current_user_can( 'manage_woocommerce' )
-				|| ! current_user_can( 'install_plugins' )
-				|| ! current_user_can( 'activate_plugins' )
-			) {
```
jeffstieler: Let's add a comment here or below explaining why we're checking what we're checking and how it relates to both single site and multi site installations.

classes/class-wc-connect-nux.php
marcinbot: Good catch, fixed in 63c6d55

classes/class-wc-connect-nux.php
marcinbot: Added in 704eba0

PR #1288, opened 2018-01-24
Admin notice calling out missing product weight and/or dimensions

dechov:
Shows a notice when a product is saved with missing weight or dimensions.

screen shot 2018-01-24 at 11 49 17 am

TODO:
- [ ] check to see if there's actually a live rates shipping method enabled
- [ ] also hook into `wc_connect_shipping_zone_method_added`, `wc_connect_shipping_zone_method_status_toggled`, and `wc_connect_shipping_zone_method_deleted`

This can later be modified to:
- be dismissible
- list the products that lack details
- mention that flat rates will still work as long as they don't exceed the maximum (if only weight is missing, and if we support that in the future)

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -657,12 +657,39 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			add_action( 'admin_init', array( $this, 'load_admin_dependencies' ) );
 			add_filter( 'wc_connect_shipping_service_settings', array( $this, 'shipping_service_settings' ), 10, 3 );
 			add_action( 'woocommerce_email_after_order_table', array( $this, 'add_tracking_info_to_emails' ), 10, 3 );
+			add_action( 'save_post', array( $this, 'maybe_trigger_product_notice' ), 10, 2 );
 
 			$tracks = $this->get_tracks();
 			$tracks->init();
 
 			$this->taxjar->init();
 			$this->paypal_ec->init();
+		}
+
+		public function maybe_trigger_product_notice( $post_id, $post ) {
+			if (
+				'publish' !== $post->post_status ||
+				'product' !== $post->post_type ||
+				! ( $product = wc_get_product( $post ) )
+			) {
+				return;
+			}
+
+			if ( $product->get_weight() && $product->get_length() && $product->get_width() && $product->get_height() ) {
+				// If details are all present, check all other products to see if notice should be disabled
+				$all_products_ready = true;
+				foreach ( wc_get_products( array( 'limit' => -1 ) ) as $product ) {
+					if ( ! ( $product->get_weight() && $product->get_length() && $product->get_width() && $product->get_height() ) ) {
+						$all_products_ready = false;
+						break;
+					}
+				}
+				if ( $all_products_ready ) {
+					WC_Connect_Options::update_option( 'product_notice', false );
+				}
+			} else {
+				WC_Connect_Options::update_option( 'product_notice', true );
+			}
 		}
 
 		public function tos_rest_init() {
--- a/classes/class-wc-connect-error-notice.php
+++ b/classes/class-wc-connect-error-notice.php
@@ -43,6 +43,8 @@ if ( ! class_exists( 'WC_Connect_Error_Notice' ) ) {
 			if ( $this->notice_enabled() ) {
 				$this->show_notice();
 			}
+
+			$this->render_product_notice();
 		}
 
 		private function notice_enabled() {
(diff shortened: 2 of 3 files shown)


woocommerce-services.php
```diff
@@ -645,13 +645,40 @@ public function attach_hooks() {
...
 			$tracks = $this->get_tracks();
 			$tracks->init();
 
 			$this->taxjar->init();
 		}
 
+		public function maybe_trigger_product_notice( $post_id, $post ) {
+			if (
+				'publish' !== $post->post_status ||
+				'product' !== $post->post_type ||
```
jeffstieler: You can avoid this check by hooking into `save_product` instead of `save_post`. See: https://github.com/WordPress/WordPress/blob/master/wp-includes/post.php#L3724

woocommerce-services.php
```diff
@@ -645,13 +645,40 @@ public function attach_hooks() {
...
 
 			$tracks = $this->get_tracks();
 			$tracks->init();
 
 			$this->taxjar->init();
 		}
 
+		public function maybe_trigger_product_notice( $post_id, $post ) {
+			if (
+				'publish' !== $post->post_status ||
```
jeffstieler: Any reason to restrict this to published products? This could be helpful to know even if the product is still in draft.

woocommerce-services.php
```diff
@@ -645,13 +645,40 @@ public function attach_hooks() {
...
+				'product' !== $post->post_type ||
+				! ( $product = wc_get_product( $post ) )
+			) {
+				return;
+			}
+
+			if ( $product->get_weight() && $product->get_length() && $product->get_width() && $product->get_height() ) {
+				// If details are all present, check all other products to see if notice should be disabled
+				$all_products_ready = true;
+				foreach ( wc_get_products( array( 'limit' => -1 ) ) as $product ) {
```
jeffstieler: This feels like **a lot** of work to do on product save. `limit => -1` always raises a flag for me in `WP_Query`s. _(This should be limited to only non-virtual products, for instance)_

I think a store-wide look at product shipping readiness makes more sense as part of our Status page.

woocommerce-services.php
```diff
@@ -645,13 +645,40 @@ public function attach_hooks() {
...
+					if ( ! ( $product->get_weight() && $product->get_length() && $product->get_width() && $product->get_height() ) ) {
+						$all_products_ready = false;
+						break;
+					}
+				}
+				if ( $all_products_ready ) {
+					WC_Connect_Options::update_option( 'product_notice', false );
+				}
+			} else {
+				WC_Connect_Options::update_option( 'product_notice', true );
```
jeffstieler: What do you think about storing this as post meta so it only shows on products that need dimensions/weight?

dechov:
@jeffstieler The implementation review is definitely helpful — to respond to just the high-level direction of your comments:

I'm fine with only showing it on the individual product page that needs dimensions or weight —  in that case, is it even worth storing any state, or should we simplify it to just a check for dimensions and weight on each page load?

I think it might also be nice, upon saving the USPS shipping method, to point out that some products aren't compatible (and listing them — or not). But let me know if you think we should tackle that part of it later, and keep the scope of this PR to just the product page notice.

If we don't show a notice anywhere else though, I don't think there's much need for it to be dismissible.

jeffstieler:
> I'm fine with only showing it on the individual product page that needs dimensions or weight — in that case, is it even worth storing any state, or should we simplify it to just a check for dimensions and weight on each page load?

That simpler check sounds great.

> I think it might also be nice, upon saving the USPS shipping method, to point out that some products aren't compatible (and listing them — or not). But let me know if you think we should tackle that part of it later, and keep the scope of this PR to just the product page notice.

How much work would it be to add a "Products" (or similarly named) section to the Status page with this product info? Perhaps with a notice on the shipping method that links to it?

We should also experiment with setting fallback values for weight and dimensions - I believe our other USPS extension will default to 1 lb, but I'm not sure how it handles dimensions.

dechov:
Rebased on latest master on the off-chance that it would fix the remote build but it didn't help.

jeffstieler:
@dechov what do you think about considering this closed in https://github.com/Automattic/woocommerce-services/pull/1303 ?

screen shot 2018-02-12 at 4 34 32 pm

dechov:
@jeffstieler Hm, it's definitely less of an issue now, but while I think putting the information in the "customer" context is important, putting it in a proper "merchant" context is an opportunity as well. That is, unless it's totally obvious to anyone with live rates set up that they have to put in those measurements, I expect there's a chance to make the requirement clear earlier — at the point when they're creating / editing / focused on the product. If so, an admin notice may not be the best way to do that, but that can (and should) be a separate design discussion.

I imagine this is especially true for a new WooCommerce user, who is probably most likely to miss the requirement to begin with. It may take them a while to figure out that WCS debug mode is what will help them get rates working, and where to find it, so presenting the requirement upfront on the edit product page may be useful. What do you think?

v18:
> How much work would it be to add a "Products" (or similarly named) section to the Status page with this product info? Perhaps with a notice on the shipping method that links to it?

A drive-by idea I thought of while reading this: we could show the weight, height, and width as new columns right on the table on the products page for easy editing.

Surfacing this error is so awesome, btw!

v18:
@dechov - hey I'm looking at PRs to review. Looks like this PR is still in progress and does not need a review right now? Is that right? I removed the `Needs Review` label, feel free to re-add it once you'd like another review.

PR #1290, opened 2018-01-26
Stripe Connect: Clear keys on stripe deauthorization

### LINKED ISSUE #1289: Stripe Connect: Clear keys on deauthorize
When a merchant deauthorizes their connection to our Stripe Connect platform (i.e. via the POST connect/stripe/account/deauthorize endpoint), I think we should also clear the Publishable Keys and Secret Keys in their settings in wp-admin (we don't presently)

From what I can tell empirically and from reading Stripe's docs ( https://stripe.com/docs/connect/oauth-reference ) the keys provided when the user connects to our platform are revoked when the user deauthorizes the connection.

> PARAMETER: access_token, DESCRIPTION: The access token you can use to make requests on behalf of this Stripe account. Use it as you would any Stripe secret API key. This key does not expire, but may be revoked by the user at any time (you'll get a account.application.deauthorized webhook event when this happens).

allendav:
Fixes #1289 

To test:
* Install on a WordPress.com Automated Transfer (Store) site with the WooCommerce Stripe Gateway plugin also installed and active
* Use Stripe Connect to connect to a Stripe account (can use Calypso for this)
* Use Stripe Connect to deauthorize / disconnect that account
* Navigate to /wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe, and verify the keys (for production) have been cleared.

### DIFF
--- a/classes/class-wc-connect-stripe.php
+++ b/classes/class-wc-connect-stripe.php
@@ -24,7 +24,7 @@ if ( ! class_exists( 'WC_Connect_Stripe' ) ) {
 		private $logger;
 
 		const STATE_VAR_NAME = 'stripe_state';
-		const WOOCOMMERCE_CONNECT_STRIPE_OPTION = 'woocommerce_stripe_settings';
+		const SETTINGS_OPTION = 'woocommerce_stripe_settings';
 
 		public function __construct( WC_Connect_API_Client $client, WC_Connect_Options $options, WC_Connect_Logger $logger ) {
 			$this->api = $client;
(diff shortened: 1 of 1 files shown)


jeffstieler:
@allendav rebasing with `master` should fix the CI build.

classes/class-wc-connect-stripe.php
```diff
@@ -24,6 +24,7 @@ class WC_Connect_Stripe {
 		private $logger;
 
 		const STATE_VAR_NAME = 'stripe_state';
+		const WOOCOMMERCE_CONNECT_STRIPE_OPTION = 'woocommerce_stripe_settings';
```
marcinbot: It was a bit confusing at first, but this key is not a part of the WCS data, so the variable name probably shouldn't include the `CONNECT` bit.

jeffstieler: @allendav let's rename this constant and 🛳 it!

allendav: Fixed in https://github.com/Automattic/woocommerce-services/pull/1290/commits/56936c36bcc97bacd3c9086af1a55a96aee635dd

PR #1292, opened 2018-02-01
Fixed broken DB migration script

marcinbot:
Fixes the Hydra-related issue in p8yzl4-1fK-p2

An argument can be made to remove the migration script completely, and the only other place where `get_enabled_services_by_ids()` is used is already making a check for an empty array, but I think it's a good practice to do that check inside this function to avoid this kind of errors if someone decides to use it somewhere else.


jeffstieler:
A rebase should fix the build here.

PR #1294, opened 2018-02-02
Set the latest Calypso rev as the dependency, fixed image imports

marcinbot:
Sets the Calypso dependency to the commit https://github.com/Automattic/wp-calypso/commit/3c8a7c8b5b31a652bb7c531954cb566a47c01d89 (latest at the time of this PR) which should fix the tests/builds randomly failing after `npm install`

Also fixes the image imports broken by the recent changes in Calypso.

It should also fix the tests failing in #1292

### DIFF
--- a/assets/stylesheets/_colors.scss
+++ b/assets/stylesheets/_colors.scss
@@ -1,50 +1,9 @@
-// Blues
-$blue-wordpress:         #0087be;
-$blue-light:             #99d9ed;
-$blue-medium:            #00a0d2;
-$blue-dark:              #005082;
-
-// Grays
-$gray:                   #87a6bc;
-$gray-light:             lighten( $gray, 33% ); //#f3f6f8
-$gray-dark:              darken( $gray, 38% ); //#2e4453
-
-// $gray-text: ideal for standard, non placeholder text
-// $gray-text-min: minimum contrast needed for WCAG 2.0 AA on white background
-$gray-text:              $gray-dark;
-$gray-text-min:          darken( $gray, 18% ); //#537994
+@import "~wp-calypso/assets/stylesheets/shared/_colors";
 
 //Purples
 $purple:                 #96588a;
 $purple-dark:            #a36597;
 $purple-light:           #bb77ae;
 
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -159,12 +160,18 @@ const config = {
...
-			resource.request = resource.request
+			let newPath = resource.request
 				.replace(
 					/^.+calypso\/images/,
 					path.resolve( __dirname, 'node_modules', 'wp-calypso', 'public', 'images' )
 				)
 				.replace( /is-([^/]+)$/, '$1' );
+
+			if ( ! fs.existsSync( newPath ) ) {
+				newPath = newPath.replace( /cc-([^/]+)$/, '$1' );
```
jeffstieler: Can you comment on why this is needed? Looks like the credit card images?

marcinbot: There was a bug in Calypso that has been fixed in https://github.com/Automattic/wp-calypso/pull/22121 where the Calypso CSS referenced non-existent payment method images.

assets/stylesheets/_colors.scss
```diff
@@ -1,50 +1,9 @@
...
-// Grays
-$gray:                   #87a6bc;
-$gray-light:             lighten( $gray, 33% ); //#f3f6f8
-$gray-dark:              darken( $gray, 38% ); //#2e4453
-
-// $gray-text: ideal for standard, non placeholder text
-// $gray-text-min: minimum contrast needed for WCAG 2.0 AA on white background
-$gray-text:              $gray-dark;
-$gray-text-min:          darken( $gray, 18% ); //#537994
+@import "~wp-calypso/assets/stylesheets/shared/_colors";
```
dechov: Can be shortened to `"shared/_colors"`, or is it better to be explicit?

assets/stylesheets/_colors.scss
```diff
@@ -1,50 +1,9 @@
...
 
 //Purples
```
dechov: The purples don't appear to be used — can we just remove them?

Issue #1115, opened 2017-08-07
Edit Order: automatically go back to order page after entering credit card


v18:
Scenario:
1. Go to edit order page for the first time
2. See that you can print a label after adding a credit card
3. Click the add credit card link
4. Select a credit card
5. 😕 <- what now?
Solution: after saving the credit card, we automatically redirect the user back to the edit order page

Related to: #1086

olaolusoga:
Thought I had commented on the related issue (#1086). We should add a link to the confirmation notification that appears in the top right when a users selects a credit card.

**Reason**: Redirecting back will most likely be jarring for users who land on the credit card step outside the edit order page path. Showing a text link that says "Create a label now" (or something along those lines) should keep it vague enough to nudge people who haven't seen the create a label widget, and those who came from there

dechov:
Is that still a concern if we redirect only in the case where you were referred from the edit order page? I do see how the redirect may be jarring anyway, and how the link may be preferable, but I'm not sure.

Also, is rendering the credit card portion of the shipping label settings in a modal right on the edit order page another option worth considering?

Separate but adjacent issue: as @msdesign21 [highlighted](https://heyhydra.wordpress.com/2017/08/10/printing-shipping-labels-with-the-dymo-label-writer-wcs/), there's no obvious next step after successfully adding a card in Calypso. Is there anything we can do about that, and is there any overlap with this issue? (Somewhat relevant discussion in https://github.com/Automattic/woocommerce-services/issues/1014: in Store on WP.com, it will show a modal for adding a card instead of redirecting to /me/billing)

dechov:
Another option is to put the card options right in the "create new label" view, which I think would make some sense from a "checkout flow" perspective. (If one is already configured, it'll just show a line of text confirming and offering to change, as worked out in https://github.com/Automattic/wp-calypso/issues/16570.) This could possibly show up in the bare right-side pane — past ideas about what else to do with that space (https://github.com/Automattic/woocommerce-services/issues/819) also focused on "checkout details", and I would think that the card info would be complementary. (cc: @kellychoffman)

Related items: a) the "Add another credit card" href should be tweaked as per https://github.com/Automattic/woocommerce-services/issues/1014#issuecomment-308284702, and b) it might help to have an obvious way to refresh the list — either a button or periodic polling — since as of now you have to know to refresh the shipping label settings page.

olaolusoga:
[...]

PR #1300, opened 2018-02-05
Add redirect link from label settings screen back to order

dechov:
Implements "Solution 1" referred to in https://github.com/Automattic/woocommerce-services/issues/1115#issuecomment-325632978.

If (and only if) the label settings screen was arrived at from an order screen (i.e. an edit post screen), and settings are saved with a payment method selected, a "Back to order" link appears in the success notification (which also persists in this case, instead of disappearing after 5 seconds).

screen shot 2018-02-05 at 2 41 09 pm

Can change "Back to order" text if there are better ideas — "return" instead of "back"? to "order page"? or "Order ##"?

### DIFF
--- a/classes/class-wc-connect-settings-pages.php
+++ b/classes/class-wc-connect-settings-pages.php
@@ -77,12 +77,9 @@ if ( ! class_exists( 'WC_Connect_Settings_Pages' ) ) {
 			}
 
 			$extra_args = array();
-			$referer_url = wp_get_referer();
-			$referer = parse_url( $referer_url );
-			if ( preg_match( '/\/wp-admin\/post\.php$/', $referer['path'] ) ) {
-				parse_str( $referer['query'], $query );
-				$extra_args['order_id'] = $query['post'];
-				$extra_args['order_href'] = $referer_url;
+			if ( isset( $_GET['from_order'] ) ) {
+				$extra_args['order_id'] = $_GET['from_order'];
+				$extra_args['order_href'] = get_edit_post_link( $_GET['from_order'] );
 			}
 
 			do_action( 'enqueue_wc_connect_script', 'wc-connect-account-settings', $extra_args );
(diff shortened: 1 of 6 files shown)


classes/class-wc-connect-settings-pages.php
```diff
@@ -76,6 +76,14 @@ public function output_account_screen() {
 				<?php
 			}
 
+			$referer_url = wp_get_referer();
+			$referer = parse_url( $referer_url );
+			if ( preg_match( '/\/wp-admin\/post\.php$/', $referer['path'] ) ) {
+				WC_Connect_Options::update_option( 'label_settings_redirect', $referer_url );
```
marcinbot: This has a potential of breaking if there are two or more admins looking at the settings page.

I think we can avoid using the DB entirely and pass an extra parameter to the `enqueue_wc_connect_script` action below. Then it would be passed to the exported function in `apps/account-settings/index.js` as an argument which can then be passed as prop to the `ViewWrapper`

dechov: Much better — thanks!

client/apps/account-settings/view-wrapper.js
```diff
@@ -33,8 +33,12 @@ class LabelSettingsWrapper extends Component {
 	}
 
 	onSaveSuccess = () => {
-		const { noticeActions, translate, orderHref } = this.props;
-		const options = orderHref ? { button: translate( 'Back to order' ), href: orderHref } : { duration: 5000 };
+		const { noticeActions, translate, orderHref, paymentMethodSelected } = this.props;
+		const options =
+			orderHref && paymentMethodSelected
+				? { button: translate( 'Back to order' ), href: orderHref }
```
jeffstieler: Is it possible to make this contain the order number? E.g. "Back to Order #38"?

dechov: Done — just parsed the post ID out of the URL. This could alternatively be done on the client to avoid passing around redundant data.

dechov:
@jeffstieler Sounds good — using query param `from_order` now.

By the way, here's how it looks now:

screen shot 2018-02-07 at 1 48 47 pm

Issue #1163, opened 2017-09-25
Include more details about the shipment in the Debug Log 


juliaamosova:
Currently, WC Services Debug Log contains some useful information about the shipment but often not enough to troubleshoot the issue when the rates on the site don't match USPS rates (the most common request received in support when it comes to using USPS). I don't have a lot of details in my current debug log but here is how it looks after placing an order and selecting one of the WC USPS shipping methods to ship the product:

![](https://cloudup.com/cfOkPXgUd-R+)
Image Link: https://cloudup.com/cfOkPXgUd-R

As you can see, there are no details of the products being quoted and which package had been used. I know that the log does show errors when the shipping can't be calculated but often times knowing what had been recorded by the plugin as far as product's weight and dimensions and what kind of package it had been packed into is the key in troubleshooting the rates issue. 

It will be great to include the following details to the debug log:

- Product's weight and dimensions
- The type of package the products have been packed to
- Origin and shipping zip code
- Rates returned by the plugin

All these details are available via USPS Shipping extension and extremely helpful in troubleshooting:

![](https://cloudup.com/cP7UqYvExPj+)
Image Link: https://cloudup.com/cP7UqYvExPj

**USPS REQUEST** looks as follows:

`API=RateV4&XML=<RateV4Request USERID="150WOOTH2143"><Revision>2</Revision><Package ID="0:1:12:10.25:5:2">       <Service>ALL</Service>      <ZipOrigination>10075</ZipOrigination>      <ZipDestination>33139</ZipDestination>      <Pounds>2</Pounds>  <Ounces>0.00</Ounces>       <Container />     <Size>REGULAR</Size>        <Width>10.25</Width>        <Length>12</Length> <Height>5</Height>  <Girth>31</Girth>   <Machinable>true</Machinable>       <ShipDate>26-Sep-2017</ShipDate></Package></RateV4Request>`

As you can see, product's weight and dimensions are there as well as the origin and destination zip codes. This allows support to confirm that the plugin is using correct product's information when requesting rates and use the same details on the USPS calculator to see if the rates are matching. 

**USPS RESPONSE** looks as follows:

![](https://cloudup.com/c13Tth3QWom+)
Image Link: https://cloudup.com/c13Tth3QWom

The response includes details of the rates being returned which also allows us to confirm whether or not the plugin is working correctly or if there is an issue with user's settings.

jeffstieler:
[...]

PR #1303, opened 2018-02-07
Output shipping debug messages on the Cart and Checkout pages.

jeffstieler:
Addresses #1163.

Optional PR: https://github.com/Automattic/woocommerce-services/pull/1304 - feature separated logging.

screen shot 2018-02-07 at 2 11 25 pm

This PR seeks to do the following:
* Separate the notion of "logging" and "debugging" based on front-end visibility
* Output USPS-extension-style debug messages in cart/checkout to help merchants troubleshoot their shipping setup

**Note: this can only output rates that are returned by the server - no packaging info will be displayed if no rates are found.**

screen shot 2018-02-06 at 9 57 37 am

Messages that can be displayed:
* Request error
* Response is missing `rates` property
* Received rate: <info>
* Skipping rate calculation - missing country
* Skipping rate calculation - invalid postcode
* Skipping rate calculation - invalid/unsupported state
* No rates found, adding fallback

_Note: if the fallback rate was added server side, the "Received rate" message will contain a note about it._

### DIFF
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -250,7 +250,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 				return;
 			}
 
-			$this->debug( 'No rates found, adding fallback' );
+			$this->debug( 'No rates found, adding fallback', 'success' );
 
 			$rate_to_add = array(
 				'id'        => self::format_rate_id( 'fallback', $this->id, 0 ),
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -393,7 +393,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 						sprintf(
 							'Received rate: %s (%s)<br/><ul><li>%s</li></ul>',
 							$rate_to_add['label'],
-							$rate->rate,
+							wc_price( $rate->rate ),
 							implode( '</li><li>', $package_names )
 						),
 						'success'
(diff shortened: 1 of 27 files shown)


jeffstieler:
cc: @joelwills @juliaamosova

juliaamosova:
Thank you @jeffstieler and the team!!

classes/class-wc-connect-shipping-method.php
```diff
@@ -258,6 +263,8 @@ private function add_fallback_rate( $service_settings ) {
 
 		public function calculate_shipping( $package = array() ) {
 
+			$this->debug( 'WooCommerce Services debug mode is on - to hide these messages, turn debug mode off in the settings.' );
```
dechov: Should we be more clear about what "the settings" is? (This could be a bit misleading since the merchant wouldn't actually find it under WooCommerce -> Settings.) Perhaps even a link?

classes/class-wc-connect-shipping-method.php
```diff
@@ -438,5 +470,19 @@ public static function format_rate_title( $rate_title ) {
...
 		}
 
+		/**
+		 * Log debug by printing it as notice.
+		 *
+		 * @param string $message Debug message.
+		 * @param string $type    Notice type.
+		 */
+		public function debug( $message, $type = 'notice' ) {
+			if ( is_cart() || is_checkout() ) {
```
dechov: We probably don't want to show anything when updating the cart from other pages, but what about upon "Update cart"? Not totally sure how we'd support it since this check seems to exclude it.

jeffstieler: I think https://github.com/Automattic/woocommerce-services/pull/1303/commits/4f07b2323da4a944f3ee6f1a81f18c6c686ccb2c should do it - only the cart update handler uses `update_cart` in the post body.

classes/class-wc-connect-shipping-method.php
jeffstieler: Added a link in https://github.com/Automattic/woocommerce-services/pull/1303/commits/3e12096187780664bd05d195279bd0c7db6e6d88.

PR #1304, opened 2018-02-07
Add feature separated logging

jeffstieler:
Separate log files by feature type - "taxes", "shipping", "payments", etc.

Display taxes, shipping, and generic logs on Status page.

screen shot 2018-02-07 at 4 48 01 pm

### DIFF
--- a/classes/class-wc-connect-help-view.php
+++ b/classes/class-wc-connect-help-view.php
@@ -466,7 +466,7 @@ if ( ! class_exists( 'WC_Connect_Help_View' ) ) {
 
 			$debug_items[] = $this->add_log_view( __( 'Taxes Log', 'woocommerce-services' ), 'taxes' );
 			$debug_items[] = $this->add_log_view( __( 'Shipping Log', 'woocommerce-services' ), 'shipping' );
-			$debug_items[] = $this->add_log_view( __( 'Debug Log', 'woocommerce-services' ) );
+			$debug_items[] = $this->add_log_view( __( 'Other Log', 'woocommerce-services' ) );
 
 			return $debug_items;
 		}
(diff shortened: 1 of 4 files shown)


classes/class-wc-connect-help-view.php
```diff
@@ -412,39 +464,9 @@ protected function get_debug_items() {
...
-				'key' => 'wcc_debug_log_tail',
-				'title' => __( 'Debug Log', 'woocommerce-services' ),
-				'type' => 'textarea',
-				'description' => $description,
-				'readonly' => true,
-				'value' => $log_tail
-			);
+			$debug_items[] = $this->add_log_view( __( 'Taxes Log', 'woocommerce-services' ), 'taxes' );
+			$debug_items[] = $this->add_log_view( __( 'Shipping Log', 'woocommerce-services' ), 'shipping' );
+			$debug_items[] = $this->add_log_view( __( 'Debug Log', 'woocommerce-services' ) );
```
dechov: This old title could be revised to disambiguate among the new distinctions — it's not clear that this view is independent/exclusive of a) the other logs, and b) the "Debug" option above. Not sure what to call it though — "Other log"? "Misc log"?

jeffstieler: Ah, good call. I'll change it to "Other".

jeffstieler:
@dechov - I've renamed "debug log" to "other log" - this good to 🚢 ?

dechov:
@jeffstieler LGTM, but a couple questions:

- would making the tax logging more concise be an alternative solution to the same problem, or independent?
- on a minor note, might we want Shipping first, since it's a more prominent feature?

jeffstieler:
> would making the tax logging more concise be an alternative solution to the same problem, or independent?

I'm hesitant to deviate too much from the TaxJar plugin codebase here. I think coupled with https://github.com/Automattic/woocommerce-services/pull/1308, we'll see a lot less "noise" in the Tax log.

> on a minor note, might we want Shipping first, since it's a more prominent feature?

That makes sense to me - I'll reorder before merging.

PR #1310, opened 2018-02-15
Taxes - Moved the total price calulation until after the taxes are added

### LINKED ISSUE #1309: Automated taxes - the tax row and total don't match when shipping is not taxable
To reproduce:
* have a store set in NC
* add a Flat Rate shipping method and set the tax status to none
* enable automated taxes
* checkout an item to an NC address
* the taxes will come back and be listed, but the cost of items+shipping+tax will be less than total

This is because shipping in NC is taxable, but the shipping method is marked as not taxable in the store. The returned tax should either respect that setting and not calculate the taxes for the method or make sure that the total is updated with taxes of both the items and shipping.

marcinbot:
Fixes #1309

`WC_Connect_TaxJar_Integration` was telling WC to calculate the totals using the old data, and then modifying some of the rows that came back as a result of that calculation but not updating the grand total. This in turn would lead to prices+taxes not adding up.

To test:
* have a store set in NC
* add a Flat Rate shipping method and set the tax status to none
* enable automated taxes
* checkout an item to an NC address
* the taxes row should not include the shipping tax and the total should be correct
* set the shipping method to taxable
* the taxes row on checkout should include the shipping tax and the total should be correct

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -268,12 +268,6 @@ class WC_Connect_TaxJar_Integration {
 			'line_items' => $line_items,
 		) );
 
-		if ( class_exists( 'WC_Cart_Totals' ) ) { // Woo 3.2+
-			do_action( 'woocommerce_cart_reset', $wc_cart_object, false );
-			do_action( 'woocommerce_before_calculate_totals', $wc_cart_object );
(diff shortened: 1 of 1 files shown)


jeffstieler:
Parking this re: https://github.com/taxjar/taxjar-woocommerce-plugin/pull/59

PR #1311, opened 2018-02-15
 Forward specific address normalization errors, as success response

dechov:
In the event of address normalization validation errors, passes specific address field validation errors back to the client. ~(Depends on server branch `add/address-normalization-field-errors`.)~

Also, changes this response from an error to a success response (to match the WCS server response): the data only gets passed back to the WP.com client in the event of a successful request.

### DIFF
--- a/classes/class-wc-rest-connect-address-normalization-controller.php
+++ b/classes/class-wc-rest-connect-address-normalization-controller.php
@@ -39,7 +39,7 @@ class WC_REST_Connect_Address_Normalization_Controller extends WC_REST_Connect_B
 			$this->logger->log( 'Address validation errors: ' . implode( '; ', array_values( (array) $response->field_errors ) ), __CLASS__ );
 			return array(
 				'success' => true,
-				'fieldErrors' => $response->field_errors,
+				'field_errors' => $response->field_errors,
 			);
 		}
 
--- a/classes/class-wc-rest-connect-address-normalization-controller.php
+++ b/classes/class-wc-rest-connect-address-normalization-controller.php
@@ -35,18 +35,13 @@ class WC_REST_Connect_Address_Normalization_Controller extends WC_REST_Connect_B
 			return $error;
 		}
 
-		if ( isset( $response->error ) ) {
-			$error = new WP_Error(
-				$response->error->code,
-				$response->error->message,
-				array(
-					'status' => 400,
-					'message' => $response->error->message,
-					'fieldErrors' => $response->fieldErrors,
-				)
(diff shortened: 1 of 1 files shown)


classes/class-wc-rest-connect-address-normalization-controller.php
```diff
@@ -35,14 +35,13 @@ public function post( $request ) {
...
-				$response->error->message,
-				array( 'message' => $response->error->message )
+		if ( isset( $response->fieldErrors ) ) {
+			$response = array(
+				'success' => true,
+				'fieldErrors' => $response->fieldErrors,
 			);
-			$this->logger->debug( $error, __CLASS__ );
-			return $error;
+			$this->logger->debug( $response, __CLASS__ );
```
jeffstieler: This causes a warning: `PHP Notice:  Array to string conversion in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/classes/class-wc-connect-logger.php on line 39`

classes/class-wc-rest-connect-address-normalization-controller.php
```diff
@@ -35,14 +35,12 @@ public function post( $request ) {
 			return $error;
 		}
 
-		if ( isset( $response->error ) ) {
-			$error = new WP_Error(
-				$response->error->code,
-				$response->error->message,
-				array( 'message' => $response->error->message )
+		if ( isset( $response->field_errors ) ) {
+			$this->logger->debug( 'Address validation errors: ' . implode( '; ', array_values( (array) $response->field_errors ) ), __CLASS__ );
```
DanReyLop: Note that this will need to be changed to `$this->logger->log` after https://github.com/Automattic/woocommerce-services/pull/1303 is merged.

dechov:
Rebased to catch up with logging changes.

classes/class-wc-rest-connect-address-normalization-controller.php
```diff
@@ -35,14 +35,12 @@ public function post( $request ) {
...
-		if ( isset( $response->error ) ) {
-			$error = new WP_Error(
-				$response->error->code,
-				$response->error->message,
-				array( 'message' => $response->error->message )
+		if ( isset( $response->field_errors ) ) {
+			$this->logger->log( 'Address validation errors: ' . implode( '; ', array_values( (array) $response->field_errors ) ), __CLASS__ );
+			return array(
+				'success' => true,
+				'fieldErrors' => $response->field_errors,
```
jeffstieler: Why is this getting converted back to camel case?

PR #1318, opened 2018-03-01
Add package and product measurements to shipping rate debug messages

dechov:
Fixes https://github.com/Automattic/woocommerce-services/issues/1163#issuecomment-368528627

Enhances shipping rate front-end debug messages by adding dimensions and weight for packages and their products:

screen shot 2018-03-01 at 4 34 46 pm

### DIFF
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -349,35 +349,56 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 				$packaging_lookup = $this->service_settings_store->get_package_lookup();
 
 				foreach ( (array) $instance->rates as $rate_idx => $rate ) {
-					$package_names = array();
-					$service_ids   = array();
+					$package_summaries = array();
+					$service_ids       = array();
+
+					$dimension_unit      = get_option( 'woocommerce_dimension_unit' );
+					$weight_unit         = get_option( 'woocommerce_weight_unit' );
+					$measurements_format = '%s x %s x %s ' . $dimension_unit . ', %s ' . $weight_unit;
 
 					foreach ( $rate->packages as $rate_package ) {
 						$package_format = '';
-						$items          = array();
+						$item_summaries = array();
 						$service_ids[]  = $rate_package->service_id;
 
 						foreach ( $rate_package->items as $package_item ) {
 							/** @var WC_Product $product */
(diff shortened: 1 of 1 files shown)


dechov:
Just realized package dimensions aren't returned in the store's dimension unit, so we may need a server change to convert them or at least report which units the dimensions are in. (Same may be true for box weight.) Marking in-progress for now.

jeffstieler:
> Just realized package dimensions aren't returned in the store's dimension unit

I don't think that's a problem - we are converting the dimensions into inches/pounds for USPS.

jeffstieler:
@dechov - so in your screenshot the debug message should use inches and pounds? Seems like we're missing a conversion back for the response..

dechov:
I think the package dimensions should use inches by passing the unit back with the dimensions, or it should convert to cm. Kg is fine in this case, since it's just the sum of the items, but I'm not sure whether it would be correct if the box had a weight.

dechov:
Removing in-progress label, but we'll need to capture the preset box dimensions edge case in a new issue if merged as is.

dechov:
Ready for review:

- Conversion has been fixed on the server
- Now grouping items by product as well

screen shot 2018-04-19 at 12 29 01 pm

classes/class-wc-connect-shipping-method.php
```diff
@@ -349,35 +349,66 @@ public function calculate_shipping( $package = array() ) {
...
+						foreach ( $product_counts as $product_id => $count ) {
 							/** @var WC_Product $product */
-							$product = $this->lookup_product( $package, $package_item->product_id );
+							$product = $this->lookup_product( $package, $product_id );
 							if ( $product ) {
-								$items[] = WC_Connect_Compatibility::instance()->get_product_name( $product );
+								$item_name = WC_Connect_Compatibility::instance()->get_product_name( $product );
+								$item_measurements = sprintf(
+									$measurements_format,
+									$package_item->length,
```
dechov: This causes a bug where the wrong measurements show up for the product. `$package_item` is left over from the previous loop and shouldn't be used here.

dechov: Fixed in https://github.com/Automattic/woocommerce-services/pull/1318/commits/b879a615aa85f34dd4bfe207d5ec784b7fe9f207.

Assuming items with the same product ID would always have the same measurements.

May also have been fixed by getting measurements directly of the product, but this would be less useful than the values from the server.

Issue #1306, opened 2018-02-12
Automated taxes: create tax classes for supported tax categories


jeffstieler:
As discussed with @mikkamp in p6q7sZ-3ZV-p2, TaxJar supports certain product categories and tax exemptions. Their documentation for setup is here: https://support.taxjar.com/article/309-overriding-tax-rates-and-exempting-products-in-woocommerce

Using the [`/v2/categories`](https://developers.taxjar.com/api/reference/#get-list-tax-categories) endpoint, we can retrieve a list of supported tax categories and create tax classes based on them as a convenience for our users.

PR #1321, opened 2018-03-06
Hide the tax settings and show a notice when automated taxes are enabled

### LINKED ISSUE #1296: Prevent manual tax rate input when "automated taxes" is enabled.
Take a page out of the TaxJar plugin:

screen shot 2018-02-02 at 2 59 30 pm

vs:

screen shot 2018-02-02 at 3 00 30 pm

### LINKED ISSUE #1297: Automated taxes: handle user choosing to enter prices inclusive of tax
Automated taxes requires that prices be entered *exclusive* of tax.

screen shot 2018-02-02 at 3 04 16 pm

We should either warn the user with a notice if they change that, or prevent them from changing that setting while automated taxes are enabled.

I think I prefer the latter.

marcinbot:
Fixes #1296
Fixes #1297

TaxJar shows all the options when enabled, but reverts them on save anyway. I decided to hide the options since they are always ignored when calculating the taxes.

screen shot 2018-03-06 at 13 04 27

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -95,10 +95,13 @@ class WC_Connect_TaxJar_Integration {
 	 * @return array
 	 */
 	public function add_tax_settings( $tax_settings ) {
+		$enabled = $this->is_enabled();
+
 		$automated_taxes = array(
 			'title'    => __( 'Automated taxes', 'woocommerce-services' ),
 			'id'       => self::OPTION_NAME, // TODO: save in `wc_connect_options`?
 			'desc_tip' => __( 'Automate your sales tax calculations with WooCommerce Services, powered by Jetpack.', 'woocommerce-services' ),
+			'desc'     => $enabled ? '<p>' . __( 'Powered by WooCommerce Services ― Your tax rates and settings are automatically configured.', 'woocommerce-services' ) . '</p>' : '',
 			'default'  => 'no',
 			'type'     => 'select',
 			'class'    => 'wc-enhanced-select',
@@ -111,13 +114,9 @@ class WC_Connect_TaxJar_Integration {
 		// Insert the "automated taxes" setting at the top (under the section title)
 		array_splice( $tax_settings, 1, 0, array( $automated_taxes ) );
 
-		if ( $this->is_enabled() ) {
+		if ( $enabled ) {
 			// If the automated taxes are enabled, disable the settings that would be reverted in the original plugin
 			foreach ( $tax_settings as $index => $tax_setting ) {
-				if ( 'tax_options' === $tax_setting['id'] && 'title' === $tax_setting['type'] ) {
-					$tax_settings[$index]['desc'] = __( 'Powered by WooCommerce Services ― Your tax rates and settings are automatically configured.', 'woocommerce-services' );
-				}
-
 				if ( in_array( $tax_setting['id'], array( 'tax_options', 'woocommerce_tax_classes', self::OPTION_NAME ) ) ) {
 					continue;
 				}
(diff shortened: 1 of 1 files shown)


jeffstieler:
Let's not remove the "additional tax classes" setting - it can be used by TaxJar - see:
* https://support.taxjar.com/article/309-overriding-tax-rates-and-exempting-products-in-woocommerce
* https://github.com/Automattic/woocommerce-services/issues/1306

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -104,9 +111,92 @@ public function add_tax_settings( $tax_settings ) {
...
+	public function sanitize_tax_option( $value, $option ) {
+		if ( ! is_array( $option ) || ! $this->is_enabled() && 'yes' != $_POST[self::OPTION_NAME] ) {
+			return $value;
+		}
+
+		switch( $option['id'] ) {
+			case 'woocommerce_calc_taxes':
+				// If automated taxes are enabled and user disables taxes we re-enable them
+				return 'yes';
+			case 'woocommerce_calc_taxes':
```
jeffstieler: Copy pasta? Should this be `woocommerce_tax_based_on`?

Could we reuse the `configure_tax_settings()` method somehow here instead of filtering value on save? _(Just looking to avoid duplicating this logic if we can)_

marcinbot: Fixed in ef522e3

I didn't notice `configure_tax_settings`. I missed it because it only happens once after the plugin has been installed and automated taxes enabled. If they are ever disabled and reenabled, the function wouldn't be called. The original plugin, on the other hand, called this code every time the class was instantiated.

I removed it in favor of my approach. It has two advantages: no unnecessary DB calls and the settings controls are updated on page save, so there's no need to refresh the page to see that they were reverted.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -104,9 +111,93 @@ public function add_tax_settings( $tax_settings ) {
...
+		}
+
+		return $value;
+	}
+
+	/**
+	 * Hack to hide the tax sections for additional tax class rate tables.
+	 */
+	public function output_sections_before() {
+		if ( ! $this->is_enabled() ) {
```
jeffstieler: Rather than check enabled status here, can we move the hook attachment to after the check in `init()`? _(Same for the _after() callback too)_

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -104,9 +111,93 @@ public function add_tax_settings( $tax_settings ) {
...
+			return;
+		}
+
+		?>
+		<div class="updated">
+			<p>
+				<b><?php esc_html_e( 'Powered by WooCommerce Services', 'woocommerce-services' ) ?></b>
+				<?php esc_html_e( ' ― Your tax rates and settings are automatically configured.', 'woocommerce-services' ) ?>
+			</p>
+		</div>
```
jeffstieler: Let's not add another notice for this - can we move this text to the `desc` of the automated tax settings field instead?

marcinbot: Agreed. A notice is easy to miss.

At first I added a desc to the control, but it didn't look great:
screen shot 2018-03-21 at 12 34 22

So I added it to the section title and it appears only when the taxes are enabled:
screen shot 2018-03-21 at 12 41 15

classes/class-wc-connect-taxjar-integration.php
marcinbot: Tried moving them, but then they look at the `is_enabled` value from the time when `init()` was called and it can change between that and the time the hooks are activated.

This makes the sections to still render right after the taxes were enabled (they only disappear after a refresh) and be hidden if they were disabled (need to refresh to make them appear).

classes/class-wc-connect-taxjar-integration.php
jeffstieler: You can wrap the `desc` in a paragraph, which results in:

screen shot 2018-03-21 at 8 33 36 am

I personally prefer the message to appear under the automated taxes control, but that in no way blocks this for merge. You can choose to move the message or not - :shipit: 👍

PR #1322, opened 2018-03-06
Updated the versions matrix to test our code in CI

### LINKED ISSUE #780: Make CI run with PHP7 and more versions of WordPress
Right now, CI runs with the folowing versions of WordPress:
* 4.4
* LATEST

We should aim for it to run in all the major versions of WordPress we support (that is, 4.4, 4.5, 4.6 and LATEST [4.7]).

Also, CI runs with these versions of PHP:
* 5.2
* 5.3
* 5.4
* 5.5
* 5.6

We should add PHP 7.0 and 7.1.

There's no need to include all the combinations in the TravisCI `matrix`. I think a sensible list of runs will be something like this:
* PHP 5.2 - WP 4.4
* PHP 5.2 - WP 4.5
* PHP 5.2 - WP 4.6
* PHP 5.2 - WP 4.7
* PHP 5.3 - WP 4.7
* PHP 5.4 - WP 4.7
* PHP 5.5 - WP 4.7
* PHP 7.0 - WP 4.7
* PHP 7.1 - WP 4.7

DanReyLop:
Closes #780

The new build matrix includes a variety of these versions:
WordPress:
- 4.4 (earliest supported by WooCommerce 2.6)
- 4.5
- 4.6
- 4.7
- 4.8
- `latest` (4.9 at the moment of writing this)

WooCommerce:
- 2.6.14 (earliest supported by WCS for now)
- 3.0.9
- 3.1.3
- 3.2.6
- 3.3.3 (latest)

PHP:
- 5.2 (earliest we support)
- 5.3
- 5.4
- 5.5
- 5.6
- 7.0
- 7.1
- 7.2 (latest)

I added a sensible matrix so we don't hog TravisCI resources for no reason. I made sure that these situations are covered:
- Earliest PHP version with latest WC and WP versions.
- Earliest PHP, WC and WP versions.
- Latest PHP versions with earliest WC and WP versions.
- Latest PHP, WC and WP versions.
- Earliest WC version with latest WP version.
- Earliest WP version with latest WC version.

If you ask me, that's already overkill, but the new matrix is slightly shorter than before (1 case less), so I think that's OK. Thoughts?

No testing instructions, just check that CI passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,8 @@ matrix:
 # Clones WordPress and configures our testing environment.
 before_script: bash tests/bin/install-wc-tests.sh wordpress_test root '' localhost $WP_VERSION $WC_VERSION
 
+script: phpunit --version
+
 sudo: false
 
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,7 @@
 # Tell Travis CI we're using PHP
 language: php
 
-# Next we define our matrix of build configurations to test against.
-
+# Define our matrix of build configurations to test against.
 matrix:
   include:
    - php: 5.2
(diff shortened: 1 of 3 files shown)


DanReyLop:
Build passes now, this is ready for review.

PR #1325, opened 2018-03-07
Added integration with WooCommerce Shipment Tracking

### LINKED ISSUE #1156: Integrate with WooCommerce Shipment Tracking

marcinbot:
Fixes #1156 

Automatically adds the tracking number when a new label is purchased and checks for WooCommerce Shipment Tracking existence when sending the "order complete" email. If WCST is going to append its data to the email, WCS will not do that.

Will need PR 87 to be shipped on the `woocommerce-shipment-tracking` repo to be fully functional, but will work without it. The PR allows us to refresh the tracking data without reloading the page.

I also wanted to lay some ground work for future integration with other official WooCommerce extensions:
* added a static `WC_Connect_Extension_Compatibility` class which should contain the PHP integration code
* added per-JS-app Redux middleware. It allows to listen for actions and extend them beyond just modifying the state, which could give us more flexibility without needing to go to Calypso with every change

### DIFF
--- a/client/apps/shipping-label/redux-middleware.js
+++ b/client/apps/shipping-label/redux-middleware.js
@@ -1,0 +1,26 @@
+/**
+ * Internal dependencies
+ */
+//from calypso
+import {
+	WOOCOMMERCE_SERVICES_SHIPPING_LABEL_EXIT_PRINTING_FLOW,
+} from 'woocommerce/woocommerce-services/state/action-types';
+
+const middlewareActions = {
+	[  WOOCOMMERCE_SERVICES_SHIPPING_LABEL_EXIT_PRINTING_FLOW ]: () => {
+		window.wc_shipment_tracking_refresh && window.wc_shipment_tracking_refresh();
+	},
+};
+
+export default () => ( next ) => ( action ) => {
+	// let the action go to the reducers
+	next( action );
+
+	const middlewareAction = middlewareActions[ action.type ];
+	if ( ! middlewareAction ) {
+		return;
+	}
+
+	// perform the action
+	setTimeout( () => middlewareAction( action ) );
+};
--- a/client/apps/shipping-label/redux-middleware.js
+++ b/client/apps/shipping-label/redux-middleware.js
@@ -3,11 +3,14 @@
  */
 //from calypso
 import {
-	WOOCOMMERCE_SERVICES_SHIPPING_LABEL_EXIT_PRINTING_FLOW,
+	WOOCOMMERCE_SERVICES_SHIPPING_LABEL_PURCHASE_RESPONSE,
 } from 'woocommerce/woocommerce-services/state/action-types';
 
 const middlewareActions = {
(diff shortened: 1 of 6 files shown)


client/apps/shipping-label/redux-middleware.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Internal dependencies
+ */
+//from calypso
+import {
+	WOOCOMMERCE_SERVICES_SHIPPING_LABEL_EXIT_PRINTING_FLOW,
+} from 'woocommerce/woocommerce-services/state/action-types';
+
+const middlewareActions = {
+	[  WOOCOMMERCE_SERVICES_SHIPPING_LABEL_EXIT_PRINTING_FLOW ]: () => {
```
DanReyLop: Using the `WOOCOMMERCE_SERVICES_SHIPPING_LABEL_PURCHASE_RESPONSE` action (and checking that there isn't any `error`) would be more efficient, and will finish fetching even before the user closes the modal most of the time.

client/apps/shipping-label/redux-middleware.js
```diff
@@ -0,0 +1,26 @@
...
+	// let the action go to the reducers
+	next( action );
+
+	const middlewareAction = middlewareActions[ action.type ];
+	if ( ! middlewareAction ) {
+		return;
+	}
+
+	// perform the action
+	setTimeout( () => middlewareAction( action ) );
```
DanReyLop: Why is `setTimeout` needed? `wc_shipment_tracking_refresh()` doesn't block the thread, and order doesn't matter.

marcinbot: I think the article about it I found was delaying this action so the state gets fully updated, but you're right, it's not need here. I removed it.

client/apps/shipping-label/redux-middleware.js
marcinbot: Yep, it's better. I changed it.

PR #1328, opened 2018-03-08
Merged the packages and labels settings

### LINKED ISSUE #1162: Rename "Packages" tab to avoid confusion with ASP "Packages" tab
Currently, if a user has WC Services activated and also using [Advanced Shipping Packages](https://docs.woocommerce.com/document/woocommerce-advanced-shipping-packages/) extension, the Shipping menu has 2 identical tabs called **Packages** next to each which is confusing:

Image Link: https://cloudup.com/cXicgDDoic0

**ADVANCED SHIPPING PACKAGES** tab looks as follows:

Image Link: https://cloudup.com/cSXJ8uCJYVd

**WC SERVICES** tab looks as follows:

Image Link: https://cloudup.com/cWwrlGJ5T7q

I see 3 solutions here:

1. We can rename WC Services Packages section to something like "Services Packages"
2. Developers of Advanced Shipping Packages extension could rename their tab to something like "Advanced Packages"
3. Both Packages settings would be displayed under the same tab called "Packages" but will have different Sections (similar to how VAT settings section is being added to the general Tax section upon VAT Number extension activation).

marcinbot:
Fixes #1162

Merges the non-service shipping settings pages into one page called "WooCommerce Services":
screen shot 2018-04-22 at 10 27 31

To test:
* all controls on the page should work
* the settings should save
* for non-Jetpack master users a notification should appear that they are not able to manage payment methods. They should still be able to manage packages


jeffstieler:
I'm not too fond of calling this "Services packaging"..

How difficult would it be to combine the packages+labels settings like in Calypso? Perhaps then we can just call the section "WooCommerce Services".

cc: @kellychoffman for design input.

kellychoffman:
> I'm not too fond of calling this "Services packaging"..

+1

> How difficult would it be to combine the packages+labels settings like in Calypso? Perhaps then we can just call the section "WooCommerce Services".

Yes, please.

PR #1337, opened 2018-03-14
Fixed the page crash when PPEC has not initialized successfully

### LINKED ISSUE #1336: Fatal error when PPEC plugin is installed but its dependencies are not
Originally reported here: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/395

PPEC has a bunch of dependencies that have to be met or the plugin's `settings` will not be initialized. [See here](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/326e0718f13d99d372c7fa6a86bf690d0006f785/includes/class-wc-gateway-ppec-plugin.php#L253)

If any of those are not met or the PPEC failed to initialize in some other way, WCS will cause the page to crash:
screen shot 2018-03-14 at 13 48 49

CC @danass

marcinbot:
Fixes #1336 

To test:
* have the PPEC installed and activated
* make one of the dependencies to break (use WooCommerce<2.5 or disable/remove the `curl` library in PHP). Or open `plugins/woocommerce-gateway-paypal-express-checkout/includes/class-wc-gateway-ppec-plugin.php` and modify it to throw an exception around line 155
* load any admin page
* the page should load

### DIFF
--- a/classes/class-wc-connect-paypal-ec.php
+++ b/classes/class-wc-connect-paypal-ec.php
@@ -38,11 +38,16 @@ if ( ! class_exists( 'WC_Connect_PayPal_EC' ) ) {
 				return;
 			}
 
+			$ppec_plugin = wc_gateway_ppec();
+			if ( ! property_exists( $ppec_plugin, 'settings' ) || empty( $ppec_plugin->settings ) ) {
+				return;
+			}
+
 			$this->maybe_set_reroute_requests();
 
 			add_filter( 'woocommerce_paypal_express_checkout_settings', array( $this, 'adjust_form_fields' ) );
 			$this->initialize_settings();
-			$settings = wc_gateway_ppec()->settings;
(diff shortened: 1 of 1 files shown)


jeffstieler:
Should we check for `ips` as well?

https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-paypal-ec.php#L134

dechov:
@jeffstieler I think `settings` should be enough to verify PPEC dependencies are satisfied: if `settings` is there, then [`_run`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/326e0718f13d99d372c7fa6a86bf690d0006f785/includes/class-wc-gateway-ppec-plugin.php#L160) has run, which will initialize `ips` as well — unless you have in mind a scenario in which it wouldn't be.

PR #1338, opened 2018-03-14
Include warning message in response upon failure to retrieve payment methods

dechov:
Returns the following message if payment methods could not be fetched when requesting label settings:

screen shot 2018-03-14 at 4 58 57 pm

This is to notify the merchant if their payment methods were not updated in the current page load, as implemented in https://github.com/Automattic/wp-calypso/pull/23307.

To test, make WCS unreachable by setting `WOOCOMMERCE_CONNECT_SERVER_URL` to a URL in which there is no WCS server, and load the label settings page.

### DIFF
--- a/classes/class-wc-connect-payment-methods-store.php
+++ b/classes/class-wc-connect-payment-methods-store.php
@@ -47,6 +47,7 @@ if ( ! class_exists( 'WC_Connect_Payment_Methods_Store' ) ) {
 			$this->update_payment_methods( $payment_methods );
 
 			$this->potentially_update_selected_payment_method_from_payment_methods( $payment_methods );
+			return true;
 		}
 
 		protected function potentially_update_selected_payment_method_from_payment_methods( $payment_methods ) {
--- a/classes/class-wc-connect-payment-methods-store.php
+++ b/classes/class-wc-connect-payment-methods-store.php
@@ -28,19 +28,24 @@ if ( ! class_exists( 'WC_Connect_Payment_Methods_Store' ) ) {
 
 		}
 
+		/**
+		 * Fetch stored payment methods from server and store in options.
+		 *
+		 * @return bool Were payment methods successfully retrieved?
+		 */
 		public function fetch_payment_methods_from_connect_server() {
 
 			$response_body = $this->api_client->get_payment_methods();
 
 			if ( is_wp_error( $response_body ) ) {
 				$this->logger->log( $response_body, __FUNCTION__ );
-				return;
+				return false;
 			}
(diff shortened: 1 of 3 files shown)


classes/class-wc-connect-payment-methods-store.php
```diff
@@ -47,6 +47,7 @@ public function fetch_payment_methods_from_connect_server() {
 			$this->update_payment_methods( $payment_methods );
 
 			$this->potentially_update_selected_payment_method_from_payment_methods( $payment_methods );
+			return true;
```
jeffstieler: Let's explicitly return `false` from the two `WP_Error` cases. Also, please add a PHP Docblock for this function. 🙂

classes/class-wc-rest-connect-account-settings-controller.php
```diff
@@ -45,7 +45,8 @@ public function get() {
 				'master_user_login' => is_a( $master_user, 'WP_User' ) ? $master_user->user_login : '',
 				'master_user_email' => $email,
 				'payment_methods' => $this->payment_methods_store->get_payment_methods(),
-			)
+				'warnings' => ! $payment_methods_fetched ? array( 'payment_methods' => 'Encountered an error while retrieving stored payment methods.' ) : array(),
```
jeffstieler: The WP way is to test the positive case, not the negative. ([#](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#ternary-operator))

Please wrap the error message string for translation.

How about "There was a problem fetching your saved credit cards." instead?

dechov:
Done:

screen shot 2018-03-15 at 2 36 34 pm

classes/class-wc-rest-connect-account-settings-controller.php
```diff
@@ -45,7 +45,8 @@ public function get() {
 				'master_user_login' => is_a( $master_user, 'WP_User' ) ? $master_user->user_login : '',
 				'master_user_email' => $email,
 				'payment_methods' => $this->payment_methods_store->get_payment_methods(),
-			)
+				'warnings' => array( 'payment_methods' => $payment_methods_fetched ? false : __( 'There was a problem fetching your saved credit cards.', 'woocommerce-services' ) ),
```
jeffstieler: Let's move this conditional up above the return to reduce line length.

dechov:
Fixed warning message wording, as decided in https://github.com/Automattic/wp-calypso/pull/23307#issuecomment-374332092. Will merge this as soon as that PR is approved.

Issue #1313, opened 2018-02-20
Output tax debug messages on the Cart and Checkout pages


dechov:
Do for taxes what https://github.com/Automattic/woocommerce-services/pull/1303 did for shipping: when in debug mode, show front-end messages with more details on the rate calculations.

Might want to show certain breakdown data always, and the rest when it's non-zero?

Can either strive to diverge from the TaxJar plugin as little as possible, or to contribute front-end debug output back to the TaxJar plugin as well.

#### Example tax calculation responses

US:
```
[...]

PR #1339, opened 2018-03-15
Tweaks to generic API error messaging

dechov:
_Aims to make sure error messaging is adequate in all cases where the server couldn't be reached or returns a `500` status._

#### If the server couldn't be reached

Instead of:

screen shot 2018-02-06 at 6 09 09 pm

...the API client responds with a more appropriate message — though if and how it shows up on the client depends on the context. Examples:

screen shot 2018-03-15 at 11 23 17 am

screen shot 2018-03-14 at 6 29 51 pm

_Edit:_ this now says:
> The WooCommerce Services API encountered an error. Please try again.

The reference to the server is changed from "WooCommerce Services server" (kind of inelegant as a result of the renaming to "WooCommerce Services") to "WooCommerce Services API", based on a precedent in the connection test tool:
https://github.com/Automattic/woocommerce-services/blob/3624a2c7c6181667af76f316d7e7f65e47a51909/classes/class-wc-connect-debug-tools.php#L28-L35

<s>(Some discussion in p1521044361000197-slack-hydra of the suggestion to check whether requests can go out at all. If possible, we should consider only showing this if we have particular reason to believe this might be the case — perhaps only if no server requests have yet succeeded.)</s>

#### Internal server error

If a 500 error comes back, there is generally no information specific to the error we can communicate via the client, so instead of:

screen shot 2018-03-15 at 2 43 57 pm

...we can simplify the message:

screen shot 2018-03-15 at 1 43 50 pm

Edit: other 5xx errors (such as 502 during server downtime) now yield:
> The WooCommerce Services API is unavailable. Please try again in just a moment.

#### 4xx errors

If available, the raw error message is passed through, without the HTTP status or meta-error message "Error: The WooCommerce Services server returned [...]":

screen shot 2018-02-13 at 2 38 49 pm

### <s>Error communication per context</s>

<s>Manually tested all relevant cases by searching for `$this->api_client->` in the codebase. Some e2e tests might be helpful so as to be confident errors are and will continue to be properly communicated.</s>

To test global notices:

- For _internal server error_ message, can point payment proxy on local server to a URL where there is no server responding, and then trigger any of these:
  - Label status fetch on order page load
  - Label refund request

- For _unreachable server_ message, can set `WOOCOMMERCE_CONNECT_SERVER_URL` to a URL where there is no WCS server responding, and then trigger any of the above, or:
  - Shipping rate fetch (on cart/checkout page, with debug mode enabled) 
  - Label rate fetch
  - Label purchase request

<s>TODO:
- Show some error if manual service schema fetch fails
- TaxJar proxy fails silently if WCS down. Might best be addressed by https://github.com/Automattic/woocommerce-services/issues/1313
- PPEC gives only an unhelpful message if WCS proxy is down: https://cloudup.com/c3OTHZxIMaj
- Test print shows a print dialog with only a `WP_Error` in JSON form. Show a notice instead
</s>

_Edit:_ now captured in https://github.com/Automattic/woocommerce-services/issues/1364

<s>Also:
- On failure to fetch payment methods, might want to communicate more about how it failed</s>

### DIFF
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -376,8 +376,8 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API could not be reached.' );
 				} elseif ( 500 == $response_code ) {
 					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API encountered an error. Please try again.' );
-				} elseif ( 200 != $response_code ) {
-					return new WP_Error( 'wcc_server_error', sprintf( 'The WooCommerce Services API returned HTTP status code: %d', $response_code ) );
+				} elseif ( 400 <= $response_code ) {
+					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API could not process the request.' );
 				}
 				return $response;
 			}
@@ -387,15 +387,9 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 				$response_body = json_decode( $response_body );
 			}
 
-			if ( 200 != $response_code ) {
+			if ( 400 <= $response_code ) {
 				if ( empty( $response_body ) ) {
-					return new WP_Error(
-						'wcc_server_empty_response',
-						sprintf(
-							'The WooCommerce Services API returned HTTP status code %d and an empty response body.',
-							$response_code
-						)
-					);
+					return new WP_Error( 'wcc_server_empty_response', 'The WooCommerce Services API could not process the request.' );
 				}
 
 				$error   = property_exists( $response_body, 'error' ) ? $response_body->error : '';
@@ -406,16 +400,7 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 					return new WP_Error( 'wcc_server_error_response', 'The WooCommerce Services API encountered an error. Please try again.' );
 				}
 
-				return new WP_Error(
-					'wcc_server_error_response',
-					sprintf(
-						'The WooCommerce Services API returned HTTP status code %d: %s. %s',
-						$response_code,
-						$error,
-						$message
-					),
-					$data
-				);
+				return new WP_Error( 'wcc_server_error_response', $message, $data );
 			}
 
 			return $response_body;
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -376,6 +376,8 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API could not be reached.' );
 				} elseif ( 500 == $response_code ) {
 					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API encountered an error. Please try again.' );
+				} elseif ( 500 < $response_code ) {
+					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API is unavailable. Please try again in just a moment.' );
 				} elseif ( 400 <= $response_code ) {
 					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API could not process the request.' );
 				}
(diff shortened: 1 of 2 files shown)


classes/class-wc-connect-api-client.php
```diff
@@ -372,11 +372,14 @@ protected function request( $method, $path, $body = array() ) {
...
-						sprintf( 'Error: The WooCommerce Services server returned HTTP code: %d', $response_code )
-					);
+				if ( ! $response_code ) {
+					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API could not be reached.' );
+				} elseif ( 500 == $response_code ) {
+					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API encountered an error. Please try again.' );
+				} elseif ( 500 < $response_code ) {
+					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API is unavailable. Please try again in just a moment.' );
+				} elseif ( 400 <= $response_code ) {
+					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API could not process the request.' );
```
jeffstieler: Can you explain your reasoning for these error messages? _(and the ones below)_

At a glance it looks like all of these could be one of two messages - either "could not be reached" or "encountered an error".

Why do some not prompt the user to try again?

classes/class-wc-connect-api-client.php
```diff
@@ -372,11 +372,14 @@ protected function request( $method, $path, $body = array() ) {
...
-						'wcc_server_error',
-						sprintf( 'Error: The WooCommerce Services server returned HTTP code: %d', $response_code )
-					);
+				if ( ! $response_code ) {
+					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API could not be reached.' );
+				} elseif ( 500 == $response_code ) {
+					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API encountered an error. Please try again.' );
+				} elseif ( 500 < $response_code ) {
+					return new WP_Error( 'wcc_server_error', 'The WooCommerce Services API is unavailable. Please try again in just a moment.' );
+				} elseif ( 400 <= $response_code ) {
```
jeffstieler: I don't know how common/possible it would be in non-development environments, but I see 404s a lot on the order details page when switching between my local WCS server and staging/production.

This change removes the meaningful "not found" I saw before.

jeffstieler (changes requested):
Questions about message verbiage.

classes/class-wc-connect-api-client.php
dechov: This is definitely wide open to revision but here's the understanding on which I based it:

- No error code means the server wasn't reached at all, for reasons unrelated to WCS. Even though it may be something like a network hiccup, there's no WCS-specific reason to think trying again might help.
- `500` means something went wrong on the server, perhaps a bug in the server application, and there's no way to know whether trying again will help but there's nothing else we can suggest without manual investigation.
- `502` means the WCS server is down. If due to a deploy, it'll be back up within a few seconds, and otherwise, it'll be back up some time soon — trying again is the only thing you can do and it should eventually work.
- `4xx` means there's something about the request the server can't or won't process — no need to try again with the same request.

I'm actually a little bit uneasy about showing "Please try again." for a 500 error over and over each time they try, since it could well be that it will never work. I think — ideally — this response would indicate a very exceptional case, in which contacting support would be justified.

classes/class-wc-connect-api-client.php
dechov: :+1: Will bring back that string for `4xx` responses at least.

jeffstieler (changes requested):

jeffstieler:
Ping @dechov 👋

classes/class-wc-connect-api-client.php
dechov: @jeffstieler This had fallen off my radar – sorry about that.

I'm good with those changes except that I think "The WooCommerce Services API encountered an error." could be a little bit misleading, at least for the `502` case: as far as I understand, it's the HTTP server that "encountered an error", the error being that the WCS application hasn't properly started (yet?).

It seems worth communicating that it has nothing to do with content of the request itself. For 502 responses, proposing "The WooCommerce Services API is momentarily unavailable." – what do you think? If 👎, I'll go ahead with the change as you'd described.

classes/class-wc-connect-api-client.php
```diff
@@ -386,31 +389,22 @@ protected function request( $method, $path, $body = array() ) {
...
-					),
-					$data
-				);
+				if ( 500 == $response_code ) {
+					return new WP_Error( 'wcc_server_error_response', 'The WooCommerce Services API encountered an error. Please try again.' );
+				} elseif ( 500 < $response_code ) {
+					return new WP_Error( 'wcc_server_error_response', 'The WooCommerce Services API is unavailable. Please try again in just a moment.' );
+				}
+
+				return new WP_Error( 'wcc_server_error_response', $message, $data );
```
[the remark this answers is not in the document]
dechov: Just to be sure, are you referring to the repetition of the code, the message, or both?

PR #1347, opened 2018-03-22
Added labels reports section

### LINKED ISSUE #1277: Label receipts - show/link to a centralized place allowing to view all and export receipts
We should either add a new page or link to an improved page on Calypso that allows to browse and export all receipts.

### LINKED ISSUE #1327: Summary of total spent on labels
A customer was asking if it's possible to get a summary of all the postage purchased through WooCommerce Services.

While you can get it from your credit card company you have to sum up all transactions.

It would be nice to see a report totaling amount of money spent (and perhaps saved?!) through using WooCommerce Services label printing.

marcinbot:
Fixes #1277
Fixes #1327

Adds a simple reports page listing all labels successfully purchased during a selected time period:
screen shot 2018-03-22 at 16 37 24

The page can be accessed under `WooCommerce > Reports > Labels`
* Doesn't list labels with errors
* Labels are listed by package, so if two labels are purchased with one request, it will result in two rows
* Only lists time, price, order ID and service. Do we need to display any other data?

### DIFF
--- a/classes/class-wc-connect-label-reports.php
+++ b/classes/class-wc-connect-label-reports.php
@@ -18,7 +18,7 @@ if ( ! class_exists( 'WC_Connect_Label_Reports' ) ) {
 			?>
 			<a
 				href="#"
-				download="report-<?php echo esc_attr( $current_range ); ?>-<?php echo date_i18n( 'Y-m-d', current_time( 'timestamp' ) ); ?>.csv"
+				download="report-shipping-labels-<?php echo esc_attr( $current_range ); ?>-<?php echo date_i18n( 'Y-m-d', current_time( 'timestamp' ) ); ?>.csv"
 				class="export_csv"
 				data-export="table"
 			>
--- a/classes/class-wc-connect-label-reports.php
+++ b/classes/class-wc-connect-label-reports.php
@@ -4,7 +4,7 @@ if ( ! class_exists( 'WC_Connect_Label_Reports' ) ) {
 	include_once( WC()->plugin_path() . '/includes/admin/reports/class-wc-admin-report.php' );
 
 	class WC_Connect_Label_Reports extends WC_Admin_Report {
-		const LABELS_CACHE_KEY = 'wcs_label_reports';
+		const LABELS_TRANSIENT_KEY = 'wcs_label_reports';
 
 		/**
 		 * @var WC_Connect_Service_Settings_Store
@@ -60,10 +60,11 @@ if ( ! class_exists( 'WC_Connect_Label_Reports' ) ) {
 		}
 
 		private function query_labels() {
-			$all_labels =  wp_cache_get( self::LABELS_CACHE_KEY );
+			$all_labels =  get_transient( self::LABELS_TRANSIENT_KEY );
 			if ( false === $all_labels ) {
 				$all_labels = $this->get_all_labels();
-				wp_cache_set( self::LABELS_CACHE_KEY, $all_labels, '', 1800 );
+				//set transient with ttl of 30 minutes
+				set_transient( self::LABELS_TRANSIENT_KEY, $all_labels, '', 1800 );
 			}
 
 			// translate timestamps to JS timestapms
--- a/classes/class-wc-connect-label-reports.php
+++ b/classes/class-wc-connect-label-reports.php
@@ -4,6 +4,8 @@ if ( ! class_exists( 'WC_Connect_Label_Reports' ) ) {
 	include_once( WC()->plugin_path() . '/includes/admin/reports/class-wc-admin-report.php' );
 
 	class WC_Connect_Label_Reports extends WC_Admin_Report {
+		const LABELS_CACHE_KEY = 'wcs_label_reports';
+
 		/**
 		 * @var WC_Connect_Service_Settings_Store
 		 */
@@ -58,7 +60,11 @@ if ( ! class_exists( 'WC_Connect_Label_Reports' ) ) {
 		}
 
 		private function query_labels() {
-			$all_labels = $this->get_all_labels();
+			$all_labels =  wp_cache_get( self::LABELS_CACHE_KEY );
+			if ( false === $all_labels ) {
+				$all_labels = $this->get_all_labels();
+				wp_cache_set( self::LABELS_CACHE_KEY, $all_labels, '', 1800 );
+			}
 
 			// translate timestamps to JS timestapms
 			$start_date = $this->start_date * 1000;
--- a/classes/class-wc-connect-label-reports.php
+++ b/classes/class-wc-connect-label-reports.php
@@ -88,10 +88,10 @@ if ( ! class_exists( 'WC_Connect_Label_Reports' ) ) {
 					continue;
 				}
 
-				//ignore labels with complete or requested refunds, but not the rejected ones
+				//ignore labels with complete refunds
 				if ( isset( $label['refund'] ) ) {
 					$refund = ( array ) $label['refund'];
-					if ( isset( $refund['status'] ) && 'rejected' !== $refund['status'] ) {
+					if ( isset( $refund['status'] ) && 'completed' === $refund['status'] ) {
 						continue;
 					}
 				}
@@ -129,6 +129,21 @@ if ( ! class_exists( 'WC_Connect_Label_Reports' ) ) {
 			return '<a href="' . WC_Connect_Compatibility::instance()->get_edit_order_url( $order ) . '">' . $order->get_order_number( $order ) . '</a>';
 		}
 
+		private function get_label_refund_status( $label ) {
+			if ( ! isset( $label['refund'] ) ) {
+				return '';
+			}
+
+			$refund = ( array ) $label['refund'];
+
+			if ( isset( $refund['status'] ) &&
+				( 'rejected' === $refund['status'] || 'complete' === $refund['status'] ) ) {
+				return '';
+			}
+
+			return __( 'Requested', 'woocommerce-services' );
+		}
+
 		/**
 		 * Get the main chart.
 		 */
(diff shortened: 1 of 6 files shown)


classes/class-wc-connect-label-reports.php
```diff
@@ -0,0 +1,163 @@
...
+				<?php if ( ! empty( $labels ) ) : ?>
+					<tbody>
+						<?php
+						foreach ( $labels as $label ) {
+							?>
+							<tr>
+								<th scope="row"><?php echo date_i18n( get_option( 'date_format' ) . ' g:i a', $label['created'] / 1000 ); ?></th>
+								<td><a href="<?php echo admin_url( 'post.php?post=' . $label['order_id'] . '&action=edit' ); ?>"><?php echo $label['order_id']; ?></a></td>
+								<td><?php echo wc_price( $label['rate'] ); ?></td>
+								<td><?php echo $label['service_name'] ?></td>
```
jeffstieler: Nit: let's add some line breaks here to reduce the overall line lengths.

classes/class-wc-connect-label-reports.php
```diff
@@ -0,0 +1,163 @@
...
+					</tr>
+				</thead>
+				<?php if ( ! empty( $labels ) ) : ?>
+					<tbody>
+						<?php
+						foreach ( $labels as $label ) {
+							?>
+							<tr>
+								<th scope="row"><?php echo date_i18n( get_option( 'date_format' ) . ' g:i a', $label['created'] / 1000 ); ?></th>
+								<td><a href="<?php echo admin_url( 'post.php?post=' . $label['order_id'] . '&action=edit' ); ?>"><?php echo $label['order_id']; ?></a></td>
```
jeffstieler: There are extensions that allow merchants to change their order numbers ([#](https://woocommerce.com/products/sequential-order-numbers-pro/), [#](https://booster.io/features/woocommerce-custom-order-numbers/)) - I think we'll need to retrieve the `WC_Order` so that we're displaying the correct order number here (it doesn't always match the post ID).

classes/class-wc-connect-label-reports.php
```diff
@@ -0,0 +1,163 @@
...
+						<th><?php esc_html_e( 'Time', 'woocommerce-services' ); ?></th>
+						<th><?php esc_html_e( 'Order', 'woocommerce-services' ); ?></th>
+						<th><?php esc_html_e( 'Price', 'woocommerce-services' ); ?></th>
+						<th><?php esc_html_e( 'Service', 'woocommerce-services' ); ?></th>
+					</tr>
+				</thead>
+				<?php if ( ! empty( $labels ) ) : ?>
+					<tbody>
+						<?php
+						foreach ( $labels as $label ) {
```
jeffstieler: Nit: this `foreach` should use the PHP alternate syntax, like the other control flow statements mixed with markup.

classes/class-wc-connect-label-reports.php
```diff
@@ -0,0 +1,163 @@
...
+			$current_range = ! empty( $_GET['range'] ) ? sanitize_text_field( $_GET['range'] ) : '7day';
+
+			if ( ! in_array( $current_range, array( 'custom', 'year', 'last_month', 'month', '7day' ) ) ) {
+				$current_range = '7day';
+			}
+
+			$this->check_current_range_nonce( $current_range );
+			$this->calculate_current_range( $current_range );
+
+			$hide_sidebar = true;
```
jeffstieler: Looks like we aren't using the sidebar - should this be `false`?

classes/class-wc-connect-label-reports.php
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		private function compare_label_dates_desc( $label_a, $label_b ) {
+			return $label_b['created'] - $label_a['created'];
+		}
+
+		private function get_all_labels() {
+			global $wpdb;
+			$query = "SELECT post_id, meta_value FROM {$wpdb->postmeta} WHERE meta_key = 'wc_connect_labels'";
+			$db_results = $wpdb->get_results( $query );
```
jeffstieler: Can we cache this result? Since we're retrieving all of the rows, caching them could significantly speed up running subsequent reports with different filters as well as just avoiding unnecessary work.

classes/class-wc-connect-label-reports.php
```diff
@@ -0,0 +1,163 @@
...
+			$results = array();
+
+			foreach( $db_results as $meta ) {
```
jeffstieler: Nit: whitespace.

classes/class-wc-connect-label-reports.php
```diff
@@ -0,0 +1,163 @@
...
+		public function __construct( WC_Connect_Service_Settings_Store $settings_store ) {
+			$this->settings_store = $settings_store;
+		}
+
+		public function get_export_button() {
+			$current_range = ! empty( $_GET['range'] ) ? sanitize_text_field( $_GET['range'] ) : '7day';
+			?>
+			<a
+				href="#"
+				download="report-<?php echo esc_attr( $current_range ); ?>-<?php echo date_i18n( 'Y-m-d', current_time( 'timestamp' ) ); ?>.csv"
```
jeffstieler: I know this is following the core pattern, but can we add `shipping` or `shipping-labels` to the filename?

dechov:
Fixes https://github.com/Automattic/woocommerce-services/issues/1327 — also, listing discounts was suggested there.

woocommerce-services.php
```diff
@@ -870,6 +878,28 @@ public function localize_and_enqueue_service_script( $method_id, $instance_id =
...
 
+		/**
+		 * Filter function for adding the report tabs
+		 *
+		 * @param array $reports - report tabs meta
+		 * @return array report tabs with WCS tabs added
+		 */
+		public function reports_tabs( $reports ) {
+			$reports[ 'wcs_labels' ] = array(
+				'title' => __( 'Labels', 'woocommerce-services' ),
```
jeffstieler: I don't think `Labels` is too obvious - can we change it to `Shipping Labels`?

classes/class-wc-connect-label-reports.php
marcinbot: I added calls to `wp_cache_set` in a73443f, but this cache [won't be persistent unless the site uses a persistent cache plugin](https://codex.wordpress.org/Class_Reference/WP_Object_Cache#Persistent_Caching). Without it the data will be reloaded on each page refresh anyway.

classes/class-wc-connect-label-reports.php
marcinbot: No, if it's not set to false it will result in a huge empty space on the left of the table

woocommerce-services.php
marcinbot: 3c4fb76

classes/class-wc-connect-label-reports.php
marcinbot: 6856dc6

classes/class-wc-connect-label-reports.php
marcinbot: d66270e

classes/class-wc-connect-label-reports.php
marcinbot: d66270e

classes/class-wc-connect-label-reports.php
jeffstieler: >  if it's not set to false

Do you mean if it **is** set to `false`? The current value is `true`.

classes/class-wc-connect-label-reports.php
```diff
@@ -0,0 +1,221 @@
...
+			$this->calculate_current_range( $current_range );
+
+			$hide_sidebar = true;
+
+			include( WC()->plugin_path() . '/includes/admin/views/html-report-by-date.php' );
+		}
+
+		private function get_order_url( $post_id ) {
+			$order = wc_get_order( $post_id );
+			return '<a href="' . WC_Connect_Compatibility::instance()->get_edit_order_url( $order ) . '">' . WC_Connect_Compatibility::instance()->get_order_id( $order ) . '</a>';
```
jeffstieler: This should be retrieving the order number, not the order id. See: [`WC_Order::get_order_number()`](https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-order.php#L441)

jeffstieler:
> I will look into adding a discount column
> Edit: it might not be possible to add that without DB changes or at least response changes on the server side

Let's punt on the discount column.

> Added a refund column which contains the refund status: one of Completed, Rejected or Requested. But I feel it might be better to list the refunded amount instead and then sum it up at the bottom in the Totals row. But should it only show the amount for completed refunds or requested ones as well?

I'm not sure that seeing the refund status is useful when it's "completed" or "rejected". If a label has been completely refunded (or not refunded in the case of rejects), should it be included in the report at all?

classes/class-wc-connect-label-reports.php
marcinbot: Thanks, I corrected it in 4c4deea

jeffstieler:
> If the label had the refund rejected, I think it should be included, but not included if it was refunded or a refund was requested.

Just curious - what's your reasoning for this?

To me: the only case I know of where a refund is "rejected" is if the package was already shipped. That's as good as "not refunded" in my mind.

Considering legit refunds take 14+ days to process, it might be nice to know that you have a pending refund when you run the report.

classes/class-wc-connect-label-reports.php
marcinbot: No, it's `true` because we want to hide the sidebar. If it's missing or `false` it results in this:
screen shot 2018-03-27 at 14 41 45

classes/class-wc-connect-label-reports.php
marcinbot: :+1: Added in 1caf315

jeffstieler:
> I can see why it can be confusing for the users, especially if they don't visit this page often and the labels might disappear/appear there seemingly for no reason. Do you have any recommendations for how the refunds should be handled?

I think labels that have been refunded (completed) should be excluded from the report.

Labels that had their refunds rejected should be included in the report, but I think we should hide the "rejected" status in the refund column - it's not particularly useful here. (It's still an expenditure to be counted in this context)

Labels that are currently pending refunds should be included in the report, with the appropriate refund status column value. (The idea here being that it's up to the merchant if they'd like to include that label in their taxes, considering it's status and the lack of refund guarantee)

PR #1354, opened 2018-03-28
Send a no-cache header for WCS REST API responses.

jeffstieler:
Prompted by cache issues on the Pantheon hosting platform.

Employs Pantheon's recommended solution: https://pantheon.io/docs/cache-control/

To test:
* Load this code on Pantheon for most points 😛, but optionally hook in yourself like: https://github.com/pantheon-systems/WordPress/blob/master/wp-content/mu-plugins/pantheon/pantheon-page-cache.php#L278
* Verify that the `Cache-Control` header on all WCS API endpoints is `no-cache, must-revalidate, max-age=0`

### DIFF
--- a/classes/class-wc-rest-connect-base-controller.php
+++ b/classes/class-wc-rest-connect-base-controller.php
@@ -59,17 +59,44 @@ abstract class WC_REST_Connect_Base_Controller extends WP_REST_Controller {
 		}
 	}
 
-	public function get_internal( $request ) {
+	/**
+	 * Consolidate cache prevention mechanisms.
+	 */
+	public function prevent_route_caching() {
 		if ( ! defined( 'DONOTCACHEPAGE' ) ) {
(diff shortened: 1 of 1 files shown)


classes/class-wc-rest-connect-base-controller.php
```diff
@@ -59,17 +59,44 @@ public function register_routes() {
...
+		$this->prevent_route_caching();
+
 		return $this->get( $request );
 	}
 
 	public function post_internal( $request ) {
-		if ( ! defined( 'DONOTCACHEPAGE' ) ) {
-			define( 'DONOTCACHEPAGE', true ); // Play nice with WP-Super-Cache
-		}
+		$this->prevent_route_caching();
```
DanReyLop: None of this is needed for `POST` requests (not even the `DONOTCACHEPAGE` constant) because caches like Varnish ignore `POST` requests, but I'm fine leaving this just in case.

PR #1359, opened 2018-04-02
Remove extensionsMiddleware, use the plain wpcomApiMiddleware instead

DanReyLop:
I managed to fix this without touching `wp-calypso`.

**The bug:** In the `self-help` page, you can't interact with the debugging toggles.

**The cause:** We're using Calypso's `extensions-middleware` to load some `woocommerce`-related data-layer handlers. That middleware maintains a global state, so when it's called 2 times with different Redux stores, the first one will be overwritten. That means that the second Redux store (in this case, the `print-test-label` section) will receive all the Redux actions, including the ones that don't correspond with that "app".

**The solution:** Avoid `extensions-middleware`. Its only purpose is to allow extensions (which in Calypso are different chunks loaded on-demand) to register handlers **after** the Redux store is configured. We don't need that, since we know ahead of time which handlers we'll need for a given "app". I've changed it to use the plain `wpcom-api-middleware` instead.

PS: To test, try the `self-help` page and check that both sections (self-help and print-test-label) work. Also try opening the `Rates` step on the Buy Label modal and check that the user-selected rate appears correctly.

### DIFF
--- a/client/apps/shipping-label/index.js
+++ b/client/apps/shipping-label/index.js
@@ -15,12 +15,10 @@ import labelSettingsReducer from 'woocommerce/woocommerce-services/state/label-s
 import reduxMiddleware from './redux-middleware';
 import ordersReducer from 'woocommerce/state/sites/orders/reducer';
 import { combineReducers } from 'state/utils';
-import { addHandlers } from 'state/data-layer/extensions-middleware';
 import orders from 'woocommerce/state/data-layer/orders';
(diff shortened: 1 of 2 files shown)


DanReyLop:
> what do you think about renaming reduxMiddleware in the ShippingLabel route class since it's no longer the only middleware? (It's also not descriptive 🙂)

Meh, I think it's good enough. It's *the* middleware that's gonna have all kinds of handlers specific to that route, I'm ok with a generic name for now.

PR #1308, opened 2018-02-13
Fix waay too many automated tax requests

jeffstieler:
Fixes #1307.

Tax calculation was being kicked off even when core WC taxes were disabled, and on every admin page load. This was a side effect of having items in cart as an admin user, and by Jetpack's JITM request.

To test:
* Ensure that debug logging is enabled
* Ensure that you have items in a cart
* With tax calc **on** + automated taxes on - verify that no tax requests are logged while navigating the admin
* With tax calc **off** + automated taxes on - verify that no tax requests are logged while navigating the admin

(cc: @allendav)

PR #1361, opened 2018-04-02
Fix too many rate requests

### LINKED ISSUE #1360: Rates are calculated when WP-Admin and unrelated AJAX requests are received
To reproduce:
* log in as an admin
* add items to the cart
* make sure WooCommerce remembers a checkout address for your user that is a valid rates address
* open WP-Admin
* enable debug mode in `WooCommerce > Settings > Shipping > Shipping Options`
* rates request is sent to the server in the background

This is making the Store on WP.com almost unusable since the requests are timing out when requesting rates.

marcinbot:
Fixes #1360 

* Copied the solution from #1308 to skip the requests when not on the checkout page or the request is a Jetpack request
* ~~added rates response memory caching - requests can attempt to calculate rates more than once and our code would send the rates request every time for this~~

To test:
* enable debug mode in `WooCommerce > Settings > Shipping > Shipping Options` to ensure that rates are not cached
* WP Admin requests and non-checkout AJAX requests should not generate a rate request - this can be checked either by monitoring the server console output or placing a debug breakpoint in `class-wc-connect-shipping-method.php`
* checkout rates should display and update when cart contents/address are updated
* ~~only one rates request should be sent per WP request - added a debug message to monitor when a request has been retrieved from cache~~

### DIFF
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -262,6 +262,9 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 		}
 
 		public function calculate_shipping( $package = array() ) {
+			if ( ! WC_Connect_Functions::should_send_cart_api_request() ) {
+				return;
+			}
 
 			$this->debug( sprintf(
 				'WooCommerce Services debug mode is on - to hide these messages, turn debug mode off in the <a href="%s" style="text-decoration: underline;">settings</a>.',
--- a/classes/class-wc-connect-functions.php
+++ b/classes/class-wc-connect-functions.php
@@ -1,0 +1,23 @@
+<?php
+
+if ( ! class_exists( 'WC_Connect_Functions' ) ) {
+	class WC_Connect_Functions {
+		/**
+		 * Checks if the potentially expensive Shipping/Tax API requests should be sent
+		 * based on the context in which they are initialized
+		 * @return bool true if the request can be sent, false otherwise
+		 */
+		public static function should_send_cart_api_request() {
+			return ! (
+				// Skip for carts loaded from session in the dashboard
+				( is_admin() && did_action( 'woocommerce_cart_loaded_from_session' ) ) ||
+				// Skip during Jetpack API requests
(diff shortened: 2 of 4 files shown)


jeffstieler:
Can you provide more detailed testing instructions?

I'm unable to reproduce this on my local VVV setup, as well as a Poopy.life test install. Note: without automated taxes enabled.

I'll test with automated taxes on Poopy.life and report back.

classes/class-wc-connect-shipping-method.php
```diff
@@ -262,6 +297,25 @@ private function add_fallback_rate( $service_settings ) {
...
+
+			// Skip during REST API or XMLRPC requests
+			if ( defined( 'REST_REQUEST' ) || defined( 'REST_API_REQUEST' ) || defined( 'XMLRPC_REQUEST' ) ) {
+				return;
+			}
+
+			// Skip during Jetpack REST API proxy requests
+			if ( isset( $_GET['rest_route'] ) && isset( $_GET['_for'] ) && ( 'jetpack' === $_GET['_for'] ) ) {
+				return;
+			}
```
jeffstieler: Rather than directly copy-paste this code, can we move it somewhere it can be shared, avoiding duplication?

classes/class-wc-connect-shipping-method.php
```diff
@@ -301,37 +355,19 @@ public function calculate_shipping( $package = array() ) {
...
+			);
+			$response_body = wp_cache_get( $cache_key );
+			if ( false !== $response_body ) {
+				$this->debug( 'Rates response retrieved from cache' );
+			} else {
+				$response_body = $this->api_client->get_shipping_rates( $services, $package, $custom_boxes, $predefined_boxes );
+				if ( $this->check_and_handle_response_error( $response_body ) ) {
+					return;
+				}
+				wp_cache_set( $cache_key, $response_body, '', 3600 );
```
jeffstieler: I'd rather us separate the addition of `/wp-admin` cart detection and shipping rate request caching so they can be tested/vetted in isolation. Can you put these changes in to a new PR?

jeffstieler:
Forgot to comment - the changes work well to prevent unnecessary rates requests in the admin. There are some organizational changes I'd like to see made, as the inline comments indicate.

PR #1369, opened 2018-04-06
Status page: REST API check and toggle state error handling

dechov:
Add a health check to the status page, that (in contrast with the existing checks) is added by the client dynamically, to verify the ability to make REST API requests.

In addition, improve the state management of the toggles on this page (in case of REST API failure), by disabling controls while saving, and reverting the state if saving failed.

&nbsp; | Success | Failure
-- | -- | --
REST API check | rest-api-check-succeeded | rest-api-check-failed
Debug toggle | debug-toggle-succeeded | debug-toggle-failed

Also, if the failure was a 403 (which encodes a rejection of the request that is inexplicable by the client), show a notice:

screen shot 2018-04-06 at 7 06 28 pm

Can test failure cases by adding a filter (e.g. in `attach_hooks`):
```
add_filter( 'rest_authentication_errors', function() {
    $message = 'REST API is disabled.';
    return new WP_Error(
        'rest_api_disabled',
        $message,
        array( 'status' => 403, 'message' => $message )
    );
} );
```

Thoughts / concerns:
- Title and messages need editing
- Should we suggest checking the JS console in the event of failure?
- Should we check for `page=wc-status&tab=connect` in the request module, and omit Status page directions in this case?

### DIFF
--- a/client/api/request.js
+++ b/client/api/request.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import { translate as __ } from 'i18n-calypso';
+import React from 'react';
 import { startsWith, endsWith } from 'lodash';
 
 /**
@@ -51,9 +52,8 @@ const _request = ( url, data, method, namespace = '' ) => {
 
 			if ( 403 === response.status ) {
 				const message = __(
-					'The request was not fulfilled. ' +
-					'Please see the status page for diagnostics: ' +
-					'WooCommerce » Status » WooCommerce Services.'
+					'The request was not fulfilled. Please see the {{a}}status page{{/a}} for diagnostics.',
+					{ components: { a: <a href="?page=wc-status&tab=connect" /> } }
 				);
 				throw { data: { message } };
 			}
--- a/client/api/request.js
+++ b/client/api/request.js
@@ -53,7 +53,7 @@ const _request = ( url, data, method, namespace = '' ) => {
 			if ( 403 === response.status ) {
 				const message = __(
 					'The request was not fulfilled. Please see the {{a}}status page{{/a}} for diagnostics.',
-					{ components: { a: <a href="?page=wc-status&tab=connect" /> } }
+					{ components: { a: <a href="admin.php?page=wc-status&tab=connect" /> } }
 				);
 				throw { data: { message } };
 			}
--- a/client/api/request.js
+++ b/client/api/request.js
@@ -49,6 +49,15 @@ const _request = ( url, data, method, namespace = '' ) => {
 				return;
 			}
 
+			if ( 403 === response.status ) {
+				const message = __(
+					'The request was not fulfilled. ' +
+					'Please see the status page for diagnostics: ' +
+					'WooCommerce » Status » WooCommerce Services.'
+				);
+				throw { data: { message } };
+			}
+
 			throw json;
 		} );
 	} );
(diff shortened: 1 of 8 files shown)


client/api/request.js
```diff
@@ -49,6 +49,15 @@ const _request = ( url, data, method, namespace = '' ) => {
 				return;
 			}
 
+			if ( 403 === response.status ) {
+				const message = __(
+					'The request was not fulfilled. ' +
+					'Please see the status page for diagnostics: ' +
+					'WooCommerce » Status » WooCommerce Services.'
```
jeffstieler: Can we make this more friendly? Is it possible to link this to the status page?

client/apps/plugin-status/state/actions.js
```diff
@@ -32,16 +36,57 @@ export const save = () => ( dispatch, getState ) => {
...
+		wcc_debug_on: state.debug_enabled,
+		wcc_logging_on: state.logging_enabled,
+	};
+
+	api.post( api.url.selfHelp(), data )
+		.then( () => {
+			dispatch( {
+				type: PLUGIN_STATUS_REST_RESPONSE,
+				success: true,
+				message: translate( 'Can manage settings and print shipping labels' ),
```
jeffstieler: Seeing how we're only testing that the debug/logging saving here - it seems like a stretch to claim that settings and labels have been checked.

While it's better than no test, what do you think about changing this to be a bit more generic? Something like "WooCommerce REST API is working properly"?

dechov: Changed in https://github.com/Automattic/woocommerce-services/pull/1369/commits/85f931d70106d1209b632538f90211055d9b8bfd to "WooCommerce REST API is working properly" vs "Encountered an issue with the WooCommerce REST API".

client/api/request.js
jeffstieler: Ping @dechov - let's change this to "The request was not fulfilled. Please see the [status page](/wp-admin/admin.php?page=wc-status&tab=connect) for diagnostics." (with the correct URL)

client/api/request.js
```diff
@@ -51,9 +52,8 @@ const _request = ( url, data, method, namespace = '' ) => {
 
 			if ( 403 === response.status ) {
 				const message = __(
-					'The request was not fulfilled. ' +
-					'Please see the status page for diagnostics: ' +
-					'WooCommerce » Status » WooCommerce Services.'
+					'The request was not fulfilled. Please see the {{a}}status page{{/a}} for diagnostics.',
+					{ components: { a: <a href="?page=wc-status&tab=connect" /> } }
```
jeffstieler: Do we not need to specify `admin.php` here?

Ref: https://github.com/Automattic/woocommerce-services/blob/04e81b836abe0fd01c5e707be5a251dbcacd37d3/client/apps/plugin-status/health.js#L21

dechov:
Rebased to resolve conflict.

client/api/request.js
dechov: Fixed in https://github.com/Automattic/woocommerce-services/pull/1369/commits/800aed16a7dfee1d89d23317d339a000ced1ff54

dechov:
@jeffstieler Not sure why that would be or how I'd reproduce that. Did any trace info come with that error output?

jeffstieler:
> Not sure why that would be or how I'd reproduce that. Did any trace info come with that error output?

For posterity - re: our Slack convo - I saw this error in the console on non-Status pages in the admin. (like WooCommerce > Settings > WooCommerce Services)

PR #1370, opened 2018-04-06
Error message on manual service data refresh failure

dechov:
screen shot 2018-04-06 at 6 40 24 pm

Test by adding `define( 'WP_HTTP_BLOCK_EXTERNAL', true );` to wp-config, and click 'Refresh' on status page under "WooCommerce Services Data".

While this is an improvement over no indication of the refresh having failed, what could make this message more informative or actionable?

### DIFF
--- a/classes/class-wc-connect-help-view.php
+++ b/classes/class-wc-connect-help-view.php
@@ -308,8 +308,8 @@ if ( ! class_exists( 'WC_Connect_Help_View' ) ) {
 		 */
 		public function page() {
 			if ( isset( $_GET['refresh'] ) && 'true' === $_GET['refresh'] ) {
-				$this->service_schemas_store->fetch_service_schemas_from_connect_server();
-				$url = remove_query_arg( 'refresh' );
+				$fetched = $this->service_schemas_store->fetch_service_schemas_from_connect_server();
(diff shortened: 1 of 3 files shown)


jeffstieler:
How hard would it be to display the error inline with the `refresh` control?

The admin notices all blue together..

screen shot 2018-04-20 at 5 48 56 pm

dechov:
@jeffstieler Done:

screen shot 2018-04-23 at 5 54 43 pm

jeffstieler:
Seeing a success icon and error notice side by side is a bit confusing to me. Looping in @kellychoffman for some design feedback here.

Should we perhaps hide the "up-to-date" success message if the user manually attempts a refresh and it fails? Mock up:

screen shot 2018-05-08 at 12 08 46 pm

kellychoffman:
> Seeing a success icon and error notice side by side is a bit confusing to me.

Agree. 

> Should we perhaps hide the "up-to-date" success message if the user manually attempts a refresh and it fails? Mock up:

Smart. Let's go with that.

dechov:
Nice and simple:

screen shot 2018-05-23 at 12 17 06 am

It depends directly on the query parameter, so when navigating to this page again, the refresh error will be forgotten.

Issue #1376, opened 2018-04-14
Renovate "An error occurred in WooCommerce Services" admin notice


dechov:
This what appears on all admin pages after an "error" has been logged — which is currently [only](https://github.com/Automattic/woocommerce-services/search?utf8=%E2%9C%93&q=%22logger-%3Eerror%22&type=) done for [shipping](https://github.com/Automattic/woocommerce-services/blob/856e19b049b22d6b2da7ea454d71568f68b9589c/classes/class-wc-connect-shipping-method.php#L156) and [tax](https://github.com/Automattic/woocommerce-services/blob/32a1f6bc5cb0f406b17f76392c722d83333cc7b9/classes/class-wc-connect-taxjar-integration.php#L277) rate retrieval/calculation errors:

<img width="607" alt="screen shot 2018-04-14 at 1 21 51 pm" src="https://user-images.githubusercontent.com/1867547/38771013-74aa4146-3fe9-11e8-99ff-62e2f51b496f.png">

How can this notice be improved upon to be more informative? Currently, it links to the status page (indirection level 1), where the logs can be found but only once you scroll past the other status views (indirection level 2). Note that plugin version 1.12 has a bug in which the message could appear without logs being written, but it is fixed in master and unrelated to this issue.

Also note: it seems in practice there is a quite a typical scenario in which this notice will appear, which is missing product weight and/or dimensions — https://github.com/Automattic/woocommerce-services/pull/1288 seeks to preempt this case, while https://github.com/Automattic/woocommerce-services/pull/1303 made this error accessible from the front-end, making the admin notice relatively less useful.

---

Lower priority, but another possible improvement is to the dismiss button triggering a page load:

<img width="701" alt="screen shot 2018-04-14 at 1 22 42 pm" src="https://user-images.githubusercontent.com/1867547/38771015-7720ec5e-3fe9-11e8-85d0-59ac7b2d515f.png">

kellychoffman:
I ran into this today when setting up a test site. Took me a while to figure out what was happening. Turns out it was because my product didn't have dimensions and weight. We need to fix this problem more holistically (ie. someone shouldn't be able to save a physical product without adding weight and dimensions), but a quick win would be to surface that particular problem in the error message itself. Right now, you have to dig through the logs to see that weight and dimensions need to be added.

madeincosmos:
[...]

PR #1380, opened 2018-04-20
Surface missing weight error directly in admin notice

dechov:
Addresses https://github.com/Automattic/woocommerce-services/issues/1376#issuecomment-382153745 by enhancing the admin notice — in the case of a shipping rate missing weight error — with a description and link to the relevant product:

screen shot 2018-04-20 at 7 05 06 pm

Questions:
- Should this code live somewhere other than `WC_Connect_Error_Notice`?
- Should clicking the link dismiss the notice?
- Are there other errors that should get a similar treatment?

### DIFF
--- a/classes/class-wc-connect-error-notice.php
+++ b/classes/class-wc-connect-error-notice.php
@@ -60,7 +60,7 @@ if ( ! class_exists( 'WC_Connect_Error_Notice' ) ) {
 				$product_name = WC_Connect_Compatibility::instance()->get_product_name( wc_get_product( $product_id ) );
 
 				$message = sprintf(
-					__( 'An error occurred in WooCommerce Services: <strong><a href="%s">%s</a> did not include a weight.</strong> Shipping rates cannot be calculated.', 'woocommerce-services' ),
+					__( '<strong>%2$s does not have have a weight defined.</strong><br />Shipping rates cannot be calculated. <a href="%1$s">Add a weight for %2$s</a> so your customers can purchase this item.', 'woocommerce-services' ),
 					get_edit_post_link( $product_id ), $product_name
 				);
 			} else {
--- a/classes/class-wc-connect-error-notice.php
+++ b/classes/class-wc-connect-error-notice.php
@@ -55,9 +55,8 @@ if ( ! class_exists( 'WC_Connect_Error_Notice' ) ) {
 
 			$error = $this->notice_enabled();
 			if ( is_wp_error( $error ) && 'product_missing_weight' === $error->get_error_code() ) {
-				$matches = array();
-				preg_match( '/ID:\s([0-9]+)/', $error->get_error_message(), $matches );
-				$product_id = $matches[1];
+				$error_data = $error->get_error_data();
+				$product_id = $error_data['product_id'];
 				$product_name = WC_Connect_Compatibility::instance()->get_product_name( wc_get_product( $product_id ) );
 
 				$message = sprintf(
--- a/classes/class-wc-connect-error-notice.php
+++ b/classes/class-wc-connect-error-notice.php
@@ -75,10 +75,16 @@ if ( ! class_exists( 'WC_Connect_Error_Notice' ) ) {
 					$link_status, $link_dismiss
 				);
 			}
+
+			$allowed_html = array(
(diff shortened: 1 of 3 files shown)


kellychoffman:
How about this instead: 

**Red Shirt does not have have a weight defined.** 
Shipping rates cannot be calculated. [Add a weight for Red Shirt](#) so your customers can purchase this item. 

> Should clicking the link dismiss the notice?

Would it reappear if you didn't fix the weight?

dechov:
Sounds good – will change the text.

> Would it reappear if you didn't fix the weight?

Yup, it would reappear when a rate is requested with that product again. As it currently works, only one WCS error notice will appear at a time, so if there was also a tax issue or multiple shipping issues, they would overwrite each other. Errors other than missing weight will still show up with the same admin notice as before (i.e. "An error occurred in WooCommerce Services" with a link.)

kellychoffman:
> Yup, it would reappear when a rate is requested with that product again

Awesome. Thanks for confirming.

DanReyLop:
> Should clicking the link dismiss the notice?

I think the better UX here would be to only show the notice if the weight is still a problem. That is, at the moment of displaying it, check if the `product_id` has weight. If it does, just dismiss the notice server-side.

kellychoffman:
Agree. (I actually thought that was what Paul meant)

dechov:
Thanks — done and done:

screen shot 2018-04-30 at 6 17 21 pm

Disappears once the product is saved with a weight.

Side-note: due to WooCommerce treating the weight as a string, a weight of `0` _is not_ considered having a weight, while a weight of `0.0` _is_ considered having a weight. Not sure if this is a problem at all, and also not sure if the WC products table work will affect this.

classes/class-wc-connect-error-notice.php
```diff
@@ -53,10 +53,23 @@ private function show_notice() {
...
 			$link_dismiss = add_query_arg( array( 'wc-connect-error-notice' => 'disable' ) );
 
-			$message = sprintf(
-				__( 'An error occurred in WooCommerce Services. Details are logged <a href="%s">here</a>.', 'woocommerce-services' ),
-				$link_status, $link_dismiss
-			);
+			$error = $this->notice_enabled();
+			if ( is_wp_error( $error ) && 'product_missing_weight' === $error->get_error_code() ) {
+				$matches = array();
+				preg_match( '/ID:\s([0-9]+)/', $error->get_error_message(), $matches );
```
jeffstieler: Can we avoid the regexp by adding data to the `WP_Error` here: https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-api-client.php#L97 ?

classes/class-wc-connect-error-notice.php
```diff
@@ -53,10 +53,23 @@ private function show_notice() {
...
+				$product_id = $matches[1];
+				$product_name = WC_Connect_Compatibility::instance()->get_product_name( wc_get_product( $product_id ) );
+
+				$message = sprintf(
+					__( 'An error occurred in WooCommerce Services: <strong><a href="%s">%s</a> did not include a weight.</strong> Shipping rates cannot be calculated.', 'woocommerce-services' ),
```
jeffstieler: Let's use `wp_kses()` to restrict what HTML we allow in the translated string, like: https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-paypal-ec.php#L135

dechov:
Fixed in
https://github.com/Automattic/woocommerce-services/commit/dfa10e4601a22c9d6e2190e3b84fc7df45045956

classes/class-wc-connect-error-notice.php
dechov: This was fixed in https://github.com/Automattic/woocommerce-services/commit/dfa10e4601a22c9d6e2190e3b84fc7df45045956

classes/class-wc-connect-error-notice.php
dechov: Done in https://github.com/Automattic/woocommerce-services/pull/1380/commits/3acb4d389ba63d88e21e6b7c830051bc603e4b05

PR #1382, opened 2018-04-24
Add support for international shipping labels

DanReyLop:
This PR improves the experience of purchasing international shipping labels. Note that, by design, this PR is **not** required for the feature to work on Calypso. Both Calypso and the server are backwards-compatible, meaning that if the merchant hasn't updated their plugin in a while, they will still be able to purchase international shipping labels, although the experience won't be optimal.

In this PR:
* Send the `selected_payment_method_id` when requesting label rates. Before, this was only required for the actual purchase, but now the server will get the merchant's legal name from the selected credit card when building the shipment object.
* Show the "Purchase Label" metabox if the merchant's country has a USPS office 8that includes the US and many territories like Puerto Rico or American Samoa). Any country in the world is accepted as a destination too.
* Return all the countries / states to the client instead of only US & PR, so it can show a proper dropdown.
* For each item on the order, retrieve and return metadata required for the Customs form, if any.
* When requesting label rates, the same way the origin/destination addresses are stored for future convenience, the per-product customs metadata (description, tariff code, and origin country) are stored on `postmeta` so they'll be used for subsequent orders.

This PR can't be tested by itself, it can only be tested through Calypso.

TO DO: Update the `wp-calypso` version dependency once the Calypso part of the feature is merged to `master`. Don't merge this PR before that.


classes/class-wc-connect-compatibility-wc26.php
```diff
@@ -116,6 +116,22 @@ public function get_product_name_from_order( $product_id, $order ) {
...
+		 */
+		public function get_product_price_from_order( $product_id, $order ) {
+			foreach ( $order->get_items() as $line_item ) {
+				if ( (int) $line_item[ 'product_id' ] === $product_id || (int) $line_item[ 'variation_id' ] === $product_id ) {
+					return round( floatval( $line_item[ 'total' ] ) / $line_item[ 'qty' ], 2 );
+				}
+			}
+			return null;
+		}
+
```
jeffstieler: I think we agreed to stop supporting WC 2.x after 3.4 was released?

classes/class-wc-connect-compatibility-wc30.php
```diff
@@ -130,6 +130,33 @@ public function get_product_name_from_order( $product_id, $order ) {
...
+		 * @return float The product (or variation) price, or NULL if it wasn't found
+		 */
+		public function get_product_price_from_order( $product_id, $order ) {
+			foreach ( $order->get_items() as $line_item ) {
+				$line_product_id = $line_item->get_product_id();
+				$line_variation_id = $line_item->get_variation_id();
+
+				if ( ! $line_product_id ) {
+					$line_product_id = (int) get_metadata( 'order_item', $line_item->get_id(), '_product_id', true );
+				}
```
jeffstieler: Why are these "backup" retrievals necessary?

classes/class-wc-connect-shipping-label.php
```diff
@@ -202,9 +175,18 @@ public function get_selected_packages( WC_Order $order ) {
 							$formatted = WC_Connect_Compatibility::instance()->get_formatted_variation( $product, true );
 							$product_data['attributes'] = $formatted;
 						}
+						$customs_info = get_post_meta( $product_data['product_id'], 'customs_info', true );
```
jeffstieler: We should probably use a more unique meta_key, perhaps `wc_connect_customs_info`?

classes/class-wc-connect-shipping-label.php
```diff
@@ -202,9 +175,18 @@ public function get_selected_packages( WC_Order $order ) {
...
+						if ( $customs_info ) {
+							$product_data = array_merge( $product_data, $customs_info );
+						}
 					} else {
 						$product_data['name'] = WC_Connect_Compatibility::instance()->get_product_name_from_order( $product_data['product_id'], $order );
 					}
+					$product_data['value'] = WC_Connect_Compatibility::instance()->get_product_price_from_order( $product_data['product_id'], $order );
+					if ( ! $product_data['value'] ) {
```
jeffstieler: Should this be a `null` check? Also.. if the `null` case isn't used by the callers, why not default to `0`?

classes/class-wc-connect-shipping-label.php
```diff
@@ -202,9 +175,18 @@ public function get_selected_packages( WC_Order $order ) {
...
+						$product_data['value'] = 0;
+					}
+					$product_data['parent_product_id'] = ( $product && $product->is_type( 'variation' ) ) ? $product->get_parent_id() : $product_data['product_id'];
```
jeffstieler: This can use `WC_Connect_Compatibility::instance()->get_parent_product_id()`. Also - where is the `parent_product_id` used?

classes/class-wc-rest-connect-shipping-rates-controller.php
```diff
@@ -18,13 +18,38 @@ class WC_REST_Connect_Shipping_Rates_Controller extends WC_REST_Connect_Base_Con
...
+		// Update the customs information on all this order's products
+		$updated_product_ids = array();
+		foreach ( $payload[ 'packages' ] as $package ) {
+			if ( ! isset( $package[ 'contents_type' ] ) ) {
+				break; // No customs information in this order, bail
+			}
+			foreach ( $package[ 'items' ] as $item ) {
+				if ( ! isset( $updated_product_ids[ $item[ 'product_id' ] ] ) ) {
+					$updated_product_ids[ $item[ 'product_id' ] ] = true;
+					update_post_meta( $item[ 'product_id' ], 'customs_info', array(
```
jeffstieler: Same note about meta key.

classes/class-wc-rest-connect-shipping-rates-controller.php
```diff
@@ -18,13 +18,38 @@ class WC_REST_Connect_Shipping_Rates_Controller extends WC_REST_Connect_Base_Con
...
+						'hs_tariff_code' => $item[ 'hs_tariff_code' ],
+						'origin_country' => $item[ 'origin_country' ],
+					) );
+				}
+			}
+		}
+		foreach ( $payload[ 'packages' ] as $package_id => $package ) {
+			foreach ( $package[ 'items' ] as $index => $item ) {
+				// The server doesn't require the "product_id property
+				unset( $payload[ 'packages' ][ $package_id ][ 'items' ][ $index ][ 'product_id' ] );
```
jeffstieler: Could this be incorporated into the loop above?

Also - is this necessary because of https://github.com/Automattic/wp-calypso/pull/24974/files#diff-09d52b7e14fcd8c2a349cabb5e073666R215 ?

classes/class-wc-connect-compatibility-wc26.php
DanReyLop: Yes, we did (I started this before 3.4 was released haha).

Do you think we should stop supporting it right now, or should we do a release where we explicitly remove support? (May be reasonable to do a `2.0.0` version in that case).

classes/class-wc-connect-compatibility-wc30.php
DanReyLop: I just copied the logic from `get_product_name_from_order()`, I assume it was like this for a good reason :/

classes/class-wc-connect-shipping-label.php
DanReyLop: Agreed, changed in https://github.com/Automattic/woocommerce-services/pull/1382/commits/f9e4f28cbf5117c8e08a2206b335020603063cf2

classes/class-wc-connect-shipping-label.php
DanReyLop: It's not used anymore. It was used to display a link to each product in the Customs form, but that column is no longer present in the UI. I've removed this in https://github.com/Automattic/woocommerce-services/pull/1382/commits/3920357bd3baa7236d74be395735db1f9e5635c1

classes/class-wc-connect-shipping-label.php
DanReyLop: I've changed the check to be more explicit in https://github.com/Automattic/woocommerce-services/pull/1382/commits/02adfbffebef69866a3d68a2b21b4d1258e99d6e
I still think defaulting to `null` makes sense for the `get_product_price_from_order` function, since `0` is a valid price. The fact that in this case, if the price isn't founf it defaults to `0`, it should be part of the business logic, not the "compatibility" layer.

classes/class-wc-rest-connect-shipping-rates-controller.php
DanReyLop: > Could this be incorporated into the loop above?

I found it a bit clearly to split it into 2 loops because in the first loop there's some logic to bail early if the form doesn't contain customs information.

> Also - is this necessary because of https://github.com/Automattic/wp-calypso/pull/24974/files#diff-09d52b7e14fcd8c2a349cabb5e073666R215 ?

Yes, Calypso needs to send back the `product_id` so the `meta` properties of the product can be updated (`tariff code`, `origin country` and `description`). But the WCS server doesn't care about that (in fact, it will fail `joi` validation if `product_id` is present).

v18:
> TO DO: Update the wp-calypso version dependency once the Calypso part of the feature is merged to master. Don't merge this PR before that.

@DanReyLop - can you explain a bit why this PR can't be merged before Calypso's feature is fully merged? The Calypso feature requires for the changes in this PR in order to display the full country list, so if we don't merge these plugin changes first, then we'll be in a 'backwards compatibility' state (e.g. showing only the country code in the drop-down) until a new plugin is released.

DanReyLop:
> The Calypso feature requires for the changes in this PR in order to display the full country list, so if we don't merge these plugin changes first, then we'll be in a 'backwards compatibility' state (e.g. showing only the country code in the drop-down) until a new plugin is released.

The idea is to do things in this order:
- Merge the `woocommerce-connect-server` PR, it's fully backwards-compatible.
- Merge the `wp-calypso` PRs to `master`, with the feature flag disabled. `wordpress.com` won't be stuck in "backwards-compatibility mode" because the only countries that matter at this point are US and Puerto Rico, and those are already returned by the plugin.
- Update the `wp-calypso` dependency in this PR (or in a different PR, I don't care as long as we don't do a plugin release in the meantime). Since the JS code in the plugin doesn't respect feature flags, it will have all the international bells and whistles enabled.
- At this point, we can cut a new version of the plugin, with international labels enabled.
- Wait for 1 day or so, so we see if there are any catastrophic bugs, and to let the Atomic sites auto-update to the new version of the plugin.
- Enable the feature flag in `wp-calypso`.

The only instance when a Calypso merchant will get the hacky "backwards-compatibility experience" is if, after going through all those steps, including enabling the feature flag in Calypso, his/her site hasn't updated yet to the newest `woocommerce-services` version.

Does it sound reasonable?

v18:
Wanted to mention that the work in Calypso will be refactored a bit, which will affect this PR going forward:

https://github.com/Automattic/wp-calypso/pull/24976#issuecomment-394189175

cc @jeffstieler

jeffstieler:
I'm getting these errors trying create a label for an existing domestic order:

`PHP Notice:  Undefined index: items in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/classes/class-wc-rest-connect-shipping-rates-controller.php on line 48`

`PHP Warning:  Invalid argument supplied for foreach() in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-services/classes/class-wc-rest-connect-shipping-rates-controller.php on line 48`

DanReyLop:
> I'm getting these errors trying create a label for an existing domestic order

Argh, turns out you were right when requesting https://github.com/Automattic/woocommerce-services/pull/1382#discussion_r190767164 , since the `items` property is only submitted when the order has customs information.

Fixed in https://github.com/Automattic/woocommerce-services/pull/1382/commits/9d595cc02634d40433688de205bb87ee1d89784a

classes/class-wc-connect-compatibility-wc26.php
jeffstieler: Fine by me to isolate the back-compat changes. 👍

classes/class-wc-connect-compatibility-wc30.php
jeffstieler: Alright - it would be great if we knew why it was done in the first place - does this jog your memory at all? https://github.com/Automattic/woocommerce-services/commit/ec4e72a72fde21789a5570bd23237d1ddcf09907

DanReyLop: Wow, it was me... plot twist!
Seriously, I don't have any recollection of that code. But I wouldn't have added that fallback if I hadn't encountered a case where it made sense. I can't find what it could possibly be now.

v18:
@DanReyLop - hey are you able to `npm install` using this branch? I'm getting a bunch of errors

jeffstieler:
> are you able to `npm install` using this branch? I'm getting a bunch of errors

I just succesfully ran: `rm -rf node_modules && npm install && npm run dist`

jeffstieler:
@DanReyLop the code changes look good here.. but I'd like to test all of the functionality (including the Calypso changes) on my local setup using `npm link`.. is that too much to put together now?

I tried my hand at it briefly, had to install a lot of npm packages.. and ultimately the label metabox contents won't render. 😕

jeffstieler:
@DanReyLop I ran into the same issues with your method .. lots of missing modules, etc.

v18:
> the code changes look good here

@jeffstieler - just FYI some changes are still required for the countries data ([current code is here](https://github.com/Automattic/woocommerce-services/pull/1382/files#diff-cc0cefd4ac82fa521d8d0ed20fc65bbbR307)). In Calypso, we'll be using the `v3/continents` endpoint. However, this is not available to users not hosted on WordPress.com yet. So, I think we'll be adding an endpoint in our plugin if it's missing, and remove the code mentioned above.

v18:
> rm -rf node_modules && npm install && npm run dist

@DanReyLop @jeffstieler what version of npm are you using when `npm install` works?

DanReyLop:
@v18 My bad, this branch hadn't been rebased with `master` in a looong time, even before upgrading to `npm 6`. It should be fine now.

`npm --version` should be `6.1.0`.

@jeffstieler After reinstalling everything (`rm -rf node_modules && npm install`, same in `wp-calypso`), `npm link ../wp-calypso` is working for me just fine now. In case it doesn't, I tried the "nuclear option" that I posted before and it works too.

v18:
After pulling in the latest changes on this branch, I was able to get it installed with `node v10.4.0` and `npm v6.1.0`.

classes/class-wc-connect-compatibility-wc30.php
```diff
@@ -130,6 +130,33 @@ public function get_product_name_from_order( $product_id, $order ) {
 			return sprintf( __( '#%d - [Deleted product]', 'woocommerce-services' ), $product_id );
 		}
 
+		/**
+		 * For a given product ID, it tries to find its price inside an order's line items.
+		 *
+		 * @param int $product_id Product ID or variation ID
+		 * @param WC_Order $order
+		 * @return float The product (or variation) price, or NULL if it wasn't found
```
v18: What are the circumstances in which the product price is not found?

DanReyLop: If the item was removed from the order, for example. Super-edge case, but it's a possibility.

v18: Doesn't line_items have the latest items in the order?

v18:
@DanReyLop you mentioned to me that this PR includes saving the tariff code for each product. I noticed that this doesn't happen when a user confirms the customs information in the customs step, only later. Does it make sense to add a separate endpoint to save the customs info?

DanReyLop:
> I noticed that this doesn't happen when a user confirms the customs information in the customs step, only later.

It happens when the `Rates` are fetched, and that should happen immediately after the `Customs` step is confirmed. Isn't it working like that for you?

classes/class-wc-rest-connect-shipping-rates-controller.php
```diff
@@ -18,13 +18,34 @@ class WC_REST_Connect_Shipping_Rates_Controller extends WC_REST_Connect_Base_Con
...
+		// Update the customs information on all this order's products
+		$updated_product_ids = array();
+		foreach ( $payload[ 'packages' ] as $package_id => $package ) {
+			if ( ! isset( $package[ 'contents_type' ] ) ) {
+				break; // No customs information in this order, bail
+			}
+			foreach ( $package[ 'items' ] as $index => $item ) {
+				if ( ! isset( $updated_product_ids[ $item[ 'product_id' ] ] ) ) {
+					$updated_product_ids[ $item[ 'product_id' ] ] = true;
+					update_post_meta( $item[ 'product_id' ], 'wc_connect_customs_info', array(
```
v18: Should we use or hook into the update_options work in https://github.com/Automattic/woocommerce-services/blob/6158b29c0dc3af4f5964942e507d2febf02ee143/classes/class-wc-connect-options.php so that these can be cleaned up neatly during uninstall?

classes/class-wc-connect-shipping-label.php
```diff
@@ -209,9 +175,17 @@ public function get_selected_packages( WC_Order $order ) {
 							$formatted = WC_Connect_Compatibility::instance()->get_formatted_variation( $product, true );
 							$product_data['attributes'] = $formatted;
 						}
+						$customs_info = get_post_meta( $product_data['product_id'], 'wc_connect_customs_info', true );
```
v18: I ran into an issue with this: When I add a tariff number for a product, I see it saved in the db in post_meta, but this information is not loaded on subsequent label purchases with that product. This happens on both Calypso and wp-admin.

v18:
> It happens when the Rates are fetched, and that should happen immediately after the Customs step is confirmed. Isn't it working like that for you?

It's working for me 👌 thanks!

classes/class-wc-rest-connect-shipping-rates-controller.php
v18: Ok! Sounds good to me. Great point about not wanting to clean up the customs info anyway!

classes/class-wc-connect-shipping-label.php
DanReyLop: I can't reproduce that bug. Are you sure the product was the same? The tariff number is per-variation, not per-product (because people use variations for all sorts of things, it's conviebable that different variations of the same product have a different code).

v18: I think I see why - the customs info only gets added if there are packages returned from `get_packaging_metadata()` in https://github.com/Automattic/woocommerce-services/blob/ec8db8844c05f33d44d66b9de2077bc48727c839/classes/class-wc-connect-shipping-label.php#L142

If there are none, then we send the default box, where the customs info isn't added:

https://github.com/Automattic/woocommerce-services/blob/ec8db8844c05f33d44d66b9de2077bc48727c839/classes/class-wc-connect-shipping-label.php#L147

So, we should add the customs info in there too.

v18:
The tariff numbers are supposed to be per product variation, not product. However, the API endpoint `connect/label/${ orderId }` isn't sending the product variation ID in `selected_packages` in 

https://github.com/Automattic/woocommerce-services/blob/ec8db8844c05f33d44d66b9de2077bc48727c839/classes/class-wc-connect-shipping-label.php#L141

We need to make sure to add the `variation_id` for the `$items` array for both of these cases in `get_selected_packages`;

1. When the items in an order are packaged (e.g. the customer used WCS for checkout rates)
2. When the items in the order are not packaged

From my initial reading, it looks like we are sometimes relying on `WC_Product_Variation` objects instead of `WC_Order_Item_Product` objects. This is problematic since deleted products and products with discounts in the order will not be handled correctly, and we should rely on the order line items to make sure that we're using the correct information about each product as it exists in the order.

This happens, for example, in in https://github.com/Automattic/woocommerce-services/blob/ec8db8844c05f33d44d66b9de2077bc48727c839/classes/class-wc-connect-shipping-label.php#L167

classes/class-wc-connect-shipping-label.php
```diff
@@ -209,9 +175,17 @@ public function get_selected_packages( WC_Order $order ) {
...
+						if ( $customs_info ) {
+							$product_data = array_merge( $product_data, $customs_info );
+						}
 					} else {
 						$product_data['name'] = WC_Connect_Compatibility::instance()->get_product_name_from_order( $product_data['product_id'], $order );
 					}
+					$product_data['value'] = WC_Connect_Compatibility::instance()->get_product_price_from_order( $product_data['product_id'], $order );
```
v18: We need to make sure to send the value even when `packages` is not defined. More details:

https://github.com/Automattic/woocommerce-services/issues/1465

classes/class-wc-rest-connect-shipping-rates-controller.php
```diff
@@ -18,13 +18,32 @@ class WC_REST_Connect_Shipping_Rates_Controller extends WC_REST_Connect_Base_Con
...
 
 		// This is the earliest point in the printing label flow where we are sure that
 		// the merchant wants to ship from this exact address (normalized or otherwise)
 		$this->settings_store->update_origin_address( $payload[ 'origin' ] );
 		$this->settings_store->update_destination_address( $order_id, $payload[ 'destination' ] );
 
+		// Update the customs information on all this order's products
+		$updated_product_ids = array();
+		foreach ( $payload[ 'packages' ] as $package_id => $package ) {
+			if ( ! isset( $package[ 'contents_type' ] ) ) {
```
v18: Relying on `contents_type` seems a bit arbitrary, similar to using a magic number. Would be cool to extract this out to a constant or function like `packageNeedsCustoms()`. That way, the comment in the line below (`// No customs information in this order, bail) would not be needed, since the function will be self-documenting.

DanReyLop:
> This is problematic since deleted products and products with discounts in the order will not be handled correctly, and we should rely on the order line items to make sure that we're using the correct information about each product as it exists in the order.

@v18 Let's consider "deleted products" an edge case that we won't handle for now. Even if we want to handle it before release, it shouldn't block merging this PR and testing the feature in general. As I said in Slack, the order line items don't have enough information to fill the Customs form. A more detailed analysis:

We have currently 3 "sources of truth" for items in an order: packaging metadata (only stored if the user purchased an WCS shipping method at checkout), the order line items, and the products stored in the database. If a customer chose Free Shipping (for example), there's no packaging metadata. If the merchant deleted the product from their store, it's not in the database. If the merchant edited the order, then the product may not be in the order line items. If 2 out of these 3 sources are present, we can get all the information required, but if 2 sources are missing, it can't be done.

> We need to make sure to add the variation_id for the $items array for both of these cases in get_selected_packages

Apart from the case where there's no packaging data (which is a legitimate bug), we are returning the product variation already. This is, in fact, the variation ID:
https://github.com/Automattic/woocommerce-services/blob/3cad87ad6e1e8d34eb820187db0c239ce6352422/classes/class-wc-connect-shipping-label.php#L67
And the `parent_product_id` is the product ID. We use `product_id` assuming it's the product ID for simple products **or** variation ID for variable products all through the app, and if you find otherwise then it's a bug.

classes/class-wc-rest-connect-shipping-rates-controller.php
DanReyLop: Done in https://github.com/Automattic/woocommerce-services/pull/1382/commits/73be1a127ea1375ffb868e10aed3139db799b926

DanReyLop:
> Apart from the case where there's no packaging data (which is a legitimate bug), we are returning the product variation already. This is, in fact, the variation ID. [...] We use product_id assuming it's the product ID for simple products or variation ID for variable products all through the app, and if you find otherwise then it's a bug.

Hah! Turns out that was bs. `$item['product_id']` was in fact the product ID, but that code was only running when there is no packaging metadata, so that threw me off balance. Fixed in https://github.com/Automattic/woocommerce-services/pull/1382/commits/b5c39573538223d29d64d43b40068c33c99796da

classes/class-wc-connect-shipping-label.php
DanReyLop: Fixed in https://github.com/Automattic/woocommerce-services/pull/1382/commits/806e5820ba6c95ff24e2ff76b98e46e099f7c377 (I'm glad some bugs are actually easier than expected to fix)

classes/class-wc-connect-shipping-label.php
DanReyLop: Fixed in 806e582, thanks for hunting this bug :)

v18:
> [deleted products] shouldn't block merging this PR and testing the feature in general

agreed 👍

v18:
> If the merchant edited the order, then the product may not be in the order line items

What makes you say this? The line_items will contain the current order information.

v18:
> 3 "sources of truth" for items in an order: packaging metadata (only stored if the user purchased an WCS shipping method at checkout), the order line items, and the products stored in the database

For everything except weight and dimensions, why not use the data from `line_items`? So that we only have one source of truth for that data.

v18:
> Fixed in b5c3957

Works for me. Just needs some UI to let the user know we're not going to able to get them labels for their deleted products.

DanReyLop:
> If the merchant edited the order, then the product may not be in the order line items
> > What makes you say this? The line_items will contain the current order information.

What I mean is that if the user deleted some lines from the order, then those deleted lines are no longer there. It's silly since the order line items are "the ultimate truth", but I was just reasoning why we need at least 2 of the 3 "sources of truth" to have the product we're looking for.

> I spoke with @claudiulodro and I found out that if a product is deleted, then we won't be able to gets its weight or dimensions from line_items. So - in order to support deleted products, we need to get their weight and dimensions by asking the user. We can not support deleted products for v1. cc @jeffstieler, we also talked about this.

Yes, exactly. The exception to that is if we have packaging metadata (i.e. a second source of truth), because there we store the dimensions and weight.

> For everything except weight and dimensions, why not use the data from line_items? So that we only have one source of truth for that data.

Yes, that's a reasonable change. I implemented some logic in the client before I realized it was pointless, I just need to port it to PHP:
* Take the order line items, store the items in a map of the form `{ variationId: { name, quantity, totalValue }, ... }`
* For each variation, get its "declared value" by dividing `totalValue / quantity`, that should cancel out any noise of "buy 1, get 1 free" offers (I'm not so clear on this part).
* Take the packaging metadata (if any), iterate over all of the packages and all of their items, decreasing the `productsMap[variationId].quantity` field for every one. If that `quantity` field is already `0`, remove the item from the package instead.
* After this, if the merchant removed some items from the order, there won't be in the packaging metadata, and some of the boxes will be half-empty instead.
* For all the products that were extracted from the `line items` and `quantity > 0`, add them to the packaging metadata in individual boxes. For those items, we'll need to query the store's database to get their weight & dimensions.
* (Nice to have) If the packaging metadata was at all changed during this process, send a flag to the client so the Packages step can auto-expand and maybe show a notice.

classes/class-wc-connect-shipping-label.php
v18: The fix works 👍

DanReyLop:
I've squashed and merged the `v3/continents` PR (https://github.com/Automattic/woocommerce-services/pull/1448) into this one. Please keep it in mind so you folks don't review the same code twice.

DanReyLop:
I'm merging this now, I'll drop a ZIP in the channel before actually cutting a release.

PR #1392, opened 2018-05-09
Added personal data export functionality

### LINKED ISSUE #1384: Privacy - support personal data export functionality
See p7bbVw-2vA-p2

An upcoming WordPress core will add hooks for the plugins to provide data to be exported when site users request it.

Identify the data that needs to be exported and use the hooks to add it to the report.

marcinbot:
Fixes #1384 

Adds the label tracking numbers to the personal data exports:
screen shot 2018-05-09 at 13 35 08

To test:
* Either use the latest [WordPress Core SVN trunk](https://develop.svn.wordpress.org/trunk/) or latest [master WordPress Core from GitHub](https://github.com/WordPress/WordPress)
* Use the latest [WooCommerce code](https://github.com/woocommerce/woocommerce)
* Under `Tools > Export personal data` you should be able to initiate the personal data export for a given email address
* Export the data for orders that have labels with tracking numbers purchased
* The data should include the tracking number

### DIFF
--- a/classes/class-wc-connect-privacy.php
+++ b/classes/class-wc-connect-privacy.php
@@ -5,10 +5,26 @@ if ( ! class_exists( 'WC_Abstract_Privacy' ) ) {
 }
 
 class WC_Connect_Privacy extends WC_Abstract_Privacy {
-	public function __construct() {
+	/**
+	 * @var WC_Connect_Service_Settings_Store
+	 */
+	protected $settings_store;
+
+	public function __construct( WC_Connect_Service_Settings_Store $settings_store ) {
 		parent::__construct( 'WooCommerce Services' );
+
+		$this->settings_store = $settings_store;
+
+		$this->add_exporter(
+			'woocommerce-services-labels',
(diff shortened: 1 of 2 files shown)


jeffstieler:
This tests well, but I have some questions:
* Should we include tracking numbers for labels that were refunded?
* Can we get the label data to come after the core Order data in the export? _(it's currently above it - see screenshot)_

screen shot 2018-05-15 at 3 18 01 pm

jeffstieler:
> Yeah, this seems to be a problem with how WP core wp_privacy_personal_data_exporters filter is implemented at the moment - if the priority passed to add_filter is lower, then the exported data appears before higher priority data. Seems like it should be the other way around, i.e. higher priority means the data being first on the list.

Ok, lets take the initiative on this. Please open a [Trac](https://core.trac.wordpress.org/component/Privacy) ticket for the data export merge order and submit a patch.

If we can't get a fix in core, we can try modifying the order in our code.

PR #1395, opened 2018-05-10
Backend Taxes: Use item, not product, for price and tax status

### LINKED ISSUE #1394: Taxes: calculating backend taxes should not result in error for orders with deleted products
# Steps to Repro

1. Create an order with a product you can delete later
2. Delete the product permanently
3. Go to the Edit Order page and try to click the 'update' button. See fatal error.

`
Fatal error: Uncaught Error: Call to a member function get_price() on boolean in [...]/woocommerce-services/classes/class-wc-connect-taxjar-integration.php:451
`

This happens because `wc_get_product( $id )` ([ref](https://docs.woocommerce.com/wc-apidocs/function-wc_get_product.html)) returns `false` in https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-taxjar-integration.php#L450

This came up in 1146884-zen

# Questions

1. What are the use-cases for calculating taxes every time a merchant updates the order?

From what I can tell, we need to calculate taxes for orders that are manually created, and orders that whose items are being changed. I see that I can edit an order's items when the order status is `processsing` - is that something that merchants do?

2. Should we even be looking up the price from the product, instead of the order line items? 

Right now, we use the product's current price to calculate taxes. What if the price of an item changes between the purchase and the time we calculate the product taxes? What if the merchant chooses to charge the customer less money for the product? In those cases, should we use the price listed in the order for calculating taxes?

# Possible Fix

Right now, we need two things from `$product = wc_get_product( $id );` (which may not be available if the product is deleted):
1. `$unit_price = $product->get_price();` ([L451](https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-taxjar-integration.php#L451))
2. `$product->is_taxable()` ([L454](https://github.com/Automattic/woocommerce-services/blob/master/classes/class-wc-connect-taxjar-integration.php#L454))

**For unit price**, it is available from the order line items, so we could use the following:

```
$unit_price = $item->get_subtotal();
```

I checked, and these unit price fixes work for:
1. products added as-is to the order
2. products that have been deleted since the order was made
3. products where the merchant applied a discount

Are there any cases I'm missing?

**For taxable status**, we can use `'taxable' !== $item->get_tax_status()` to see if the item is taxable. The rest remains unchanged.

See PR #1395 for full code changes.

cc @jeffstieler - what do you think?
also cc @Automattic/proton for use-cases I've possibly missed and/or any errors in understanding core WooCommerce or taxes.

v18:
Fixes #1394

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -447,11 +447,10 @@ class WC_Connect_TaxJar_Integration {
 				$tax_class = explode( '-', $item['tax_class'] );
 			}
 
-			$product = wc_get_product( $id );
-			$unit_price = $product->get_price();
+			$unit_price = $item->get_subtotal();
 			$tax_code = '';
 
-			if ( ! $product->is_taxable() ) {
+			if ( 'taxable' !== $item->get_tax_status() ) {
 				$tax_code = '99999';
 			}
 


jeffstieler:
@v18 is this ready for review, or still in progress?

v18:
@jeffstieler - not ready yet. I need to do this to move it forward:
1. see how the issue manifests in TaxJar the standalone plugin (current version)
2. see if I can make a patch there
3. open a PR against @DanReyLop's TaxJar update PR https://github.com/Automattic/woocommerce-services/pull/1398

v18:
Solved by @DanReyLop  in https://github.com/Automattic/woocommerce-services/pull/1450

PR #1398, opened 2018-05-14
Updated TaxJar integration

DanReyLop:
I've updated our TaxJar integration to their latest version (currently `1.7.0`). For that, I checked which was the last version we updated the plugin with (`9d8e725`), I opened [this GitHub view](https://github.com/taxjar/taxjar-woocommerce-plugin/compare/9d8e725...master), collapsed every file except `class-wc-taxjar-integration.php`, and copied along. We were more than 3 months behind them, diong it PR by PR would have been much more tedious.

Noteworthy:
- Their `smartcalcs_cache_request` function doesn't use that weird TLS transient system anymore. Now it can be directly copied.
- There's a new filter function: `override_woocommerce_tax_rates`. I just copied it over.
- The `calculate_totals` and `calculate_backend_totals` have been refactored into smaller functions. Those huge functions were directly copied over before, so I just copied over the new ones.
- Some minor changes like removing the `global $woocommerce;` usages.

I don't know how the integration works, but I'm fairly confident in this update, it wasn't very involved. I tried it and it returns taxes in the checkout, this is ready for a more comprehensive review.

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -267,7 +267,7 @@ class WC_Connect_TaxJar_Integration {
 			'taxjar_city_setting'     => WC()->countries->get_base_city(),
 		);
 
-		return apply_filters( 'taxjar_store_settings', $store_settings, $this->settings );
+		return apply_filters( 'taxjar_store_settings', $store_settings, array() );
 	}
 
 	/**
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -587,7 +587,8 @@ class WC_Connect_TaxJar_Integration {
 	public function override_woocommerce_tax_rates( $taxes, $price, $rates ) {
 		if ( isset( $this->response_line_items ) && array_values( $rates ) ) {
 			// Get tax rate ID for current item
-			$tax_rate_id = array_keys( $taxes )[0];
+			$keys = array_keys( $taxes );
+			$tax_rate_id = $keys[0];
 			$line_items = array();
 
 			// Map line items using rate ID
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -41,6 +41,9 @@ class WC_Connect_TaxJar_Integration {
 	) {
 		$this->api_client = $api_client;
 		$this->logger = $logger;
+
+		// Cache rates for 1 hour.
+		$this->cache_time = HOUR_IN_SECONDS;
 	}
 
 	public function init() {
@@ -86,6 +89,8 @@ class WC_Connect_TaxJar_Integration {
 
 		// Set customer taxable location for local pickup
 		add_filter( 'woocommerce_customer_taxable_address', array( $this, 'append_base_address_to_customer_taxable_address' ), 10, 1 );
+
+		add_filter( 'woocommerce_calc_tax', array( $this, 'override_woocommerce_tax_rates' ), 10, 3 );
 	}
 
 	/**
@@ -250,7 +255,7 @@ class WC_Connect_TaxJar_Integration {
 	 * Gets the store's location settings.
 	 *
 	 * Modified version of TaxJar's plugin.
-	 * See: https://github.com/taxjar/taxjar-woocommerce-plugin/blob/82bf7c58/includes/class-wc-taxjar-integration.php#L796
+	 * See: https://github.com/taxjar/taxjar-woocommerce-plugin/blob/4b481f5/includes/class-wc-taxjar-integration.php#L910
 	 *
 	 * @return array
 	 */
@@ -262,7 +267,7 @@ class WC_Connect_TaxJar_Integration {
 			'taxjar_city_setting'     => WC()->countries->get_base_city(),
 		);
 
-		return $store_settings;
+		return apply_filters( 'taxjar_store_settings', $store_settings, $this->settings );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


v18:
Just noting that there's a bug with some new code as of v1.6, detailed here (see Jeff's comment):

https://github.com/taxjar/taxjar-woocommerce-plugin/issues/68

DanReyLop:
> Just noting that there's a bug with some new code as of v1.6, detailed here (see Jeff's comment):

So before, the plugin would crash on deleted products, now it just ignores them. I'd say that's a net improvement.

If we want to solve that issue on our own, I suggest we do that in a separate PR :)

DanReyLop:
Nice, our CI caught some PHP 5.2 and 5.3 incompatibility! Fixed it and opened https://github.com/taxjar/taxjar-woocommerce-plugin/pull/70 upstream.

v18:
I don't think introducing a new bug and calculating taxes incorrectly is the right move here.

v18:
> For what I understood is not a new bug, it's just a bug that was already there and hasn't been completely solved (just mitigated).

Sorry for not explaining more fully, it's a bit complicated :P. Here's the deal:

1. In TaxJar version 1.5.x (and this is the code we currently have in our plugin), deleted products would cause a PHP fatal when updating the order.
2. In version 1.6, they fixed this bug in TaxJar by not calculating tax for deleted products. This is a problem. We have not ported over this code.
4. This is the code that we're about to port over to our plugin, meaning that our plugin would not calculate taxes correctly for deleted products in an order.

DanReyLop:
Ok I understood it correctly then :)

> This is the code that we're about to port over to our plugin, meaning that our plugin would not calculate taxes correctly for deleted products in an order.

At the very least, this behaviour is *not* worse than a PHP fatal. I don't think it should be a blocker for updating the integration. The issue (https://github.com/taxjar/taxjar-woocommerce-plugin/issues/68) is already created so it's documented.

Are you worried that we'll merge this, then fix the bug in TaxJar, and then wait another 6 months to update the integration again? That shouldn't be a problem. We should try to do this chore more often so it's easier, it would be just cherry-picking changes.

v18:
> Are you worried that we'll merge this [...]

I'm worried that we merge this PR, and then not fix the bug in our plugin before the next release, thereby releasing this new bug to users.

> At the very least, this behaviour is not worse than a PHP fatal

I think it is, because the tax rate is likely to be incorrectly calculated without any indication. And it's probably a headache to fix for merchants.

v18:
Noting here that I'm going to write a PR with the fix.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -262,7 +267,7 @@ public function get_store_settings() {
 			'taxjar_city_setting'     => WC()->countries->get_base_city(),
 		);
 
-		return $store_settings;
+		return apply_filters( 'taxjar_store_settings', $store_settings, $this->settings );
```
v18: I'm getting an error on this line:

`
Notice: Undefined property: WC_Connect_TaxJar_Integration::$settings in woocommerce-services/classes/class-wc-connect-taxjar-integration.php on line 270
`

DanReyLop: My bad, should have tested with `WP_DEBUG` on.

The filter was added [here](https://github.com/taxjar/taxjar-woocommerce-plugin/commit/c8db821cc91536fb0ee6337eb785603b8e76f514).

I fixed it in https://github.com/Automattic/woocommerce-services/pull/1398/commits/73fb20e8134928379d3cb9af1eb872ec13b13d9f, I added a dummy empty array to keep the filter "signature".

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -713,34 +825,35 @@ public function create_or_update_tax_rate( $line_item_key, $location, $rate, $ta
...
-		$cache_key = 'wcs_tax_' . hash( 'md5', $json );
+		$cache_key = 'tj_tax_' . hash( 'md5', $json );
 		$response  = get_transient( $cache_key );
 
 		if ( false === $response ) {
 			$response = $this->smartcalcs_request( $json );
 
 			if ( 200 == wp_remote_retrieve_response_code( $response ) ) {
-				set_transient( $cache_key, $response, HOUR_IN_SECONDS );
+				set_transient( $cache_key, $response, $this->cache_time );
```
jeffstieler: Why differ from the TaxJar plugin here?

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -657,16 +754,31 @@ public function calculate_tax( $options = array() ) {
 			);
 
 			// Add line item tax rates
-			foreach ( $this->line_items as $line_item_key => $line_item ) {
+			foreach ( $taxes['line_items'] as $line_item_key => $line_item ) {
 				$line_item_key_split = explode( '-', $line_item_key );
```
jeffstieler: The TaxJar plugin has since renamed this to `$line_item_key_chunks`, let's port that over too. See: https://github.com/taxjar/taxjar-woocommerce-plugin/blob/master/includes/class-wc-taxjar-integration.php#L350

classes/class-wc-connect-taxjar-integration.php
DanReyLop: We were differing *before*, now it's exactly the same as the Taxjar plugin.

jeffstieler: Whoops! I missed that. 👍

jeffstieler:
@DanReyLop @v18 I think we're good to merge this. In testing, deleted products aren't sent to TaxJar when recalculating, but they do get correct tax applied using the tax rate table entry that the integration creates. We can apply https://github.com/taxjar/taxjar-woocommerce-plugin/pull/72 when it's ready as well to include the deleted product line item info in the TaxJar request.

PR #1401, opened 2018-05-16
Fix "Undefined variable: service_settings" error

marcinbot:
Fixes a PHP error that appears when the server cannot be reached.

To reproduce:
* Make sure the client cannot reach the server (for example stop the server if using a local env)
* Try to request rates
* Warnings will appear

To test:
* Repeat the reproduce steps
* No PHP warnings should appear

This might also require a point release since it breaks the fallback rate functionality :/

### DIFF
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -245,7 +245,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 			return is_string( $preset_id );
 		}
 
-		private function check_and_handle_response_error( $response_body ) {
+		private function check_and_handle_response_error( $response_body, $service_settings ) {
 			if ( is_wp_error( $response_body ) ) {
 				$this->debug(
 					sprintf(
@@ -349,7 +349,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 				$this->debug( 'Rates response retrieved from cache' );
 			} else {
 				$response_body = $this->api_client->get_shipping_rates( $services, $package, $custom_boxes, $predefined_boxes );
-				if ( $this->check_and_handle_response_error( $response_body ) ) {
+				if ( $this->check_and_handle_response_error( $response_body, $service_settings ) ) {
 					return;
 				}
 				wp_cache_set( $cache_key, $response_body, '', 3600 );


v18:
Hey @marcinbot, I'm unable to repro the warnings you see. This is what I tried:

1. Check out `master`
2. Go to 'Edit Order'
3. Open shipping labels modal, and make sure that addresses are already validated on order
4. Stop server
5. Change the package
5. See this error "Error: The WooCommerce Services server returned HTTP code: 0"
screen shot 2018-05-16 at 7 37 16 am

I didn't see any PHP warnings show up on that page, or in my log.

Could you write out more detailed steps on how you were able to repro the error from `master`?

v18:
@marcinbot coolio, thanks 👍 

Here's what I tried on `master`:
0. WCS debugging is turned off 
1. Go to cart with an item that will use a USPS WCS shipping method
2. Stop the server
3. Calculate the shipping rates with a new zipcode
4. See these PHP warnings in my php error log:

`
Error. Unable to get shipping rate(s) for wc_services_usps instance id 5. (check_and_handle_response_error)
`

`wcc_server_error Error: The WooCommerce Services server returned HTTP code: 0 (check_and_handle_response_error)`

`PHP Notice:  Undefined variable: service_settings in [...]/woocommerce-services/classes/class-wc-connect-shipping-method.php on line 269`

`PHP Warning:  First parameter must either be an object or the name of an existing class in /Applications/MAMP/htdocs/wp.localhost/wp-content/plugins/woocommerce-services/classes/class-wc-connect-shipping-method.php on line 284`
5. See no error on the frontend, though

After checking out `fix/undefined-service-settings` and following steps 0-3 these are the errors I see:

`Error. Unable to get shipping rate(s) for wc_services_usps instance id 5. (check_and_handle_response_error)`

`wcc_server_error Error: The WooCommerce Services server returned HTTP code: 0 (check_and_handle_response_error)`

So, it seems like the original error is fixed 👍 

> This might also require a point release since it breaks the fallback rate functionality :/

This fix didn't fix fallback rate functionality for me - in either case, the fallback rate was not used unless it had already been returned from the server in a previous shipping rate calculation for that cart.

v18:
> Can you turn enable debug mode under WooCommerce > Settings > Shipping > Shipping options and try again?

Cool, I didn't know about this option! Here I was trying to bust the cache manually, and that obviously wasn't working.

I enabled debug mode, and saw the fallback rate 👍

v18:
Incidentally, this also fixes another issue: when a merchant has a product without dimensions/weight, then we return no available rates. Instead, this fix will return the fallback rate for them. 🎊
cc @stephhjacq for bringing this up :D

PR #1406, opened 2018-05-18
Admin breaking when product with missing weight is deleted

dechov:
Fixes crash scenario (introduced by https://github.com/Automattic/woocommerce-services/pull/1380) when product featured in missing weight admin notice is deleted. To reproduce (from @v18):
1. get rates for a product without ~dimensions~ weight
2. delete product
3. in `master`, see fatal on any admin screen. In this branch, admin notice should just disappear


v18:
I was able to repro the error using the steps. This PR fixed the error for me. Smart & clear solution, 🚢

PR #1408, opened 2018-05-18
Add personal data erasure

marcinbot:
Erases personal data (tracking numbers) from the client database and calls the server to erase the same data.

Note: the affected labels will currently disappear from the order page. A separate Calypso PR will follow which will render the anonymized labels in the log, but not show the reprint link.

To test:
* Make sure you're using WordPress 4.9.6+
* Use the latest [WooCommerce code](https://github.com/woocommerce/woocommerce)
* Use the server PR - 1089-gh-woocommerce-connect-server
* Set WooCommerce to erase the order data by navigating to `WooCommerce > Settings > Accounts & Privacy` and enabling the `Remove personal data from orders` checkbox
* Place an order with an email address
* Purchase a label for that order 
* Navigate to `Tools > Erase Personal Data` and erase the data using the email with which the order was placed
* The labels should have their tracking number removed in the database and status set to `ANONYMIZED`
* Same should happen on the server, the label image cache should not exist for the affected labels

### DIFF
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -252,6 +252,16 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 		}
 
 		/**
+		 * Gets the shipping label status (refund status, tracking code, etc)
+		 *
+		 * @param $order_id integer
+		 * @return object|WP_Error
+		 */
+		public function anonymize_order( $order_id ) {
+			return $this->request( 'POST', '/privacy/order/' . $order_id . '/anonymize' );
+		}
+
+		/**
 		 * Request a refund for a given shipping label
 		 *
 		 * @param $label_id integer
--- a/classes/class-wc-connect-privacy.php
+++ b/classes/class-wc-connect-privacy.php
@@ -10,8 +10,14 @@ class WC_Connect_Privacy {
 	 */
 	protected $settings_store;
 
-	public function __construct( WC_Connect_Service_Settings_Store $settings_store ) {
+	/**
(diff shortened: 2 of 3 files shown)


DanReyLop:
> A separate Calypso PR will follow which will render the anonymized labels in the log, but not show the reprint link.

Will this be implemented after the 25th?

> Either use the latest WordPress Core SVN trunk or latest master WordPress Core from GitHub

Is `4.9.6` enough, or does this depend on something yet to be released?

DanReyLop:
> this will require some changes to make the latest Calypso master be compatible with the code here (or the other way around)

Could you elaborate on this? Right now the `ANONYMIZED` labels simply dissapear, I think that's a decent backwards-compatible behaviour. In fact, I'd say that the Calypso PR isn't a hard requirement for shipping this feature.

The Calypso PR that handles `ANONYMIZED` labels will need to be merged and then the `wp-calypso` dependency will need to be updated before releasing the new plugin version. Do you want to do that updating as part of this PR, or a different one?

In any case, code looks good, I'll test it before giving you a thumbs up but it looks good.

DanReyLop:
> That's why I would like to perform the Calypso dependency update in a separate PR.

That's reasonable :)

> Namely, some of the API layer code was recently updated in Calypso and it leads to JS errors which are unrelated to the feature implemented in this PR.

What are you referring to? We can talk it in Slack if you want. I updated the `wp-calypso` dependency recently [here](https://github.com/Automattic/woocommerce-services/commit/a39a560bf7561add6a823fef1d5a34fd5ccc93e1) and it was pretty straightforward.

Issue #1350, opened 2018-03-26
Remove or change the blue "tutorial" notice that appears on top of the shipping method config page


DanReyLop:
This notice:

![screen shot 2018-03-26 at 17 12 04](https://user-images.githubusercontent.com/1715800/37918492-10b81cd2-3119-11e8-9454-b3707ef94218.png)

It made sense when we started, but now when adding a shipping method, the ZIP code and some services are automatically prefilled. So, either remove it entirely, or tweak its text.

jeffstieler:
Let's remove it. 👍

PR #1410, opened 2018-05-21
Use Calypso state and components for the USPS and Canada Post settings

DanReyLop:
Fixes https://github.com/Automattic/woocommerce-services/issues/1350

This PR looks scary but ~80% of the added lines are for the shrinkwrap, and all the removed lines are, well, removed files.

This was the last part of code that was still duplicated between Calypso and the plugin. I managed to use the Calypso components and reducers/selectors to work in WP-Admin.

It's better to review this PR commit-by-commit:
- https://github.com/Automattic/woocommerce-services/commit/a39a560bf7561add6a823fef1d5a34fd5ccc93e1 Update the `wp-calypso` dependency, because I needed https://github.com/Automattic/wp-calypso/pull/24868 to make the port work.
- https://github.com/Automattic/woocommerce-services/commit/3ca962f92523a165097b6ae31b2c0daffe55591e Remove all the code on the `settings` app.
- https://github.com/Automattic/woocommerce-services/commit/4f8b00622006211b224bc777e5a7925199a00fe2 The core of this PR. Add a new `index.js` and view wrapper that just uses Calypso components and reducers.
- https://github.com/Automattic/woocommerce-services/commit/1856ec4f54b166b7806700b5f9830514825b41bb Massive cleanup of unused code.
- https://github.com/Automattic/woocommerce-services/commit/240d652fb692c9c6d20b34c9b7eb76d40c77d7bd Removes all the logic related to the NUX notice that appeared on top of the USPS/Canada settings, since that's no longer used. See https://github.com/Automattic/woocommerce-services/issues/1350 for more context.

The `client/apps/settings/*` files are the most important ones in this PR. I had to recreate more of the Calypso state tree than in other "apps", due to the way the Shipping Zones UI is implemented in Calypso. Let me know if you think it's not worth it to jump through these hoops just to reuse the code.

I tested an USPS shipping instance and smoke-tested the rest of the UI (self-help page, Package Manager, Label Settings, Label Purchase). This PR does a lot of cleanup so testing a bit all our UI makes sense.


DanReyLop:
In https://github.com/Automattic/woocommerce-services/pull/1410/commits/a8b928c457c55f4c35ee0c961cb994e4d993ce20 I changes the `npm test` config so it doesn't fail if it couldn't find any test files. Now that almost all our code is in Calypso, turns out we don't have any test files left. Still, the `npm test` task is still important to be run in TravisCI because it runs Webpack/Babel and `eslint`, so it's still valuable to catch some bugs.

DanReyLop:
> I noticed that a React warning pops up on the settings pages, but it doesn't seem to be blocking and needs to be fixed in Calypso:

Yeah I saw that but it happens in Calypso too, I didn't want to go off the rails.

Thanks for the CSS fixes, I hadn't noticed it!

v18:
So.. I just saw this was merged. This was the comment I was in the middle of writing. For posterity:

Hey @DanReyLop, so I'm going to be the bearer of a possibly annoying request :), bear with me. I think this PR should be broken into several smaller PRs. I understand the reasons why it's easier to make this only one PR:
- it's all related
- a lot of the code is deletion, so it should be simple to review
- a lot of this code is shrinkwrap changes, the rest of the changes are not so big
- it's not easy to decide on which code to merge in first
Maybe there are more reasons which I'm missing.

I do see that you outlined that the review can be done in pieces following the commits. Those commits could be the separate PRs (more or less, I understand they may not be atomic enough for a PR as-is).

However... it's really difficult to read this large a PR, and it can result in a few problems. The main problem is that things are not reviewed in enough detail, causing problems later down the road. Another big problem is that it takes too much effort or time to give attention to the possible side-effects of this PR on other related code, and on the overall architecture we have. Reviewing in smaller pieces minimizes these problems. Lastly, each PR will give more context - you could explain the changes atomically in each PR's description, making the reviewers job easier, and helping those who will read this code in the future.

As for the specific details of this PR, sure we can ignore the shrinkwrap changes when reviewing, but these things still apply:
- the deletion changes should be paid attention to - is each change necessary, how does it affect other parts of the code, etc. It could be a PR on its own (with any relevant pieces added in as needed)
- the commit 4f8b0062 -  'Use calypso state and UI for the USPS / CanadaPost settings pages' - is +322, that's pretty large already, and should be its own PR

PR #1413, opened 2018-05-22
Fix log file retrieval on Status page to work with new WC 3.4 naming conventions

jeffstieler:
Fix log file retrieval on Status page to work with new WC 3.4 format that includes the date.

https://github.com/woocommerce/woocommerce/commit/e9b0c6deec21c7befa7805e40c84cebc451d83a8 introduced a date string `YYYY-MM-DD` into the log file names.

To test:
* Verify that an "old style" log still loads on the WCS status page (name like `wc-services-c4fa6ec2cfff7f708e89c42ac4b5a28c.log`)
* Delete that log file
* Perform an action that adds entries to the file, like rates retrieval/taxes
* Verify that entries are displayed on the Status page


classes/class-wc-connect-help-view.php
```diff
@@ -228,7 +228,7 @@ protected function get_debug_log_data( $feature = '' ) {
 			$latest_file_date = 0;
 
 			foreach ( $logs as $log_key => $log_file ) {
-				if ( ! preg_match( '/' . $log_prefix . '\-[0-9a-f]{32}\-log/', $log_key ) ) {
+				if ( ! preg_match( '/' . $log_prefix . '\-(?:\d{4}\-\d{2}\-\d{2})?\-[0-9a-f]{32}\-log/', $log_key ) ) {
```
marcinbot: This line should read:
`if ( ! preg_match( '/' . $log_prefix . '\-(?:\d{4}\-\d{2}\-\d{2}\-)?[0-9a-f]{32}\-log/', $log_key ) ) {`
i.e. move the dash following the date into the optional match, so the legacy file names are matching. Otherwise the regexp expects a filename like `prefix--guid.log`

jeffstieler: Fixed in https://github.com/Automattic/woocommerce-services/pull/1413/commits/6c0108730cc9da5ca08987f241bab92358d962b0.

PR #1428, opened 2018-05-31
Respond with real HTTP errors when the shipping settings can't be validated

DanReyLop:
There's more context in https://github.com/Automattic/wp-calypso/pull/25101

We had an error in which service schema validation would fail in the server if the adjustment fields were touched in any way. The error has been fixed (something something `coerceValues`), but I dug deeper on "Why doesn't this fail in Calypso"?

In Calypso, a validation failure in the plugin side will happily show a "success" notice, but the shipping method won't be updated at all. That's because the Calypso HTTP layer logic checks for non-200 HTTP responses to trigger the error notice, and the plugin responds with a `HTTP 200` response with `{ success: false }` in the body.

This PR changes the plugin logic so it responds with a real HTTP error. All our other REST endpoints already return a `WP_Error` on failure, this one was the only one that used the `{ success: false }` trick.

To test:
- With this branch of the plugin installed on your site, go to Calypso (can be `wordpress.com` or a local install).
- Go to the Shipping settings, and add or edit a USPS shipping method.
- Before saving, stop your local Hydra server (or make the `WOOCOMMERCE_CONNECT_SERVER_URL` constant to point to a non-existing server).
- Save the changes in Calypso.
- Check that you get an error notice. Before, it would just fail silently and display a Success notice.

### DIFF
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -447,22 +447,12 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 			if ( ! empty( $instance ) ) {
 				$service_schema = $this->service_schemas_store->get_service_schema_by_instance_id( $instance );
 				if ( ! $service_schema ) {
-					wp_send_json_error(
-						array(
-							'error' => 'bad_instance_id',
-							'message' => __( 'An invalid service instance was received.', 'woocommerce-services' )
-						)
-					);
+					return new WP_Error( 'bad_instance_id', __( 'An invalid service instance was received.', 'woocommerce-services' ) );
 				}
 			} else {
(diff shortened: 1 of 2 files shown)


DanReyLop:
@marcinbot I've fixed that in https://github.com/Automattic/woocommerce-services/pull/1428/commits/8181e0a3388361f1c280a0607db000558a3e0f00. At the time I had `WP_DEGUG` disabled so it worked for me. `WP_Error->get_error_data()` returns `NULL` when it hasn't been explicitly set, and `array_merge($array, NULL)` works but throws a warning (making the response invalid JSON).

v18:
Putting this back on 'Needs Review' since I'm not 100% convinced about sending HTTP status codes for 'successful' responses. Brb with more notes.

v18:
Nvm, all good 👍

PR #1430, opened 2018-06-04
Don't show "add method to zone" pointer if shipping was opted into

### LINKED ISSUE #1334: Don't show "add method to zone" pointer if a WCS method has already been setup
screen shot 2018-03-12 at 2 28 22 pm

This pointer still appears after using the WooCommerce Setup Wizard to setup a WCS-powered shipping method.

jeffstieler:
Fixes #1334.

Don’t show the shipping method NUX pointer if shipping was opted into in the core Setup Wizard.

To test:
* Set up a new WooCommerce store using the OBW
* Opt in to either domestic or international live rates
* Connect Jetpack
* Verify that the "add a WCS method" pointer *does not* show on the Shipping settings page

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -113,6 +113,23 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 			return array();
 		}
 
+		/**
+		 * Dismiss a WP pointer for the current user.
+		 *
+		 * @param string $pointer_to_dismiss Pointer ID to dismiss for the current user
+		 */
+		public function dismiss_pointer( $pointer_to_dismiss ) {
+			$dismissed_pointers = $this->get_dismissed_pointers();
+
+			if ( in_array( $pointer_to_dismiss, $dismissed_pointers, true ) ) {
+				return;
+			}
+
+			$dismissed_pointers[] = $pointer_to_dismiss;
+			$dismissed_data = implode( ',', $dismissed_pointers );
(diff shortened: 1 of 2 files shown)


classes/class-wc-connect-nux.php
```diff
@@ -149,9 +166,7 @@ public function register_order_page_labels_pointer( $pointers ) {
 
 			// If the user is not new to labels, we should just dismiss this pointer
 			if ( ! $this->is_new_labels_user() ) {
-				$dismissed_pointers[] = 'wc_services_labels_metabox';
```
dechov: Do we still need `$dismissed_pointers` declared above, considering that the pointers are checked for having been dismissed at [a later point](https://github.com/Automattic/woocommerce-services/blob/02cad08415eca8b1321dff4f037e52cda38917ee/classes/class-wc-connect-nux.php#L94-L96) anyway?

jeffstieler: Nice catch - I'll clean it up.

jeffstieler: Fixed in bdbdd2fc.

dechov:
Should have noted what I did to test, especially since I didn't precisely follow the testing directions:

- Test site with the WCS shipping method pointer showing
- Made sure ordinary dismissal still worked
- Restored pointer by editing usermeta table
- Deleted shipping zone, went to "Shipping" step of OBW wizard, and hit "Continue" with Live rates enabled
- Jetpack and WCS was already set up, and shipping method pointer no longer appeared :+1:
- Made sure shipping label pointer still worked as before

PR #1434, opened 2018-06-07
Refresh service schemas when specific WooCommerce settings are updated

### LINKED ISSUE #1273: Immediately (re) fetch schemas if store country is changed
i.e. don't make a merchant wait for the services schemas cache to expire (1 day) if they change their store country (e.g. from US to CA - that way they can set immediately set up Canada Post and not have to wait)

jeffstieler:
Fixes #1273.

Refresh service schemas when the following options are updated: store post code, currency, weight unit, and dimension unit.

**To Test:**

- Load this branch
- Enable logging
- Make sure your frequent fetch flag is NOT enabled
- Change any one of the mentioned settings via `wp-admin`
- Observe that a fetch_service_schemas_from_connect_server event is logged in the logs
- Change any one of the mentioned settings via the REST API
- Observe that a fetch_service_schemas_from_connect_server event is logged in the logs

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -663,11 +663,15 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 				}
 			}
 
-			add_action( 'rest_api_init', array( $this, 'rest_api_init' ) );
+			// Changing the postcode, currency, weight or dimension units affect the returned schema from the server.
+			// Make sure to update the service schemas when these options change.
+			// TODO: Add other options that change the schema here, or figure out a way to do it automatically.
 			add_action( 'update_option_woocommerce_store_postcode',  array( $schemas_store, 'fetch_service_schemas_from_connect_server' ) );
 			add_action( 'update_option_woocommerce_currency', array( $schemas_store, 'fetch_service_schemas_from_connect_server' ) );
 			add_action( 'update_option_woocommerce_weight_unit', array( $schemas_store, 'fetch_service_schemas_from_connect_server' ) );
 			add_action( 'update_option_woocommerce_dimension_unit', array( $schemas_store, 'fetch_service_schemas_from_connect_server' ) );
+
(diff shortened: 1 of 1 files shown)


woocommerce-services.php
```diff
@@ -664,7 +664,10 @@ public function attach_hooks() {
 			}
 
 			add_action( 'rest_api_init', array( $this, 'rest_api_init' ) );
-			add_action( 'woocommerce_settings_saved', array( $schemas_store, 'fetch_service_schemas_from_connect_server' ) );
+			add_action( 'update_option_woocommerce_store_postcode',  array( $schemas_store, 'fetch_service_schemas_from_connect_server' ) );
```
dechov: Any way to ensure this is kept in sync with what the server consumes? I'm not sure how but maybe at least a code comment?

jeffstieler: > Any way to ensure this is kept in sync with what the server consumes?

I'm not sure what you mean by this.

dechov: Referring to hypothetical future changes: if we were to update the server to act on more (or fewer) of the settings, we would likely want to modify this set of actions as well.

jeffstieler: Gotcha, I'll add a comment explaining this.

jeffstieler: Done in 32c198b. If we start making more changes on the server based on option values, we could look at adding those option keys to the schema and dynamically updating them in the future.

jeffstieler:
> Noticed that if you save multiple settings at once (such as weight and dimensions units), schemas will be fetched multiple times – I don't think it's a blocker, but it would be worth preventing if there's an easy way to do it.

👍 I'll look into this.

jeffstieler:
Added a queueing mechanism in https://github.com/Automattic/woocommerce-services/pull/1434/commits/59f092aaa0117163e565068db96a7641d8a3aedb that uses the `shutdown` hook to fetch new schemas.

This needs a retest.

PR #1437, opened 2018-06-08
Order fulfillment options when purchasing label

### LINKED ISSUE #1274: Handle emailing of the tracking info better
We append the tracking numbers to the emails that are sent on order completion: #1173

This feature is not exposed very well, and also clashes with the "email details" checkbox visible in Calypso (two emails get sent containing this information, one as a plaintext note and one as part of the order status update)

We should:
* show the same checkbox in WP-Admin as we do in Calypso
* handle emailing in the backend as opposed to the note added to the order (might be tricky to make it backwards-compatible in Calypso)
* only email if the checkbox is set
* merge the emails if the merchant fulfills the order and opts in to email the tracking info

Related: #1245

jeffstieler:
Fixes #1274.

This PR shows the "mark order fulfilled" and "email details to customer" checkboxes on the label modal:

screen shot 2018-06-08 at 4 35 09 pm
 
It also adds the backend provisions for store-wide defaults for these - view in Calypso:

screen shot 2018-06-08 at 3 55 50 pm

_Note: Without landing the Calypso PR and updating the dependency here, the "email details" checkbox is effectively inert._

**To test:**
- Check out https://github.com/Automattic/wp-calypso/pull/25400 and `npm link` it
- Verify that checkboxes on the order screen default to the values set store-wide
- Verify that the "mark order fulfilled" checkbox changes the order from "processing" to "complete" when purchasing a label
- Verify that the "email details" checkbox is hidden when "mark fulfilled" is unchecked
- Verify that tracking info is in "completed order" email and no customer note is created when "email details" is checked
- Verify that transitioning an order to "complete" outside of the labels flow sends tracking info if the order has labels and the store-wide setting for "email details" is checked

### DIFF
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -52,16 +52,15 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 		public function get_account_settings() {
 			$default = array(
 				'selected_payment_method_id' => 0,
-				'enabled' => true,
+				'enabled'                    => true,
+				'email_receipts'             => true,
+				'mark_orders_complete'       => true,
+				'email_tracking_info'        => true,
 			);
 
 			$result = WC_Connect_Options::get_option( 'account_settings', $default );
 			$result['paper_size'] = $this->get_preferred_paper_size();
 			$result = array_merge( $default, $result );
-
-			if ( ! isset( $result['email_receipts'] ) ) {
-				$result['email_receipts'] = true;
-			}
 
 			return $result;
 		}
--- a/client/apps/shipping-label/style.scss
+++ b/client/apps/shipping-label/style.scss
@@ -32,11 +32,6 @@
 	font-size: 15px;
 }
 
-.label-purchase-modal__option-email-customer,
-.label-purchase-modal__option-mark-order-fulfilled {
-	display: none;
-}
-
 .order-activity-log {
 	.foldable-card .foldable-card__content {
 		padding-left: 0;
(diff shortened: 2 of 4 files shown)


classes/class-wc-connect-shipping-label.php
```diff
@@ -440,11 +440,13 @@ public function get_label_payload( $post_order_or_id ) {
...
 			$payload = array(
 				'orderId'            => $order_id,
-				'paperSize'          => $this->settings_store->get_preferred_paper_size(),
+				'paperSize'          => $account_settings['paper_size'],
 				'formData'           => $this->get_form_data( $order ),
 				'labelsData'         => $this->settings_store->get_label_order_meta_data( $order_id ),
 				//for backwards compatibility, still disable the country dropdown for calypso users with older plugin versions
 				'canChangeCountries' => true,
+				'markOrdersComplete' => $account_settings['mark_orders_complete'],
+				'emailTrackingInfo'  => $account_settings['email_tracking_info'],
```
marcinbot: See https://github.com/Automattic/wp-calypso/pull/25400#discussion_r196472217 - no need to include the settings data in the labels response. Possibly we could also remove `paperSize` because of that.

jeffstieler:
> Was there a decision made about having to refresh the page to see that it was marked as complete?

Not really - @dechov and I briefly discussed it on Slack and I made a quick attempt to refresh the page, but did so in a naive way that caused the dismissal of the print dialog before it was interacted with.

Looking at the SO links, I think using the [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) tactic is a good route to pursue for browsers that support it.

joshuaja:
This looks to be a great UX improvement to the original email tracking feature. When is this scheduled to be merged and deployed?

jeffstieler:
@c-shultz I'm going through and closing old PRs - I assume it's safe to close this eh?

PR #1440, opened 2018-06-11
Fixes the "Save changes" button on USPS/Canada settings staying in "busy" state after failing validation

DanReyLop:
If there are invalid fields when you open the USPS/Canada settings form, but you haven't interacted with those fields, then they won't show any error until you click "Save changes". Clicking that button will trigger a full validation and show an error notice if it fails (the logic of that is in `wp-calypso`, including the error notice). Since all that logic is in `wp-calypso`, the `<ViewWrapper>` component's state wasn't being updated, so it would keep showing the `busy` animation forever.

To reproduce:
* Get to the service settings page having some fields be incorrect. The easiest way is to change the Store base address to have an invalid ZIP code, and then add a new shipping method (it will take that bogus ZIP code as default).
* Interact with any other field (such as the title, or services list).
* Click the `Save changes` button.
* In `master`, you'll see the error notice but the button will keep being "busy". With this change, it will become disabled but it will be clickable again once you fix the errors.

### DIFF
--- a/client/apps/settings/view-wrapper.js
+++ b/client/apps/settings/view-wrapper.js
@@ -48,10 +48,12 @@ class ViewWrapper extends Component {
 		};
 		const failureAction = ( dispatch, getState, { error } ) => {
 			this.setState( { isSaving: false } );
-			dispatch( errorNotice( isString( error )
-				? error
-				: translate( 'There was a problem with one or more entries. Please fix the errors below and try saving again.' )
-			) );
+			if ( error ) {
+				dispatch( errorNotice( isString( error )
+					? error
+					: translate( 'There was a problem with one or more entries. Please fix the errors below and try saving again.' )
+				) );
+			}
 		};
 
 		this.setState( { isSaving: true } );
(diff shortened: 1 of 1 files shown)


client/apps/settings/view-wrapper.js
```diff
@@ -116,6 +118,7 @@ const saveChanges = ( successAction, failureAction ) => ( dispatch, getState ) =
 	// Trigger a full validation
 	dispatch( closeShippingZoneMethod( siteId ) );
 	if ( getCurrentlyOpenShippingZoneMethod( getState() ) ) {
+		dispatch( failureAction );
```
jeffstieler: It's not apparent to me why we need to dispatch our failure action here.

Also, I know this wasn't introduced in your changes, but do you know why we aren't using the `method` variable in the `if`?

DanReyLop: I've added some comments to clarify that behaviour. Basically, in this function we are "emulating" the Calypso UI and behaviour.

When dispatching the `closeShippingMethod()` action, it triggers a full validation of the shipping method, because that's what happens when you close the shipping method modal in Calypso. The Redux state tree knows which method is currently opened so there's no need to specify `method.id`.

If the validation fails, the shipping method isn't closed (in Calypso that would mean that the modal would keep being opened), so if after dispatching the `close` action there is still a method marked as being edited, we know it failed.

PR #1442, opened 2018-06-12
Updated nodeJS version to 10.4.0 and npm version to 6.1.0

DanReyLop:
See p4TIVU-8XM-p2 for more context.

`npm 6` seems to have less bugs than `npm 5`, so Calypso has decided to jump to `node 10` and `npm 6`. `node 10.4.0` comes by default paired with `npm 6.1.0`, so installing `node 10.4.0` should take care of updating `npm` too.

Note that, after this change is merged, `wp-calypso` and `woocommerce-services` work will need to use `npm 6` and `node 10`, whereas `woocommerce-connect-server` will keep using the LTS versions (`node 8` and `npm 5`) until at least October (when `node 10` becomes the LTS version). I don't believe it's worth it to update the server version yet, the dependency tree is much smaller and `npm` hasn't given us any problems there.

Once this is merged, you'll need to run `nvm install` on the `woocommerce-services` folder. There's some cool integrations with shells like `zsh` to do that automatically, but I don't know enough about them. Just make sure than, when you're developing on `woocommerce-services` or `wp-calypso`, `node --version` and `npm --version` give the expected results.

In this PR:
- Updated `.nvmrc` so `nvm` can install the new version automatically. `TravisCI` should also honor the version in `.nvmrc`.
- Updated `node-sass` because `4.9.0` is its first version to support `node 10`.
- Updated the `wp-calypso` dependency. [This](https://github.com/Automattic/wp-calypso/pull/25393) is the only new `Store Services` PR included (@dechov I did a quick test and the ellipsis menu works fine).
- Regenerated the `shrinkwrap`, using `npm run update-deps`.


dechov:
Thanks, I was trying to bring in that commit last night but had some trouble with `npm install` stalling out for some reason. Haven't tried this branch yet but when I tried last week, I thought I needed to import component styles into `assets/stylesheets/_components.scss`: `components/ellipsis-menu/style`, `components/popover/style` – is it not necessary?

DanReyLop:
@dechov This is ready for review. You can see the ellipsis menu animating in all its glory :)

PR #1448, opened 2018-06-25
Use the wc/v3/data/continents endpoint

DanReyLop:
See https://github.com/Automattic/wp-calypso/pull/25723 for the Calypso part of this PR.

This PR removes the `countriesData` property that was returned in the `labels` payload. Instead, now the countries/states data is fetched from the `wc/v3/data/continents` REST endpoint. Since a site using `woocommerce-services` from WP-Admin may not have the `wc/v3/*` endpoints (right now they're only available through the `wc-api-dev` plugin), the entire endpoint is copied in the plugin, unchanged.

To test:
* Disable or uninstall the `wc-api-dev` plugin.
* Update your `wp-calypso` dependency to the branch `update/store-labels-use-continents-endpoint`. `npm link ../wp-calypso` should work now that we're using `npm 6`.
* Go to an order.
* Open the `Purchase Label` modal.
* See that all the "USPS" countries and territories appear in the `Origin address` `Country` dropdown.
* Select `United States`, see that the military states appear in the `State` dropdown.

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -820,8 +820,12 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		 * Delete this when the "v3" REST API is included in all the WC versions we support.
 		 */
 		public function wc_api_dev_init() {
-			if ( ! class_exists( 'WC_REST_Dev_Data_Continents_Controller' ) ) {
-				require_once( plugin_basename( 'classes/wc-api-dev/class-wc-rest-dev-data-controller.php' ) );
+			$rest_server = rest_get_server();
+			$existing_routes = $rest_server->get_routes();
+			if ( ! isset( $existing_routes['/wc/v3/data/continents'] ) ) {
+				if ( ! class_exists( 'WC_REST_Dev_Data_Continents_Controller' ) ) {
+					require_once( plugin_basename( 'classes/wc-api-dev/class-wc-rest-dev-data-controller.php' ) );
+				}
 				require_once( plugin_basename( 'classes/wc-api-dev/class-wc-rest-dev-data-continents-controller.php' ) );
 				$continents = new WC_REST_Dev_Data_Continents_Controller();
 				$continents->register_routes();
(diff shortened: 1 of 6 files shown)


DanReyLop:
@jeffstieler I've copied over the base class too. It works now, even with `wc-api-dev` disabled (I guess I didn't disable the correct plugin when I was testing this).

v18:
@DanReyLop can you rebase this branch on top of `add/intl-labels`? It's hard to test this otherwise, since the `hs_tariff_code` commit isn't in here.

DanReyLop:
@v18 Done, rebased with `add/intl-labels`

woocommerce-services.php
```diff
@@ -820,8 +820,12 @@ public function rest_api_init() {
...
 		public function wc_api_dev_init() {
-			if ( ! class_exists( 'WC_REST_Dev_Data_Continents_Controller' ) ) {
-				require_once( plugin_basename( 'classes/wc-api-dev/class-wc-rest-dev-data-controller.php' ) );
+			$rest_server = rest_get_server();
+			$existing_routes = $rest_server->get_routes();
+			if ( ! isset( $existing_routes['/wc/v3/data/continents'] ) ) {
+				if ( ! class_exists( 'WC_REST_Dev_Data_Continents_Controller' ) ) {
+					require_once( plugin_basename( 'classes/wc-api-dev/class-wc-rest-dev-data-controller.php' ) );
+				}
 				require_once( plugin_basename( 'classes/wc-api-dev/class-wc-rest-dev-data-continents-controller.php' ) );
```
jeffstieler: Shouldn't this be inside of the `class_exists()` check?

DanReyLop: It doesn't need to. If the `'/wc/v3/data/continents'` route doesn't exist, then neither `wc-api-dev` or the `v3` API exist.

In fact, the `class_exists` check isn't needed at all, I've just removed it in https://github.com/Automattic/woocommerce-services/pull/1448/commits/45a67a089627623fab8970aa8513b76af958e14a

PR #1451, opened 2018-06-28
Stub out wpcom middleware to prune unneeded code from the compiled JS

marcinbot:
`wpcom-api-middleware.js` file in Calypso imports a lot of other middlewares that we have no use for. They import other files and bloat our bundle size and translatable strings. An easy way of checking this is to run `npm run i18n` and checking the `i18n/strings.php` file. It will contain a lot of strings that are not related to our project.

Stubbing `wpcom-api-middleware.js` and `wpcom-http.js` with only the minimum of what we need reduced the bundle size from 2.1MB to 1.5MB. Also, only relevant strings now show as an output of the `i18n` script.

To test:
* restart the webpack server if it's already running
* verify that the label modal still shows the price that the customer selected at the checkout
* verify that the shipping settings (payment method and packages) still save after making some changes
* verify that the shipping service settings still save


client/calypso-stubs/state/data-layer/wpcom-api-middleware.js
```diff
@@ -0,0 +1,95 @@
+/** @format */
+
```
jeffstieler: Since this is largely a copy-paste from https://github.com/Automattic/wp-calypso/blob/master/client/state/data-layer/wpcom-api-middleware.js, can you comment as such - pointing out the modifications? _(like we do with TaxJar)_

client/calypso-stubs/state/data-layer/wpcom-api-middleware.js
```diff
@@ -0,0 +1,95 @@
...
+ * Internal dependencies
+ */
+
+import { bypassDataLayer } from 'state/data-layer/utils';
+import { mergeHandlers } from 'state/action-watchers/utils';
+import httpHandlers from 'state/http';
+
+const mergedHandlers = mergeHandlers(
+	httpHandlers,
+);
```
jeffstieler: Do we need to call `mergeHandlers()` with only one argument?

client/calypso-stubs/state/data-layer/wpcom-api-middleware.js
marcinbot: Added in 15c5ba4

client/calypso-stubs/state/data-layer/wpcom-api-middleware.js
marcinbot: Using `httpHandlers` instead of `mergedHandlers` works as well. I changed it in 102250a

client/calypso-stubs/state/data-layer/wpcom-http.js
```diff
@@ -0,0 +1,7 @@
+/**
```
DanReyLop: This file should be `calypso-stubs/state/data-layer/wpcom-http/index.js`, same path as in Calypso. If works anyway, but it's confusing.

PR #1470, opened 2018-07-31
Store the submitted product declared value and weight in the product metadata

DanReyLop:
Since we're changing the Calypso UI for the Customs form so the `weight` and `declared value` are editable by the merchant, this PR implements logic to persist those properties when the merchant gets label rates (same as with the current `tariffNumber`, `originCountry` and `description` props).

The `declared value` is stored in the product's customs meta-data, since the declared value may not match the product price, sale price, or any other price.

The `weight` property, on the other hand, updates the product's `weight`. I figured that a product's weight is the same in the store and for a Customs inspector, I don't see a use case where we wouldn't want to keep the product weight and the product "customs weight" in sync. I'm willing to be convinced otherwise.

There are testing instructions in https://github.com/Automattic/wp-calypso/pull/26423

### DIFF
--- a/classes/class-wc-rest-connect-shipping-rates-controller.php
+++ b/classes/class-wc-rest-connect-shipping-rates-controller.php
@@ -28,11 +28,11 @@ class WC_REST_Connect_Shipping_Rates_Controller extends WC_REST_Connect_Base_Con
 
 		// Update the customs information on all this order's products
 		$updated_product_ids = array();
-		foreach ( $payload[ 'packages' ] as $package ) {
+		foreach ( $payload[ 'packages' ] as $package_id => $package ) {
 			if ( ! isset( $package[ 'contents_type' ] ) ) {
 				break; // No customs information in this order, bail
 			}
-			foreach ( $package[ 'items' ] as $item ) {
+			foreach ( $package[ 'items' ] as $index => $item ) {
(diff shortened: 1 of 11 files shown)


classes/class-wc-rest-connect-shipping-rates-controller.php
```diff
@@ -43,7 +43,13 @@ public function post( $request ) {
 						'description' => $item[ 'description' ],
 						'hs_tariff_number' => $item[ 'hs_tariff_number' ],
 						'origin_country' => $item[ 'origin_country' ],
+						'value' => $item[ 'value' ],
 					) );
+					$product = wc_get_product( $item[ 'product_id' ] );
+					if ( $product && $product->get_weight() !== wc_format_decimal( $item[ 'weight' ] ) ) {
+						$product->set_weight( $item[ 'weight' ] );
+						$product->save();
```
marcinbot: Why do we update the weight on the actual product, but not the price? Also, should we update these at all? This seems like a bad UX where the products get overwritten without the user's knowledge.

jeffstieler:
After some discussion with @marcinbot, I think the changes in this PR should be abandoned. It's not clear that modifying the product's weight from the customs info form is obvious/valuable for merchants, and the price entered into the customs form could be order-specific (discounted) and probably doesn't need to persist either.

PR #1474, opened 2018-08-07
Remove "(WooCommerce Services)" from shipping method names

DanReyLop:
This PR removes the "___ (WooCommerce Services)" string from the "Add shipping method" list. It keeps it only when there is another extension that offers the shipping method for the same carrier, to not confuse the merchant.

## UI:
screen shot 2018-08-07 at 16 44 26

## Questions:
* What happens if there is only one extension that clashes with one of our methods? For example, in the screenshot, "FedEx" comes from the FedEx extension, "FedEx (WooCommerce Services)" comes from WooCommerce Services (duh), and "USPS" comes from WooCommerce Services too. Should we add the "(WooCommerce Services)" string to all of our methods in that case?
* Do we want to keep the existing descriptions? `USPS Shipping Rates, Powered by WooCommerce Services`, or do we want to hide the "WooCommerce Services" brand as much as possible?
* I've abused the current structure we have on the shipping schemas in the server, so `id: "usps", method_id: "wc_services_usps", method_title: "USPS (WooCommerce Services)", carrier_name: "USPS"`. Is that too fragile? Should we have this logic coded in the client side instead, and start relying less on the server-side schemas?

## To test: 
* Install an extension that provides a shipping method that we offer too, like `woocommerce-shipping-fedex`.
* Check that the UI when adding a shipping method to a shipping zone looks like the screenshot.

### DIFF
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -365,8 +365,8 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 				$service_settings = $this->get_service_settings( $method->method_id, $method->instance_id );
 				if ( is_object( $service_settings ) && property_exists( $service_settings, 'title' ) ) {
 					$title = $service_settings->title;
-				} else if ( is_object( $service_schema ) && property_exists( $service_schema, 'method_title' ) ) {
-					$title = $service_schema->method_title;
+				} else if ( is_object( $service_schema ) && property_exists( $service_schema, 'carrier_name' ) ) {
+					$title = $service_schema->carrier_name;
 				} else {
 					$title = _x( 'Unknown', 'A service with an unknown title and unknown method_title', 'woocommerce-services' );
 				}
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -24,7 +24,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 		 */
 		protected $api_client;
 
-		public function __construct( $id_or_instance_id = null ) {
+		public function __construct( $id_or_instance_id, $existing_shipping_methods = array() ) {
 			parent::__construct( $id_or_instance_id );
 
 			// If $arg looks like a number, treat it as an instance_id
@@ -60,7 +60,11 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 				$this->title = '';
 			} else {
 				$this->id = $this->service_schema->method_id;
-				$this->method_title = $this->service_schema->method_title;
+				// If there is already a shipping method with the "$service_schema->id" ID ("usps", "canada_post", etc),
+				// use the longer "method_title" property for the method title, which has the "XXX (WooCommerce Services)" disambiguation
+				$this->method_title = isset( $existing_shipping_methods[ $this->service_schema->id ] )
+					? $this->service_schema->method_title
+					: $this->service_schema->carrier_name;
 				$this->method_description = $this->service_schema->method_description;
 				$this->supports = array(
 					'shipping-zones',
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -24,7 +24,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 		 */
 		protected $api_client;
 
-		public function __construct( $id_or_instance_id, $existing_shipping_methods = array() ) {
+		public function __construct( $id_or_instance_id = null, $existing_shipping_methods = array() ) {
 			parent::__construct( $id_or_instance_id );
 
 			// If $arg looks like a number, treat it as an instance_id
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -68,7 +68,7 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 				);
 
 				// Set title to default value
-				$this->title = $this->service_schema->method_title;
+				$this->title = $this->service_schema->carrier_name;
 
 				// Load form values from options, updating title if present
 				$this->init_form_settings();
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -645,7 +645,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			$schemas = $schemas_store->get_service_schemas();
 
 			if ( $schemas ) {
-				add_filter( 'woocommerce_shipping_methods', array( $this, 'woocommerce_shipping_methods' ) );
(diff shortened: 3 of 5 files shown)


classes/class-wc-connect-service-settings-store.php
```diff
@@ -365,8 +365,8 @@ public function get_enabled_services_by_ids( $service_ids ) {
 				$service_settings = $this->get_service_settings( $method->method_id, $method->instance_id );
 				if ( is_object( $service_settings ) && property_exists( $service_settings, 'title' ) ) {
 					$title = $service_settings->title;
-				} else if ( is_object( $service_schema ) && property_exists( $service_schema, 'method_title' ) ) {
-					$title = $service_schema->method_title;
+				} else if ( is_object( $service_schema ) && property_exists( $service_schema, 'carrier_name' ) ) {
```
v18: Where is this used? I can see the function `get_enabled_services` is called in the helper, but I don't see any mention of the service names on the helper page.

classes/class-wc-connect-shipping-method.php
```diff
@@ -60,15 +60,19 @@ public function __construct( $id_or_instance_id = null ) {
...
+					: $this->service_schema->carrier_name;
 				$this->method_description = $this->service_schema->method_description;
 				$this->supports = array(
 					'shipping-zones',
 					'instance-settings'
 				);
 
 				// Set title to default value
-				$this->title = $this->service_schema->method_title;
+				$this->title = $this->service_schema->carrier_name;
```
v18: If another plugin is added after a WCS shipping method is added, then the title will still show the carrier name instead of the method title, for example with `USPS`:

screen shot 2018-08-13 at 4 44 37 pm
screen shot 2018-08-13 at 4 44 31 pm

Those are WCS shipping methods, and can be confusing if USPS is installed later.

Probably not a huge issue, and would require a lot of work to fix (dynamically getting the method title when displaying), so ok without fixing it for now.

woocommerce-services.php
```diff
@@ -647,8 +645,7 @@ public function attach_hooks() {
 			$schemas = $schemas_store->get_service_schemas();
 
 			if ( $schemas ) {
-				add_filter( 'woocommerce_shipping_methods', array( $this, 'woocommerce_shipping_methods' ) );
-				add_action( 'woocommerce_load_shipping_methods', array( $this, 'woocommerce_load_shipping_methods' ) );
+				add_filter( 'woocommerce_shipping_methods', array( $this, 'woocommerce_shipping_methods' ), 9999 );
```
v18: Why `9999`?

woocommerce-services.php
```diff
@@ -1071,26 +1049,12 @@ public function woocommerce_shipping_methods( $shipping_methods ) {
...
 
 			return $shipping_methods;
 		}
 
-		/**
-		 * Registers shipping methods for use in things like the Add Shipping Method dialog
-		 * on the Shipping Zones view
-		 *
-		 */
-		public function woocommerce_load_shipping_methods() {
```
v18: Why do we no longer need to load shipping methods?

v18:
Tested with the `USPS` extensions, works well 👌. 

> Should we add the "(WooCommerce Services)" string to all of our methods in that case?

Nah, I think the explanatory text ("powered by WCS" at the bottom of the screenshot) should suffice for now.

> Do we want to keep the existing descriptions? USPS Shipping Rates, Powered by WooCommerce Services

I don't think we should hide "WooCommerce Services" since that's the name of the plugin that's installed on users' sites that's powering this. However, I would update the capitalization since title case isn't appropriate here:

`USPS shipping rates, powered by WooCommerce Services`

Maybe add a plugin blurb?

`USPS shipping rates, powered by the WooCommerce Services plugin`

> I've abused the current structure we have on the shipping schemas in the server

What do you mean by abused? 

> Should we have this logic coded in the client side instead, and start relying less on the server-side schemas?

I think we should keep it server-side.

woocommerce-services.php
DanReyLop: Just so the filter is executed after all other shipping methods have been loaded. Would `PHP_INT_MAX` for example make it clearer?

classes/class-wc-connect-service-settings-store.php
DanReyLop: In the Self-Help page (`WooCommerce -> Status -> WooCommerce Services`) all the WCS configured methods are listed, and for each one it shows the shipping method name (its title), last successful rate fetch timestamp, etc. Here: https://github.com/Automattic/woocommerce-services/blob/ee8b7445c3d8d4b1d893d7938ab6f0ec0553b5d8/classes/class-wc-connect-help-view.php#L202

DanReyLop:
> What do you mean by abused?

This is the schema returned by the WCS server for one of our shipping methods:

```js
{
		id: 'fedex',
		method_id: 'wc_services_fedex',
		method_description: 'FedEx Shipping Rates, Powered by WooCommerce Services',
		method_title: 'FedEx (WooCommerce Services)',
		carrier_name: 'FedEx',
		...
}
```

So, as it is right now, I'm following these assumptions:
* The `id` field is the ID of the shipping method as provided by other extensions. The FedEx extension's shipping method has the `fedex` ID, Canada Post has `canada_post`, and USPS has `usps`, so I'm using that `id` field to detect if there's an installed extension that clashes with our methods.
* `carrier_name` is the human-readable name for the carrier, and `method_title` is the human-readable name with the `(WCS)` disambiguation.

These assumptions happen to be correct right now, but it feels somewhat abusing the current schema structure.

woocommerce-services.php
v18: Nope, it's all good in this PR. I am just not familiar with the different filters so wasn't sure what we were removing.

woocommerce-services.php
v18: The `9999` thing got me wondering what other filters we could use that would remove the need for setting priorities manually.

Do you think using `woocommerce_shipping_method_title` would work? I found it in the [list of WC hooks](https://docs.woocommerce.com/wc-apidocs/hook-docs.html).

Here's some reasoning on why I went digging:
The way I'm reading the WC core code, there's a way that even if we're the last filter, it may not work in some cases is because `woocommerce_shipping_methods` is used in `get_shipping_method_class_names` (https://docs.woocommerce.com/wc-apidocs/source-class-WC_Shipping.html#131). `get_shipping_method_class_names` is  used in `load_shipping_methods` (https://docs.woocommerce.com/wc-apidocs/source-class-WC_Shipping.html#156) where more shipping methods can be loaded after using `woocommerce_load_shipping_methods`, so we may not be getting all of them even if we're the last ones using the filter. It's a bit edgecase, so if another filter doesn't work, I think we can go with what you already have here.

v18:
> so I'm using that id field to detect if there's an installed extension that clashes with our methods

Would it make sense to check against the method titles, instead of their IDs? What we're trying to make sure if that what the user sees (method titles) is not duplicated.

v18:
> carrier_name is the human-readable name for the carrier, and method_title is the human-readable name with the (WCS) disambiguation.

I think using those for human-readable titles is exactly the right usage, so we don't need to change anything there. Maybe we could consider changing the naming of the properties in the API, but I think they make sense for now.

PR #1489, opened 2018-08-24
Do not log invalid tax to_zip errors and show them to the customer

### LINKED ISSUE #1427: Report automated tax "Bad Request" error to customer
When the postcode and state don't match (e.g. `NY`, `90210`), the shipping rate request works fine since it only uses the postcode, but the tax rate request fails with such an error:

```
Error retrieving the tax rates. Received (400): {"error":"Bad Request","detail":"to_zip 90210 is not used within to_state NY","status":400} (WCS Tax)
```

Can reproduce from cart or checkout page – the above is logged but the front-end is unaffected.

Since this often leads to an incorrect tax rate, and since we have the error on the backend, there's a chance to provide a better experience for the customer in this case.

We have https://github.com/Automattic/woocommerce-services/issues/1313 open for showing notices in debug mode, but we could go further along those lines, and a) show errors regardless of debug mode, and b) possibly block checkout on `4xx` errors. (May want to restrict this to errors pertaining to `to_` fields.)

Thanks @melindahelt for surfacing this and reporting user expectations.

marcinbot:
Fixes #1427 

To test:
* Enable automated taxes
* Make sure the "An error occurred in WooCommerce Services" notice is closed in wp-admin
* On checkout, enter a mismatching zip/state combination
* A notice should appear on the checkout page (see below) and proceeding with the checkout should not be possible
* No notice should appear in wp-admin and no error should be logged in the tax logs on the status page

screen shot 2018-08-24 at 13 14 15

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -300,7 +300,10 @@ class WC_Connect_TaxJar_Integration {
 			} else {
 				$message = sprintf( _x( 'Invalid %s entered.', '%s - Postcode/Zip checkout field label', 'woocommerce-services' ), $postcode_field_name );
 			}
-			wc_add_notice( $message, 'error' );
+
+			if ( ! wc_has_notice( $message, 'error' ) ) {
+				wc_add_notice( $message, 'error' );
+			}
 
 			return;
 		}
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -286,8 +286,8 @@ class WC_Connect_TaxJar_Integration {
 		$formatted_message = is_scalar( $message ) ? $message : json_encode( $message );
 
 		//ignore error messages caused by customer input
-		$state_zip_mismatch = 1 === preg_match( '/to_zip [^"}]* is not used within to_state/', $formatted_message );
-		$invalid_postcode = 1 === preg_match( '/isn\'t a valid postal code for/', $formatted_message );
+		$state_zip_mismatch = false !== strpos( $formatted_message, 'to_zip' ) && false !== strpos( $formatted_message, 'is not used within to_state' );
+		$invalid_postcode = false !== strpos( $formatted_message, 'isn\'t a valid postal code for' );
 		if ( $state_zip_mismatch || $invalid_postcode ) {
 			$fields = WC()->countries->get_address_fields();
 			$postcode_field_name = __( 'Postcode / ZIP', 'woocommerce-services' );
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -288,7 +288,7 @@ class WC_Connect_TaxJar_Integration {
 		//ignore error messages caused by customer input
 		$state_zip_mismatch = false !== strpos( $formatted_message, 'to_zip' ) && false !== strpos( $formatted_message, 'is not used within to_state' );
 		$invalid_postcode = false !== strpos( $formatted_message, 'isn\'t a valid postal code for' );
-		if ( $state_zip_mismatch || $invalid_postcode ) {
+		if ( ! is_admin() && ( $state_zip_mismatch || $invalid_postcode ) ) {
 			$fields = WC()->countries->get_address_fields();
 			$postcode_field_name = __( 'Postcode / ZIP', 'woocommerce-services' );
 			if ( isset( $fields['billing_postcode'] ) && isset( $fields['billing_postcode']['label'] ) ) {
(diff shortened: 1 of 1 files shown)


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -285,6 +285,12 @@ public function _log( $message ) {
 	public function _error( $message ) {
 		$formatted_message = is_scalar( $message ) ? $message : json_encode( $message );
 
+		//ignore error messages caused by customer input
+		if ( 1 === preg_match( '/to_zip .* is not used within to_state/', $formatted_message ) ) {
```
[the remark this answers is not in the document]
marcinbot: I replaced it with `[^"}]*` in 19696da so it should only match within one JSON string field

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -285,6 +285,12 @@ public function _log( $message ) {
...
+			wc_add_notice( __( 'ZIP does not match the selected state.', 'woocommerce-services' ), 'error' );
```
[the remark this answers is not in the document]
marcinbot: Added in 19696da

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -286,8 +286,22 @@ public function _error( $message ) {
 		$formatted_message = is_scalar( $message ) ? $message : json_encode( $message );
 
 		//ignore error messages caused by customer input
-		if ( 1 === preg_match( '/to_zip .* is not used within to_state/', $formatted_message ) ) {
-			wc_add_notice( __( 'ZIP does not match the selected state.', 'woocommerce-services' ), 'error' );
+		$state_zip_mismatch = 1 === preg_match( '/to_zip [^"}]* is not used within to_state/', $formatted_message );
```
jeffstieler: Since the error message is something like "to_zip 20112 is not used within to_state AL" I don't see a reason to try and detect the end of the JSON encoded string. Can we just look for alphanumeric characters and whatever other special characters are allowed for postcodes? Something like `[\w\-]+`?

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -286,8 +286,22 @@ public function _error( $message ) {
...
+			$fields = WC()->countries->get_address_fields();
+			$postcode_field_name = __( 'Postcode / ZIP', 'woocommerce-services' );
+			if ( isset( $fields['billing_postcode'] ) && isset( $fields['billing_postcode']['label'] ) ) {
+				$postcode_field_name = $fields['billing_postcode']['label'];
+			}
+
+			if ( $state_zip_mismatch ) {
+				$message = sprintf( __( '%s does not match the selected state.', '%s - Postcode/Zip checkout field label', 'woocommerce-services' ), $postcode_field_name );
+			} else {
+				$message = sprintf( __( 'Invalid %s entered.', '%s - Postcode/Zip checkout field label', 'woocommerce-services' ), $postcode_field_name );
```
jeffstieler: These calls to [`__()`](https://developer.wordpress.org/reference/functions/__/) seem to be incorrect - the second argument is the text domain. Were these second arguments meant to be comments for translators?

classes/class-wc-connect-taxjar-integration.php
marcinbot: I think we are overthinking this. I'm trying to follow the general logic from the [core postcode validator](https://github.com/woocommerce/woocommerce/blob/d0491072e8a7d0a4535db0ec2bae4a6d43ff64fd/includes/class-wc-validation.php#L47).  While `[\w\-]+` would catch all the postcode formats that core checks against, the validation also returns true for unimplemented country/format combinations.

The changes in this PR are too simple to go through every post code format in the world to include every possible character to match against. A very simple regex that would not return false positives, and a simple error message should be sufficient.

classes/class-wc-connect-taxjar-integration.php
marcinbot: Fixed in d9a8990

classes/class-wc-connect-taxjar-integration.php
jeffstieler: Perhaps we're overthinking it. I'm really just trying to keep an eye on performance. Since this isn't critical functionality, what do you think about using `strpos()` instead of `preg_match()` for both checks since it's faster?

marcinbot: Performance is a very valid point - I replaced regex with `strpos()` in 7e6c66f

jeffstieler:
The code looks good here, but I seem to be getting a multiple errors showing in cart/checkout:

screen shot 2018-08-31 at 9 34 35 am

I also tested using a bad zip code in a "backend" order and didn't see any sort of error messaging. Is that expected?

PR #1494, opened 2018-08-30
Control the TaxJar plugin using hooks

marcinbot:
See https://github.com/taxjar/taxjar-woocommerce-plugin/pull/85

This PR intends to replace the copied code from the TaxJar plugin with an actual integration using actions and filters.

The intention is to install the TaxJar plugin alongside WCS and do the following:
* stop the API key verification logic
* stop order download
* remove nexus functionality
* rewrite API requests to hit our proxy
* add UI to “switch” to full taxjar

To test:
* Install the taxjar plugin from https://github.com/taxjar/taxjar-woocommerce-plugin/pull/85
* Verify an option is present on the taxjar settings page to configure the plugin using WCS
* Enable that option
* Automated taxes should work without having to enter the API key


classes/class-wc-connect-taxjar-control.php
```diff
@@ -0,0 +1,213 @@
...
+	public function taxjar_method_description( $description ) {
+		if ( ! $this->is_enabled() ) {
+			return $description;
+		}
+
+		return __( 'TaxJar is currently powered by WooCommerce Services. Use the controls below to disable the WooCommerce Services integration and enter your own API token (<a href="https://app.taxjar.com/api_sign_up/" target="_blank">click here to get a token</a>).', 'woocommerce-services' );
+	}
+
+	public function should_check_status( $value ) {
+		return ! $this->is_enabled();
```
DanReyLop: You should use `$value` here, just to be consistent.

classes/class-wc-connect-taxjar-control.php
```diff
@@ -0,0 +1,213 @@
...
+		WC_Connect_API_Client $api_client,
+		WC_Connect_Logger $logger
+	) {
+		$this->api_client = $api_client;
+		$this->logger = $logger;
+	}
+
+	public function init() {
+		if ( ! class_exists( 'WC_Taxjar' ) ) {
+			return;
```
DanReyLop: How are we going to handle this? Right now, the TaxJar plugin isn't auto-installed in the Onboarding Wizard. With this change, the TaxJar integration would break for existing users.

I'm ok handling this in a different PR, but I just want to point out that we can't release this change as it is now.

classes/class-wc-connect-taxjar-control.php
marcinbot: Added in 80224e7

classes/class-wc-connect-taxjar-control.php
marcinbot: We will need to handle taxjar in a similar way we do the payment gateway plugins - install the original plugin alongside WCS. Also need to handle people who are updating from an older version and had automated taxes enabled already.

These are the decisions we need to make outside of this PR, but before it is shipped.

DanReyLop:
@marcinbot Code looks good. Should we hod off testing/merging this until https://github.com/taxjar/taxjar-woocommerce-plugin/pull/85 is accepted and merged? Or until we figure out how to handle automatically installing TaxJar?

Issue #1319, opened 2018-03-03
Add flow for existing Stripe accounts in the WC Wizard


v18:
The flow can be similar to what exists on WP.com:

Step 1: User lets us know they already have an account

<img width="528" alt="user-selects-option-already-has-stripe-account" src="https://user-images.githubusercontent.com/11487924/36935497-84f3bfa8-1ec6-11e8-87c5-109ac298e70f.png">

Step 2: Stripe login + oauth confirmation

<img width="346" alt="stripe-oauth-login" src="https://user-images.githubusercontent.com/11487924/36935498-8a65e89e-1ec6-11e8-968e-5b1ce2ba2567.png">

<img width="606" alt="stripe-authorize-connection" src="https://user-images.githubusercontent.com/11487924/36935502-96e0c7c4-1ec6-11e8-94e8-528d6a5ffa66.png">

Step 3: confirmation while we set up the connection

<img width="563" alt="user-sees-completing-connection-progress-bar" src="https://user-images.githubusercontent.com/11487924/36935509-a1cf0cc2-1ec6-11e8-918e-1cf5a95f985e.png">

Step 4: success notice + settings page (? should this be in the wizard)

<img width="547" alt="stripe-success-notice-and-settings" src="https://user-images.githubusercontent.com/11487924/36935514-a8df2c72-1ec6-11e8-9bca-bee085febb30.png">

A question: is this issue redundant once we get a dashboard in wp-admin akin to the one on store on wp.com? cc @kellychoffman 

One issue is to which step in the wizard we can add it. We can probably add it right after the Payments step, by adding a new step to the wizard on the fly after the Stripe plugin is installed.

kellychoffman:
This issue isn't redundant, but instead is a good flow to work into the dashboard project. I think we should have the dashboard in place (at least a design for it) before this work gets done.

v18:
Ok, sounds good! Removing it from the hack week milestone, then!

PR #1495, opened 2018-08-30
Add Stripe banner for connecting existing accounts

dechov:
In the event that deferred Stripe account creation fails due to an account already existing with that email address, this banner will now show up in wp-admin:

screen shot 2018-08-30 at 3 26 28 pm

Clicking "Connect" will initiate the OAuth flow on stripe.com, starting with requesting certain business information if it's missing, and culminating in:

screen shot 2018-08-30 at 6 31 25 pm

(Note that if you click api.woocommerce.com, you see: https://cloudup.com/cFf9cnOhm5q)

Upon completing this flow, the user arrives at their Stripe settings page with the API keys filled in and the banner gone ([screenshot](https://cloudup.com/cvPc_EQGAMk)).

Fixes broken flow for the existing account case, which is now explicitly supported as of https://github.com/woocommerce/woocommerce/pull/21224. Should address https://github.com/Automattic/woocommerce-services/issues/1319 in part. Based on static design in p7bbVw-2dB-p2.

To test:
- In the Payments step of the onboarding wizard (`/wp-admin/index.php?page=wc-setup&step=payment`), enable Stripe, check the checkbox, enter the email for an existing Stripe account that you can log into, and "Continue"
- Back in the wp-admin dashboard, see the Stripe banner on any of these screens: Dashboard, Plugins, WooCommerce » Settings » Payments, WooCommerce » Extensions » Payments
 click "Connect"
- After returning from Stripe, verify that the publishable key and secret key are filled ("test" keys if using staging server, "live" keys if using production server)

Concerns:
- Do any failure cases call for special handling?
- Could / should we have a special `stripe_state` value instead of adding a new `banner_stripe` option?


kellychoffman:
> Just a thought on the copy – would "you'll need to" be preferable to "you need to"?

Yeah, good idea.

dechov:
👍

screen shot 2018-09-04 at 5 44 54 pm

kellychoffman:
Nice.

dechov:
Added a one-time success notice upon returning from OAuth flow and obtaining keys. Copy suggestions welcome.

> **Ready to accept Stripe payments!** Your Stripe account has been successfully connected. Additional settings can be configured on this screen.

screen shot 2018-09-10 at 1 04 08 pm

(Not sure what failure handling might be needed, but currently it just keeps the banner up if the keys weren't successfully retrieved.)

dechov:
> Something appears to be wrong 

May have just needed to re-run `npm start`, but it works fine after rebase.

jeffstieler:
How difficult would it be to show a slightly different banner message for account creations that fail due to an existing account? (My inner monologue asks "I thought an account was being created for me? Why do I need to connect one?)

jeffstieler:
> we could acknowledge that we know [given email] already has an account associated with it

+1

cc @kellychoffman for any design input - this banner only shows if an email already exists for a Stripe account.. the user will have checked a box that says "Set up Stripe for me using this email".

jeffstieler:
How about this for the banner text:

> **Connect your Stripe account**
> 
> Good news - your email [email] already has a Stripe account! Click connect to activate Stripe payments on your store.
> 
> [Connect]

dechov:
Thanks, I think it's a good start.

> Good news [...]

In my view this isn't ideal for those who are already well aware they have a Stripe account, which may well be a majority.

> [...] Click connect to [...]

I'd avoid "click" – since it implicitly assumes use of a mouse – in favor of just focusing on the action ("Connect to [...]" or "Connect your account in order to [...]").

> [...] to activate Stripe payments on your store.

There's something I prefer about "To start accepting payments with Stripe", I think because it sounds a little more clear to me about what will change when you connect.

How's this for an alternative:

> It looks like there is an existing Stripe account at [email]. To start accepting payments with Stripe, you'll need to connect it to your site.

jeffstieler:
> How's this for an alternative:

Looks good - but can we say store instead of site?

dechov:
Done:

screen shot 2018-09-20 at 3 53 50 pm

classes/class-wc-connect-stripe.php
```diff
@@ -156,5 +162,33 @@ private function get_default_stripe_config() {
 
 			return $result;
 		}
+
+		public function maybe_show_banner() {
+			if ( 'yes' !== WC_Connect_Options::get_option( 'banner_stripe', null ) ) {
+				return;
+			}
+
+			if ( isset( $_GET[ 'wcs_stripe_code' ] ) ) {
```
jeffstieler: Should there be an `isset()` check for `wcs_stripe_state` as well?

woocommerce-services.php
```diff
@@ -543,8 +543,12 @@ public function init_core_wizard_payments_config() {
 				$response = $this->stripe->create_account( $email, $country );
 
 				if ( is_wp_error( $response ) ) {
-					// TODO handle case of existing account
-					$this->logger->debug( $response, __CLASS__ );
+					// Handle existing account case.
+					if ( false !== strpos( $response->get_error_message(), 'Account already exists for the provided email.' ) ) {
+						WC_Connect_Options::update_option( 'banner_stripe', 'yes' );
+					} else {
+						$this->logger->log( $response, __CLASS__ );
```
jeffstieler: What do you think about logging even when the message triggers the banner?

assets/stylesheets/banner.scss
```diff
@@ -19,6 +19,16 @@
 			width: 220px;
 			height: 121px;
 		}
+
+		&.is-compact {
```
jeffstieler: Can you elaborate on why you went with this generic styling versus something Stripe specific? Do you see the potential for many "compact logo" banners where the logo should be `100px`?

classes/class-wc-connect-paypal-ec.php
```diff
@@ -184,13 +184,13 @@ public function maybe_show_banner() {
...
-				|| ( // WooCommerce settings.
+				|| ( // WooCommerce payments settings.
 					'woocommerce_page_wc-settings' === $screen->base
 					&& isset( $_GET['tab'] ) && 'checkout' === $_GET['tab']
 					)
-				|| ( // WooCommerce payment gateway extension page
+				|| ( // WooCommerce payment gateway extension page.
 					'woocommerce_page_wc-addons' === $screen->base
-					&& isset( $_GET['section'] ) && 'payment_gateways' === $_GET['section']
+					&& isset( $_GET['section'] ) && 'payment-gateways' === $_GET['section']
```
jeffstieler: Nice catch. 👍

classes/class-wc-connect-stripe.php
```diff
@@ -163,14 +163,24 @@ private function get_default_stripe_config() {
...
+		public function maybe_show_notice() {
+			$setting = WC_Connect_Options::get_option( 'banner_stripe', null );
+			if ( is_null( $setting ) ) {
 				return;
 			}
 
-			if ( isset( $_GET[ 'wcs_stripe_code' ] ) ) {
-				$this->connect_oauth( $_GET[ 'wcs_stripe_state' ], $_GET[ 'wcs_stripe_code' ] );
+			if ( 'success' === $setting ) {
+				add_action( 'admin_notices', array( $this, 'connection_success_notice' ) );
```
jeffstieler: Could `wc_add_notice()` be used here, or is the styling different?

classes/class-wc-connect-stripe.php
```diff
@@ -206,5 +216,13 @@ public function banner() {
 				'compact_logo'   => true,
 			) );
 		}
+
+		public function connection_success_notice() {
+			?>
+				<div class="notice notice-success">
+					<p><?php echo wp_kses( '<strong>Ready to accept Stripe payments!</strong> Your Stripe account has been successfully connected. Additional settings can be configured on this screen.', array( 'strong' => array() ) ); ?></p>
```
jeffstieler: This is missing translation.

classes/class-wc-connect-stripe.php
```diff
@@ -205,11 +205,18 @@ public function maybe_show_notice() {
 		}
 
 		public function connection_banner() {
+			$result = $this->get_oauth_url( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe' ) );
+
+			if ( is_wp_error( $result ) ) {
+				$this->logger->log( $result, __CLASS__ );
```
jeffstieler: 👍

classes/class-wc-connect-stripe.php
```diff
@@ -218,9 +218,14 @@ public function connection_banner() {
...
 			}
 
+			$options = get_option( self::SETTINGS_OPTION, array() );
+
 			$this->nux->show_nux_banner( array(
-				'title'          => __( 'Connect your account to activate Stripe', 'woocommerce-services' ),
-				'description'    => esc_html( __( 'To start accepting payments with Stripe, you\'ll need to connect your site to a Stripe account.', 'woocommerce-services' ) ),
+				'title'          => __( 'Connect your account', 'woocommerce-services' ),
+				'description'    => wp_kses(
+					sprintf( __( 'It looks like there is an existing Stripe account at <strong>%s</strong>. To start accepting payments with Stripe, you\'ll need to connect it to your store.', 'woocommerce-services' ), $options['email'] ),
```
jeffstieler: We should check for the `email` key before using it to avoid PHP warnings.

classes/class-wc-connect-paypal-ec.php
dechov: Yeah, I'm pretty sure it changed at some point on WC side, which can happen at any time since it loads those from a remote endpoint.

classes/class-wc-connect-stripe.php
dechov: Isn't that only for front-end notices?

jeffstieler: But perhaps `WC_Admin_Notices::add_notice()` might reduce some code duplication?

woocommerce-services.php
dechov: Done in 6b4bed8

classes/class-wc-connect-stripe.php
dechov: I didn't anticipate a case where the banner option would be set but the `email` wouldn't be present, but it's good to add a check. Done in 6b4bed8

classes/class-wc-connect-stripe.php
dechov: Yeah, looks like `WC_Admin_Notices::add_custom_notice` is a good way to add a persistent/dismissible notice, but I'd wired up the current notice to only appear when redirected from stripe.com and not on subsequent page loads – open to feedback on revising that.

classes/class-wc-connect-stripe.php
dechov: Whoops, done in 6069e5a

classes/class-wc-connect-stripe.php
dechov: Done in 6b4bed8

woocommerce-services.php
dechov: Ah, I originally decided not to, as the way the current flow is, it's not really an error from the merchant's perspective. But of course, this isn't strictly an error log. Changed in 922dc1d

assets/stylesheets/banner.scss
dechov: In concept, there's nothing Stripe specific about `100px`, so I thought there was potential to easily use it for other smaller graphics. If there's another "compact" graphic in the future that can't use this rule, it could be differentiated in the future.

That was my thinking – do you think specifically evoking `stripe` in these rules would be preferable?

dechov:
Rebased on `master`, now that https://github.com/Automattic/woocommerce-services/pull/1497 is in. This makes testing this PR easier, in that there is now UI for verifying and disconnecting the connected account.

PR #1497, opened 2018-08-31
Import connected account UI into Stripe settings

dechov:
screen shot 2018-08-30 at 7 14 51 pm

Imports the connected account view from Calypso ([screenshot of original Calypso context](https://cloudup.com/c4X-8pqE1rT)). Changes contain some style tweaks to match appearance in Calypso as well as static design in p7bbVw-2dB-p2.

A page refresh is triggered after "Disconnect" so that the keys show up as cleared.

Example for a newly created account:

screen shot 2018-08-31 at 1 19 27 pm

Implementation concern: in order to make the UI render here, I filtered the Stripe settings to add a title, the description for which was the container into which WooCommerce Services renders the React tree. This isn't ideal because:
  - <s>We can't pass data since `data-args` is filtered out by `kses` so we have to fetch the account details on first render even if we're fetching them earlier (and a small change to `main.js` was necessary to account for the missing attribute)</s>
  - A small change to the `wp_register_script` call was necessary, in order to have the script in the footer
  - We may not really want the title, as it's larger than what was designed and so appears to include settings lower down in its scope

We might be able to add an action in the Stripe plugin to be done before the settings table, though I'm not sure how. _Edit:_ overriding and mostly duplicating [`admin_options`](https://github.com/woocommerce/woocommerce/blob/dc55b9b002be6e102aca47a06e19c4e952d514de/includes/abstracts/abstract-wc-payment-gateway.php#L177-L186) is the only thing that comes to mind.

TODO:
- [x] Fix misalignment between "Disconnect" button and "Disconnecting" indicator
- [x] Indicate page refresh more clearly
- [ ] Perhaps revert pre-fetching account data, so as to not block page load

### DIFF
--- a/classes/class-wc-connect-stripe.php
+++ b/classes/class-wc-connect-stripe.php
@@ -173,25 +173,11 @@ if ( ! class_exists( 'WC_Connect_Stripe' ) ) {
 		}
 
 		/**
-		 * Make kses exception for attribute containing preloaded state.
-		 */
-		public function allow_args_attribute( $allowed ) {
-			$allowed['div']['data-args'] = true;
-			return $allowed;
-		}
-
-		/**
 		 * Add to settings page container for dynamically rendered connected account view.
 		 */
 		public function show_connected_account( $settings ) {
-			$account_details = $this->get_account_details();
-			if ( is_wp_error( $account_details ) ) {
-				return $settings;
-			}
-
 			ob_start();
-			do_action( 'enqueue_wc_connect_script', 'wc-connect-stripe-connect-account', $account_details );
-			add_filter( 'wp_kses_allowed_html', array( $this, 'allow_args_attribute' ) );
+			do_action( 'enqueue_wc_connect_script', 'wc-connect-stripe-connect-account' );
 
 			$new_settings = array(
 				'connection_status' => array(
--- a/classes/class-wc-connect-stripe.php
+++ b/classes/class-wc-connect-stripe.php
@@ -156,5 +156,23 @@ if ( ! class_exists( 'WC_Connect_Stripe' ) ) {
 
 			return $result;
 		}
+
+		public function show_connected_account( $settings ) {
+			if ( is_wp_error( $this->get_account_details() ) ) {
+				return $settings;
+			}
+
+			ob_start();
+			do_action( 'enqueue_wc_connect_script', 'wc-connect-stripe-connect-account' );
+
+			$new_settings = array(
+				'connection_status' => array(
+					'type'        => 'title',
+					'title'       => __( 'Connected Stripe account', 'woocommerce-gateway-stripe' ),
+					'description' => ob_get_clean(),
+				),
+			);
+			return array_merge( $new_settings, $settings );
+		}
 	}
 }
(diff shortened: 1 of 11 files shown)


jeffstieler:
> We can't pass data since data-args is filtered out by kses so we have to fetch the account details on first render even if we're fetching them earlier (and a small change to main.js was necessary to account for the missing attribute)

Is there a chance to use the `wp_kses_allowed_html` filter to allow `data-args` ?

dechov:
> Is there a chance to use the `wp_kses_allowed_html` filter to allow `data-args` ?

Ah, thanks! Done.

dechov:
The filter is needed for the eventual settings form output, rather than the output buffering of the form field description. If I can work out the proper action on which to do it, I can remove the filter right after (and perhaps add it right before) – does that sound good?

jeffstieler:
> If I can work out the proper action on which to do it, I can remove the filter right after (and perhaps add it right before) – does that sound good?

Sounds good.

> Perhaps revert pre-fetching account data, so as to not block page load

Is the worst case scenario a FOUC for the length of a WP remote request timeout? If so, we should revert to a post-page-load fetch.

dechov:
Not sure if it's exactly what you mean, but the loading of the whole settings screen is prolonged by the length of the external request – ideally we should avoid that, but what I forgot to mention is that I was originally making that call just to know if there's a connected account at all. Can get around that by handling it in the view code: "No account connected" message, followed by https://github.com/Automattic/woocommerce-services/issues/1500 in future.

jeffstieler:
Ah ok, so it seems like the best of both worlds to avoid the connected account request that blocks page render. 👍

dechov:
No longer loading the account data before the client renders. State when no account is connected is now handled in the UI, and account deactivation state is handled more explicitly.

stripe-connect-deactivation

Some minor copy/design concerns:
- The heading looks like it applies to the settings below, but it doesn't. Could solve this by giving those settings their own heading, in this case where we have something rendered before them. "General"?
- Is it OK that it still says "Sign up for a Stripe account, and get your Stripe account keys." in the description at top? Is there anything we can do about it, considering that (as of latest commits) we no longer know if an account is connected before page render?

classes/class-wc-connect-stripe.php
```diff
@@ -156,5 +171,22 @@ private function get_default_stripe_config() {
...
+		 * Add to settings page container for dynamically rendered connected account view.
+		 */
+		public function show_connected_account( $settings ) {
+			ob_start();
+			do_action( 'enqueue_wc_connect_script', 'wc-connect-stripe-connect-account' );
+
+			$new_settings = array(
+				'connection_status' => array(
+					'type'        => 'title',
+					'title'       => __( 'Connected Stripe account', 'woocommerce-gateway-stripe' ),
```
jeffstieler: The translation domain should be `woocommerce-services`.

client/apps/stripe-connect-account/style.scss
```diff
@@ -0,0 +1,30 @@
...
+		vertical-align: bottom;
+	}
+}
+
+.stripe-connect-account__placeholder-container {
+	@extend .stripe__connect-account;
+
+	.stripe-connect-account__placeholder-body {
+		@extend .stripe__connect-account-body;
+		@include placeholder();
```
jeffstieler: TIL! 👍

Issue #1315, opened 2018-02-23
Automatically select newly added credit card


dechov:
If the merchant added a card, we can probably assume that's the one they intend to use.

If a card wasn't already selected (or if the selected one got removed), we can almost certainly assume it.

We should automatically select an added card in the list of payment methods after it's added, to save a small intermediate step for the merchant, and also to remove the risk of confusion due to not noticing a card was added but not actually selected.

(If multiple cards were added at once — no longer possible in Calypso, only in wp-admin — we can either pick one or leave the selection unchanged.)

msdesign21:
> Should we go further and "Save" this selection automatically
As long as it's clear what card is selected, I think this would be ideal.

Abdalsalaam:
Shipping related work now happens in WooCommerce Shipping and we recommend using it: https://wordpress.org/plugins/woocommerce-shipping/

Issue #1073, opened 2017-06-29
Confusing UX when activating WCS with card already on file


pmaiorana:
I already have a credit card on file with WordPress.com, but after activating WCS, Jetpack, etc. when I go to an Order page I see this:

<img src="https://cldup.com/KmdwFh-z6K.png" width="450">

When I click the "Add a credit card" link it takes me here—but the "Save changes" button is disabled, so I don't know what to do.

<img src="https://cldup.com/g5XQkixwwB.png" width="450">

After toggling some settings I can now "Save changes", and can then proceed to printing labels.

It sounds like, from a discussion with @v18, we need the user to confirm that they want to use the credit card on file.

Ideally we could confirm that right from the Order page—but as a baby-step in that direction, if that's not feasible, it would be great to at least change the copy in the first screenshot to acknowledge that we do already have a card on file, and do something to address the Save issue so they can at least confirm that's the credit card they'd like to use.

PR #1507, opened 2018-09-24
Automatically select credit card if just one is available

dechov:
Context: the first available credit card was automatically selected...
- ...if there wasn't a selected card that was still available, as of https://github.com/Automattic/woocommerce-services/pull/488
- ...only if a previously selected card is no longer available, as of https://github.com/Automattic/woocommerce-services/pull/1076

This PR reverts the primary change in https://github.com/Automattic/woocommerce-services/pull/1076 – again automatically selecting a card so there are never available cards with none selected – as it is unnecessary for fixing the main issue in https://github.com/Automattic/woocommerce-services/issues/1073. _Edit:_ **now only does so there's just one card available.**

(I see now that @v18 was mentioned in that issue as the source for "we need the user to confirm that they want to use the credit card on file", so CC in case there's a reason not to do this that's being overlooked.)

<s>Addresses https://github.com/Automattic/woocommerce-services/issues/1315 (at its root)</s> _edit:_ there's more to do there.
Addresses https://github.com/Automattic/wp-calypso/pull/22504#issuecomment-369946283

### DIFF
--- a/classes/class-wc-connect-payment-methods-store.php
+++ b/classes/class-wc-connect-payment-methods-store.php
@@ -72,12 +72,9 @@ if ( ! class_exists( 'WC_Connect_Payment_Methods_Store' ) ) {
 				return;
 			}
 
-			// Has the stored method ID been removed? Select the first available one
+			// Is there no stored method that's in the list? Select the first available one
 			$selected_payment_method_id = $this->service_settings_store->get_selected_payment_method_id();
-			if (
-				$selected_payment_method_id &&
-				! in_array( $selected_payment_method_id, $payment_method_ids )
-			) {
+			if ( ! in_array( $selected_payment_method_id, $payment_method_ids ) ) {
 				$this->service_settings_store->set_selected_payment_method_id( $payment_method_ids[ 0 ] );
 			}
 		}
--- a/classes/class-wc-connect-payment-methods-store.php
+++ b/classes/class-wc-connect-payment-methods-store.php
@@ -72,9 +72,12 @@ if ( ! class_exists( 'WC_Connect_Payment_Methods_Store' ) ) {
 				return;
 			}
 
-			// Is there no stored method that's in the list? Select the first available one
+			// Has the stored method ID been removed, or is there only one available? Select the first available one
 			$selected_payment_method_id = $this->service_settings_store->get_selected_payment_method_id();
-			if ( ! in_array( $selected_payment_method_id, $payment_method_ids ) ) {
+			if (
+				( $selected_payment_method_id || 1 === count( $payment_method_ids ) ) &&
+				! in_array( $selected_payment_method_id, $payment_method_ids )
+			) {
 				$this->service_settings_store->set_selected_payment_method_id( $payment_method_ids[ 0 ] );
 			}
 		}


v18:
Can you list out some of the user scenarios where this change would affect UX? For example, would the following work:

User already has a WP.com account with a credit card on file, they install a new WP site and connect to JP. Then they install WC and go through the onboarding wizard, selecting a WCS feature. WCS is installed and they agree to WCS's TOS. They are able to immediately purchase a label using the one credit card on file, without having to manually go to the credit card page and select it.

I'm not 100% sure we should use a user's credit card without being more explicit.

At the very least, we should only pre-select the card if there's only one card on the list and it's not selected. If there are multiple cards, we really don't know which the user intends to use.

dechov:
@v18 That's right – this scenario should work, along with any other way of arriving at Jetpack + WCS with a card already added, and I see it as the main benefit of this change.

From what I've seen so far, I don't really understand why we shouldn't pre-select the first card among multiple. Not precisely the same flow, but the WP.com checkout screen defaults the selection to the first (i.e. most recently added) available card, which it will use if you take no other action before proceeding. If what we selected isn't the one which the user intends to use, they can change it based on the selected card notice already in place.

I do agree we should be more explicit about which card is used, by having it indicated – or better yet, configurable – as the last element of the label purchase flow before the purchase button. I think this could be a follow-on PR, but I also wouldn't mind if it were deemed a blocker for this change.

DanReyLop:
Can we get a consensus on this one? This seems like a good UX enhancement if we make it non-surprising for the user. I personally think putting an extra step in the Purchase Label flow just for this is too much.

Right now, this is what the user sees before getting into the Purchase Label modal:
screenshot 2018-10-30 at 22 58 05

Would you folks say that's explicit enough? If not, @v18's idea (only defaulting the credit card if it's the only one on file) seems like a reasonable compromise. Almost all new merchants will have zero or one card on file anyway.

dechov:
Agreed we should at least auto-select a card if there's only one, perhaps accompanied by a calypso PR to add the card info (if not a selector) to the bottom of the sidebar. It might be easier to implement a selector than reach consensus on whether cards should be auto-selected when there's no selector.

> _[...]_ selecting the credit card "as the last element of the label purchase flow". I don't know if that's a candidate for hack week, but if you're up for it I think it would be cool.

I'd be up for working on that or supporting anyone else wanting to work on it. https://github.com/Automattic/wp-calypso/pull/22890 was an initial attempt at this but may have been doing too much, since at the time I figured "if we want this setting in the sidebar, why not all of the settings?"

classes/class-wc-connect-payment-methods-store.php
```diff
@@ -72,10 +72,10 @@ protected function potentially_update_selected_payment_method_from_payment_metho
 				return;
 			}
 
-			// Has the stored method ID been removed? Select the first available one
+			// Has the stored method ID been removed, or is there only one available? Select the first available one
 			$selected_payment_method_id = $this->service_settings_store->get_selected_payment_method_id();
 			if (
-				$selected_payment_method_id &&
+				( $selected_payment_method_id || 1 === count( $payment_method_ids ) ) &&
```
v18: The requirements of this PR changed, so I think this code no longer works how we want it to.

From reading the PR conversation, we only want to set the payment method ID automatically if there is just one available payment method.

At this point we have the following information:

* `$payment_method_ids`: list of currently available payment method IDs
* `$selected_payment_method_id`: this the last selected payment ID. it may not exist in the list of payment IDs anymore (for example, if a user deleted it)

We want to select the payment ID automatically for the user in these two cases:
1. there was a payment ID selected before, but it's no longer available. there is only one available payment method
2. there is no currently selected payment method, and there is only one available.

For both of these we could check `1 === count( $payment_method_ids )` and if that's true, we can select the payment ID:

```
if ( 1 === count( $payment_method_ids ) {
  $this->service_settings_store->set_selected_payment_method_id( $payment_method_ids[ 0 ] );
}
```

Am I missing a scenario we want to cover?

DanReyLop: > For both of these we could check 1 === count( $payment_method_ids ) and if that's true, we can select the payment ID

Not quite. If there's a valid payment method selected, we don't want to do anything:
```php
if ( ! in_array( $selected_payment_method_id, $payment_method_ids ) && 1 === count( $payment_method_ids ) {
  $this->service_settings_store->set_selected_payment_method_id( $payment_method_ids[ 0 ] );
}
```

Otherwise I agree with your reasoning, and I can't think on any other scenario we're missing.

dechov: The other scenario is the status quo in `master`: if there was already a selected payment method, we can just select the first stored one without worrying if it's the first out of multiple. This way, it never goes from selection to no selection unless there are no stored cards.

I interpreted the conversation above to refer to expanding the auto-selection to the case where there isn't an existing selection, but not restricting the condition for the case where there is one.

I'm not necessarily opposed to only changing to a new selection if there's no doubt in which card to select – do you all think that's best?

DanReyLop: > if there was already a selected payment method, we can just select the first stored one without worrying if it's the first out of multiple.

I don't understand this. Suppose we have these available payment methods: `[1234, 5678, 9012]`, and the selected method ID is `5678`. Why would you mark "the first one" as selected? That would reset it to `1234`.

dechov: @DanReyLop Sorry, should have mentioned for completeness: as you mentioned, it would only change it if the method already selected is not available [[code](https://github.com/Automattic/woocommerce-services/blob/4a401a0a1482ed6f29245a79ea1f672cd8b9c5b7/classes/class-wc-connect-payment-methods-store.php#L78-L79)].

v18: > it would only change it if the method already selected is not available

Is this what we want, though?

That means we're blindly selecting a payment method on behalf of the user without any action on their part.

In fact I think it would doubly-strange because at some point this user had selected a different method that's no longer available. I don't see why they'd expect that when their previous method is no longer available, we'd choose a new one.

I did see that Dani mentioned this was probably ok because we say which credit card is being used before starting. I'm not so sure about this. Users who are not the credit card holders can still print labels (iirc) so that means that someone not familiar with the last 4 digits of the card may go through.

To play devil's advocate on my own arguments though: We've had things work this way for a while and heard no complaints :D. Maybe this means this does help our users by not asking them to choose a credit card every time there is more than one. Or maybe the number of users with multiple cards is small, so no one's run into a scenario they don't like.

In the off-chance this causes a problem there are steps to remediate:
0. update the selected credit card
1. re-charge the user using a BD manual payment for the amount of the receipts they don't like
2. refund the charges on the incorrect credit card

So, if people are leaning towards keeping the "choose a credit card for me out of many" behaviour, I've justified it to myself as well and let's go for it 👍

dechov: > at some point this user had selected a different method that's no longer available

Is this to say we should reconsider a) auto-selecting one card out of multiple, b) auto-selecting a card when there was a previous selection, or c) auto-selecting a card in general?

Unless it's (c), I'd be in favor of going forward with this PR as is, since (at least from my perspective), it is scoped to expanding the auto-selection criteria to the no-card-selected case, not restricting the existing criteria – for which we could open another issue.

v18: > Is this to say we should reconsider a) auto-selecting one card out of multiple, b) auto-selecting a card when there was a previous selection, or c) auto-selecting a card in general?

In that specific phrase I was referring to (b).

v18:
Works great, 🚢 

To summarize for future reference, this is the expected behaviour:

Scenario 1: no selected card yet
a. if there is only one card in the list of available cards, select it
b. if there are multiple cards, do not select any and let user decide

Scenario 2: there is a selected card already
a. if the selected card is in the list of available cards, do nothing
b. if the selected card is not in the list of available cards, select a new card from the list (the first one, list[0])

PR #1509, opened 2018-10-10
Clean-up dev tools and logic that has never quite worked

DanReyLop:
I've simplified the code around React hot-loading. By "simplified", I mean that I've removed it entirely. I don't think we've ever got hot module reloading working.

I've also removed `redbox-react` because:
* I don't recall seeing it working since a long time ago (maybe because I've done most of my dev work directly in Calypso).
* When something is obviously not working in development, we all know to open the browser console. We don't need the entire screen flashing blood-red and slapping us in the face with a full-screen error. I don't feel strongly about this, so if anyone finds it useful I'll revert this part of the PR.

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -7,8 +7,6 @@ import ReactDOM from 'react-dom';
 import { applyMiddleware, createStore, compose } from 'redux';
 import thunk from 'redux-thunk';
 import { Provider } from 'react-redux';
-import { translate as __ } from 'i18n-calypso';
-import _ from 'lodash';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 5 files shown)


client/main.js
```diff
@@ -85,55 +83,21 @@ Array.from( document.getElementsByClassName( 'wcc-root' ) ).forEach( ( container
...
+	window.addEventListener( 'beforeunload', () => {
 		const state = store.getState();
 
 		if ( window.persistState ) {
 			storageUtils.setWithExpiry( persistedStateKey, Route.getStateForPersisting( state ) );
-			return;
 		}
-
-		//this only handles the shipping service settings page. Other pages use ProtectFormGuard and Component state
-		//TODO: remove once the service settings are Calypso-fied
```
DanReyLop: This happen like a year ago :)

I can confirm that the "You have unsaved changes" prompt works in the USPS service settings page, even without this code.

Issue #1500, opened 2018-09-04
Add entry point for Stripe OAuth flow into settings


dechov:
<s>Once https://github.com/Automattic/woocommerce-services/pull/1495 and https://github.com/Automattic/woocommerce-services/pull/1497 land, there will be</s> There is an opportunity to render a "Connect" button (initiating the OAuth flow) right in the context of the Stripe settings, in the event that no Stripe account is connected.

I would assume we want to carry over the description and button from the banner...

<img width="858" alt="screen shot 2018-08-30 at 3 26 28 pm" src="https://user-images.githubusercontent.com/1867547/45046340-dc4aff00-b043-11e8-8551-d48de4913e96.png">

...into the settings, replacing the connected account view here:

<img width="780" alt="screen shot 2018-08-30 at 7 14 51 pm" src="https://user-images.githubusercontent.com/1867547/45046373-f389ec80-b043-11e8-93ae-3d8dc7e14e61.png">

...which currently appears this way when there is no connected account:

<img width="390" alt="screen shot 2018-09-26 at 6 47 58 pm" src="https://user-images.githubusercontent.com/1867547/46113523-c9e46100-c1bc-11e8-975d-a6397ce4a45a.png">

This way, the onboarding wizard + (dismissible) banner is not the only way to connect an account.

Are there any further design considerations at this point? Otherwise, will seek design review directly on code changes.

On the other hand, a broader approach here may also support a way to create an account from this screen. Do we want to import the existing Calypso UI for this in some shape or form?

dechov:
Updated the post above with a screenshot of the "no account connected" state, now merged into `master`.

PR #1513, opened 2018-10-15
Add Stripe OAuth connect action into settings

dechov:
Fixes https://github.com/Automattic/woocommerce-services/issues/1500

_Edit:_ <s>Fixes https://github.com/Automattic/woocommerce-services/issues/1526</s> (already fixed by https://github.com/Automattic/woocommerce-services/pull/1547)

Closes the loop for the "Connected Stripe account" section of the Stripe settings (added by WooCommerce Services) by adding an entry point for the merchant to connect their account into the Stripe settings, when no account is connected – replacing the "No account connected via WooCommerce Services" notice:

screen shot 2018-10-13 at 2 17 39 pm

(The only existing ways to enter the OAuth flow and connect an account are a) via Calypso and b) via the wp-admin NUX banner that shows up after completing the OBW when a Stripe account already exists at the provided email.)

To test:
- With WooCommerce Services enabled, go to WooCommerce » Settings » Payments » Stripe, and see the "Connected Stripe account" section near the top
  - If a Stripe account is already connected, disconnect it
- Verify that a card with a "Connect" action appears
- Click "Connect" and complete the OAuth connection flow
- Verify that the connected account view appears, and that API keys are filled
- Disconnect the account
- Verify that the "Connect your account" action card appears again after auto-refresh

Concerns:
- Do we want a "create account" action here as well, using our deferred account creation mechanism? The OAuth connection provides for creating an account on the spot, but demands much much information from the merchant upfront.
- The copy is taken from the banner. Does it work here, or is it a bit out of step with the description at the top of the settings? Should it depend on whether keys are already filled? Maybe it should at least say "you can" (not "you'll need to") connect your account?

### DIFF
--- a/assets/stylesheets/_components.scss
+++ b/assets/stylesheets/_components.scss
@@ -6,7 +6,6 @@
 // stylesheet order does not matter.
 // ==========================================================================
 
-@import 'components/action-card/style';
 @import 'components/button/style';
 @import 'components/button-group/style';
 @import 'components/card/style';
--- a/assets/stylesheets/_components.scss
+++ b/assets/stylesheets/_components.scss
@@ -6,6 +6,7 @@
 // stylesheet order does not matter.
 // ==========================================================================
 
+@import 'components/action-card/style';
 @import 'components/button/style';
 @import 'components/button-group/style';
 @import 'components/card/style';
--- a/classes/class-wc-connect-stripe.php
+++ b/classes/class-wc-connect-stripe.php
@@ -187,9 +187,21 @@ if ( ! class_exists( 'WC_Connect_Stripe' ) ) {
 		}
 
 		/**
-		 * Trigger Stripe connection-related admin notice if one is meant to be shown.
+		 * Handle connection and just-in-time messaging around OAuth flow.
 		 */
 		public function maybe_show_notice() {
+			// Handle redirect back from OAuth flow.
+			if ( isset( $_GET[ 'wcs_stripe_code' ] ) && isset( $_GET[ 'wcs_stripe_state' ] ) ) {
+				$response = $this->connect_oauth( $_GET[ 'wcs_stripe_state' ], $_GET[ 'wcs_stripe_code' ] );
+				if ( ! is_wp_error( $response ) ) {
+					WC_Connect_Options::update_option( 'banner_stripe', 'success' );
+				}
+
+				wp_safe_redirect( remove_query_arg( array( 'wcs_stripe_state', 'wcs_stripe_code' ) ) );
+				exit;
+			}
+
+			// Trigger connection-related admin notice if triggered and if on relevant screen.
 			$setting = WC_Connect_Options::get_option( 'banner_stripe', null );
 			if ( is_null( $setting ) ) {
 				return;
@@ -199,16 +211,6 @@ if ( ! class_exists( 'WC_Connect_Stripe' ) ) {
 				add_action( 'admin_notices', array( $this, 'connection_success_notice' ) );
 				WC_Connect_Options::delete_option( 'banner_stripe' );
 				return;
-			}
-
-			if ( isset( $_GET[ 'wcs_stripe_code' ] ) && isset( $_GET[ 'wcs_stripe_state' ] ) ) {
-				$response = $this->connect_oauth( $_GET[ 'wcs_stripe_state' ], $_GET[ 'wcs_stripe_code' ] );
-				if ( ! is_wp_error( $response ) ) {
-					WC_Connect_Options::update_option( 'banner_stripe', 'success' );
-				}
-
-				wp_safe_redirect( remove_query_arg( array( 'wcs_stripe_state', 'wcs_stripe_code' ) ) );
-				exit;
 			}
 
 			$screen = get_current_screen();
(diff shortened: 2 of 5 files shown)


dechov:
@kellychoffman I merely copied that heading from p7bbVw-2dB-p2 (bottom), but I think in some way we should hope to minimize confusion between a Stripe account with keys filled manually, and one with an OAuth connection facilitated by the WCS server (upon which the API key settings are _filled in automatically_, to answer your second question).

kellychoffman:
> we should hope to minimize confusion between a Stripe account with keys filled manually

If someone has a Stripe account with the keys already filled in, we shouldn't show this connection banner. (Right? Unless I'm missing something here....)

dechov:
@kellychoffman The connection state is only stored on the WCS server, which is only queried after the page loads, so we could show a notice indicating no account connected – that seems better than hiding the whole thing after showing a placeholder, but let me know what you think.

On the other hand, I think it could be valuable to give the option of connecting a different account than the one they'd previously input manually, in the same way that it's not necessary to clear existing keys before inputting new ones. (The copy should acknowledge the existing keys though, and revise the "to start, you'll need to" language.)

Happy to go with the first option if you still think that's best. Also, if you prefer, the action can be reduced in emphasis to something like this from the PayPal Checkout extension:

screen shot 2018-10-19 at 11 59 45 am

dechov:
We should ensure that this way of connecting can be tracked separately from the connection banner, if possible – may need a new event on the plugin side.

assets/stylesheets/_components.scss
```diff
@@ -6,6 +6,7 @@
 // stylesheet order does not matter.
 // ==========================================================================
 
+@import 'components/action-card/style';
```
DanReyLop: You should rebase with `master`. Calypso's `action-card` component now uses [modular CSS](https://github.com/Automattic/wp-calypso/blob/master/client/components/action-card/index.jsx#L19), so this `@import` is redundant.

kellychoffman:
> Also, if you prefer, the action can be reduced in emphasis to something like this from the PayPal Checkout extension

This feels like the way to go!

dechov:
OK, I like it:

screen shot 2018-11-15 at 7 44 24 pm

Text says:

> To automatically copy keys from a Stripe account, [connect](#) it to your store.

I think this text is far more appropriate for this settings screen (compared to the banner text I previously had), answering why you might want to connect and what will happen when you do. I also tried to make it work whether keys were already present or not.

Questions at this point:

- Do we want to go further still, and have slightly different text if keys are already set?
- Should we move this "Connected Stripe account" bit to just above the "Test mode" setting? That would bring it into proximity with the key fields, also easing ambiguity around the heading.
- Should we mention "WooCommerce Services" in some way? This way people are aware, and aren't surprised if this piece is gone in the event they deactivate WCS.

kellychoffman:
In the screenshot you shared above, is the desired action to still connect an account? If so, I'm worried its too de-emphasized. The screenshot here seems more desirable: "Do this preferred thing, but if you don't want to, you can take this other, not so preferred thing thing."

dechov:
> In the screenshot you shared above, is the desired action to still connect an account?

Yes, if there are no keys saved, but if there are, then there isn't necessarily a "desired action" to emphasize. Bearing that in mind, we can pass a prop to the WCS view indicating whether keys are saved.

Without keys:

screen shot 2018-11-20 at 4 36 33 pm

With keys:

screen shot 2018-11-20 at 4 39 52 pm

Does that seem best? Open to revisions on the copy.

The button is Calypso-styled because I don't see a way to render an ordinary core-styled button within a `wcc-root` container without manually overriding the styles, and I didn't want to avoid rendering the WCS view altogether since there may be a need to disconnect an account even in a case when the keys have been manually removed – open to implementation review on that.

jobthomas:
@dechov @DanReyLop any updates on this? As you can see on this issue, this is causing for a lot of confusion: the reports keep piling up. #1526

dechov:
@jobthomas I was thinking we should try to merge this PR early next week, or otherwise release a wording change that makes very clear that keys may still be input manually below.

jobthomas:
That would be awesome. Thanks @dechov

DanReyLop:
@dechov I'm getting an infinite loop of these 2 actions, because my test site is server over HTTP:
```
{
  type: 'WOOCOMMERCE_SETTINGS_STRIPE_CONNECT_ACCOUNT_OAUTH_INIT_COMPLETE',
  error: 'Your site must be served over HTTPS in order to connect your Stripe account via WooCommerce Services',
  siteId: 1
}
```
```
{
  type: 'WOOCOMMERCE_SETTINGS_STRIPE_CONNECT_ACCOUNT_DEAUTHORIZE_OAUTH_INIT',
  siteId: 1
}
```
And during all that, this is what's displayed (not helpful):
screenshot 2018-12-06 at 11 57 11

kellychoffman:
Agree with :shipit:

dechov:
Just in case it was unclear, https://github.com/Automattic/woocommerce-services/pull/1513#issuecomment-439244200 is the state it was merged/released in, and https://github.com/Automattic/woocommerce-services/pull/1513#issuecomment-440440148 (below it) is still under consideration for a subsequent release.

PR #1514, opened 2018-10-16
Use wp-calypso's Babel config

DanReyLop:
I'm trying to simplify as much as possible our "boilerplate" code and our `package.json` dependency list. The first step to do that is to catch up to what `wp-calypso` is using.

In this PR:
* All the `babel`-related dependencies are directly copied from `wp-calypso` (we were a few versions behind).
* Our `babel` config is removed, it just points to `wp-calypso`'s `babel.config.js` file.
* I had to add a few env variables to the `npm` commands to make sure Calypso's babel config works for us (disable code-splitting, etc).

To test, just run `npm start` and/or `npm run dist` and do a smoke test on some of our UI. It should be extremely obvious if something's not working.

As an added bonus, now our bundle is significantly smaller! I don't remember the exact numbers, but it went from `1.6MB` to `1.4MB` minified aprox.


DanReyLop:
I'm including these changes into #1520, closing this.

PR #1518, opened 2018-10-23
Remove “live rates” from NUX banner feature list.

jeffstieler:
Don't show new users mentions of the "live rates" feature in the NUX/connection banner.


dechov:
Oh, right – we'll need a similar treatment in https://github.com/woocommerce/woocommerce/blob/a4d99fee7c91a079c67bae8f12721432767cbe79/includes/admin/class-wc-admin-setup-wizard.php#L1981-L2021 – in that case not because it shows "live rates" but because it doesn't show label printing.

jeffstieler:
@dechov thanks for catching that! Opened a PR: https://github.com/woocommerce/woocommerce/pull/21645

PR #1520, opened 2018-10-24
Update wp-calypso dependency, includes fix for the new Calypso's CSS build system

DanReyLop:
In progress, checkboxes and dropdowns look a bit off.

cc/ @jeffstieler

### DIFF
--- a/assets/stylesheets/_components.scss
+++ b/assets/stylesheets/_components.scss
@@ -25,14 +25,12 @@
 @import 'components/forms/form-text-input/style';
 @import 'components/forms/form-text-input-with-affixes/style';
 @import 'components/forms/form-toggle/style';
-@import 'components/token-field/style';
 @import 'components/global-notices/style';
 @import 'components/image/style';
 @import 'components/info-popover/style';
 @import 'components/notice/style';
 @import 'components/payment-logo/style';
 @import 'components/screen-reader-text/style';
-@import 'components/section-header/style';
 @import 'components/segmented-control/style';
 @import 'components/spinner/style';
 
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -1,10 +1,4 @@
-// Only nest imports that conflict with WP core style or apply
-// to elements rendered as a child of our React container
-@import 'shared/functions'; // functions that we've used from Compass, ported over
-@import 'shared/functions/functions'; // sass functions for z-index, etc.
-@import 'colors'; // import wcc colors
-@import 'shared/mixins/mixins'; // sass mixins for gradients, bordius radii, etc.
-//@import 'shared/colors'; // import all of our wpcom colors
+@import 'shared/utils';
 
 .wcc-root {
 	// Buttons
@@ -30,7 +24,7 @@
 		background: $blue-wordpress;
 		border-color: $blue-dark;
 		&:hover {
-			background: $button-hover;
+			background: $blue-medium;
 		}
 		// overwrite core styles
 		&[disabled],
@@ -47,20 +41,6 @@
 			background-image: linear-gradient( -45deg, $blue-medium 28%, darken( $blue-medium, 5% ) 28%, darken( $blue-medium, 5% ) 72%, $blue-medium 72%) !important;
 			border-color: #0081a9 !important;
 		}
-	}
-
-    // Forms
-    input[type="text"],
-    input[type="search"],
-    input[type="email"],
-    input[type="number"],
-    input[type="password"],
-    input[type=checkbox],
-    input[type=radio],
-    input[type="tel"],
-    input[type="url"],
-    textarea {
-        @extend %form-field;
 	}
 
 	input[type=checkbox]:checked:before {
@@ -221,12 +201,7 @@
 
 	// Shared
 	@import 'shared/reset'; // css reset before the rest of the styles are defined
-	@import 'shared/typography'; // all the typographic rules, variables, etc.
-	@import 'shared/extends'; // sass extends for commonly used styles
-	@import 'shared/animation'; // all UI animation
-	@import 'shared/extends-forms'; // sass extends for form styles
-	@import 'shared/forms'; // form styling
-	@import 'shared/dropdowns'; // dropdown styling
+	@import 'shared/_forms'; // form styling
 
 	// Main
 	@import 'main'; // global layout and responsive styles
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     "xgettext-js": "^1.1.0"
   },
   "dependencies": {
-    "autoprefixer": "^6.7.4",
     "@babel/cli": "7.1.2",
     "@babel/core": "7.1.2",
     "@babel/plugin-proposal-class-properties": "7.1.0",
@@ -40,6 +39,7 @@
     "@babel/preset-env": "7.1.0",
     "@babel/preset-react": "7.0.0",
     "@babel/runtime": "7.1.2",
+    "autoprefixer": "^6.7.4",
     "babel-eslint": "10.0.1",
     "babel-loader": "8.0.4",
     "babel-plugin-add-module-exports": "1.0.0",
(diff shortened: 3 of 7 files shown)


assets/stylesheets/_components.scss
```diff
@@ -25,14 +25,12 @@
 @import 'components/forms/form-text-input/style';
 @import 'components/forms/form-text-input-with-affixes/style';
 @import 'components/forms/form-toggle/style';
-@import 'components/token-field/style';
```
DanReyLop: These styles are now loaded directly from the component's JSX file, same way as we load the main stylesheet (`import "./style.scss";`).

assets/stylesheets/style.scss
```diff
@@ -30,7 +24,7 @@
 		background: $blue-wordpress;
 		border-color: $blue-dark;
 		&:hover {
-			background: $button-hover;
```
DanReyLop: This was the only color we used that wasn't in the Calypso's pallete. I made an executive decision.

webpack.config.js
```diff
@@ -12,6 +13,33 @@ const isDev = ! isProd && ! isI18n;
...
+const cssLoaders = [
+	isDev ? 'style-loader?hmr=false' : MiniCssExtractPlugin.loader,
+	'css-loader',
+	{
+		loader: 'postcss-loader',
+		options: {
+			plugins: () => [
+				autoprefixer( { browsers } ),
+				url( {
+					url: ( asset ) => 'https://wordpress.com/' + asset.url,
```
DanReyLop: `css-loader` doesn't have this functionality anymore, they recommend to use this `postcss` plugin instead.

webpack.config.js
```diff
@@ -1,43 +1,50 @@
...
+const cssLoaders = [
+	isDev ? 'style-loader?hmr=false' : MiniCssExtractPlugin.loader,
+	'css-loader',
+	{
+		loader: 'postcss-loader',
+		options: {
+			plugins: () => [
+				autoprefixer( { browsers } ),
+				url( {
+					url: ( asset ) => asset.url.startsWith( 'data:' ) ? asset.url : ( 'https://wordpress.com/' + asset.url ),
```
DanReyLop: css-loader doesn't have this functionality anymore, they recommend to use this postcss plugin instead.

package.json
```diff
@@ -83,7 +81,15 @@
     "redux": "^3.6.0",
     "redux-thunk": "^2.2.0",
     "reselect": "^2.5.4",
+    "sass-loader": "6.0.7",
```
DanReyLop: Calypso is using version `7.*.*` of `sass-loader`, but until [this regression](https://github.com/webpack-contrib/sass-loader/issues/556) is fixed we must stay in `6.*.*`.

package.json
```diff
@@ -2,12 +2,12 @@
   "name": "woocommerce-services",
   "version": "1.16.1",
   "scripts": {
-    "start": "cross-env NODE_ENV=development webpack-dev-server --hot --inline --watch --content-base dist --port 8085",
-    "dist": "cross-env NODE_ENV=production webpack",
+    "start": "cross-env NODE_ENV=development DISABLE_FEATURES=code-splitting CALYPSO_CLIENT=true webpack-dev-server --hot --inline --watch --content-base dist --port 8085",
+    "dist": "cross-env NODE_ENV=production DISABLE_FEATURES=code-splitting,wpcom-user-bootstrap CALYPSO_CLIENT=true webpack",
```
DanReyLop: Leaving `wpcom-user-bootstrap` on did spit a few warnings in the console, it has no relevance otherwise.

client/apps/settings/redux-middleware.js
```diff
@@ -10,7 +10,7 @@ import { fetchShippingMethodSchemaSuccess } from 'woocommerce/woocommerce-servic
 import { openShippingZoneMethod } from 'woocommerce/state/ui/shipping/zones/methods/actions';
 
 const middlewareActions = {
-	[  WOOCOMMERCE_SERVICES_SHIPPING_ZONE_METHOD_SETTINGS_REQUEST_SUCCESS ]: ( { siteId, instanceId, data }, dispatch ) => {
+	[  WOOCOMMERCE_SERVICES_SHIPPING_ZONE_METHOD_SETTINGS_REQUEST_SUCCESS ]: ( { siteId, data }, dispatch ) => {
```
DanReyLop: I guess the new version of `eslint` is better now, it didn't catch this unused variable before.

woocommerce-services.php
```diff
@@ -1140,8 +1140,7 @@ public function admin_enqueue_scripts() {
 
 			// Use the same version as Jetpack
 			$jetpack_version = defined( 'JETPACK__VERSION' ) ? JETPACK__VERSION : '0';
-			wp_register_style( 'noticons', $this->wpcom_static_url( '/i/noticons/noticons.css' ), array(), $jetpack_version . '-' . gmdate( 'oW' ) );
```
DanReyLop: `wp-calypso` now uses a SVG for the checkbox checkmarks, so the `noticons` font can finally be removed 🎉

webpack.config.js
```diff
@@ -107,10 +133,22 @@ const config = {
...
 				enforce: 'post',
-				use: {
-					loader: 'babel-loader',
-					options: babelSettings,
-				},
+				use: [
+					{
+						loader: 'thread-loader',
+						options: {
+							workers: Math.max( 2, Math.floor( os.cpus().length / 2 ) ),
```
DanReyLop: This is what makes this baby fly now, try it in one of those new fancy 6-core Macbook Pros for maximum awesomeness :)

package.json
```diff
@@ -19,62 +19,60 @@
...
-    "style-loader": "^0.13.1",
-    "svg-url-loader": "^2.0.2",
-    "url-loader": "^0.5.7",
-    "webpack": "^2.2.1",
-    "webpack-dev-server": "^2.4.1",
-    "webpack-node-externals": "^1.5.4",
-    "webpack-notifier": "^1.5.0",
     "xgettext-js": "^1.1.0"
   },
   "dependencies": {
```
DanReyLop: I'm putting all the dependencies I touched in `dependencies` because before we weren't being consistent in our use of depencencies vs devDependencies. I think it's better to not separate them at all, than to separate them arbitrarily 🤷‍♂️

webpack.config.js
```diff
@@ -107,10 +133,22 @@ const config = {
...
+						},
+					},
+					{
+						loader: 'babel-loader',
+						options: {
+							configFile: path.resolve( __dirname, 'node_modules', 'wp-calypso', 'babel.config.js' ),
+							cacheDirectory: true,
+							cacheIdentifier: require( 'wp-calypso/server/bundler/babel/babel-loader-cache-identifier' ),
```
jeffstieler: Do we need our own cache identifier? If not, should this be "client" instead of "server"?

jeffstieler:
This built fine and all (React) plugin functionality seems to be working. Just have one question about the webpack caching.

webpack.config.js
DanReyLop: This is the identifier that Calypso uses on its **client** config, see: [[#](https://github.com/Automattic/wp-calypso/blob/master/webpack.config.js#L167)] [[#](https://github.com/Automattic/wp-calypso/blob/master/webpack.config.js#L26)]. The identifier itself is quite simple, it's like the default one but it also takes into account the version of a custom babel transform that Calypso uses: [see here](https://github.com/Automattic/wp-calypso/blob/cf1d19c8e494c110eb400d0c8fdf43a94c81f9b3/server/bundler/babel/babel-loader-cache-identifier/index.js).

It's confusing that it's under `server/bundler`, but it's really used by the client. It may be overkill for us, but I prefer to apply the "do whatever Calypso does and forget about it" rule of thumb.

package.json
DanReyLop: The line is not as clear-cut as you think it is :)

I'd say that testing and linting are `devDependencies`, but since we (currently) use the ESLint Webpack loader, so `npm start` or even `npm run dist` would fail without `eslint`.

I would consider transpiling (babel) a `dependency`, because it's needed in order to produce a `production` bundle.

Anyway, I'm trying to remove almost all our dependencies entirely and ge tthe build to use the ones under `node_modules/wp-calypso/node_modules` instead, so in the near future we may end up with just a handful of dependencies, and we can decide then.

webpack.config.js
jeffstieler: Ah. yeah - I totally missed that this was in a `require()`. Nevermind!

PR #1527, opened 2018-10-29
Prevent Stripe from attempting to connect if the site is not using ssl

marcinbot:
See server PR 1165 for extra detail

Currently, if the site is not using SSL in the admin url the plugin will try to connect to the server to generate the OAuth url and fail every time.

This prevents the site from attempting to connect in the first place, reducing the traffic and increasing performance.

### DIFF
--- a/classes/class-wc-connect-stripe.php
+++ b/classes/class-wc-connect-stripe.php
@@ -237,7 +237,10 @@ if ( ! class_exists( 'WC_Connect_Stripe' ) ) {
 			$result = $this->get_oauth_url();
 
 			if ( is_wp_error( $result ) ) {
-				$this->logger->log( $result, __CLASS__ );
+				//do not log the invalid url protocol error when attempting to render the banner
+				if ( 'invalid_url_protocol' !== $result->get_error_code() ) {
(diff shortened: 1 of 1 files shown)


dechov:
Good to prevent the requests – this would still flood the logs, though – do we want to prevent that by dismissing the banner at this point, considering that https://github.com/Automattic/woocommerce-services/pull/1513 will allow them to connect at a later point anyway?

classes/class-wc-connect-stripe.php
```diff
@@ -46,6 +46,11 @@ public function get_oauth_url( $return_url = '' ) {
 			if ( empty( $return_url ) ) {
 				$return_url = admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe' );
 			}
+
+			if ( substr( $return_url, 0, 8 ) !== 'https://' ) {
+				return new WP_Error( 'invalid_url_protocol', __( 'Your site requires SSL in order to connect your Stripe account via WooCommerce Services', 'woocommerce-services' ) );
```
DanReyLop: Nitpick: I think saying `HTTPS` will be easier to understand than `SSL` for a merchant.

PR #1528, opened 2018-10-29
Remove all the dependencies from package.json if they are already dependencies of wp-calypso

DanReyLop:
In progress


v18:
I can't find any docs about this, but I think best practices say that each project should list its own dependencies, even when those are already going to be installed by other deps in the project. that makes sense for a few reasons, for example seeing what dependencies your project uses at a glance, not accidentally breaking things when another project removes a dep, etc.

The way we're using wp-calypso is unique, but I don't think that we should rely on the dependencies listed there as our own.

PR #1536, opened 2018-11-08
Show admin notice for missing product dimension errors

### LINKED ISSUE #1386: Make sure there is an error when dimensions for a product are not set
@dechov says "Note that missing dimensions (as I've finally realized) don't cause an error AFAICT — the dimension value is just treated as 0." in https://github.com/Automattic/woocommerce-services/issues/1376#issuecomment-383245781

This is a problem for the box packer since it needs to make sure the items actually fit in a box. We should make sure to log an error for products with no dimensions set, similar to what happens when the weight is not set.

jeffstieler:
Fixes #1386.

Prevents rate requests from dispatching when a product is missing dimension value(s). Surfaces the error in `/wp-admin` in the form of a notice and on the cart page if debugging is enabled.

screen shot 2018-11-08 at 4 20 43 pm

screen shot 2018-11-08 at 4 24 10 pm

**To test:**
* Add a product to your cart that is either missing a dimension value (length, width, height) or has `0` as the value
* Verify that the fallback rate is used (if configured)
* Verify that the admin notice is displayed on the dashboard
* Verify that entering dimensions for the product dismissed the notice


classes/class-wc-connect-error-notice.php
```diff
@@ -70,6 +73,28 @@ private function show_notice() {
...
+					$this->disable_notice();
+					return;
+				}
+
+				$product_name = WC_Connect_Compatibility::instance()->get_product_name( $product );
+				$product_id = is_a( $product, 'WC_Product_Variation' ) ? $product->get_parent_id() : $id;
+				$message = sprintf(
+					__( '<strong>"%2$s" is missing length, width, or height.</strong><br />Shipping rates cannot be calculated. <a href="%1$s">Enter dimensions for %2$s</a> so your customers can purchase this item.', 'woocommerce-services' ),
+					get_edit_post_link( $product_id ), $product_name
+				);
```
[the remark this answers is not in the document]
jeffstieler: Done in https://github.com/Automattic/woocommerce-services/pull/1536/commits/ee142c039d6d766686bc93840deaf2f65291f987

classes/class-wc-connect-api-client.php
```diff
@@ -101,17 +101,23 @@ public function build_shipment_contents( $package ) {
...
-						$length = $product->get_length();
-						$width  = $product->get_width();
+					if (
+						! $product->get_length() ||
+						! $product->get_height() ||
+						! $product->get_width()
+					) {
+						return new WP_Error(
+							'product_missing_dimension',
+							sprintf( "Product ( ID: %d ) is missing a dimension value. Shipping rates cannot be calculated.", $product->get_id() ),
```
marcinbot: Do we want to translate this? There's no clear precedent with other parts of the code as we sometimes translate the error message and sometimes don't. And in this case, the message in the notice is translated and this only ends up in the logs.

jeffstieler: Done in https://github.com/Automattic/woocommerce-services/pull/1536/commits/a230a8baa179ce0ad2dfd607d17005ad5a1cec39.

PR #1538, opened 2018-11-09
Don't show admin notice for "unactionable" events

### LINKED ISSUE #1376: Renovate "An error occurred in WooCommerce Services" admin notice

jeffstieler:
Fixes #1376.

This PR removes the display of the "An error occurred in WooCommerce Services" admin notice in the following scenarios (that I'm proposing are "unactionable"):

1. A request to a WCS endpoint (local to the store) fails (e.g. `/wc/v1/connect/*`)
1. A tax rate request fails
1. A shipping rate request fails

**To test:**
* Point your store to a locally running WCS server
* Keep the local server **off**
* Visit the cart page, with items in cart
* Verify that either the tax rate or shipping rates are unable to be fetched (or both)
* Verify that no admin notice appears in the dashboard
* Verify that the failures are still logged on the Status page

### DIFF
--- a/classes/class-wc-connect-error-notice.php
+++ b/classes/class-wc-connect-error-notice.php
@@ -50,14 +50,20 @@ if ( ! class_exists( 'WC_Connect_Error_Notice' ) ) {
 		}
 
 		private function show_notice() {
-			$link_status = admin_url( 'admin.php?page=wc-status&tab=connect' );
+			$link_status  = admin_url( 'admin.php?page=wc-status&tab=connect' );
 			$link_dismiss = add_query_arg( array( 'wc-connect-error-notice' => 'disable' ) );
+			$error        = $this->notice_enabled();
 
-			$error = $this->notice_enabled();
-			if ( is_wp_error( $error ) && 'product_missing_weight' === $error->get_error_code() ) {
+			if ( ! is_wp_error( $error ) ) {
+				return;
+			}
+
+			$message = false;
+
+			if ( 'product_missing_weight' === $error->get_error_code() ) {
 				$error_data = $error->get_error_data();
-				$id = $error_data['product_id'];
-				$product = wc_get_product( $id );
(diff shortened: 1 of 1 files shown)


jeffstieler:
> But replacing `error()` with l`og()` has a side effect where we won't log errors if debug is disabled.

Good catch.

> How about we remove the generic "An error occurred in WooCommerce Services" notice from `class-wc-connect-logger.php` instead?

Works for me!

PR #1539, opened 2018-11-16
Avoid WCS bundle loading unnecessarily on WC settings screens

dechov:
Now that a Calypso view is being rendered on the Stripe settings page, the WCS JS bundle is enqueued on every WooCommerce » Settings screen (i.e. whenever the Stripe settings filter fires).

<s>The changes in this PR support a **separation** between delivering **markup** and enqueuing the **script**. In this case, the purpose is for the latter to occur conditionally on the buffered markup being displayed.</s>

Note: JS assets intended for specific payment method settings forms – e.g. PayPal Standard, PayPal Checkout, Stripe – are also currently enqueued on all settings pages, but those are under 1 kB each.

(This addresses the secondary cause of the "missing shipping methods UI when both WCS and Stripe extensions are activated" issue, as was described in https://github.com/Automattic/woocommerce-services/pull/1522.)

_Edit:_ To test, verify that "woocommerce-services.{js|css}" is loaded at WooCommerce » Settings » Payments » Stripe (and WooCommerce » Settings » Shipping » WooCommerce Services) but not other WooCommerce settings pages.

### DIFF
--- a/classes/class-wc-connect-stripe.php
+++ b/classes/class-wc-connect-stripe.php
@@ -285,8 +285,7 @@ if ( ! class_exists( 'WC_Connect_Stripe' ) ) {
 		 */
 		public function show_connected_account( $settings ) {
 			ob_start();
-			do_action( 'print_wc_connect_container', 'wc-connect-stripe-connect-account' );
-			add_action( 'admin_enqueue_scripts', array( $this, 'maybe_enqueue_script' ) );
+			do_action( 'render_wc_connect_view', 'wc-connect-stripe-connect-account' );
 
 			$new_settings = array(
 				'connection_status' => array(
@@ -296,21 +295,6 @@ if ( ! class_exists( 'WC_Connect_Stripe' ) ) {
 				),
 			);
 			return array_merge( $new_settings, $settings );
-		}
-
-		/**
-		 * On Stripe settings screen, enqueue script for rendering connected account view.
-		 */
-		public function maybe_enqueue_script() {
-			$screen = get_current_screen();
-
-			if ( // WooCommerce » Settings » Payments » Stripe.
-				'woocommerce_page_wc-settings' === $screen->base
-				&& isset( $_GET['tab'] ) && 'checkout' === $_GET['tab']
-				&& isset( $_GET['section'] ) && 'stripe' === $_GET['section']
-			) {
-				do_action( 'enqueue_wc_connect_script' );
-			}
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


classes/class-wc-connect-stripe.php
```diff
@@ -285,7 +285,8 @@ public function connection_success_notice() {
 		 */
 		public function show_connected_account( $settings ) {
 			ob_start();
-			do_action( 'enqueue_wc_connect_script', 'wc-connect-stripe-connect-account' );
```
DanReyLop: I don't get why all this PR is needed. I think changing this line with:
```php
			$screen = get_current_screen();
 			if ( // WooCommerce » Settings » Payments » Stripe.
				'woocommerce_page_wc-settings' === $screen->base
				&& isset( $_GET['tab'] ) && 'checkout' === $_GET['tab']
				&& isset( $_GET['section'] ) && 'stripe' === $_GET['section']
			) {
				do_action( 'enqueue_wc_connect_script', 'wc-connect-stripe-connect-account' );
			}
```
Should fix the problem. Unless I'm missing something obvious?

dechov: I was going to test this to make sure but didn't get around to it: my understanding was that `show_connected_account` was run on a hook that was too early for `get_current_screen`, which returns `null` until screen is initialized (before `current_screen` hook [[doc](https://codex.wordpress.org/Function_Reference/get_current_screen)]).

That said, I believe it can still be done by checking for `'wc-settings' === $_GET['page']`.

DanReyLop: The problem for me is that you're changing the logic to be more complex (hooks upon callbacks upon more hooks). If changing this line to add the conditional that checks for `'wc-settings' === $_GET['page']` is enough to solve this issue, we should change just that. Do you agree, or do you see any problem relying on the `$_GET['page']` variable?

dechov: I don't see a problem – I must have just been thinking "if there is a `get_current_screen` function, it should be used". I've updated it with that change.

dechov:
This was treated with low priority, but is still an issue AFAIK (unless there was a change in approach I'm unaware of): large bundle being loaded unnecessarily on all WC settings pages.

laurendavissmith:
No change in approach, I went with the i'll get a ping if its still relevant approach when doing some of this cleanup 😄 

Do we have an issue for this or just what you described here in the PR?

dechov:
@laurendavissmith Sorry for the late reply here. There's no issue for it AFAIK – just what I described above, simply a case of a script being loaded on many more screens than the one on which it's used.

PR #1542, opened 2018-11-21
Display packing log in frontend debug and on order details

### LINKED ISSUE #1541: Display packing log in order details and in-cart debugger
Server PR 1177 is adding a "packing log" to each rate returned from USPS. The data is an array of strings:

```json
[
    "Fitting these product IDs: 29, 74, 92",
    "medium_tube fits product IDs: 74, 92, 29",
    "priority_7 fits product IDs: 74, 92, 29",
    "priority_7 has better fit than medium_tube",
    "small_tube fits product IDs: 74, 92, 29",
    "small_tube has better fit than priority_7",
    "priority_1095 fits product IDs: 74, 92, 29",
    "priority_1095 has better fit than small_tube",
    "shoe_box fits product IDs: 74, 92",
    "priority_1097 fits product IDs: 74, 92",
    "priority_4 does not fit any products",
    "priority_1096L fits product IDs: 74, 92",
    "priority_tyvek_envelope does not fit any products",
    "Winner: priority_1095 with product IDs packed: 74, 92, 29"
]
```

Add the packing log to the front end debugger (depicted below) and to the order details page.

48634618-fa15d800-e993-11e8-8a44-ef1b2fe370e6

jeffstieler:
Fixes #1541.

Introduces the "packing log" to both the Order Details screen (as a metabox) and the frontend debugger:

fireshot capture 20 - edit order local woocommerc_ - http___local wordpress test_wp-admin_post php

**Note: this PR assumes the packing log will be returned on the `packing_log` property of each rate.**


woocommerce-services.php
```diff
@@ -1205,10 +1205,27 @@ public function add_meta_boxes() {
 			if ( $this->shipping_label->should_show_meta_box() ) {
 				add_meta_box( 'woocommerce-order-label', __( 'Shipping Label', 'woocommerce-services' ), array( $this->shipping_label, 'meta_box' ), null, 'side', 'default' );
 			}
+
+			add_meta_box( 'woocommerce-services-shipping-debug', __( 'Shipping Debug', 'woocommerce-services' ), array( $this, 'shipping_rate_packaging_debug_log_meta_box' ), 'shop_order', 'normal', 'default' );
+		}
+
+		public function shipping_rate_packaging_debug_log_meta_box( $post ) {
```
v18: What happens if the shipping log is non-existant? We should display a message explaining this to the user.

classes/class-wc-connect-shipping-method.php
```diff
@@ -437,7 +437,8 @@ public function calculate_shipping( $package = array() ) {
 						'cost'      => $rate->rate,
 						'meta_data' => array(
 							'wc_connect_packages' => $rate->packages,
-							__( 'Packaging', 'woocommerce-services' ) => $packaging_info
+							__( 'Packaging', 'woocommerce-services' ) => $packaging_info,
+							'wc_connect_packing_log' => $rate->packing_log,
```
v18: The property is called `box_packing_log` on the server. Did you want to change it?

classes/class-wc-connect-shipping-method.php
```diff
@@ -448,10 +449,11 @@ public function calculate_shipping( $package = array() ) {
 						} else {
 							$this->debug(
 								sprintf(
-									'Received rate: <strong>%s</strong> (%s)<br/><ul><li>%s</li></ul>',
+									'Received rate: <strong>%s</strong> (%s)<br/><ul><li>%s</li></ul><br/>Packing log: <ul><li>%s</li></ul>',
```
v18: Can we add i18n here? And also break up this string for readability. Something like this could work:

```
'Received rate: <strong>%s</strong> (%s)<br/>' .
	'<ul><li>%s</li></ul><br/>' .
	'Packing log: ' .
	'<ul><li>%s</li></ul>',
```

v18:
Can we add the product ID to the list of products (this is part of the previous debug log we had)? So that our final logs looks something like this:

debug_log_with_ids

woocommerce-services.php
jeffstieler: What do you think about hiding the metabox instead?

jeffstieler:
> Can we add the product ID to the list of products

Done in 3dd205cf.

classes/class-wc-connect-shipping-method.php
jeffstieler: Reworked in https://github.com/Automattic/woocommerce-services/commit/3dd205cf22692fa4e10cbbfbf9356a756e245bf3.

classes/class-wc-connect-shipping-method.php
```diff
@@ -446,15 +447,21 @@ public function calculate_shipping( $package = array() ) {
...
-								),
-								'success'
+							$rate_debug = sprintf(
+								/* translators: 1: name of shipping service, 2: shipping rate (price) */
+								__( 'Received rate: <strong>%1$s</strong> (%2$s)', 'woocommerce-services' ),
+								$rate_to_add['label'],
+								wc_price( $rate->rate )
 							);
+
+							$rate_debug .= "\n<ul><li>" . implode( '</li><li>', $package_summaries ) . '</li></ul>';
```
v18: Why do need the newline character for?

jeffstieler: It's not! 😛 Removed in b7ce0d0c.

classes/class-wc-connect-shipping-method.php
```diff
@@ -446,15 +447,21 @@ public function calculate_shipping( $package = array() ) {
...
+
+							$this->debug( $rate_debug, 'success' );
+
+							$packing_debug  = __( 'Packing log:', 'woocommerce-services' );
+							$packing_debug .= '<ul><li>' . implode( '</li><li>', $rate->box_packing_log ) . '</li></ul>';
+
+							$this->debug( $packing_debug, 'success' );
```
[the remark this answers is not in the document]
jeffstieler: Recombined in b7ce0d0c.

woocommerce-services.php
```diff
@@ -1205,10 +1205,27 @@ public function add_meta_boxes() {
...
+			$order            = wc_get_order( $post );
+			$shipping_methods = $order->get_shipping_methods();
+
+			foreach ( $order->get_shipping_methods() as $method ) {
+				if ( empty( $method['wc_connect_packing_log'] ) ) {
+					continue;
+				}
+
+				echo '<strong>', $method->get_name(), '</strong> <em>(', $method->get_method_id(), ':', $method->get_id(), ')</em>';
```
v18: The HTML and PHP mixed in together here is a bit messy. Possible to mix using `<?php` and `?>` as outlined [here](https://make.wordpress.org/core/handbook/best-practices/coding-standards/html/#indentation)?

v18: Can we add this information to the packing log, so it's clear to the user what's happening:

- the rate was retrieved using WooCommerce Shipping live rates (so that users have context as to why this packing log is showing up)
- an indication of what `$method->get_name()` means, for example `This is the shipping method used for this rate: $method->get_name()`
- the cost of the rate received (so that users don't have to scroll back up to the packing cost line item)
- the box used for the live rate, so it's easy to find when reading from the top

classes/class-wc-connect-shipping-method.php
```diff
@@ -446,15 +447,18 @@ public function calculate_shipping( $package = array() ) {
...
+							$rate_debug  = '<strong>';
+							$rate_debug .= sprintf(
+								/* translators: 1: name of shipping service, 2: shipping rate (price) */
+								__( 'Received rate: %1$s (%2$s)', 'woocommerce-services' ),
+								$rate_to_add['label'],
+								wc_price( $rate->rate )
 							);
+							$rate_debug .= '</strong><ul><li>' . implode( '</li><li>', $package_summaries ) . '</li></ul>';
+							$rate_debug .= '<strong>' . __( 'Packing log:', 'woocommerce-services' ) . '</strong>';
+							$rate_debug .= '<ul><li>' . implode( '</li><li>', $rate->box_packing_log ) . '</li></ul>';
```
v18: When the box packing log does not exist, the packing log looks a bit weird. Possible to make sure it exists before displaying it?

screen shot 2018-11-26 at 7 32 10 pm

jeffstieler: Done in https://github.com/Automattic/woocommerce-services/pull/1542/commits/14166345792bea49b5184b13ee6af3fd901c9625.

woocommerce-services.php
jeffstieler: Done in https://github.com/Automattic/woocommerce-services/pull/1542/commits/41e3bdadcaab8e9980e5c99dd205d7f1cfa855c6.

With the exception of:
> the box used for the live rate, so it's easy to find when reading from the top

Can you elaborate on what you're wanting to see here? We could list all of the boxes used in the packing, but that information is already presented in the packing log ("winners").

v18: Yeh, I meant all the boxes - the same info as the "Winners". If you think it's too much text, I'm ok to skip that 👍

woocommerce-services.php
v18: @jeffstieler - when the log is nonexistent, can we either remove the debug meta box or add an explanation why it's empty?

woocommerce-services.php
```diff
@@ -1205,10 +1205,43 @@ public function add_meta_boxes() {
...
+					continue;
+				}
+
+				$method_instance = new WC_Connect_Shipping_Method( $method->get_instance_id() );
+			?>
+				<p>
+					<strong>Shipping Method:</strong> <?php echo $method_instance->get_method_title(); ?>
+				</p>
+				<p>
+					<strong>Instance: </strong>
```
v18: Language nitpick here - "Instance" might not mean much to users. How about using "Shipping Method ID" here, and "Shipping Method Name:" in the line above instead of "Shipping Method"?

woocommerce-services.php
```diff
@@ -1205,10 +1205,43 @@ public function add_meta_boxes() {
...
+
+			foreach ( $shipping_methods as $method ) {
+				if ( empty( $method['wc_connect_packing_log'] ) ) {
+					continue;
+				}
+
+				$method_instance = new WC_Connect_Shipping_Method( $method->get_instance_id() );
+			?>
+				<p>
+					<strong>Shipping Method:</strong> <?php echo $method_instance->get_method_title(); ?>
```
v18: Let's translate these strings, too.

v18:
Looking great! Left a few comments and then it's good to go:

1. translate all remaining strings: https://github.com/Automattic/woocommerce-services/pull/1542#pullrequestreview-179591711
2. language nitpick for the language we use to indicate a shipping method instance: https://github.com/Automattic/woocommerce-services/pull/1542#discussion_r237322272
3. what we should do when the log is nonexistent: https://github.com/Automattic/woocommerce-services/pull/1542#discussion_r237321731

jeffstieler:
Updated to address all feedback.

v18:
Looks great! 🚢 I've added an issue to help the `pre` text wrap in smaller screens. I just realized this, so I didn't want to tack that on to this PR.

woocommerce-services.php
```diff
@@ -1201,14 +1201,62 @@ public function shipping_zone_method_status_toggled( $instance_id, $service_id,
 			}
 		}
 
-		public function add_meta_boxes() {
+		public function should_show_shipping_debug_meta_box( $post ) {
+			$order            = wc_get_order( $post );
+			$shipping_methods = $order->get_shipping_methods();
```
v18: The function `should_show_shipping_debug_meta_box()` gets run for all post types, so `$order` can end up being false here ([doc](https://docs.woocommerce.com/wc-apidocs/function-wc_get_order.html)).

Let's either return `false` if `$order` is false, or use `should_show_shipping_debug_meta_box` only if `$post_type` is `shop_order` in `add_meta_boxes()`.

classes/class-wc-connect-shipping-method.php
```diff
@@ -437,7 +437,8 @@ public function calculate_shipping( $package = array() ) {
 						'cost'      => $rate->rate,
 						'meta_data' => array(
 							'wc_connect_packages' => $rate->packages,
-							__( 'Packaging', 'woocommerce-services' ) => $packaging_info
+							__( 'Packaging', 'woocommerce-services' ) => $packaging_info,
+							'wc_connect_packing_log' => $rate->box_packing_log,
```
v18: Can we also handle the case where `$rate->box_packing_log` is NULL? (i.e. the server does not respond with a `box_packing_log` property)

jeffstieler: Done in f9865626.

v18:
Hey @jeffstieler given that this PR wasn't ready to merge, let's also fix the CSS. Here's a screenshot showing what's going wrong with the `<pre>` tag:

screen shot 2018-12-03 at 12 37 53 pm

Feel free to do in a new PR, or in this one.

jeffstieler:
> given that this PR wasn't ready to merge, let's also fix the CSS

Done in 93c3b9dc.

RadoslavGeorgiev:
I added the escaping and @marcinbot took a look at it (no issues found) so I will merge this now.

Issue #1526, opened 2018-10-27
Don't display notice in Stripe settings if account is connected directly


jobthomas:
Currently, when an account is directly configured through Stripe, and not through WCS, this notice shows:

"No account connected via WooCommerce Services."

This is quite confusing especially if the customer already was struggling to configure Stripe. See for example 1504690-zen.

So either that notice needs to be a button that allows them to connect their existing account that way, or it needs to be hidden if Stripe creds are added below. 

![http://cld.wthms.co/X5AlVs](http://cld.wthms.co/X5AlVs+)
Link to image here: http://cld.wthms.co/X5AlVs.

madeincosmos:
Also reported in 1526195-zen

EricaKuschel:
Also reported in 1538409-zen

dougaitken:
Adding a few I've had 
1535582-zen
1523915-zen

orhanfernandes:
Got one here: 1558937(zen)

yukikatayama:
reported 1562196-zen

vajrasar:
One here: #5564151-hc

orhanfernandes:
got another one, here: 1573006-zen

harishankerr:
Another report in 6176348-hc (1544871-zen)

dphyled:
Also reported in 1586237-zen

mouligreenlaw:
And another one here 1589814-zen

madeincosmos:
Bumping priority as there's a lot of users confused about this message

AtrumGeost:
Another report here 7627673-hc

orhanfernandes:
1601981 - Zen

harishankerr:
Another report in 8441312-hc.

PR #1547, opened 2018-11-28
Replace alarming Stripe settings notice with helpful text

dechov:
Fixes https://github.com/Automattic/woocommerce-services/issues/1526

(Note: previously considered in scope for https://github.com/Automattic/woocommerce-services/pull/1513 but was likely a poor match for the issue's priority. This PR serves as a stop-gap ahead of that one.)

For the case when there is no connected Stripe account (see https://github.com/Automattic/woocommerce-services/pull/1497), this change removes the "No account connected via WooCommerce Services." info notice and replaces it with a line of text directing the merchant down the screen to manually set up keys:

> Stripe account keys may be configured below.

`master` | this PR
-- | --
 | 

I was tempted to keep the "No account connected via WooCommerce Services." message as a parenthetical, but I suppose it would still result in more confusion than clarity.

I believe the new text is adequate but would certainly welcome feedback, and if necessary can also tweak the message conditionally on whether keys are already saved, though it would add some complexity.

(The reason we don't just hide the whole "Connected Stripe account" bit is that we don't know if an account is connected without making a request to the WCS server, which we weren't willing to hold up the page load for, so that element is already present, filtered into the settings.)

cc @jobthomas

### DIFF
--- a/classes/class-wc-connect-stripe.php
+++ b/classes/class-wc-connect-stripe.php
@@ -290,7 +290,7 @@ if ( ! class_exists( 'WC_Connect_Stripe' ) ) {
 			$new_settings = array(
 				'connection_status' => array(
 					'type'        => 'title',
-					'title'       => __( 'Connected Stripe account', 'woocommerce-services' ),
+					'title'       => __( 'Stripe Account (connected to WooCommerce Services)', 'woocommerce-services' ),
 					'description' => ob_get_clean(),
 				),
 			);
--- a/client/apps/stripe-connect-account/view-wrapper.js
+++ b/client/apps/stripe-connect-account/view-wrapper.js
@@ -60,7 +60,7 @@ class StripeConnectAccountWrapper extends Component {
 		}
 
 		if ( ! stripeConnectAccount.connectedUserID ) {
-			return translate( 'Stripe account keys may be configured below.' );
+			return translate( 'No account connected via WooCommerce Services. Stripe account keys may be configured below.' );
 		}
 
 		return (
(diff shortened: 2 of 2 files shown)


jobthomas:
@dechov I want to propose a slightly different wording: 

`Connected Stripe Account` -> `Stripe connected to WooCommerce Services`

`No account connected via WooCommerce Services.` -> `No account connected via WooCommerce Services. Stripe account keys may be configured below.`

And I do like the change of not using the notice.

Another big question I have, why are we showing this in the first place if there's no way to create an action out of this? If there's no way for a customer to connect through WCS from that page, why can't we just hide the whole section if they only can add manual keys after that?

dechov:
@jobthomas 👍 That wording change sounds fine to me, though I would opt for "Stripe account connected via WooCommerce Services" or "Stripe Account (connected via WooCommerce Services)" – what do you think?

> If there's no way for a customer to connect through WCS from that page, why can't we just hide the whole section if they only can add manual keys after that?

We can do that but not without either a) making an earlier request to see if there's an account connected (impacting the page load), or b) reverting https://github.com/Automattic/woocommerce-services/pull/1497 until https://github.com/Automattic/woocommerce-services/pull/1513 is ready.

dechov:
@jobthomas Updated title to "Stripe Account (connected to WooCommerce Services)" and description to "No account connected via WooCommerce Services. Stripe account keys may be configured below."

screen shot 2018-11-30 at 12 13 49 pm

jobthomas:
Happy with that change @dechov :shipit:

PR #1549, opened 2018-11-29
Add specificity to root selector to avoid conflicts in Calypsoified wp-admin.

jeffstieler:
This PR seeks to fix style conflicts when running within a "Calypsoified" `/wp-admin`. See: p58i-7nL-p2 (`#comment-40377`)

To test:
* `npm run dist`
* Clear local style cache
* Verify that all WCS screens and components have the correct styling (be sure to test hover, click, tooltips, etc)

### DIFF
--- a/assets/stylesheets/_components.scss
+++ b/assets/stylesheets/_components.scss
@@ -33,6 +33,7 @@
 @import 'components/screen-reader-text/style';
 @import 'components/segmented-control/style';
 @import 'components/spinner/style';
+@import 'components/token-field/style';
 
 &.woocommerce {
 	@import 'woocommerce/components/bulk-select/style';
--- a/assets/stylesheets/_components.scss
+++ b/assets/stylesheets/_components.scss
@@ -33,7 +33,6 @@
 @import 'components/screen-reader-text/style';
 @import 'components/segmented-control/style';
 @import 'components/spinner/style';
-@import 'components/token-field/style';
 
 &.woocommerce {
 	@import 'woocommerce/components/bulk-select/style';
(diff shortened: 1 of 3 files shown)


assets/stylesheets/_components.scss
```diff
@@ -33,6 +33,7 @@
 @import 'components/screen-reader-text/style';
 @import 'components/segmented-control/style';
 @import 'components/spinner/style';
+@import 'components/token-field/style';
```
DanReyLop: This shouldn't be needed, the stylesheet for `token-field` is already included as "modular CSS" (see [here](https://github.com/Automattic/wp-calypso/blob/b5b88e6b061b790e237135ee8ac46c0dcc7dca2d/client/components/token-field/index.jsx#L21)).

DanReyLop: If specificity is a problem, you should add the `.wp-core-ui.wp-admin` prefix to [here](https://github.com/Automattic/woocommerce-services/blob/master/webpack.config.js#L125).

jeffstieler: Fixed in 9d1f95d4. Thank you!

assets/stylesheets/style.scss
```diff
@@ -1,6 +1,6 @@
 @import 'shared/utils';
 
-.wcc-root {
```
DanReyLop: Sorry I missed this. This solution would break horribly in Store-in-dotcom, because Calypso doesn't add the `wp-admin wp-core-ui` classes to the `</body>`, so no CSS would be applied.

PR #1560, opened 2018-12-10
Prevent multiple stripe oauth init requests when the site is not on ssl

### LINKED ISSUE #1558: Infinite loop on the Stripe settings page if the site is HTTP
After merging https://github.com/Automattic/woocommerce-services/pull/1513, and related to the check implemented on https://github.com/Automattic/woocommerce-services/pull/1527, I noticed this bug.

When your site is HTTP (like my test site), and you go to the `Payments -> Stripe` WP-Admin page, you get a JS infinite loop of these Redux actions:
```
{
  type: 'WOOCOMMERCE_SETTINGS_STRIPE_CONNECT_ACCOUNT_OAUTH_INIT_COMPLETE',
  error: 'Your site must be served over HTTPS in order to connect your Stripe account via WooCommerce Services',
  siteId: 1
}
```
```
{
  type: 'WOOCOMMERCE_SETTINGS_STRIPE_CONNECT_ACCOUNT_DEAUTHORIZE_OAUTH_INIT',
  siteId: 1
}
```
And during all that, this is what's displayed (just the placeholder, not helpful):
screenshot 2018-12-06 at 11 57 11

The solution should be to receive and correctly display any server-side errors that happened while fetching the OAuth init URL, instead of blindly retrying.

Additionally, `api-staging.woocommerce.com` allows `http` sites, so maybe we shouldn't check the protocol if the `WOOCOMMERCE_CONNECT_SERVER_URL` is defined, what do you folks think?

cc/ @marcinbot @dechov

marcinbot:
Fixes #1558 

Attempts to generate OAuth url only once and then shows the error if it failed:
screen shot 2018-12-10 at 15 47 28

To test:
* On a site that's using HTTP navigate to the stripe settings page with dev tools open
* Verify that only one `oauth/init` request is sent and then the error appears
* Either upload the plugin to a site using HTTPS (such as wpcom) or modify `class-wc-rest-connect-stripe-oauth-init-controller.php` to always return some url
* Verify that the connection prompt appears as expected

### DIFF
--- a/client/apps/stripe-connect-account/view-wrapper.js
+++ b/client/apps/stripe-connect-account/view-wrapper.js
@@ -14,6 +14,7 @@ import StripeConnectAccount from 'woocommerce/app/settings/payments/stripe/payme
 import Notice from 'components/notice';
 import { getSelectedSiteId } from 'state/ui/selectors';
 import {
+	getError,
 	getIsRequesting,
 	getStripeConnectAccount,
 	getIsDeauthorizing,
@@ -33,8 +34,9 @@ class StripeConnectAccountWrapper extends Component {
 	}
 
 	componentDidUpdate( prevProps ) {
-		const { siteId, isLoading, connectedUserID, oauthURL } = this.props;
-		if ( ! isLoading && prevProps.isLoading && ! connectedUserID && ! oauthURL ) {
+		const { siteId, isLoading, connectedUserID, oauthURL, isOAuthInitializing, stripeError } = this.props;
+		const isInitError = ! isOAuthInitializing && prevProps.isOAuthInitializing && stripeError;
+		if ( ! isLoading && prevProps.isLoading && ! connectedUserID && ! oauthURL && ! isInitError ) {
 			this.props.oauthInit( siteId );
 		}
 	}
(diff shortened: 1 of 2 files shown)


dechov:
I'm wary of using a notice here – though it isn't the same situation as https://github.com/Automattic/woocommerce-services/issues/1526, it would still be calling attention to a piece of information they don't actually need to be concerned about. That is, while they may in fact need to be concerned (for a different reason) about their site not being served over HTTPS, connecting their Stripe account via WooCommerce Services isn't something they need to do.

Haven't tested yet but the rest of the changes look good to me.

dechov:
Looks fine other than wrapping unnecessarily on anything but a narrow screen – we should probably revert the CSS change [here](https://github.com/Automattic/woocommerce-services/pull/1513/files#diff-fdb987836a65b6d207ad7214a1ac3982), which was meant to intended to cover the width of the `<ActionCard>` that didn't end up being used. Would you consider that in scope for this PR?

client/apps/stripe-connect-account/view-wrapper.js
```diff
@@ -47,13 +49,23 @@ class StripeConnectAccountWrapper extends Component {
...
+			stripeError,
 			isDeauthorizing,
 			isReloading,
 			oauthURL,
 			translate,
 		} = this.props;
 
+		if ( ! isOAuthInitializing && stripeError ) {
+			return (
+				<div className="stripe-connect-account__placeholder-container">
```
dechov: Can this `<div>` wrapper be omitted? (Though the same goes for the `. stripe-connect-account__connect-action` below.)

client/apps/stripe-connect-account/view-wrapper.js
```diff
@@ -47,13 +49,23 @@ class StripeConnectAccountWrapper extends Component {
...
+			isOAuthInitializing,
 			stripeConnectAccount,
+			stripeError,
 			isDeauthorizing,
 			isReloading,
 			oauthURL,
 			translate,
 		} = this.props;
 
+		if ( ! isOAuthInitializing && stripeError ) {
```
dechov: If this were moved below the other conditionals, would checking for `stripeError` be enough?

marcinbot: Good catch, moved in 4be68b2

dechov:
:shipit:

Issue #1576, opened 2019-02-11
"Invalid Zip Entered" Error when Automated Taxes are enabled.


zwiesman31:
Howdy!

When "Automated Taxes" are enabled: 

![http://cld.wthms.co/Na5KQS](http://cld.wthms.co/Na5KQS+)
Image link: http://cld.wthms.co/Na5KQS

And a user enters an invalid zipcode at checkout, then navigates back to the cart page, it is impossible for them to get back to the checkout page, as this error is displayed instead: 

![http://cld.wthms.co/mfIghX](http://cld.wthms.co/mfIghX+)
Image link: http://cld.wthms.co/mfIghX

Now, in the settings under `WooCommerce > Settings > Shipping > Shipping Options` we could select the `Enable shipping calculator on the cart page` option, then change the zip that way, but not all users want that option enabled: 

![http://cld.wthms.co/9XWlt8](http://cld.wthms.co/9XWlt8+)
Image link: http://cld.wthms.co/9XWlt8

Here's a full screencast showing the error: 

![http://cld.wthms.co/m4dspm](http://cld.wthms.co/m4dspm+)
Screencast link: http://cld.wthms.co/m4dspm

dechov:
Looked into this a bit – the error (initially added in https://github.com/Automattic/woocommerce-services/pull/1489) shouldn't be shown on the Cart when the shipping calculator isn't enabled. Tried to prevent it by bailing out if `is_cart() && 'no' === get_option( 'woocommerce_enable_shipping_calc' )` but it doesn't seem to be so simple.

SiR-DanieL:
one more report in 4668624-hc

TMcLoone:
I've just encountered this issue. Happens if a user takes a little longer to enter in the zipcode and they get booted out. They have to clear cache to get back in. Any ideas how to fix?

zargogo:
I am encountering this issue when a customer adds a product, goes to checkout, inputs a false ZIP and then closes checkout. When it is re-opened it just throws the error message

frosso:
From a preliminary investigation, it looks like the issue is caused by the notices displayed on the checkout page when the "debug" checkout flag is "on" and both the "USPS live rates" and the "automatic taxes" functionality of the plugin are enabled.

![Screen Shot 2020-11-04 at 9 37 44 AM](https://user-images.githubusercontent.com/273592/98136561-b2413180-1e86-11eb-877b-bab58ffd0ded.png)

[...]

PR #1588, opened 2019-02-25
Fix zip code error preventing checkout

v18:
This is a partial solution to #1576

# The steps to repro this are:
1. Enable automated taxes in `wp-admin/admin.php?page=wc-settings&tab=tax`
2. Disable shipping calculator in `wp-admin/admin.php?page=wc-settings&tab=shipping&section=options`  - i.e uncheck the option `Enable the shipping calculator on the cart page`
3. Add product to cart
4. Go to checkout
5. Enter 90210 for zipcode, FL for state
6. Go back to cart
7. Try to go to checkout again. See that you cannot

# Here's an explanation of why the problem happens:
Whenever we fetch taxes from TaxJar and get an error, we check to see if this error should be just logged, or shown to the customer who's checking out.

We need to display the error we receive in two cases of customer input error:
1. The zipcode does not match the state (e.g. `90210` (a California zipcode) and the state `Florida`)
2. The zipcode is nonsense (e.g. `dkfjdk12121`)

We add a WC notice of the type "error" in these two cases https://github.com/Automattic/woocommerce-services/blob/979854d66a823fb61d8320eef7ce0df70616904b/classes/class-wc-connect-taxjar-integration.php#L288

When WC tries to load the checkout page, we make such an API request to TaxJar and possibly get this error.

When WC sees any errors on the checkout page before loading it, it loads the cart-errors page instead in https://github.com/woocommerce/woocommerce/blob/ef05bfccfc01bb2c621ef1293e61f7c57950670f/includes/shortcodes/class-wc-shortcode-checkout.php#L270

So... whenever the customer can't edit the address on the cart page (step # 2 in steps to repro), and the address is wrong according to TaxJar (step # 5 in steps to repro), and we try to let the customer know this using an error notice, then they can't do anything about it.

# There are a few problems with this approach:
1. It does not allow users to edit an address when using the PayPal Checkout buy now button. Probably a better solution is to 
2. We are updating a lot of TaxJar functions. We are moving further away from the code TaxJar has. This is problematic. At the very least, we should maintain our own fork so it's not so hard to merge their code back in to ours, and there's a better log of what's going on at each merge.
3. This isn't a huge problem per se, but when going to the checkout page for the first time with a bad zip code, we don't display the error. We should probably validate the address when it's updated as well (maybe with the `woocommerce_after_save_address_validation` hook, not 100% sure on that).

# Some other ideas, untested:
i. What about adding an address edit link to the cart page if the user has disabled the shipping calculator? Similar to the link WC add when the shipping calculator:
shipping-calculator-1
shipping-calculator-2
ii. How about not letting users disble the shipping calculator when automated taxes are enabled? Seems drastic, but we've done similar things - for example not allowing to enter prices inclusive of tax. If we go this route, we need to somehow let our current users know that we're doing this.

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -268,10 +268,21 @@ class WC_Connect_TaxJar_Integration {
 	public function _error( $message ) {
 		$formatted_message = is_scalar( $message ) ? $message : json_encode( $message );
 
-		//ignore error messages caused by customer input
+		// display these error messages directly to the user
+		// this error will be displayed on both cart and checkout pages
+		// and will block any further action
 		$state_zip_mismatch = false !== strpos( $formatted_message, 'to_zip' ) && false !== strpos( $formatted_message, 'is not used within to_state' );
 		$invalid_postcode = false !== strpos( $formatted_message, 'isn\'t a valid postal code for' );
-		if ( ! is_admin() && ( $state_zip_mismatch || $invalid_postcode ) ) {
+
+		// We should only block checkout if the user can edit the address on cart
+		$user_can_edit_address_on_cart = 'yes' === get_option( 'woocommerce_enable_shipping_calc' );
+		$should_not_add_wc_error_notice = ! $user_can_edit_address_on_cart && is_checkout();
+
+		if (
+			! is_admin()
+			&& ! $should_not_add_wc_error_notice
+			&& ( $state_zip_mismatch || $invalid_postcode )
+		) {
 			$fields = WC()->countries->get_address_fields();
 			$postcode_field_name = __( 'Postcode / ZIP', 'woocommerce-services' );
 			if ( isset( $fields['billing_postcode'] ) && isset( $fields['billing_postcode']['label'] ) ) {
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -976,6 +976,17 @@ class WC_Connect_TaxJar_Integration {
 		if ( false === $response ) {
 			$response = $this->smartcalcs_request( $json );
 
+			// Copied over from the original smartcalcs_request
+			if ( is_wp_error( $response ) ) {
+				$this->_error( 'Error retrieving the tax rates. Received (' . $response->get_error_code() . '): ' . $response->get_error_message() );
+				return;
+			} elseif ( 200 == $response['response']['code'] ) {
+				return $response;
+			} else {
+				$this->_error( 'Error retrieving the tax rates. Received (' . $response['response']['code'] . '): ' . $response['body'] );
+				return;
+			}
+
 			if ( 200 == wp_remote_retrieve_response_code( $response ) ) {
 				set_transient( $cache_key, $response, $this->cache_time );
 			}
@@ -1007,13 +1018,9 @@ class WC_Connect_TaxJar_Integration {
 			'body' => $json,
 		) );
 
-		if ( is_wp_error( $response ) ) {
-			$this->_error( 'Error retrieving the tax rates. Received (' . $response->get_error_code() . '): ' . $response->get_error_message() );
-		} elseif ( 200 == $response['response']['code'] ) {
-			return $response;
-		} else {
-			$this->_error( 'Error retrieving the tax rates. Received (' . $response['response']['code'] . '): ' . $response['body'] );
-		}
+		// Instead of doing something with the error, return the result
+		// of the API call and decide in each scenario what to do with it
+		return $response;
 	}
 
 	/**
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -83,6 +83,7 @@ class WC_Connect_TaxJar_Integration {
 		// Calculate Taxes at Cart / Checkout
 		if ( class_exists( 'WC_Cart_Totals' ) ) { // Woo 3.2+
 			add_action( 'woocommerce_after_calculate_totals', array( $this, 'maybe_calculate_totals' ), 20 );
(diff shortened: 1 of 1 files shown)


DanReyLop:
> When WC sees any errors on the checkout page before loading it, it loads the cart-errors page instead in https://github.com/woocommerce/woocommerce/blob/ef05bfccfc01bb2c621ef1293e61f7c57950670f/includes/shortcodes/class-wc-shortcode-checkout.php#L270

That's no longer the case, as of WC `3.5.4`: https://github.com/woocommerce/woocommerce/pull/22337

That regression (redirecting to the `cart-errors` page when there are notices) was introduced in `3.5.0`: https://github.com/woocommerce/woocommerce/pull/19794

So the behaviour that you're trying to patch was only a problem between 3.5.0 and 3.5.3. What do you think about dropping this PR? As you said, seems like this solution is not ideal, and if I understood it correctly there's no problem that needs to be solved anymore.

If we go ahead with this solution or a nother similar one, let's be prepared to fix this in the main Taxjar plugin too, maybe even contribute there first and then backport the changes to WCS, like we've done a few times in the past.

v18:
@DanReyLop hey hey, I'm still able to repro the issues with WC 3.5.4 using these steps (same as in the PR description):

1. Enable automated taxes in `wp-admin/admin.php?page=wc-settings&tab=tax`
2. Disable shipping calculator in `wp-admin/admin.php?page=wc-settings&tab=shipping&section=options`  - i.e uncheck the option `Enable the shipping calculator on the cart page`
3. Add product to cart
4. Go to checkout
5. Enter 90210 for zipcode, FL for state
6. Go back to cart
7. Try to go to checkout again. See that you cannot

Are you not able to repro this in WC 5.4+?

I still see the same code that loads the cart errors page in `master` here:

https://github.com/woocommerce/woocommerce/blob/d69f014807f29e6ba7514dc85ce22ac1e77ecb11/includes/shortcodes/class-wc-shortcode-checkout.php#L274

DanReyLop:
You're totally right. [This line](https://github.com/woocommerce/woocommerce/blob/d69f014807f29e6ba7514dc85ce22ac1e77ecb11/includes/shortcodes/class-wc-shortcode-checkout.php#L256) clears all the error notices, but if there's any other error between that and [this line](https://github.com/woocommerce/woocommerce/blob/d69f014807f29e6ba7514dc85ce22ac1e77ecb11/includes/shortcodes/class-wc-shortcode-checkout.php#L272), then the `cart-errors` page will show. Sorry for the noise, looks like the fix in WC 3.5.4 didn't really address this particular scenario.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -268,10 +331,21 @@ public function _log( $message ) {
...
+		// display these error messages directly to the user
+		// this error will be displayed on both cart and checkout pages
+		// and will block any further action
 		$state_zip_mismatch = false !== strpos( $formatted_message, 'to_zip' ) && false !== strpos( $formatted_message, 'is not used within to_state' );
 		$invalid_postcode = false !== strpos( $formatted_message, 'isn\'t a valid postal code for' );
-		if ( ! is_admin() && ( $state_zip_mismatch || $invalid_postcode ) ) {
+
+		// We should only block checkout if the user can edit the address on cart
+		$user_can_edit_address_on_cart = 'yes' === get_option( 'woocommerce_enable_shipping_calc' );
+		$should_not_add_wc_error_notice = ! $user_can_edit_address_on_cart && is_checkout();
```
DanReyLop: `$should_not_add_wc_error_notice` is confusing. Please change it to `$should_add_wc_error_notice` so the logic is easier to follow.

DanReyLop: Alternatively, what do you think about doing something like this?

```php
if ( ( $state_zip_mismatch || $invalid_postcode ) && is_checkout() && ! is_ajax() ) {
    wc_add_notice( 'blah blah blah', 'error' );
    wc_print_notices();
}
```

That will clear the notices list and prevent the `cart-errors` template from rendering.

v18:
> Sorry for the noise, looks like the fix in WC 3.5.4 didn't really address this particular scenario.

@DanReyLop do you think this is something to be fixed in core then? I'm not sure if there's some buggy behaviour in core I'm not seeing, let me know.

If we are going to keep this change, are you 👍 for the way I'm going about fixing it? Now that I've re-read my initial description, this seems like a good way to go about fixing the bug.

I'm all for writing a PR in TaxJar first, as well.

DanReyLop:
> @DanReyLop do you think this is something to be fixed in core then? I'm not sure if there's some buggy behaviour in core I'm not seeing, let me know.

I don't think this should be fixed in Core, because it's not a regression. Even if it was fixed in `3.6`, all the versions before it would still be broken. I thought this was a regression that only affected `3.5.0 - 3.5.3` but it isn't.

> I'm all for writing a PR in TaxJar first, as well.

No need to write in TaxJar first, as long as there's plans to port it over in either direction I'm happy :)

> If we are going to keep this change, are you 👍 for the way I'm going about fixing it? Now that I've re-read my initial description, this seems like a good way to go about fixing the bug.

I'm not thrilled about its shortcomings. The user won't see the validation until he submits the checkout form. But it seems like a reasonable compromise.

What do you think about this? https://github.com/Automattic/woocommerce-services/pull/1588#discussion_r272064286 It's simpler and would show the validation error even on first load, but it may have some shortcomings I'm not seeing.

v18:
@allendav this PR remained untouched for a while since I was unassigned from it in favour of payments stuff. Should I continue working on it & respond to @DanReyLop's comments?

allendav:
@v18 wrote:

> Should I continue working on it & respond to @DanReyLop's comments?

Yes please. Perhaps next week when you're janitor? (This issue is the only taxes issue on the janitor's log, fwict.)

PR #1594, opened 2019-03-25
Fix overwriting manually entered Stripe account info

### LINKED ISSUE #1592: WCS overwrites manually entered Stripe accounts
[Video](https://cld.wthms.co/FP41Cf)

During the on-boarding wizard of WooCommerce users can choose to enable the Stripe payment gateway. When enabled, they can also enter an email address that will be used to register a Stripe account for them, and to auto-configure the extension to work with that account.

For this to work, the wizard will also install and activate WooCommerce Services. At the end of the OBW users are prompted to connect their site/store to Jetpack, but that is an optional step.

### The issue

If someone skips the Jetpack step after choosing to have an account created for them, the account will neither be created, nor connected. Although this could be considered a bug by itself, there is not much we can do to fix it. WCS requires an active Jetpack connection in order to use Automattic's services to register the Stripe account.

### The symptom

When both WCS & Jetpack are active, and Jetpack gets conneced, WCS will step up, create the account, and save it locally. This could be fine as long as no Stripe account was entered yet. However, __if there is one, it will get overwritten__, and that can create a lot of trouble for our merchants, as @ryandonsullivan explained in [their tweet](https://twitter.com/ryandonsullivan/status/1108921326348046336).

### Proposed solution

Make sure that no account has already been entering before proceeding with https://github.com/Automattic/woocommerce-services/blob/master/woocommerce-services.php#L537

marcinbot:
Fixes #1592 

This is a simple fix for the issue - the underlying issue of communicating in the wizard that Jetpack is required to provision the Stripe account still remains and will require design input.

To test:
* On a fresh install, open the WooCommerce wizard
* Opt-in to have a Stripe account created
* Do NOT connect Jetpack and skip the step instead
* Finish the wizard
* On the Stripe settings page enter keys of an existing account (either live or test) and save
* Connect Jetpack now
* Stripe keys should remain the same (you can test by running `jQuery('#woocommerce_stripe_test_publishable_key').val()` in dev console or similar)

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -553,8 +553,6 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			}
 
 			if ( $user_elected_to_create_stripe_account && is_plugin_active( 'woocommerce-gateway-stripe/woocommerce-gateway-stripe.php' ) ) {
-				unset( $stripe_settings['create_account'] );
-				update_option( 'woocommerce_stripe_settings', $stripe_settings );
 				$this->tracks->record_user_event( 'core_wizard_stripe_setup' );
 
 				$email = isset( $stripe_settings['email'] ) ? $stripe_settings['email'] : wp_get_current_user()->user_email;
@@ -569,6 +567,10 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 						WC_Connect_Options::update_option( 'banner_stripe', 'connection' );
 					}
 				}
+
+				unset( $stripe_settings['email'] );
+				unset( $stripe_settings['create_account'] );
+				update_option( 'woocommerce_stripe_settings', $stripe_settings );
 			}
 		}
 
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -545,6 +545,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 					|| ! empty( $stripe_settings['publishable_key'] )
 					|| ! empty( $stripe_settings['secret_key'] )
 				);
+
 			if ( $user_elected_to_create_stripe_account && $stripe_already_connected ) {
 				unset( $stripe_settings['email'] );
 				unset( $stripe_settings['create_account'] );
@@ -568,9 +569,15 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 					}
 				}
 
-				unset( $stripe_settings['email'] );
-				unset( $stripe_settings['create_account'] );
-				update_option( 'woocommerce_stripe_settings', $stripe_settings );
+
+				// The Stripe settings have changed here - the keys were added,
+				// so we need to get a fresh copy.
+				$new_stripe_settings = get_option( 'woocommerce_stripe_settings', false );
+				if ( is_array( $new_stripe_settings ) ) {
+					unset( $new_stripe_settings['email'] );
+					unset( $new_stripe_settings['create_account'] );
+					update_option( 'woocommerce_stripe_settings', $new_stripe_settings );
+				}
 			}
 		}
 
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -530,11 +530,29 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 
 		public function init_core_wizard_payments_config() {
 			$stripe_settings = get_option( 'woocommerce_stripe_settings', false );
-			$stripe_enabled  = is_array( $stripe_settings )
+
+			$should_create_stripe_account  = is_array( $stripe_settings )
 				&& ( isset( $stripe_settings['create_account'] ) && 'yes' === $stripe_settings['create_account'] )
 				&& ( isset( $stripe_settings['enabled'] ) && 'yes' === $stripe_settings['enabled'] );
 
-			if ( $stripe_enabled && is_plugin_active( 'woocommerce-gateway-stripe/woocommerce-gateway-stripe.php' ) ) {
+			// In certain scenarios, the user can enter an email address but not connect Jetpack in the wizard,
+			// but instead add the Stripe keys manually and connect Jetpack after. If the existing keys are detected,
+			// forget the wizard settings and never retry
+			$stripe_already_connected = is_array( $stripe_settings )
+				&& (
+					! empty( $stripe_settings['test_publishable_key'] )
+					|| ! empty( $stripe_settings['test_secret_key'] )
+					|| ! empty( $stripe_settings['publishable_key'] )
+					|| ! empty( $stripe_settings['secret_key'] )
+				);
+			if ( $should_create_stripe_account && $stripe_already_connected ) {
+				unset( $stripe_settings['email'] );
+				unset( $stripe_settings['create_account'] );
+				update_option( 'woocommerce_stripe_settings', $stripe_settings );
+				return;
+			}
+
+			if ( $should_create_stripe_account && is_plugin_active( 'woocommerce-gateway-stripe/woocommerce-gateway-stripe.php' ) ) {
 				unset( $stripe_settings['create_account'] );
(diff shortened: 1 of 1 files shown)


woocommerce-services.php
```diff
@@ -530,11 +530,29 @@ public function init_core_wizard_shipping_config() {
 
 		public function init_core_wizard_payments_config() {
 			$stripe_settings = get_option( 'woocommerce_stripe_settings', false );
-			$stripe_enabled  = is_array( $stripe_settings )
+
+			$should_create_stripe_account  = is_array( $stripe_settings )
```
v18: When reading this, `$should_create_stripe_account` sounds like it should indicate whether a Stripe account should be created under all conditions. However, we don't want to create a Stripe account based on `$stripe_already_connected`.

I think renaming this variable from `$should_create_stripe_account` to `$user_elected_to_create_stripe_account` would make this distinction more obvious without losing readability later in line 555:

`if ( $user_elected_to_create_stripe_account && is_plugin_active( 'woocommerce-gateway-stripe/woocommerce-gateway-stripe.php' ) ) {`

woocommerce-services.php
```diff
@@ -530,11 +530,29 @@ public function init_core_wizard_shipping_config() {
...
+			// forget the wizard settings and never retry
+			$stripe_already_connected = is_array( $stripe_settings )
+				&& (
+					! empty( $stripe_settings['test_publishable_key'] )
+					|| ! empty( $stripe_settings['test_secret_key'] )
+					|| ! empty( $stripe_settings['publishable_key'] )
+					|| ! empty( $stripe_settings['secret_key'] )
+				);
+			if ( $should_create_stripe_account && $stripe_already_connected ) {
+				unset( $stripe_settings['email'] );
```
v18: Why is `$stripe_settings['email']` getting deleted? I don't see it deleted when going through with the Stripe account creation.

v18:
I tested this out and the fix works for me - the Stripe API keys do not get overridden if they've already been entered. I left a couple of minor comments.

woocommerce-services.php
marcinbot: Good point, I renamed it in dfb4904

woocommerce-services.php
marcinbot: I was debating whether it should be deleted or not. It's created as part of the wizard, and is not used by Stripe extension, just here. I think once we determine that we don't need the wizard options anymore, we should clean up everything. 

On the other hand, the option name `email` could be used in the stripe extension in the future. Deleting it here could break it.

I'm a bit on the fence. What do you think? Should it be deleted?

v18: > I think once we determine that we don't need the wizard options anymore, we should clean up everything.

Agreed.

So I looked it up and the Stripe extension does not using the `email` setting at all, for at least the last 3 versions. Same with WooCommerce itself.

The reason the `email` option got introduced is for this wizard set up. I don't think that Stripe should rely on that, but if you want to double-check with the team, though, that seems reasonable, too.

Given this, I don't think it will cause problems if we delete it. It would be great to be able to test it while developing though, so if you want to delete it in a separate PR I'm game for that as well, so it doesn't go out in the next release.

marcinbot: Ok, I think we should leave it as is here and delete `email` together with everything else.

v18: Cool - shall we also delete `email` in the case that we create the account for the user as well then? Within this `if`:

https://github.com/Automattic/woocommerce-services/blob/dfb4904468ee663cf3431117649cd0623f84b814/woocommerce-services.php#L555

marcinbot: Good catch, I unset it in d8cc385

woocommerce-services.php
```diff
@@ -530,13 +530,29 @@ public function init_core_wizard_shipping_config() {
...
+			// but instead add the Stripe keys manually and connect Jetpack after. If the existing keys are detected,
+			// forget the wizard settings and never retry
+			$stripe_already_connected = is_array( $stripe_settings )
+				&& (
+					! empty( $stripe_settings['test_publishable_key'] )
+					|| ! empty( $stripe_settings['test_secret_key'] )
+					|| ! empty( $stripe_settings['publishable_key'] )
+					|| ! empty( $stripe_settings['secret_key'] )
+				);
+			if ( $user_elected_to_create_stripe_account && $stripe_already_connected ) {
```
v18: We're currently tracking the event when the account is created:

`$this->tracks->record_user_event( 'core_wizard_stripe_setup' );`

in https://github.com/Automattic/woocommerce-services/blob/d8cc3858a49e5a592786c61a56b6fd709ca5dd7a/woocommerce-services.php#L556

Should we track when we tried to set up the account but are not able to, for example using

```
$this->tracks->record_user_event( 'core_wizard_stripe_setup_failed' );
```

woocommerce-services.php
```diff
@@ -530,13 +530,29 @@ public function init_core_wizard_shipping_config() {
...
+				unset( $stripe_settings['email'] );
 				unset( $stripe_settings['create_account'] );
 				update_option( 'woocommerce_stripe_settings', $stripe_settings );
+				return;
+			}
+
+			if ( $user_elected_to_create_stripe_account && is_plugin_active( 'woocommerce-gateway-stripe/woocommerce-gateway-stripe.php' ) ) {
 				$this->tracks->record_user_event( 'core_wizard_stripe_setup' );
```
v18: The `core_wizard_stripe_setup` is getting tracked before we check for an error in the response below. Should we track it after we know the request was successful? If we decide to track the failure above, we could track the failure here as well.

woocommerce-services.php
v18: I just double-checked about the best way to track this event. We should create a new event and have `success` and `reason` props, something like this would work for this "not tried" case:

$this->tracks->record_user_event( 'core_wizard_stripe_setup_failed', array(
  'success' => false,
  'reason' => 'already set up by user'
) );

And for the failing otherwise:

$this->tracks->record_user_event( 'core_wizard_stripe_setup_failed', array(
  'success' => false,
  'reason' => 'email already exists'
) );

$this->tracks->record_user_event( 'core_wizard_stripe_setup_failed', array(
  'success' => false,
  'reason' => $wp_error
) );

And success:

$this->tracks->record_user_event( 'core_wizard_stripe_setup_failed', array(
  'success' => true,
) );

I didn't test out the array, but it looks like it should work.

v18: Btw, I think we can just fix the success tracking in this case with a new event name, be able to release this today, and then have an issue for the optimal tracking down the road to be released later. What do you think @marcinbot ?

marcinbot: Let's move out of the scope here and add any extra tracking in a separate issue, especially that we don't know how much value there is in tracking failures here. As far as I understand we use it only to gauge the funnels, not to get information about detailed events, such as success/failures.

woocommerce-services.php
marcinbot: See above re: adding tracking separately. This PR only fixes one edge case, and adding any extra logic, especially so soon before the potential release, can break more than it will add.

woocommerce-services.php
v18: Yep, happy to move this to a different PR. Can you create an issue so we don't forget?

Fwiw, we're tracking both failures and successes in one event, which is why we should at least correct that behaviour.

As for tracking failures when the user has already added their keys - that shows whether or not the features is working as intended. The goal would be for those events to be near 0 at all times. If they're not, then it tells us that something is preventing our users from using the feature to its full potential.

v18:
Stripe keys were not being saved due to a race condition of sorts - the Stripe keys were saved elsewhere with the option `woocommerce_stripe_settings` but then overwritten with the previously-gotten `woocommerce_stripe_settings` option. Fixed by getting the option `woocommerce_stripe_settings` anew before re-saving.

woocommerce-services.php
marcinbot: > Can you create an issue so we don't forget?

Opened #1616

PR #1597, opened 2019-04-01
Update shipping label button to primary

v18:
The button type was updated to primary, changing its colour, and the text was updated to `Create shipping label`. The translations were updated, since some text was changed.

Here's how I updated the translation files:

`npm run i18n`
`npm run i18njson`

Is that enough/correct? If so, I'll add it to the documentation

Before:

button-before

After:

button-after

Todo:
- [ ] add translation instructions to docs ([here's a reference to follow up on](https://github.com/Automattic/woocommerce-services/pull/1019#discussion_r119542518))

cc @aleftick

### DIFF
--- a/client/apps/shipping-label/view-wrapper.js
+++ b/client/apps/shipping-label/view-wrapper.js
@@ -116,6 +116,7 @@ class ShippingLabelViewWrapper extends Component {
 
 		const className = classNames( 'shipping-label__new-label-button', {
 			'is-placeholder': ! loaded,
+			'is-primary': loaded,
 		} );
 
 		return (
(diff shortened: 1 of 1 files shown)


v18:
@marcinbot and I talked in Slack about this. The non-translation changes are good to be merged. I removed all the translation changes.

For the translations - we need to import any translation files we have so that we can generate new json translations files for all the strings in JS. I opened an issue for this: https://github.com/Automattic/woocommerce-services/issues/1618

There were other recent changes with new strings that will need to be updated. Our plugin isn't currently well translated, so we decided it was ok to ship these changes and translate later.

PR #1601, opened 2019-04-15
Add wp-calypso as a Git submodule

DanReyLop:
In this PR:
- Added `wp-calypso` as a submodule, removed it as a `package.json` dependency.
- Use the `@automattic/calypso-color-schemes` package instead of the (now removed) old SCSS color variables.

This way, we can go back to using `wp-calypso` proper instead of an old branch, without clashing with the lerna efforts.

https://git-scm.com/book/en/v2/Git-Tools-Submodules is the official documentation about Git submodules, and [here](https://github.blog/2016-02-01-working-with-submodules/) is an easier-to-read tutorial.

**TL;DR:**
- ~Instead of `npm install`, use `npm run install-deps`, that will take care of installing the deps in the `wp-calypso` submodule too.~ That's no longer needed, as of https://github.com/Automattic/woocommerce-services/pull/1601/commits/bfbc6437c72f396062e0aad1fc5a0a9e81c1deb0. `npm install` all you want.
- You'll need to run `git submodule update --init` after every pull or branch switch that may have changed the `wp-calypso` submodule commit SHA1. Alternatively, you can enable `submodule.recurse` ([#](https://git-scm.com/docs/git-config#Documentation/git-config.txt-submodulerecurse)) like this: `git config --global submodule.recurse true`
- Inside the `wp-calypso` folder will be the full Calypso repo. To update it, you'll need to do it like this:
```sh
$ cd wp-calypso
$ git checkout master
$ git pull # Or checkout a commit SHA1 explicitly
$ cd ..
$ git commit -am "Updating Calypso dependency"
```


jrodger:
LGTM.

Testing:

* `npm install` and `npm run dist` behave as expected
* Built package size is about the same when compared to the old approach (so nothing strange going on there)
* General smoke test clicking around the plugin screens and purchasing a label

Only comments are:

* Is it worth adding the TL;DR; notes from this PR to the readme somewhere for reference?
* I like the use of the `prepublish` npm hook so that `npm install` just works. Have we considered this [deprecation notice](https://docs.npmjs.com/misc/scripts#prepublish-and-prepare) though? I think the version of npm we recommend with this project is > `4.0.0`?

DanReyLop:
> I like the use of the prepublish npm hook so that npm install just works. Have we considered this deprecation notice though? I think the version of npm we recommend with this project is > 4.0.0?

Great find! As of `npm 6.4.1`, `prepublish` still runs after `npm install`, but no need for this to bite us in a few years. I've changed the script to run on `prepare`, still works.

> Is it worth adding the TL;DR; notes from this PR to the readme somewhere for reference?

Yep. I've just added the instructions to initialize the submodules. I don't feel confident in the "how to update `wp-calypso`" part yet, since I haven't ever done it and working with submodules isn't super-intuitive.

jrodger:
> I don't feel confident in the "how to update wp-calypso" part yet, since I haven't ever done it and working with submodules isn't super-intuitive.

Fair enough, let's test the instructions next time we want to update the Calypso dependencies and add them to the readme once we know we have it correct. Better that than misleading documentation.

I think this PR is ready to merge!

PR #1604, opened 2019-04-19
Update WC tested version and release 1.19.1

### LINKED ISSUE #1603: WC tested up to 3.6
Test with 3.6 and update the `WC tested up to:` header in woocommerce-services.php

marcinbot:
Fixes #1603

I smoke tested with 3.6.1 using the following steps:
* remove node_modules and reinstall
* `npm run release`
* use the compiled plugin
* tested checkout rates, shipping settings, tax settings, status page, label purchases

No problems found.

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -7,9 +7,9 @@
  * Author URI: https://woocommerce.com/
  * Text Domain: woocommerce-services
  * Domain Path: /i18n/languages/
- * Version: 1.19.0
+ * Version: 1.19.1
  * WC requires at least: 3.0.0
- * WC tested up to: 3.5.5
+ * WC tested up to: 3.6.1
  *
  * Copyright (c) 2017 Automattic
  *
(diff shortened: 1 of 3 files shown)


v18:
Looks good. Tested the following using `npm run release` using node v10.13.0 (npm v6.4.1):

- USPS checkout label rates
- label purchase and refund
- US tax rates

:shipit:

woocommerce-services.php
```diff
@@ -7,9 +7,9 @@
  * Author URI: https://woocommerce.com/
  * Text Domain: woocommerce-services
  * Domain Path: /i18n/languages/
- * Version: 1.19.0
+ * Version: 1.19.1
  * WC requires at least: 3.0.0
- * WC tested up to: 3.5.5
+ * WC tested up to: 3.6.1
```
DanReyLop: For the next times: We don't need to put `3.6.1`. In fact, that may give the impression that we don't support anything from `3.6.2` to `3.6.*`. WooCommerce core only checks major and minor version numbers, and assumes that a WooCommerce patch release can't possibly break anything: https://github.com/woocommerce/woocommerce/blob/393271013606a43f1fed9878054470289ff0cb7f/includes/admin/plugin-updates/class-wc-plugin-updates.php#L201-L213

So, putting `3.6` or even `3.6.*` would probably be marginally clearer.

PR #1608, opened 2019-04-23
Add ability to render server notices in the admin panel

marcinbot:
Fixes server's issue 1268

Serverside PR will need to be shipped first in order not to break the request validation with the new taxes_enabled field

To test:
* Checkout PR 1270 on the server
* Set your base state to CA or TX
* Confirm the notice appears as expected

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -641,6 +641,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			$logger = $this->get_logger();
 			$this->set_help_view( new WC_Connect_Help_View( $schema, $settings, $logger ) );
 			add_action( 'admin_notices', array( WC_Connect_Error_Notice::instance(), 'render_notice' ) );
+			add_action( 'admin_notices', array( $this, 'render_schema_notices' ) );
 		}
 
 		/**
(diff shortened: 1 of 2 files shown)


woocommerce-services.php
```diff
@@ -1369,6 +1370,25 @@ function enqueue_wc_connect_script( $root_view, $extra_args = array() ) {
 				</div>
 			<?php
 		}
+
+		function render_schema_notices() {
+			$schemas = $this->get_service_schemas_store()->get_service_schemas();
+			if ( empty( $schemas ) || ! property_exists( $schemas, 'notices' ) || empty( $schemas->notices ) ) {
```
RadoslavGeorgiev: A very small nitpick, but do we need `property_exists( $schemas, 'notices' )` considering that we will have to deploy the server first anyway?

marcinbot: I think it's better to check for it and handle the situations where the array gets accidentally deleted. There were problems with missing object/array fields before.

RadoslavGeorgiev:
I like how you used a transient in combination with `MONTH_IN_SECONDS` instead of a saved option. In case somebody dismisses the notice without taking action, they'd be reminded again. Nice!

:shipit:

PR #1630, opened 2019-06-19
Try Travis CI debugging

v18:
And others...


v18:
This was a test PR to trigger CI builds without needing to enable builds on branches.

PR #1634, opened 2019-07-12
Support namespaced Jetpack tracking methods

### LINKED ISSUE #1633: Notice: jetpack_tracks_record_event deprecated since version jetpack-7.5
Similar to https://github.com/woocommerce/woocommerce/issues/24067

Screenshot 2019-07-12 at 13 10 35

To reproduce:
* enable `WP_DEBUG`
* perform any action that might trigger Jetpack tracks. For example, disconnect and reconnect Jetpack

marcinbot:
Fixes #1633 

To test:
* Install Jetpack 7.4 installed and test that no notices are given when connecting the site
* Install Jetpack 7.5 and test that no notices are given when connecting the site

### DIFF
--- a/classes/class-wc-connect-jetpack.php
+++ b/classes/class-wc-connect-jetpack.php
@@ -86,5 +86,26 @@ if ( ! class_exists( 'WC_Connect_Jetpack' ) ) {
 				'woocommerce-services-auto-authorize'
 			);
 		}
+
+		/**
+		 * Records a Tracks event
+		 * @param $user
+		 * @param $event_type
+		 * @param
+		 */
+		public static function tracks_record_event( $user, $event_type, $data ) {
+			if ( ! class_exists( 'Jetpack' ) && defined( JETPACK__VERSION ) ) {
+				return false;
+			}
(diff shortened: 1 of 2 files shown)


classes/class-wc-connect-jetpack.php
```diff
@@ -86,5 +86,26 @@ public static function build_connect_url( $redirect_url ) {
...
 		}
+
+		/**
+		 * Records a Tracks event
+		 * @param $user
+		 * @param $event_type
+		 * @param
+		 */
+		public static function tracks_record_event( $user, $event_type, $data ) {
+			if ( ! class_exists( 'Jetpack' ) && defined( JETPACK__VERSION ) ) {
```
DanReyLop: Are you sure about this `&& defined( JETPACK__VERSION )` condition? What are you trying to test for here?

marcinbot: I was basing it on the fix in woo core, but it looks like we don't need to check if Jetpack is installed here. Also, `defined( JETPACK__VERSION )` is broken and would always return false. I fixed this in 71d8e3e where I removed this check and restored the original functionality in the tracks class.

PR #1635, opened 2019-07-15
Add PHP 5.3 requirement

marcinbot:
#1634 introduces namespaces, which are required by Jetpack 7.5

We need to require at least PHP 5.3 if we want to integrate with the latest Jetpack functionalities and make sure this plugin doesn't break. This PR adds this requirement and removes PHP 5.2 from Travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,14 +6,6 @@ language: php
 # Define our matrix of build configurations to test against.
 matrix:
   include:
-   - php: 5.2
-     env: WP_VERSION=4.4 WC_VERSION=2.6.14
-     sudo: true
-     dist: precise
-   - php: 5.2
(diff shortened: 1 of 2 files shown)


woocommerce-services.php
```diff
@@ -8,6 +8,7 @@
  * Text Domain: woocommerce-services
  * Domain Path: /i18n/languages/
  * Version: 1.21.0
+ * Requires PHP: 5.3
```
DanReyLop: This isn't needed, the WordPress 5.2 logic reads the `Requires PHP` header from both the `readme.txt` file and the plugin file. The wordpress.org plugin directory only parses the `readme.txt` file though, so if we're putting it in just one place it should be there.

Unless you prefer to put the `Requires PHP` header in as many places as possible so people have more chances to see it, which I'm fine with :)

PR #1655, opened 2019-09-05
Copy WCS from Calypso to WCS root structure closes #1652

### LINKED ISSUE #1652: Pull WCS code from Calypso into this project
As a first step to remove the Calypso dependency we've decided to copy the WCS code from Calypso into this project which will allow us to iteratively diverge the features being added.

As part of this issue we'll also want to create a `develop` branch to align this codebase with our git flow release process.

roykho:
Fixes #1652 

This PR basically copies the WCS code out of Calypso and into `woocommerce-services/client/extensions`

This is to ensure new features we add will live inside WCS and not in Calypso. When time comes, it will be easier to phase out Calypso completely.


bor0:
The changes look good. Even though the PR is big I did some things to test its impact. Some notes:

1. We still have reference to `wp-calypso` gitmodule in webpack config (I don't know if this is intentional or not).
2. If you run `npm install` without initializing the gitmodule, it will loop endlessly. Initializing gitmodule solves the problem.
3. After I initialized the gitmodule, I ran `diff --recursive wp-calypso/client/extensions/woocommerce/ client/extensions/woocommerce/`. Looks good.

roykho:
> We still have reference to wp-calypso gitmodule in webpack config (I don't know if this is intentional or not).

Yes this is intentional because we still need components from Calypso.

> If you run npm install without initializing the gitmodule, it will loop endlessly. Initializing gitmodule solves the problem.

This behavior exists before this PR so I think its "normal" :)

The thing I am not 100% sure yet is how will we discern what is store.com and what is not to know which version of the WCS to load. Let's discuss this at the roundtable.

roykho:
From what I have learn asking the nice folks on Hydra, store on .com only uses WCS for the REST APIs and the Calypso UI stuff comes directly from the WP-Calypso. So our changes shouldn't affect anything once we have copied it out into the WCS root. We just need to be mindful if we change any of the REST APIs, it would have to be backwards compatible.

Other than that, to test, you would need to add a new business/eCom plan on wp.com

PR #1670, opened 2019-09-20
Add single form of string when needed closes #1663

### LINKED ISSUE #1663: Create shipping label panel text incorrect for single item
When only one item needs to be fulfilled, the text should read, `1 item needs to be fulfilled`, instead we see:

pluralization

It looks like the text would be fine for >1 items, so we probably just need some pluralization logic.

roykho:
Fixes #1663 

Ensure string detects singular and plural when needed.


roykho:
@budzanowski in case you missed this one :)

PR #1692, opened 2019-10-03
Add new/e2e test infrastructure

### LINKED ISSUE #1698: E2E testing infrastructure follow-up
We have a PR (https://github.com/Automattic/woocommerce-services/pull/1692) ready for end to end to end testing locally. 

Here are a list of follow up tasks for the remaining work for that PR:

- [x] `tests/e2e-tests/utils/components.js` Update `await page.waitFor( 2000 );` to wait for element to show up instead of a fixed time. `2000` sometimes is too fast for the product notice to show up and failed the test
- [x] `tests/e2e-tests/utils/flows.js` Update `const WP_ADMIN_EXISTING_ORDER_PAGE = baseUrl + '/wp-admin/post.php?post=82&action=edit';` so that the post ID is passed in instead of fixed.
- [x] Finish the shipping label tests `tests/e2e-tests/specs/admin/create-shipping-label.test.js` 
- [ ] ~Make sure travis-ci pass test using headless mode (`npm run test:e2e` runs in headless mode)~
- [ ] ~Some of "flows' might be duplicated of https://github.com/woocommerce/woocommerce/pull/24218/files, we may need to create follow up to clean this up. Depending on how we want to refactor them.~
- [ ] ~Make sure Travis runs PHPUnit on version 5.3 to 5.6~
- [ ] ~Check if we need PHP 7.3 coverage in Travis' tests~
- [ ] ~Update .travis.yml notification to team email~

harriswong:
Setup travis to run end to end tests. 

Closes #1698


Ferdev:
I've just pushed and update to this PR. The current test suite should pass in headless mode. It's currently pretty slow: it takes around 200s in my machine to go through the entire test suite. I wonder how fast would it be on a mac.
For those who want to give it a try, don't forget to have a test environment ready, and these environment variables populated: `WP_BASE_URL`, `WP_ADMIN_USER_NAME`, `WP_ADMIN_USER_PW`.

laurendavissmith:
> it takes around 200s

@Ferdev did you mean 200ms or is it really that slow?

Ferdev:
Now, that's correct. I just did another run in my lapton an it took 206 seconds.

tests/bin/run-e2e-travis.sh
```diff
@@ -1,12 +1,7 @@
 #!/usr/bin/env bash
-if [[ ${RUN_E2E} == 1 ]]; then
```
harriswong: @budzanowski A little background story on this script and the `RUN_E2E` check.

We added it here because we wanted it to only run on 1 of the matrices, namely, the latest PHP, https://github.com/Automattic/woocommerce-services/blob/93d2d9e7f0699275239b7a921af1b24a6d0f2054/.travis.yml#L53. 

We might want to check with @juliaamosova if we want to run on all versions.

superdav42:
Checkout that green checkbox!
However the e2e check is actually failing in the last 3 tests.
@Ferdev or @budzanowski are these know issues and do we know how to fix them or something I should troubleshoot?

superdav42:
Turns out the e2e tests pass using the storefront theme but fail using twenty twenty theme. So that gives me a place to start.

tests/bin/run-e2e-travis.sh
budzanowski: I may have done something awkward sometime ago. For sure we want to run on only one.

budzanowski:
@superdav42 Yes, this is known. Tests pass locally if you are using a "real" connect server. By real I mean local instance. Tests fail on Travis because we are using mocks instead of the connect server. WCS behaves differently with mocks and this is why e2e tests fail. Either not everything is mocked, or the mocks are missing some data. The way to approach this IMO is to observe what fails exactly and compare that to what the behavior should actually be. From what I remember the last time I was working on it the test was failing during the fulfillment process on this view:

image

When I observe this bit locally the address is validated and correct. My guess is that we are missing mocked endpoint and dataset for `/shipping/address/normalize` connect server API.

https://github.com/Automattic/woocommerce-services/pull/1692/files#diff-ea0da27fc34116fb63904f971c9632f9R2

Ferdev:
hey @superdav42, please also check this file I updated while working on the test data generation spike: https://github.com/Automattic/woocommerce-services/blob/spike/issue-1840/tests/php/class-wc-connect-api-client-local-test-mock.php#L26-L37 (full commit https://github.com/Automattic/woocommerce-services/commit/0192f5a78921283db9b2d6495c57a8cc62a8f8f7)

Some parts of WC Services consume the data from the connect server as an associative array, and the current mocked api client does not support that. That commit fixes that and also I think it also mocks all the endpoints the purchase label process uses.

superdav42:
How do you run the test with the mocked data? It seems to use the real server locally so what does it take to test the mock service so I can check if it's working locally?

budzanowski:
@superdav42 you need to define `WOOCOMMERCE_SERVICES_LOCAL_TEST_MODE`

The Travis E2E tests use this helper plugin to set things up:
https://github.com/Automattic/woocommerce-services/pull/1692/files#diff-bf35422824529e33d2dc30952227fcddR12

c-shultz:
The latest issue related to this work is #1834.

superdav42:
I suggest we merge this PR as is now. It's not 100 of the way done but will not affect the plugin in anyway and it's getting stale. So I suggest we merge it and finish the e2e work in a new branch and issues.

Ferdev:
sounds good to me, @superdav42!

.travis.yml
```diff
@@ -37,26 +37,57 @@ matrix:
...
-   - language: node_js
-     before_script: ""
+  #  - language: node_js
+  #    before_script: ""
+   - name: "E2E tests"
+     php: 7.2
+     env: WP_VERSION=latest WP_MULTISITE=0 WC_VERSION=latest RUN_E2E=1
+     script:
+         - npm run test:e2e
+         - cat $HOME/wordpress/wp-content/debug.log
```
harriswong: Not sure what the `cat` does
but to address travis passing when npm test fails, maybe we can do something like
```
- npm run test:e2e; npm_exit_code=`$?`
 - cat $HOME/wordpress/wp-content/debug.log
- exit npm_exit_code
```
something like that to capture npm code and return after it.

juliaamosova: >something like that to capture npm code and return after it.

The way we do this in Core is by using `spawnSync()` instead of `spawn()` in `tests/bin/e2e-test-integration.js` to capture `jest` exit code and then pass it to `npm`:

https://github.com/woocommerce/woocommerce/blob/master/tests/bin/e2e-test-integration.js#L23-L41

PR #1695, opened 2019-10-07
Move rates out of dropdown.

### LINKED ISSUE #1645: Move rates out of dropdown
We want to remove the rates from the dropdown and list them on the page after a package is selected.

Will need to work with @olaolusoga on how this should work given that the designs dont show it in an accordion like we have now

budzanowski:
Fixes #1645


laurendavissmith:
@budzanowski I did pull down the PR but the `Your customer..` messaging doesnt match the figma design, is that also included in the PR and the screenshot is just old?

budzanowski:
@laurendavissmith Yes, that is true. Was this the scope of the Issue? I don't mind doing that :) not a problem, at all. The issue does not mention this part. It would be good to have a screenshot, of what exactly needs to be addressed, taken directly from Figma. The banner has nothing to do code-wise with the dropdown. The dropdown was removed and replaced with the new ShippnRate components list.

So OK - I will add that bit.

laurendavissmith:
Yeah its been an issue with all of these since I wrote them prior to the designs being finalized, lets go ahead and make it as part of this PR since it seems relatively small?

> It would be good to have a screenshot

💯

laurendavissmith:
Also noting for the PR reviewer that the USPS logo is a placeholder until we have confirmation from USPS on what to use, https://a8c.slack.com/archives/CKNC7KACA/p1570464708034200

budzanowski:
PR updated to address notice change in the design:
image

budzanowski:
@olaolusoga should there be a `.` at the end of the notice sentence?

package.json
```diff
@@ -22,6 +22,7 @@
     "archiver": "^1.3.0",
     "colors": "^1.3.3",
     "confirm-simple": "^1.0.3",
+    "file-loader": "^4.2.0",
```
bor0: Might want to `git add` the lock file as well to not break CI

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/carrier-logo.js
```diff
@@ -0,0 +1,29 @@
...
+ * Internal dependencies
+ */
+
+function CarrierLogo( { carrier_id } ) {
+	switch ( carrier_id ) {
+		case "usps":
+			return <img className="rates-step__carier-logo-image-usps" src={ usps } alt="" />;
+	
+		default:
+			return <div/>
```
Ferdev: Missing semicolon ;

PR #1705, opened 2019-10-08
Remove Calypso dependency

### LINKED ISSUE #1704: [Architecture] Remove Calypso dependency
We need to only use the necessary bits of calypso and don't import all of it. We have a PR that is the result of the work at our Athens meetup.

bor0:
Fixes #1704.

Make sure we identify and test all possible flows before merging.


bor0:
Need to figure out what else we want to do for this PR.

bor0:
Closing for now because of no activity/prioritization. The PR can be used as a reference in the future.

PR #1718, opened 2019-10-10
Tweak/issue 1683

### LINKED ISSUE #1683: Update the banner look/behavior after creating a shipping label
We need to know how the banner the user sees above the order details will look/behave after a shipping label has been created

roykho:
Fixes #1683 

This adds a tracking modal.

To test:
* Click on the "Track Packages" button on the top where "Create Shipping Label" used to be. This button will only show up when you have already created a label.
* Ensure the modal pops up with relevant tracking information.

TODO:
You will see the "Create Shipping Label" button before the "Track Packages" button. I will create a separate issue for this.


client/extensions/woocommerce/woocommerce-services/state/shipping-label/reducer.js
```diff
@@ -144,6 +153,13 @@ reducers[ WOOCOMMERCE_SERVICES_SHIPPING_LABEL_EXIT_PRINTING_FLOW ] = ( state, {
 	};
 };
 
+reducers[ WOOCOMMERCE_SERVICES_SHIPPING_EXIT_TRACKING_FLOW ] = ( state ) => {
```
Ferdev: I don't have a strong opinion here, but we should probably remove the parenthesis here around the `state` argument, like the rest of reducers in this file with only one argument.

roykho:
@Ferdev that error is already present in the "Create Shipping Label" modal and not introduced by this PR. So I will open separate issue for that.

roykho:
@Ferdev another round please.

PR #1723, opened 2019-10-14
Fix/issue 1717

budzanowski:
This fixes multiple API calls in woocommerce services for Labels Purchasing screen.

This fix consist of two parts:

1. Making Labels and Tracking views use the same Redux store. Changes are done in main.js file.
2. Lots of small changes to improve synchronization. Usually isLoading check in query components and isLoading check using thunk inside actions.


client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
```diff
@@ -1093,7 +1098,12 @@ export const fetchLabelsStatus = ( orderId, siteId ) => ( dispatch, getState ) =
 		const setSuccess = json => {
 			response = json.label;
 		};
-
+		console.log( WOOCOMMERCE_SERVICES_SHIPPING_LABEL_STATUS_RETRIEVAL_IN_PROGRESS );
```
bor0: Debug code

client/main.js
```diff
@@ -29,73 +29,94 @@ if ( global.wcConnectData ) {
 	setBaseURL( global.wcConnectData.baseURL );
 }
 
-const getRouteClass = ( classNames ) => {
+const getRouteClassName = ( classNames ) => {
```
bor0: This is a weird function, wondering about the reason for refactoring

budzanowski: I needed the class name to be able to use it as a key in the object that stores different Redux stores. This way I can reuse the store for the object of the same type. Witch I think that we should actually not do either way and just use one store - I will bring this for discussion. But that is a bigger change for later.

client/main.js
```diff
@@ -29,73 +29,94 @@ if ( global.wcConnectData ) {
...
-		if ( window.persistState ) {
-			storageUtils.setWithExpiry( persistedStateKey, Route.getStateForPersisting( state ) );
-		}
-	} );
+			if ( window.persistState ) {
+				storageUtils.setWithExpiry( persistedStateKey, Route.getStateForPersisting( state ) );
+			}
+		} );
+
+		createdStores[ routeClassName ] = store;
```
harriswong: Curious why we want to use multiple stores instead of just 1 store?

Ferdev: Yep! Agree with you. @budzanowski said he'll bring it to our next roundtable for discussion.

harriswong: Oh I see! Thanks!

budzanowski: @harriswong  Yeh, this architecture is weird. This is what we have inherited - multiple stores per different components. This makes working with the whole app very counterintuitive. I have merged the store for components of the same type. Which looks like a step 1. Step 2 will be to have everything under the same Redux store. Right now different components can use different middleware - maybe this is something that is actually required - but I doubt that it is really useful. After my changes various QueryComponents located in different parts of the application started to communicate ( through a shared store ) and stopped duplicating API calls. The label purchase code is a bit more snappy now :)

client/extensions/woocommerce/woocommerce-services/state/shipping-label/reducer.js
```diff
@@ -1153,6 +1154,21 @@ reducers[ WOOCOMMERCE_SERVICES_SHIPPING_LABEL_CLOSE_REFUND_DIALOG ] = state => {
...
+	};
+
+	const newState = {
+		...state,
+		labels: [ ...state.labels ],
+	};
+	newState.labels[ labelIndex ] = labelData;
+	return newState;
+};
+
```
harriswong: Sorry, I have more questions! Do we want to have tests for reducers? They are all pure functions and probably don't need mocks.

budzanowski: I don't know :) Test as often as possible but choose your tests wisely. Let's say that we want more tests - is this a good place to tests or we should pick something different - maybe better ( in the context of wcs ). What do you think?

harriswong: Yea, there are costs for us to achieve quality for our customers. So I agree to pick a good place and not spend time to add test simply for coverage. I don't know what the good place is though. 

Tests are most valuable when added in area we lack confidence in. If we don't need confidence in this area, then yea we don't need to test.

PR #1724, opened 2019-10-15
Fix a glitch when closing the "Add Package" popover

Ferdev:
There's no need to reset the list of predefined packages when the modal
is dismissed, since it's repopulated whenever the modal is open.

Thanks to @bor0 for pointing out the root cause!

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/packages/reducer.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/packages/reducer.js
@@ -71,7 +71,6 @@ reducers[ WOOCOMMERCE_SERVICES_PACKAGES_DISMISS_MODAL ] = state => {
 		modalErrors: {},
 		showModal: false,
 		showOuterDimensions: false,
-		currentlyEditingPredefinedPackages: null,
 	} );
 };
 


bor0:
@Ferdev I approved without checking, but with this PR if you re-open the dialog will it remember what you saved last?

Ferdev:
Yes, I tested this thoroughly and the behavior is the same clearing the predefined packages list or not doing it. Every time you open the modal, that list is populated as you can see here: https://github.com/Automattic/woocommerce-services/blob/fix/issue-1714/client/extensions/woocommerce/woocommerce-services/state/packages/reducer.js#L49

My guess is it was being cleared only to reduce the size of the state object.

bor0:
My only concern is it remembering the last pick - not sure if it is a good UX. Other than that all good

Ferdev:
Yes, that's the actual behavior in `develop` and I'd say that's intended. @aleftick, when the user clicks the "Add package" link and the "Custom/Service package" modal is open, on the "Service" tab if the user selects an option, closes the modal and opens it again, the modal remember their last selection. Do you know if this is intended behaviour?

c-shultz:
My assumption is that they chose to have it prechecked because you can't add two of the same service package. The services packages are either added or not. You can actually remove the service packages by unchecking them and then pushing `Add packages` to save the changes.

image

The way that modal works currently, it would more aptly be titled `Manage packages` instead of `Add a package`

aleftick:
I think there might be something missing in the flow. Let see if @olaolusoga can clarify.

In our current flow (the one live right now), the merchant selects a package from a list of his packages. The packages that are displayed, had to be entered/selected by the merchant in WCShip settings prior. 
Screen Shot 2019-10-15 at 12 44 41 PM

The challenge with that design is that if a merchants wants to introduce a new package, he needs to go back to WCShip settings to add it and then restart the label printing process. It isn't ideal and that is what this new design is trying to address. 

However, in the new design, I don't see a way for a merchant to select a package that he previously saved as a "Custom package" or from the packages he selected from "Service package".

olaolusoga:
> If it's the "Add a package" I think it should act the same way if it was the first package, nothing "remembered".
> 
> If it's the "Package details" drop-down, it should remember the selection and let the merchant make changes if needed.

This is accurate.

---

> The packages that are displayed, had to be entered/selected by the merchant in WCShip settings prior.

A merchant doesn't need to select packages within settings before it's displayed as an option. Here's a screenshot of the default setting which show all options when creating a label—pulled from a pressable test site.
Screen Shot 2019-10-15 at 6 42 52 PM

when I select a package within settings, it's added to the package list:
Screen Shot 2019-10-15 at 6 44 59 PM

When the create label modal appears, it shows the added option as an addition to the list:
Screen Shot 2019-10-15 at 6 45 56 PM

What the new design is doing is removing the need to go to setting—excluding when a custom package needs to be deleted—and moving package selection within the create a label step. It's literally pulling the modal from the settings area into the create a label step. There are small deviations, like the way the service packages are ordered. The new design is removing the distinctions between priority mail, and priority mail express. The package selection is focusing on the package the merchant is planning to use, and then the rates deals with the speed and pricing, vs the store owner trying to figure out the package, and speed before knowing the price.

In the new design, when a package is selected, we'll default to the package being one of the options the merchant can choose. If they need to add a new package option, then they would "add a new package".

Ferdev:
Just to clarify, this is the current behavior of the "Add Package" modal in the `develop` branch: https://www.dropbox.com/s/bepprijxtv0jl56/wcs.add_package.remember.selection.gif?dl=0

As you can see, once the user selects a couple of predefined packages, the modal remembers that selection. If you open that modal again while you're creating that label, it will preselect the predefined packages the user selected previously.

@aleftick, you said:
> If it's the "Add a package" I think it should act the same way if it was the first package, nothing "remembered".

So we need to create a new issue to address this and remove this unwanted behavior, right?

On a side note, just wanted to +1 @c-shultz suggestion of renaming the link to open this modal from "Add Package" to "Manage packages".

Thanks a lot to everyone for your help!

aleftick:
First my apologies to Ola. I didn't fully understand the flow (there is a flow for a first time user that has no package settings).  Some of my earlier comments aren't relevant.

>On a side note, just wanted to +1 @c-shultz suggestion of renaming the link to open this modal from "Add Package" to "Manage packages".

I think y'all have a point here. “Add package” could mean adding a second box to the order or adding a package to the list.

I discussed this with Ola and I suggested that we move the "Add new Package" link into at the bottom of the drop-down list. Clicking on it would then open the "Add new package" modal. This would avoid users confusing "Add new package" with adding a second box to the order.

If this is not something we can do in this release, we could consider it for a future release.

Screen Shot 2019-10-17 at 3 03 10 PM

> As you can see, once the user selects a couple of predefined packages, the modal remembers that selection. If you open that modal again while you're creating that label, it will preselect the predefined packages the user selected previously.

That is the correct behaviour. No changes are needed. _Our discussion was convoluted us referring two different things_

Ferdev:
Thanks for the clarification, @aleftick! Then we can leave it as it is right now, and I'll create a new issue for moving the "Add new package" to the bottom of the drop-down. I personally wouldn't add it to this release, but I'll leave that call to @laurendavissmith.

PR #1725, opened 2019-10-15
Preselect last used package

### LINKED ISSUE #1682: Preselect the last used package
When a user opens a new shipping label modal their last used package should be preselected for them

bor0:
Fixes #1682.

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/label-settings/reducer.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/label-settings/reducer.js
@@ -39,9 +39,7 @@ reducers[ WOOCOMMERCE_SERVICES_LABELS_INIT_FORM ] = (
 			...formMeta,
 			pristine: true,
 			isLoaded: true,
-			user: {
-				...userMeta,
-			}
+			user: userMeta,
 		},
 		data: {
 			...state.data,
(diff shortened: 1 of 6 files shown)


client/extensions/woocommerce/woocommerce-services/state/label-settings/selectors.js
```diff
@@ -34,6 +34,11 @@ export const getLabelSettingsStoreOptions = ( state, siteId = getSelectedSiteId(
 	return form && form.storeOptions;
 };
 
+export const getLabelSettingsUserMeta = ( state, siteId = getSelectedSiteId( state ) ) => {
+	const form = getLabelSettingsForm( state, siteId );
```
budzanowski: Maybe use getLabelsSettingsFormMeta. It already gets meta from the form. Either that or check if meta exists. You risk of getting: `Cannot read property 'user' of undefinded`.

client/extensions/woocommerce/woocommerce-services/state/label-settings/reducer.js
```diff
@@ -39,6 +39,9 @@ reducers[ WOOCOMMERCE_SERVICES_LABELS_INIT_FORM ] = (
 			...formMeta,
 			pristine: true,
 			isLoaded: true,
+			user: {
+				...userMeta,
```
budzanowski: is userMeta also an object? If yes then do you need destructuring or simple assignment would be sufficient:
`user: userMeta`

client/extensions/woocommerce/woocommerce-services/state/label-settings/selectors.js
bor0: Good catch!

PR #1731, opened 2019-10-18
Add signature option to rate list

### LINKED ISSUE #1721: Add USPS signature required checkbox and logic
Continuation from https://github.com/Automattic/woocommerce-connect-server/pull/1307 we need to add the actual checkbox for the "Signature Required" to the USPS offerings. Additionally, we need the logic to send the input as a flag to the API to request the rate + signature.

Ref here for design https://github.com/Automattic/woocommerce-connect-server/issues/1301

c-shultz:
Fixes #1721

Adds signature required option as checkbox:

(note: signature required is not an option for express mail so the checkbox is not shown there)

### Changes
- Adds checkbox to UI (new data store entry was needed to store this state)
- Adds signature required price as a separate line item
- The signature required rates are queried by adding a second package to the request with the `signature='yes'` option, so a single request gets rates for signature _and_ no signature.
     - The second package is named with a `_wcs_rate_type_signature_required` suffix so the resulting rates can be all be nicely grouped together before sending them back to the frontend.


c-shultz:
The direction I’ve been working towards is to allow the $2.60 signature fee to be tacked onto any USPS label purchase. I got the UI in place for this (including checkboxes and adding $2.60 to the total) and am sending `signature='yes'` to the Connect Server when checked.

After getting into things on the connect server, I think this may be the wrong approach. It presents a couple problems:

- Not all USPS services allow signature requirement ([International and Priority Express are excluded](https://faq.usps.com/s/article/How-is-Signature-Confirmation-and-Signature-Confirmation-Restricted-Delivery-Used)). _We could work around this by hardcoding for certain services, but at this point we would be hardcoding the $2.60, carrier, and shipping option exclusions. That's messy/hacky._
- The connect server expects the following flow:
   1. WCS client requests rates based on package dimension/weight and service options (**like signature required**)
   2. Connect server retrieves rates from carrier(s) per the client's request, and sends these back to the WCS client
   3. Store owner picks from these rates and chooses one to purchase.
   4. WCS client sends chosen rate ID to Connect Server for purchase.

Adding the $2.60 charge in after the Connect Server has already requested rates, doesn't fit into the above flow.

To get this to work with the design and within the constraints of the Connect Server architecture, I'm now exploring a different approach. I'm thinking we should make multiple two rate requests: one for signature required and one for no signature required. This has the downside of requiring two rate requests up front, but has a number of advantages:
- No more hardcoding of $2.60 amount for USPS. We can just get latest rate from easy post
- No need to figure out how to exclude priority/international methods from signature required
- Fits into Connect Server architecture of requesting rates and then later choosing from that list of rates for purchase

This new direction requires a couple chunks of work:
- Add the double-request to WCS to get rates with/without signature required
- Modify the UI work (on this PR) so the with/without signature rate pairs are displayed together and selected between with the signature required checkbox per the design

budzanowski:
I have two comments:

1. I think that we should not use and add an additional store state for holding the is signature required parameter. We can hold the per rate state locally - this makes each rate remember what was the setting. On top of that, we should add this value to the rates update action we have already, that would remember the selection for the currently selected action inside the global store. This is just another parameter we would pass into `updateValue` function inside `shipping-rate.js`. I think that this approach should make the required change significantly smaller - but this is just a hunch. We need to modify one action, one reducer and maybe add one selector without adding other things.

2. What happens when we send a request for a rate with a signature required but for a service that does not have a signature option. Maybe  we could figure out from the response that there is no signature available and that the price is final and ew don't need to do another non signature request. Maybe we could also figure stuff out for requests for services that have signature option. Maybe the signature prices is somewhere there. We could save a lot of time and requests this way.

I don't have any objections to this PR in general. It is good the way it is. So if you don't think the my comments require any action then :shipit:

c-shultz:
>I think that we should not use and add an additional store state for holding the is signature required parameter. We can hold the per rate state locally - this makes each rate remember what was the setting. On top of that, we should add this value to the rates update action we have already, that would remember the selection for the currently selected action inside the global store. This is just another parameter we would pass into updateValue function inside shipping-rate.js. I think that this approach should make the required change significantly smaller - but this is just a hunch. We need to modify one action, one reducer and maybe add one selector without adding other things.

I like the idea of not storing the signatureRequired box for every service in the global state. The latest commit implements that suggestion. It moves that state into `ShippingRate` and only puts `signatureRequired` for the currently selected rate up into the data store. This removes the need for the additional action and reducer, since it uses `updateState` for currently selected service and `signatureRequired`.

Could you give the PR another once over after that change, @budzanowski?


>What happens when we send a request for a rate with a signature required but for a service that does not have a signature option. Maybe we could figure out from the response that there is no signature available and that the price is final and ew don't need to do another non signature request. Maybe we could also figure stuff out for requests for services that have signature option. Maybe the signature prices is somewhere there. We could save a lot of time and requests this way.

It's always requesting rates for both signature and non-signature options on this branch. It's all combined in one request by adding a dummy package with `_wcs_rate_type_signature_required` in its name. EasyPost returns rates for both even if signature required is not available for a given service. The only way I could find to tell if the signature requirement is available is to check for a price difference between the two rates. I filter out the signature option by checking whether or not there is a price difference on the front end with `requiredSignatureCost > 0`:
https://github.com/Automattic/woocommerce-services/pull/1731/files#diff-464e08994433b72877e47f9464d6d925R84

classes/class-wc-rest-connect-shipping-rates-controller.php
```diff
@@ -48,6 +62,94 @@ public function post( $request ) {
...
+		}
+
+		return array(
+			'success' => true,
+			'rates'   => $response,
+		);
+	}
+
+	/**
+	 * Get standard rates along with rates for special options
```
budzanowski: Full stop.

budzanowski:
Nice! Looks great now. It is a pity that we don't have sufficient information to do just one request.

PR #1738, opened 2019-10-24
Use localStorage for common requests.

budzanowski:
Each view of the order page requires fetching of following objects using AJAX request:
settings, continents, packages. Since the objects do not change in general, and if they change we should be able to know when we could optimize to user experience by temporary caching those values in the browsers localStorage. This PR is a quick demonstration of how this could work.
Values are stored for 1 minute so when we do a second loading of the same page it should go much faster.

The strategy of data retention should be different in the final solution - this one is just for demo purposes.


budzanowski:
TODO:
Create an issue for this so we will be able to schedule work for this.

PR #1752, opened 2019-10-29
Remove validation from PackageDialog

### LINKED ISSUE #1745: Package options can't be saved with no payment options
If you try to add a package with no payment options added, you run into an error message:
image

bor0:
Fixes #1745.

### DIFF
--- a/client/apps/shipping-settings/view-wrapper.js
+++ b/client/apps/shipping-settings/view-wrapper.js
@@ -63,8 +63,7 @@ class LabelSettingsWrapper extends Component {
 	onSaveChanges = () => {
 		this.props.createWcsShippingSaveActionList(
 			this.onSaveSuccess,
-			this.onSaveFailure,
-			this.onPaymentMethodMissing
+			this.onSaveFailure
 		);
 	}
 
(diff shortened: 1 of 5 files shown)


bor0:
@c-shultz I assigned you for a review since @budzanowski is AFK.

Unrelated to these changes but with a similar scenario, can you please confirm that you can go through the process without having any cards entered?

I am not sure if it's my environment but here's what I get:

Untitled

c-shultz:
I have seen the 500 error really intermittently (but not while testing this PR). I've always attributed it to something not working quite right with my environment. Restarting things seemed to resolve it. The environment is so complex, it's hard to tell the root cause. :\

We should keep an eye for it though in case there's a real issue...

bor0:
@c-shultz do you think we should re-introduce the logic for `To purchase shipping labels...`?

c-shultz:
@bor0, I don't think we want that anymore since one of the big changes in this release is to not require the credit card until after all the label options are selected.

bor0:
@c-shultz ahh I see, I remember now. Thanks!

bor0:
@c-shultz got more info from @DanReyLop on the errors we're seeing. It seems they are related to our dev envs only.

Reference: https://a8c.slack.com/archives/CKNC7KACA/p1572390559102600

PR #1755, opened 2019-10-29
Initial implementation of ShippingSummary

### LINKED ISSUE #1737: Weird UI for package modal
Screen Shot 2019-10-22 at 17 49 32

The part on the right side looks odd. One thing we noticed is that the Shipping summary is missing, maybe once we introduce it will look better

bor0:
Fixes #1737.

Adds shipping summary:
image

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
@@ -22,6 +22,7 @@ import {
 	pick,
 	sumBy,
 	uniqBy,
+	isNull,
 } from 'lodash';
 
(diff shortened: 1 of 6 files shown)


c-shultz:
@superdav42 ,
Here's how it's looking now:
image

>Also it seems a little odd to have the origin address in two places. This even confused @laurendavissmith. It seems we should really consider removing the origin accordion and have the edit button open a modal as it was in the original design.

Yeah, that's a good point, but I think this PR is still a good enough improvement on its own once it's ready to go. Maybe we should make a separate issue for the above?

laurendavissmith:
@c-shultz issue would be great, low pri enhancement

PR #1763, opened 2019-10-30
Build i18n files before creating a release

### LINKED ISSUE #1753: Create language files during build process
One way to do this is to `npm run i18n` in `prerelease` within package.json?

The lang files are commited so we probably have to remove and gitignore them

bor0:
Fixes #1753.


superdav42:
Actually, I think we want to remove `i18n/` files from source control and add the folder to `.gitignore` since this appears to generate them during release.

PR #1765, opened 2019-10-30
Display notices when base country/currency mismatches

### LINKED ISSUE #1751: Display WC notices when there's a configuration mismatch
We have a lot of conditionals that determine whether we should display the printing label button or not: https://github.com/Automattic/woocommerce-services/blob/abc2bdb/classes/class-wc-connect-shipping-label.php#L311-L348

In other shipping extensions if there's a configuration mismatch you will get a WC notice "Your base currency must be set to ..." or "Your base country must be set to ...".

In the case of WCS, we just hide the button silently.

Thanks to @Ferdev for helping me figuring this out on my local install.

bor0:
Fixes #1751.

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -212,7 +212,6 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		/**
 		 * environment_check function.
 		 *
-		 * @access public
 		 * @return void
 		 */
 		public function environment_check() {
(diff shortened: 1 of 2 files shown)


woocommerce-services.php
```diff
@@ -209,6 +209,33 @@ function wpcom_static_url($file) {
...
+			$admin_page = 'wc-settings';
+
+			$base_currency = get_woocommerce_currency();
+			$base_location = wc_get_base_location();
+
+			if ( ! WC_Connect_Shipping_Label::is_supported_currency( $base_currency ) ) {
+				echo '<div class="error">
+					<p>' . sprintf( __( 'WooCommerce Services requires that the <a href="%s">currency</a> is set to US Dollars.', 'woocommerce-services' ), admin_url( 'admin.php?page=' . $admin_page . '&tab=general' ) ) . '</p>
+				</div>';
+			} else if ( ! WC_Connect_Shipping_Label::is_supported_country( $base_location['country'] ) ) {
```
superdav42: using `else` here does not seem necessary since both currency and country could be unsupported so we should probably show both notices.
```suggestion
			} 
			if ( ! WC_Connect_Shipping_Label::is_supported_country( $base_location['country'] ) ) {
```

woocommerce-services.php
```diff
@@ -209,6 +209,33 @@ function wpcom_static_url($file) {
 			return set_url_scheme( $url );
 		}
 
+		/**
+		 * environment_check function.
+		 *
+		 * @access public
```
superdav42: `@public` in redundant and not necessary https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-functions-class-methods

PR #1773, opened 2019-10-30
Fix multi-package support

### LINKED ISSUE #1743: Can't select options for more than one package
Items can be moved to other packages, but you can't select shipping options for those packages, so there's no way to proceed:

c-shultz:
Closes #1743 
Brings back original package selection design:
image

Adds section headers to rates lists for separate packages:
image


laurendavissmith:
fantastic work @c-shultz!

lets make sure we either get the max 5 per package rate with load more into this or open a new issue for it

PR #1776, opened 2019-10-31
Change logic to how items are counted

### LINKED ISSUE #1767: Quantity is not being taken into account for the notice
If I have more than 1 qty of the same item to ship I would expect to see that number reflected in the notice but when I order multiple quantities of the same item I'm only seeing `1 item` in the notice

Screen Shot 2019-10-30 at 8 02 36 AM

bor0:
Fixes #1767.

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -379,14 +379,26 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 			return $product && $product->needs_shipping();
 		}
 
+		/**
+		 * Reduce items to sum their quantities.
+		 *
+		 * @param  int   $sum  Current sum.
+		 * @param  array $item Item to add to sum.
+		 * @return int
+		 */
+		protected function reducer_items_quantity( $sum, $item ) {
+			return $sum + $item->get_quantity();
+		}
+
 		public function meta_box( $post, $args ) {
 			$order = wc_get_order( $post );
 			$order_id = WC_Connect_Compatibility::instance()->get_order_id( $order );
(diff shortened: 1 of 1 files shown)


classes/class-wc-connect-shipping-label.php
```diff
@@ -379,14 +379,26 @@ protected function filter_items_needing_shipping( $item ) {
...
+		 */
+		protected function reducer_items_quantity( $sum, $item ) {
+			return $sum + $item->get_quantity();
+		}
+
 		public function meta_box( $post, $args ) {
 			$order = wc_get_order( $post );
 			$order_id = WC_Connect_Compatibility::instance()->get_order_id( $order );
 			$items = array_filter( $order->get_items(), array( $this, 'filter_items_needing_shipping' ) );
+			$items = array_reduce( $items, array( $this, 'reducer_items_quantity' ), 0 );
```
superdav42: This would be much more clear if `$items` was renamed here since we are changing its type.
```suggestion
			$items_count = array_reduce( $items, array( $this, 'reducer_items_quantity' ), 0 );
```

classes/class-wc-connect-shipping-label.php
```diff
@@ -379,14 +379,26 @@ protected function filter_items_needing_shipping( $item ) {
...
 			$payload = array(
 				'orderId' => $order_id,
 				'context' => $args['args']['context'],
-				'items'   => ! empty( $items ) ? count( $items ) : 0,
+				'items'   => $items,
```
superdav42:
```suggestion
				'items'   => $items_count,
```

bor0:
Fixed in d42caa69

PR #1781, opened 2019-11-02
Limit 5 rates per package 

### LINKED ISSUE #1778: We should only show 5 rates per package
Follow-up to #1773 (see comment https://github.com/Automattic/woocommerce-services/pull/1773#issuecomment-548152025)

We should hide rates behind a "view more" link if there are more than 5 in the rates list for a given package.
image
(the above screenshot is only showing 4 rates, but if there were more, they should be hidden)

harriswong:
Added "Show More" button to the ShippingRates component. The button will trigger an event which will update a state change on 2 arrays: 
- packageNames //represents all available rates 
- visiblePackageRates //represent rates that are showing to the user

Every time "show more" is clicked, item in `packageNames` will flow into `visiblePackageRates`. "Show more" button will disappear when `packageNames` is empty.

### To test locally with show more button
1. You can force "Show more" button to show up by changing 5 to 1. Change `MAX_VISIBLE_RATE` to 1
2. You should see "Show more" button
3. Clicking rates should update pricing summary

### Test case where there are not enough rates to have "Show more" button
1. You can either find a postal code that has less than 5 rates, or change `MAX_VISIBLE_RATE` to something like 20

### To multiple packages should have its own "Show more" button 
1. Create shipping label with multiple products
2. package them into different packges
3. Each should have its own "Show more" button.

Closes #1778

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
@@ -110,7 +110,6 @@ class PackageShippingRates extends Component {
 					</div>
 				) : null }
 				{ Object.values(
-					// TODO: figure out why this doesn't work
 					mapValues( this.state.visiblePackageRates, ( ( serviceRateObject ) => {
 						const { service_id } = serviceRateObject;
 						const rateObjectSignatureRequired = find( signatureRates, r => service_id === r.service_id );
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
@@ -140,7 +140,13 @@ class PackageShippingRates extends Component {
 						/>
 					} ) )
 				) }
-				<button onClick={ this.handleShowMore } >Show more</button>
+
+				{ this.state.packageRates.length > 0 ? (
+					<div className="rates-step__package-container-rates-show-more" role="button" onClick={ this.handleShowMore }
+						>Show more rates
+					</div>
+				) : null }
+
 				{ packageErrors.slice( 1 ).map( ( error, index ) => {
 					// Print the rest of the errors (if any) below the dropdown
 					return <FieldError type="server-error" key={ index } text={ error } />;
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
@@ -65,7 +65,7 @@ class PackageShippingRates extends Component {
 		super(props);
 
 		this.state = {
-			packageRates: this.props.availableRates.default.rates,
+			packageRates: JSON.parse(JSON.stringify(this.props.availableRates.default.rates)),
 			visiblePackageRates: []
 		}
 
@@ -125,8 +125,7 @@ class PackageShippingRates extends Component {
 				) : null }
 				{ Object.values(
 					// TODO: figure out why this doesn't work
-					// mapValues( this.state.visiblePackageRates, ( ( serviceRateObject ) => {
-					mapValues( this.props.availableRates.default.rates, ( ( serviceRateObject ) => {
+					mapValues( this.state.visiblePackageRates, ( ( serviceRateObject ) => {
 						const { service_id } = serviceRateObject;
 						const rateObjectSignatureRequired = find( signatureRates, r => service_id === r.service_id );
 						return <ShippingRate
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
@@ -3,7 +3,7 @@
 /**
  * External dependencies
  */
-import React from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { localize } from 'i18n-calypso';
 import { mapValues, find } from 'lodash';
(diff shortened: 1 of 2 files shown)


client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
```diff
@@ -29,43 +29,89 @@ const renderRateNotice = translate => {
...
+		</div>
+	}
+}
+
+class PackageShippingRates extends Component {
+	constructor(props) {
+		super(props);
+
+		this.state = {
+			packageRates: JSON.parse(JSON.stringify(this.props.availableRates.default.rates)),
```
budzanowski: What is the purpose of this line?

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
```diff
@@ -29,43 +29,89 @@ const renderRateNotice = translate => {
...
+			visiblePackageRates: []
+		}
+
+		this.state.visiblePackageRates = this.state.visiblePackageRates.concat(this.state.packageRates.splice(0, MAX_VISIBLE_RATE))
```
budzanowski: This feels very complicated.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
```diff
@@ -29,43 +29,89 @@ const renderRateNotice = translate => {
...
+		this.handleShowMore = this.handleShowMore.bind(this);
+	}
+
+	handleShowMore() {
+		this.setState({
+			visiblePackageRates: this.state.visiblePackageRates.concat(this.state.packageRates.splice(0, MAX_VISIBLE_RATE))
```
budzanowski: This also feels complicated.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
```diff
@@ -29,43 +29,89 @@ const renderRateNotice = translate => {
...
-
 				{ ! hasSinglePackage ? (
 					<div className="rates-step__package-container-rates-header">
-						{ translate( 'Choose rate: %(pckg)s', { args: { pckg: packageNames[ pckgId ] } } ) }
+						{ translate( 'Choose rate: %(pckg)s', { args: { pckg: packageNames } } ) }
 					</div>
 				) : null }
 				{ Object.values(
-					mapValues( packageRates, ( ( serviceRateObject ) => {
+					// TODO: figure out why this doesn't work
```
budzanowski: Is this TODO still valid?

budzanowski:
I don't really understand the purpose of splitting into two classes. Could you explain a bit more?

I have adjusted https://github.com/Automattic/woocommerce-services/pull/1782 to be usable without bumping React version by adding a small state providing wrapper that allows easy show more implementation. Could you please review it?

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
harriswong: This is to handle object references for price summary. I added a comment in the commit here: https://github.com/Automattic/woocommerce-services/pull/1781/commits/5847be0f4d3413f34a9db50edbf854abbb694d3e

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
harriswong: I needed a way to move the `packageRates` array elements from one end to the other. Something like this
```
[] [a, b, c, d]
[a] [b, c, d]
[a, b] c, d]
...
```
I have tried a few approaches with slice, splice, etc but none of the native JS function allows me to make this shorter. i am open to other ways.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
harriswong: Nope, we can remove this now.

harriswong:
Regarding splitting `ShippingRates` up to `ShippingRates` and `PackageShippingRates`. I was thinking more of the minimal component hierarchy. Here is how I visualize it:
image

When I needed to add the "Show more" button, I noticed the ShippingRates is a function that returns a div block that returns a map function that returns another rendering. If I was to add show more, then every time show more is clicked, this whole thing will need to be re-render.

I was thinking of breaking it down to the components in the screenshot i have above so "Show more" can live inside "PackageShippingRates". 

`ShippingRates` responsibility is to load all the packages' rates + error/feedback message
`PackageShippingRates` responsibility is to load all the rates for this package + "show more"
`ShippingRate` is responsible by its own rate rendering.

harriswong:
> I have adjusted #1782 to be usable without bumping React version by adding a small state providing wrapper that allows easy show more implementation. Could you please review it?
Yes, I will take a look at that.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
budzanowski: You store how much you need in a variable that you update whenever `Show more` is clicked. You use the value from that variable to update the number of things that you want to show.

You can `splice` or you can use `lodash` `take` for example: https://github.com/Automattic/woocommerce-services/pull/1782/commits/4aa43cd908554221401ea8694cecefc042a6c864#diff-8f6bc9adbfa669876e5a156a4c4411cdR53

budzanowski: At least that is how I think about it. This way the number of values is stored explicitly.

budzanowski: Since your version works I would not worry about this too much right now :P

bor0:
Thinking of multi-carrier I'd actually vote to put this on hold now. I propose implementing something like this with true pagination on the backend.

laurendavissmith:
@bor0 given the back and forth we keep having on this i'm ok with that

harriswong:
PR no longer relevant. Not merged. Closing.

PR #1782, opened 2019-11-03
Alternative solution to show more.

budzanowski:
Use useState for handling show more button.


harriswong:
👀

harriswong:
@budzanowski  I tested this with 2 packages. Upon clicking "Show more", both packages will add more rates. It should only add 1 rate to the package I clicked show more from. I think it has something to do with the show more state is dependent on the whole shipping rates, instead of the individual packages. I didn't look deep enough.

You might also want to test 2 different packages that have different amount of rates. For example, 1st package with 5 rates, 2nd package with 2 rates.

budzanowski:
Yep you are correct. this was a shared state issue. I have made per package state to fix that.

PR #1783, opened 2019-11-05
Fixes loading of admin tooltips.

### LINKED ISSUE #1760: Discounted shipping label tooltip is pointing at "Design your store"
First time user experience user tool tip for discounted shipping label seems to be pointing at the wrong thing

To reproduce, go create a fresh store in Jurassic ninja. 
1. Setup a new store
2. Install WooCommerce and WooCommerce Service
3. Setup store
4. Add product, checkout order
5. Go to orders page and tooltip not pointing to shipping label button

image

Ferdev:
Admin tooltips were being called to early in the loading process of the admin page. This was causing some tooltips to be displaced if the page hadn't fully loaded and the elements the tooltips are attached to change their position once the page is fully loaded.

Closes #1760

### DIFF
--- a/client/admin-pointers.js
+++ b/client/admin-pointers.js
@@ -4,7 +4,7 @@
  */
 import jQuery from 'jquery';
 
-jQuery( document ).ready( function( $ ) {
+( function( $ ) {
 	function show_pointer( pointers, i ) {
 		if ( ! ( Array.isArray( pointers ) && pointers[ i ] ) ) {
 			return;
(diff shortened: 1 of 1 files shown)


harriswong:
👀

harriswong:
To reproduce, delete the option by finding it via `SELECT * FROM wordpress.wp_options where option_name like '%storefront_nux_dismissed%';`

harriswong:
Tried to scroll down while it's loading, the tooltip does show up properly as well.

harriswong:
LGTM 🚢

PR #1786, opened 2019-11-06
Whitelist free signature confirmation for Express service.

### LINKED ISSUE #1771: Certain rates are filtering out signature required even if its included
Certain rates include signature required as part of their rate but this is being filtered out on our rates screen. We need to not filter these rates and allow them to set signature required if its included in the rate and also say in the text that its included at no charge.

**$0 doesn’t necessarily mean that is include for free. Some services might not offer it (like international). Is there a way to know from the API response? Or should look to usps sla and we hard code it?**

@aleftick we'll need the verbiage for this

Slack thread with more info: https://a8c.slack.com/archives/CAKN301C3/p1572396322381400

More info from @c-shultz 

> I specifically filtered out the signature option when the price difference was $0. I made a (maybe wrong) assumption that if there was no price difference, then signature was not available.
The fix may be as simple as removing that check if I’m understanding the option correctly

c-shultz:
Fixes #1771 

Allows 'free' signature confirmation for USPS Express service. All other signature rates that are the same as non-signature rates are assumed to be invalid rate options. I believe this logic should handle all the cases we're concerned with.

image

/cc @budzanowski

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/shipping-rate.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/shipping-rate.js
@@ -52,10 +52,21 @@ class ShippingRate extends Component {
 		} = this.props;
 		let requiredSignatureCost = null;
 		let details = 'Includes tracking';
+		let requiredSignatureCostText;
 
 		if ( null !== rateObjectSignatureRequired ) {
 			requiredSignatureCost = rateObjectSignatureRequired.rate - rateObject.rate;
+			if ( requiredSignatureCost > 0 ) {
+				requiredSignatureCostText = translate( '+%(price)s',
+					{ args: { price: formatCurrency( requiredSignatureCost, 'USD') } }
+				);
+			} else {
+				requiredSignatureCostText = translate( 'free' );
+			}
 		}
+
+		// USPS express service includes signature confirmation for free.
+		const signatureRequirementAllowed = requiredSignatureCost > 0 || ( 'Express' === service_id );
 
 		switch ( carrier_id ) {
 			case 'usps':
@@ -81,11 +92,11 @@ class ShippingRate extends Component {
 						<div className="rates-step__shipping-rate-description-title">{ title }</div>
 						<div className="rates-step__shipping-rate-description-details">
 							{ details }
-							{ null !== requiredSignatureCost && requiredSignatureCost > 0 ? (
+							{ null !== requiredSignatureCost && signatureRequirementAllowed ? (
 								<CheckboxControl
 									label={ translate(
-										'Signature Required (+%(price)s)',
-										{ args: { price: formatCurrency( requiredSignatureCost, 'USD') } }
+										'Signature required (%(price)s)',
+										{ args: { price: requiredSignatureCostText } }
 									) }
 									disabled={ ! isSelected }
 									checked={ signatureRequired }


budzanowski:
That looks great. I only have one question regarding the APO/... and other restricted areas.
I have checked the Palau Island delivery and this is what I get for standard packaging:
image

We have Express services - this means that we assume that delivery confirmation is part of the service and can be requested. At the same time, Palau ( its zip code ) is listed as not eligible for Delivery Confirmation. I am looking at https://faq.usps.com/s/article/How-is-Signature-Confirmation-and-Signature-Confirmation-Restricted-Delivery-Used for this info.

At the same time,
 @Ferdev has pointed out that maybe we have two different things here. One is the Delivery Confirmation service option. At the same time, we have a signature request option for Express Mail :) this means that we can't offer the Delivery Confirmation but a delivery option from the Express Mail is still available. And this would be OK but at the same time for custom packaging I get the following offer for Palau:
image

For me, it looks wrong! We should not get signature options for anything except Express. That is according to USPS documentation. EasyPost claims that this is just passed by them to USPS. This would mean that USPS API is showing rates with signature option. If we agree that EasyPost has nothing to do with this I say let's go with this solution bc we can't do anything if USPS is saying that they have a signature requested rate. 

Long story short 🚢 

@aleftick @c-shultz thoughts?

c-shultz:
>@c-shultz Do you maybe know that this is because both requests return the same thing or the rates requests does not return anything and we get rates response bc how we merge rates on the server?

I checked this and saw that we're getting rates for both signature and non-signature for international shipments. The rates are the same though, so it hides it. As far as I can tell, if we add the signature required option to the request, it will _always_ give us a rate whether or not it's actually available...

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/shipping-rate.js
```diff
@@ -81,11 +92,11 @@ class ShippingRate extends Component {
 						<div className="rates-step__shipping-rate-description-title">{ title }</div>
 						<div className="rates-step__shipping-rate-description-details">
 							{ details }
-							{ null !== requiredSignatureCost && requiredSignatureCost > 0 ? (
+							{ null !== requiredSignatureCost && signatureRequirementAllowed ? (
 								<CheckboxControl
 									label={ translate(
-										'Signature Required (+%(price)s)',
-										{ args: { price: formatCurrency( requiredSignatureCost, 'USD') } }
```
harriswong: I am assuming the formatCurrency is removed since the cost is coming back in the correct currency, and we shouldn't hard code it all to USD?

c-shultz: I moved the price string generation up here: https://github.com/Automattic/woocommerce-services/pull/1786/files#diff-464e08994433b72877e47f9464d6d925R61 which still uses `formatCurrency` with `USD` specified

harriswong: Ah i don't know how i didn't see that, thanks!

aleftick:
>We have Express services - this means that we assume that delivery confirmation is part of the service and can be requested. At the same time, Palau ( its zip code ) is listed as not eligible for Delivery Confirmation. I am looking at https://faq.usps.com/s/article/How-is-Signature-Confirmation-and-Signature-Confirmation-Restricted-Delivery-Used for this info.

@budzanowski Indeed there are a few zip codes that aren't available. Can we hard code to exclude those zip codes? 
I wished EasyPost did that via their API. I will bring it up to them and see if we can influence their roadmap. But don't cross your fingers on anything soon.

c-shultz:
@aleftick , I opened #1804 so we can look into addressing this on a separate issue.

PR #1794, opened 2019-11-07
Add logic for 'Choose credit card' button

### LINKED ISSUE #1792: Unable to purchase label even with CC added to account but none selected.
When Credit cards are added to the user's account but none are selected it's impossible for users to purchase a label. Here how my WooCommerce Services Settings page looked when I first setup the plugin on Jurassic ninja:
Screenshot from 2019-11-06 14-25-03
Notice 2 Credit Cards but none are checked.
When creating a label it prompts to add a new Credit Card:
Screenshot from 2019-11-06 14-25-28
Even after adding a new credit card it just keeps prompting to add one.

The only way to make it work is to go and check one of the credit cards as the default in the settings page.

Since we have made it possible to add credit cards from the order page and add packages users might not know there is a settings page or that they need to go there to fix this problem. 

This Seems to happen whenever I setup a new site. I already have a CC in my account but since we are storing the payment_method_id in a WP option WCS does not know which CC I want to use. We should change the Last Call to Action from "Add Credit Card" to "Choose Credit Card"  when CC exists but no payment_method_id is set. The call to action should redirect to the WCS Settings page so the customer can check a CC. We could avoid this if they have only one credit card and assume they want to use that.

As far as I know WordPress.com has no concept of a preferred billing method so when a customer has multiple CC we have to prompt them on which one to choose.

bor0:
Fixes #1792.

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/purchase-section/choose-credit-card-button.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/purchase-section/choose-credit-card-button.js
@@ -1,0 +1,69 @@
+/**
+ * External dependencies
+ */
+import React, { Component, Fragment } from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import { localize } from 'i18n-calypso';
+import Gridicon from 'gridicons';
+
+/**
+ * Internal dependencies
+ */
+import Button from 'components/button';
+import { fetchSettings } from 'woocommerce/woocommerce-services/state/label-settings/actions';
+
+class ChooseCreditCardButton extends Component {
+
+	onVisibilityChange = () => {
+		if ( ! document.hidden ) {
+			this.refetchSettings();
+		}
+		if ( this.chooseCreditCardWindow && this.chooseCreditCardWindow.closed ) {
+			document.removeEventListener( 'visibilitychange', this.onVisibilityChange );
+		}
+	};
+
+	refetchSettings = () => {
+		this.props.fetchSettings( this.props.siteId );
+	};
+
+	onChooseCard = () => {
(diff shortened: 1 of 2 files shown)


harriswong:
👀

harriswong:
Tested and was able to see Choose Credit Card since i didn't have any credit card setup originally.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/purchase-section/choose-credit-card-button.js
```diff
@@ -0,0 +1,69 @@
+/**
```
harriswong: I noticed this file is very similar to https://github.com/Automattic/woocommerce-services/blob/develop/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/purchase-section/add-credit-card-button.js.  

First thought was maybe we can create a single component instead of 2 components, and then pass label/description/url/functions as props instead of duplicating the same functionalities. I think same component also make sense because they are sharing quite a bit of similar functionality and layout (button design).

Thinking about trade offs, the benefit is that it helps future maintenance since it reduces duplication. The con is that these two buttons do have differences so the parent component would need to handle callback functions, translation, URL and the refactoring the "Add credit card" button, seems to be out of scope and we won't meet our timeline. 

If we are creating more of these buttons or if these 2 components are always going to show up together, then maybe it is a good idea to make 1 component instead of 2. I don't think either of these is the case for us here, so I think we are okay with 2 separate components even though they share some of the same codes.

bor0: Thanks @harriswong. I created https://github.com/Automattic/woocommerce-services/issues/1799 for this.

PR #1800, opened 2019-11-08
Refactor credit card buttons into a single component

### LINKED ISSUE #1799: DRY for Add/Choose credit card button
See https://github.com/Automattic/woocommerce-services/pull/1794#discussion_r343941463.

bor0:
Fixes #1799.


harriswong:
👀

harriswong:
I couldn't uncheck this checkbox so I couldn't get the "Choose credit card" to show up with the condition.
Since the condition around the buttons didn't change, I tested the 2 buttons by moving the code block up above the condition, verified both redirected to the correct places.

Side note, it would still be good to know if there is a way I can have it not checked by default. I tried removing my cards but it still show.

harriswong:
@bor0 i thought we are gonna put it in the backlog, didn't expect to see this changed already!
👍

PR #1802, opened 2019-11-08
Show full price breakdown to make pricing less confusing.

### LINKED ISSUE #1801: You save value no longer clear when signature option is selected. Price summary is confusing.
image

The You save value does not reflect the additional cost of signature request.

budzanowski:
Improve how prices are presented. 
Fixes #1801

4 cases:

1. No saving on price, no signature:
image

2. No saving on price, with signature:
image

3. With price saving, no signature:
image

4. With price saving, with signature:
image

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
@@ -153,6 +153,7 @@ export const getTotalPriceBreakdown = ( state, orderId, siteId = getSelectedSite
 			const price = {
 				title: foundRate.title,
 				retailRate: foundRate.retail_rate,
+				rate: foundRate.rate,
 				addons: [],
 			}
 			if ( signatureRequired && null !== foundRateSignatureRequired ) {
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/price-summary.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/price-summary.js
@@ -86,14 +86,15 @@ class PriceSummary extends Component {
 		}
 
 		const { prices, discount, total } = priceBreakdown;
-
+		
 		return (
 			<div className="label-purchase-modal__shipping-summary-section">
 				<hr />
 				{ prices.map( ( service, index ) => {
 					return (
 						<Fragment key={ index }>
-							{ this.renderRow( service.title, service.retailRate, index ) }
+							{ ( service.retailRate !== service.rate ) ? this.renderRow( service.title, service.retailRate, index ) : null }
+							{ this.renderRow( translate( "Your price"), service.rate, 'rate-' + index ) }
 							{ service.addons.map( ( addon, addonIndex ) =>
 								<div key={ 'addons-' + index } className="label-purchase-modal__price-item-addons">
 									{ this.renderRow( addon.title, addon.rate, 'addon-' + addonIndex ) }


budzanowski:
Verbiage needed.

aleftick:
Couple comments:
- We don't need `your price`.
- Currently, we use `USPS - Priority Mail` (changes base on the service). The design mock-up has `Package 1`, `Package 2`. Any reason we're using that verbiage? `USPS - Priority Mail` still works but `Package 1` is more descriptive.
- We don't need to itemize the cost of the signature option. We just need to adjust the cost (by +$3.05) and saving (by +$0.45). At retail, it would cost them $3.05 but with WCShip it is $2.60.

For example, for the 4 cases:

1. No saving on price, no signature:
Package 1  --> $AA
Total  --> $AA

2. No saving on price, with signature:
Package 1  --> $BB (where the signature option cost is included: $AA+$3.05) 
You save --> $0.45
Total  --> $CC (where CC=BB-0.45)

3. With price saving, no signature:
Package 1  --> $DD (price without saving)
You save --> $EE (where EE=label saving)
Total  --> $FF

4. With price saving, with signature:
Package 1  --> $GG (where the signature option cost is included; +$3.05) 
You save --> $HH (where HH=label saving+0.45)
Total  --> $II

budzanowski:
This requires hardcoding the difference between retails and API signature price - that is not perfect but OK and should be easy. If we hardcode that we could as well hardcode the whole signature price - but that means that the work that @c-shultz did for the connect server and plugin signature calculation is no longer needed :/ It will get more complicated when we will add options like Insurance and ADULT_SIGNATURE, we are recreating USPS logic and backtracking values form the rate $total. At this point, we also have a logical mismatch of how we think about signature rates. In our calculated discount, we don't observe the $0.45 element which means that the discount we show is not the discount from the USPS perspective. 

The price breakdown - especially with the signature option enabled together with price and signature selection screen will have a lot of different values - and this in my option is the confusing bit. People will not know what is what and how we ended up in the $total. It will be correct though - just complicated to understand. Being verbose is something that merchants like from my observation.

I don't know how we have ended up with the `USPS - Priority Mail` vs the `Package 1`. Should we change that back?

@c-shultz @aleftick Thoughts?

 I don't mind how we are doing this I just pointed out what feels wired to me :)

aleftick:
>This requires hardcoding the difference between retails and API signature price - that is not perfect but OK and should be easy. If we hardcode that we could as well hardcode the whole signature price - but that means that the work that @c-shultz did for the connect server and plugin signature calculation is no longer needed :/ It will get more complicated when we will add options like Insurance and ADULT_SIGNATURE, we are recreating USPS logic and backtracking values form the rate $total. At this point, we also have a logical mismatch of how we think about signature rates. In our calculated discount, we don't observe the $0.45 element which means that the discount we show is not the discount from the USPS perspective.
Yeah, I forgot to consider the ADULT_SIGNATURE option; it has its own cost and saving amounts. Including the savings from the signature options would be a nice-to-have but not critical. 
The discount is not based on our preferred rates but because USPS offers a discount when you purchase them electronically (vs at the post office). That is why it isn't reflected in the API.

> The price breakdown - especially with the signature option enabled together with price and signature selection screen will have a lot of different values - and this in my option is the confusing bit. People will not know what is what and how we ended up in the $total. It will be correct though - just complicated to understand. Being verbose is something that merchants like from my observation.

There are benefits of adding details; merchants can really understand the pricing breakdown. Thre are also cons. Too many details can be confusing. Imagine also having 2,3... packages and all the details for each.
WCShip is geared to be a simple solution. We need to make it clear to merchants what the cost is and remind them that they save using WCShip.
Prior we didn't make it clear the cost of signature options. Now we will as they check the option box; they can see the `+$2.60`, label cost go up and the total increase. 
I looked at how we currently do it, Ola's designs and Shopify. They all keep it simple and present it in a similar fashion.

budzanowski:
Alright, let's roll!

budzanowski:
@c-shultz what is the difference between retail and API adult-signature prices. Is there any?

c-shultz:
@budzanowski, the adult signature required rate is $6.40 more than the standard rate. I looked through the USPS documentation it looks like that's always $6.40 whether or not its retail or commercial.

c-shultz:
>In our calculated discount, we don't observe the $0.45 element which means that the discount we show is not the discount from the USPS perspective.

I personally don't think we should rush into adding in the $0.45 offset at this stage. This would be new functionality as far as I can tell. I believe the plugin has always just shown the retail rate vs the rate they're being charged as returned by EasyPost. Maybe we should open a new issue to look into this on a future release?

Should we start with removing the itemized `Signature Required` line item and just show the retail rate, what they save, and then total?

aleftick:
>@budzanowski, the adult signature required rate is $6.40 more than the standard rate. I looked through the USPS documentation it looks like that's always $6.40 whether or not its retail or commercial.
That is correct. I documented what I researched on the signature options prices and services it works with [here](https://github.com/Automattic/woocommerce-services/issues/1771#issuecomment-549642519). 
Note that `Adult Signature Required` is not free for PME.

harriswong:
👀

c-shultz:
I opened #1805 to address the missing $0.45

harriswong:
@budzanowski conflict

budzanowski:
Closing this as we will re-think/re-design the whole thing.

PR #1803, opened 2019-11-08
Add adult signature

### LINKED ISSUE #1787: 'Adult signature required' is no longer available
From @aleftick in https://github.com/Automattic/woocommerce-services/issues/1771#issuecomment-549642519 :
>In our current version, we offer the Adult Signature Required option however I don't see it in the new mockups. Hence we shouldn't stop offering it. Adult Signature Required (+$6.40) should be added under Signature Required (+$2.60). Adult Signature Required is an option for all USPS domestic services but not for First-Class Mail (however it's available for First-Class Package Service) and for Media Mail. The cost is $6.40.

c-shultz:
Fixes #1787 

Adds adult signature required to request and displays signature options as dropdowns:

image


c-shultz:
Design feedback is welcome 
/cc @laurendavissmith @aleftick

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/shipping-rate.js
```diff
@@ -47,15 +47,30 @@ class ShippingRate extends Component {
...
 		let details = 'Includes tracking';
+
+		const defaultOption = {
+			label: translate( 'No signature required' ),
+			value: false,
+		};
+		const signatureOptions = concat( defaultOption, values(
+			mapValues( signatureRates, ( r, key ) => {
+				const priceString = ( 0 === r.optionNetCost ) ? translate( 'free' ) :
+					translate( '+%s', { args: [ formatCurrency( r.optionNetCost, 'USD') ] } );
```
bor0: Hardcoding this to USD is acceptable for now, but I prefer adding a TODO here, we might want to support more stuff after multi-carriers

harriswong:
@c-shultz I got `TypeError: Cannot read property "rates" of undefined`
image

To reproduce
1. Open shipping label
2. Pick USPS Media mail, choose signature
3. Choose SIgnature/adult signature for next 2 options
4. Go to express mail, pick "Adult signature"
5. Switch back to "No signature required", got error

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/shipping-rate.js
c-shultz: I think this is fine for now since there's a bunch of places in the code with 'USD' hardcoded. These should be easy to track down and update if these become configurable later.

c-shultz:
Very good catch, @harriswong! I was able to reproduce that. Can you give it another look/test?

harriswong:
Yep! 👀

c-shultz:
@harriswong ,

My original intent was that `false` (boolean) was the default value of that property. The issue you found was that the value was getting set to `"false"` (string) by the dropdown, so the `false !== signatureRequired` check was not acting the way I expected. The last commit removes the explicit `false` (boolean) check and just checks whether or not the `signatureRequired` value is one of the expected rate types. Since we're not explicitly checking for `false` anymore than I think continue to initialize the value the same way should be good.

PR #1809, opened 2019-11-11
Adds .pot generation to tasks/i81n.js

### LINKED ISSUE #1785: Task to create .po/.mo files
`npm run release` was failing after fa39cf38e50d583b247534d5157ad283ab568dcb.

Not a big deal - I reverted it in 9b4c50c960cbbd85020a901fcafe98e7ceaf540b. However, we need to create tasks that will generate .po/.mo files. We already have something like this in `woocommerce-bookings` that can be re-used.

Ferdev:
Generates the .pot file using the already linked calypso-i18n cli libraries.

Closes #1785.

### DIFF
--- a/tasks/i18n.js
+++ b/tasks/i18n.js
@@ -9,3 +9,11 @@ i18nCli( {
 	textdomain: 'woocommerce-services',
 } );
 
+i18nCli( {
+	inputPaths: [ 'dist/woocommerce-services.js' ],
+	output: 'i18n/languages/woocommerce-services.pot',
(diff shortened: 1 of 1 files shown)


harriswong:
👀

harriswong:
Double checked against https://github.com/Automattic/i18n-calypso/tree/master/cli#from-your-code for format.
Ran `npm run i18n` and confirmed locale files modified
```
	modified:   i18n/languages/woocommerce-services.pot
	modified:   i18n/strings.php
```

Issue #1811, opened 2019-11-12
Can't create label again if refund 1 of 2 packages


harriswong:
If we create 2 labels for 2 separate packages, and refund 1 of them, the other is okay. The UI will now stay in "Track Packages" and there is no way to create the label again. The work around is to refund all of it and re-create all.

### To reproduce:
1. Add an order with multiple items
2. Create shipping label for the order
![image](https://user-images.githubusercontent.com/572862/68719179-97382c00-0568-11ea-9069-7dd7d72441d8.png)
3. Separate order to 2 packages, each with different label
![image](https://user-images.githubusercontent.com/572862/68719136-7c65b780-0568-11ea-9dc9-2723723b5346.png)
![image](https://user-images.githubusercontent.com/572862/68718825-a7034080-0567-11ea-9342-dd1503ae7684.png)
4. refund one
![image](https://user-images.githubusercontent.com/572862/68718894-d619b200-0567-11ea-9082-b41696ea902a.png)

5. Now we can't re-create the label since it's always in "Tracking"
![image](https://user-images.githubusercontent.com/572862/68718917-e6ca2800-0567-11ea-806d-355ac6020561.png)

Some discussions here: https://a8c.slack.com/archives/CAKN301C3/p1573598601324700

laurendavissmith:
@harriswong can you add the screenshots to make it easier to see here as well

harriswong:
Resolved by https://github.com/Automattic/woocommerce-services/pull/1813

PR #1813, opened 2019-11-13
Add buttons visibility logic for partial refund case

bor0:


harriswong:
Closes https://github.com/Automattic/woocommerce-services/issues/1811

PR #1821, opened 2019-11-14
Fix/issue1812

### LINKED ISSUE #1812: Some combinations of items weights are producing terrible formatted values.
image

budzanowski:
Fixes #1812. 

This adds the precision formating to `sprintf` strings inside `translate`.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** WooCommerce Services Changelog ***
 
 20XX-XX-XX - version 1.22.1
-* Fix   - Packages weigth total value formatting. 
+* Fix   - Packages weight total value formatting. 
 
 2019-11-14 - version 1.22.0
 * Add   - Display notices when base country/currency mismatches.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Services Changelog ***
 
+20XX-XX-XX - version 1.22.1
+* Fix   - Packages weigth total value formatting. 
+
 2019-11-14 - version 1.22.0
 * Add   - Display notices when base country/currency mismatches.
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Services Changelog ***
 
+20XX-XX-XX - version 1.22.1
+* Fix   - Packages weigth total value formatting.
```
c-shultz: Typo
```suggestion
* Fix   - Packages weight total value formatting. 
```

budzanowski:
Good idea, I will adjust the format based on the unit. For grams, it looks funny :) there should be no decimal points there. This also means that we should adjust the package + items weight input field bc too much precision there makes no sense.

I suggest the following resolution:
`g`     - 0 decimal places
`kg`   - 3 decimals paces ( this is the precision of USPS scales )
`oz`   - 1 decimal place ( this is the precision of USPS scale when set to oz )
`lbs`  - 3 decimals based on oz precision

You mention that 3 is too much. I agree that it does not look too great but I have decided on 3 when I found that EasyPost expects weights in `oz` using 0.1 resolution which implies single gram resolution for `kg` ( 3 decimals ).

Will you sign off this PR if I will use suggested precision and make it unit dependant?

c-shultz:
@budzanowski , now that you explain it, I see why we should have 3 digits for when it's set to kg and pounds. If it looks simple to modify the number of digits based on the unit, that's a good idea, but this PR is a definite improvement as it is, so I'm good with merging.

budzanowski:
@c-shultz display is now unit dependant. I did not adjust the input precision bc it looks like this is just not working in React - but that part is irrelevant. The display counts. Please review again.

PR #1825, opened 2019-11-15
Preselect rate when there is only one rate available.

### LINKED ISSUE #1823: Pre-select rate if there is only one option available
If there is only one rate available for a particular package we should go ahead and select that for the user to save a click.

budzanowski:
Fixes #1823.

An additional change was made around rates selection step expand/collapse logic. If there was only one rate and it was preselected the rates step was collapsing and thus hiding the signature select dropdown. I think ( this is a design decision ) that keeping the step always open makes sense as the merchant knows more clearly what was automatically selected.

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/index.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/index.js
@@ -150,7 +150,6 @@ const RatesStep = props => {
 		values,
 		available,
 		errors,
-		expanded,
 		ratesTotal,
 		translate,
 	} = props;
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/index.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/index.js
@@ -159,6 +159,20 @@ const RatesStep = props => {
 	const updateRateHandler = ( packageId, serviceId, signatureRequired ) =>
 		props.updateRate( orderId, siteId, packageId, serviceId, signatureRequired );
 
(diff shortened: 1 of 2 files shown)


client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/index.js
```diff
@@ -159,11 +159,25 @@ const RatesStep = props => {
 	const updateRateHandler = ( packageId, serviceId, signatureRequired ) =>
 		props.updateRate( orderId, siteId, packageId, serviceId, signatureRequired );
 
+	// Preselect rates for packages that have only one rate available.
+	forEach( form.packages.selected, ( selectedRate, pckgId ) => {
+		// Skip preselection for already selected values.
+		if( values[ pckgId ] !== "" ) {
```
c-shultz: I think we're generally using 'yoda' conditional check orders in this codebase 🙄, so the order should be switched

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/index.js
```diff
@@ -159,11 +159,25 @@ const RatesStep = props => {
...
+			return;
+		}
+
+		if ( ( ! isEmpty( available ) ) && ( pckgId in available ) && ( available[ pckgId ].default.rates.length === 1 ) ) {
```
c-shultz: Same comment about conditional order for length check

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/index.js
budzanowski: image

budzanowski:
Fixed.

PR #1829, opened 2019-11-18
Fix backwards compatibility with Calypso.

c-shultz:
Fixes https://github.com/Automattic/wp-calypso/issues/37638

For backwards compatibility with Calypso, disables the requesting of extra rates if `_via_calypso` parameter is present. The new response format introduced in `1.22.0` to include multiple rate types can not be parsed by Calypso frontend.

To test:
- Select package options and get rates in wp-admin
- Select package options and get rates in Store on WP.com Calypso interface

### DIFF
--- a/classes/class-wc-rest-connect-shipping-rates-controller.php
+++ b/classes/class-wc-rest-connect-shipping-rates-controller.php
@@ -38,9 +38,11 @@ class WC_REST_Connect_Shipping_Rates_Controller extends WC_REST_Connect_Base_Con
 	 * @return array|WP_Error
 	 */
 	public function post( $request ) {
-		$payload = $request->get_json_params();
+		$payload                        = $request->get_json_params();
 		$payload[ 'payment_method_id' ] = $this->settings_store->get_selected_payment_method_id();
-		$order_id = $request[ 'order_id' ];
+		$order_id                       = $request[ 'order_id' ];
+		// Don't add extra rates if request is coming from Calypso ( Store on WP.com ).
+		$add_extra_rates_to_request     = ! isset( $request['_via_calypso'] );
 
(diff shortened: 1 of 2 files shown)


classes/class-wc-rest-connect-shipping-rates-controller.php
```diff
@@ -38,9 +38,11 @@ private function has_customs_data( $package ) {
 	 * @return array|WP_Error
 	 */
 	public function post( $request ) {
-		$payload = $request->get_json_params();
+		$payload                        = $request->get_json_params();
 		$payload[ 'payment_method_id' ] = $this->settings_store->get_selected_payment_method_id();
-		$order_id = $request[ 'order_id' ];
+		$order_id                       = $request[ 'order_id' ];
+		// Don't add extra rates if request is coming from Calypso ( Store on WP.com ).
+		$add_extra_rates_to_request     = ! isset( $request['_via_calypso'] );
```
harriswong: Can the request coming in specify what version it wants? something like `$request['wcs_version']` or something? 
Then we can check for version like
`if (version_compare(CURRENT_WCS_VERSION, `1.22.0`, '>=') === true)`

Reason why I am asking is cause I think `_via_calypso` will always stay `true` from them, so now the condition becomes "if it is from wp.com calypso store, use 1.21 response"; versus "if request coming in is 1.21, we response in 1.21". 

[updated]
Just checked with @c-shultz , there is no `wcs_version` we can use :(

harriswong: That said, I think this will fix the existing problem though.

c-shultz:
Closing this since we decided to fix this in Calypso instead: https://github.com/Automattic/wp-calypso/pull/37698

PR #1844, opened 2019-11-23
Add redo fulfillment button.

### LINKED ISSUE #1843: Re-purchasing a shipping label is no longer available
As confirmed in Slack convo p1574089383152300-slack-C0D9EC4H0, re-purchasing shipping labels is no longer available.

We have a customer who reported that they need to re-purchase their labels all the time in case the originally shipped order gets lost, damaged or dies (they sell bees).

Creating this report to make sure other complaints are logged too.

Reported in:
2510063-zen
2513759-zen
2516317-zen
2491642-zen
15197880-hc
12969458-hc

budzanowski:
image
Fixes #1843

### DIFF
--- a/client/apps/shipping-label/view-wrapper-label.js
+++ b/client/apps/shipping-label/view-wrapper-label.js
@@ -125,11 +125,21 @@ class ShippingLabelViewWrapper extends Component {
 
 			// All items are packaged, just show track button
 			return (
-				<Button
-					onClick={ this.handleTrackPackagesButtonClick }
-				>
-					{ translate( 'Track Packages' ) }
-				</Button>
+				<span> 
+					<span className="shipping-label__redo-shipping-button">
+						<Button
+							borderless
+							onClick={ this.handleCreateLabelButtonClick }
+						>
+							{ translate( 'Redo Shipping' ) }
+						</Button>
+					</span>
+					<Button
+						onClick={ this.handleTrackPackagesButtonClick }
+					>
+						{ translate( 'Track Packages' ) }
+					</Button>
+				</span>
 			);
 		}
 
--- a/client/apps/shipping-label/view-wrapper-label.js
+++ b/client/apps/shipping-label/view-wrapper-label.js
@@ -123,7 +123,7 @@ class ShippingLabelViewWrapper extends Component {
 				);
 			}
 
-			// All items are packaged, just show track button
+			// All items are packaged, show track button and create shipping label button to allow redo fulfillment
 			return (
 				<span> 
 					<span className="shipping-label__redo-shipping-button">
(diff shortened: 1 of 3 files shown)


laurendavissmith:
@bor0 I've added you as a reviewer to start testing this change, we need to make sure that button plays nicely in all scenarios so be sure to check things like partial fulfillment/refund, new order, existing order, etc..

@aleftick will provide verbiage to what the text for that button should be, we'll want to change it from the current `Redo Shipping`

laurendavissmith:
@bor0 also do not merge it until we have a chance to discuss it next week

aleftick:
>@aleftick will provide verbiage to what the text for that button should be, we'll want to change it from the current Redo Shipping

Let's replace `Redo Shipping` with `Create new label` so the language is closer to our initial button `Creat shipping label`.

client/apps/shipping-label/view-wrapper-label.js
```diff
@@ -125,11 +125,21 @@ class ShippingLabelViewWrapper extends Component {
 
 			// All items are packaged, just show track button
```
bor0: Please update this comment, it is no longer true after this code change

client/apps/shipping-label/view-wrapper-label.js
```diff
@@ -125,11 +125,21 @@ class ShippingLabelViewWrapper extends Component {
...
 			return (
-				<Button
-					onClick={ this.handleTrackPackagesButtonClick }
-				>
-					{ translate( 'Track Packages' ) }
-				</Button>
+				<span>
```
bor0: @budzanowski given this logic now, I think we can get rid of L105-L124. Do you agree?

client/apps/shipping-label/view-wrapper-label.js
```diff
@@ -125,11 +125,21 @@ class ShippingLabelViewWrapper extends Component {
...
-				>
-					{ translate( 'Track Packages' ) }
-				</Button>
+				<span> 
+					<span className="shipping-label__redo-shipping-button">
+						<Button
+							borderless
+							onClick={ this.handleCreateLabelButtonClick }
+						>
+							{ translate( 'Redo Shipping' ) }
```
bor0: We need to change this label per @aleftick suggestion

```suggestion
							{ translate( 'Create new label' ) }
```

client/apps/shipping-label/view-wrapper-label.js
budzanowski: Not really. Although the functionality is virtually the same, the labels are a tiny bit different but the styling is totally different. This is done to emphasize functionality importance in different scenarios. We could add logic around scenarios to present different styling but that would be equally awkward imo. I don't want to sacrifice code readability for a couple of lines of code. I don't have a very strong opinion on this, it just feels right this way for me.

budzanowski:
@bor0 Second pass? Thanks for the direct suggestions :) I just clicked accept and magic happened.

bor0:
Skipping merge as @laurendavissmith wanted to discuss it first.

PR #1845, opened 2019-11-26
Fix metabox when it's in the sidebar.

### LINKED ISSUE #1830: Shipping Label metabox remains in the sidebar on the order edit screen
In the update to version 1.22 the Shipping Label metabox was moved from the sidebar to the main content area. However in WordPress it's possible to drag and drop metaboxes. If we do this then the order is saved in the user meta data. For example for the order edit screen it's saved in the user meta with the key `meta-box-order_shop_order`. Besides defining the order it also saves whether a metabox is displayed in the sidebar or not. If this is saved it will override the default, and can cause it to remain in the sidebar.

Steps to reproduce:

- Install WooCommerce Services 1.21
- Edit an order and drag and drop the Shipping Label meta box to a different location (ex. after the order notes)
- Upgrade to WooCommerce Services 1.22
- Notice that the Shipping Label remains in the sidebar (which makes the text go outside it's box)

image

Because of the newly added CSS rule to hide the title we are unable to drag and drop it back to the main content location. One workaround for this is to temporarily unhide the metabox `<h2>` title so we can drag it back to the right location.

If there is a different value saved in the user meta `meta-box-order_shop_order`, should we override this to force it to end up in a different location?

superdav42:
Also make style better when located on the sidebar.
Closes #1830

Shipping label box can now be dragged back to the sidebar if so desired. Display in sidebar should be same as on small screen.

### DIFF
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -273,15 +273,9 @@
 	white-space: pre-wrap;
 }
 
-#woocommerce-order-label {
-	.hndle,
-	.handlediv {
-		display: none;
-	}
-
-	.inside {
-		display: block !important;
-	}
+// When metabox in sidebar display same as small screen.
+.wp-core-ui.wp-admin #side-sortables .wcc-root .shipping-label__container {
+	display: block;
 }
 
 .wp-core-ui.wp-admin #woocommerce-order-shipment-tracking {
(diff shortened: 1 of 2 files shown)


superdav42:
> @superdav42 this works fine and appears correctly on mobile. However, when I move `Shipping Label` to the right sidebar (desktop view) the CSS becomes weird. You can either fix that as part of this PR, or If you want, you can merge this but please create an issue for that.
> 
> Screen Shot 2019-11-26 at 13 53 06

It was super easy so I fixed it.

Issue #1831, opened 2019-11-19
Spike: Utilizing Calypso as a build tool


laurendavissmith:
As part of our research into the architecture of WCS its been brought up that we should possibly leverage Calypso as a build tool. This spike needs to research this in-depth, talk to some of the teams that have done this already and determine what that would entail for this extension.

More info here: https://extendomattic.wordpress.com/2019/10/05/woocommerce-services-architecture/#comment-4001

Deliverable: p2, dev roundtable discussion

c-shultz:
I'm going to leave some notes on this before getting things more fully compiled in a P2 post:

Current situation: WCS currently extends the `wp-calypso` submodule build tooling. The dependencies are messy and the whole thing is hard to maintain, so I think it's clear that we want to move away from this.

Possible directions to go:
1. Copy build configuration from`wp-calypso` repo to WCS repo.
    - Pros:
        - This is probably the least disruptive choice in the short term since we still have a lot of Calypso component dependencies.
    - Cons:
        - We'll need to maintain this over time so we'll fall behind the latest and greatest build features.
        - Maintaining a separate set of build configurations is a waste of resources for Automattic

1. Use `calypso-build` (exploratory PR: #1846)
    - Pros:
        - Share build configuration and development resources with other teams
        - Stay more up-to-date as `calypso-build` is improved
        - Appears to include most necessary features (including the ability to import styles from .js files)
    - Cons: 
        - It's going to take some work to get it implemented

1. Use [wp-scripts](https://developer.wordpress.org/block-editor/packages/packages-scripts/) 
    - Pros:
        - Many of the same pros as using `calypso-build`
    - Cons: 
        - More limited functionality (no style imports from .js)
        - Potentially slower moving development because it's part of WordPress

c-shultz:
P2 Post: https://wcship.wordpress.com/2019/12/02/building-woocommerce-services/

PR #1846, opened 2019-11-27
Explore building with `calypso-build`

c-shultz:
This is an work-in-progress exploratory PR to try out building WCS with [calypso-build](`https://www.npmjs.com/package/@automattic/calypso-build`) as part of spike #1831

The Calypso team has been working to modularize and export the the Calypso build tooling as an npm package. `calypso-build` essentially exports a generic Webpack and Babel configuration which can be utilized for different projects, so they don't have to start from scratch to get their build tooling set up.

In this PR, the build step is now defined as: `"build": "calypso-build --config='./webpack.config.js'",`

This is currently incomplete and does not fully build WCS. The JS resources are building so I believe we're 98% there for basic functionality. There's CSS build error that needs to be resolved, but I wanted to timebox my efforts here before we make a decision about what direction to go.
Screen Shot 2019-11-27 at 3 37 07 PM


samnajian:
Too old to live!

PR #1847, opened 2019-12-02
Spike/issue 1840

### LINKED ISSUE #1840: Spike: Automating test data creation for E2E tests
Currently the e2e suite in branch `new/e2e-test-infrastructure` has a `beforeAll` step where it creates the test data the tests need by going through the WooCommerce UI. This process is very slow and prone to errors. Test data should be created and destroyed automatically in each test run, and completely independent from the test cases.

Deliverable: p2 on wcship, roundtable discussion, actionable GH issues

Ferdev:
Adds `sequelize-fixtures` to the e2e setup so products and orders can be created directly in the database through a mySQL connection.

Removes also all unneeded test code in the create shipping label test.

To test it:

Add a few environment variables to setup your mySQL connection:

```
export WP_MYSQL_E2E_USERNAME=myusername
export WP_MYSQL_E2E_PASSWORD=mypassword
export WP_MYSQL_E2E_HOST=127.0.0.1
```
Update your npm packages: `npm install`

Then to execute the e2e tests: `npm run test:e2e`

Closes #1840


budzanowski:
The MYSQL username and password concepts are already present inside the testing environment. When we execute the install script for Travis we pass user, password and the location ( localhost ) for the db. I suggest making this two things the same thing - with the same user and password:

https://github.com/Automattic/woocommerce-services/blob/spike/issue-1840/.travis.yml#L78

otherwise, it will become quite confusing.

budzanowski:
What about creating the product for example using wp cli:
```wp wc product create -–name="Test Product" --type=simple --sku=WCCLITESTP --regular_price=20 –user
```

This way we are relying on wp cli and wp internal functions to populate the database correctly. We don't need to update the schema when something changes.

tests/e2e-tests/specs/admin/create-shipping-label.test.js
```diff
@@ -18,150 +18,72 @@ describe( 'Create shipping label', () => {
...
-		await CustomerFlow.productIsInCheckout( 'Simple product', 1, 9.99 );
-		console.log( 'Product is in the order review page!' );
-		await expect( page ).toMatchElement( '.cart-subtotal .amount', { text: '$9.99' } );
-	} );
+	it( 'should create a new shipping label', async () => {
+		await db.loadFixtures( async ( models ) => {
+			await StoreOwnerFlow.login();
+			await StoreOwnerFlow.openExistingOrderPage( models.wp_posts[1].ID );
 
-	it( 'should allow customer to place order', async () => {
```
budzanowski: I would leave this as a not connected test step for future use. We might end up having product page customizations when WCS is enabled.

budzanowski:
It looks very good in general. I have not tested it though. I will start running this tomorrow though when I will finish my Janitorial rotation.

Ferdev:
> What about creating the product for example using wp cli:
> `wp wc product create -–name="Test Product" --type=simple --sku=WCCLITESTP --regular_price=20 –user `
> 
> This way we are relying on wp cli and wp internal functions to populate the database correctly. We don't need to update the schema when something changes.

That's a really good idea. We'll need to figure out though how and where to introduce it as a setup step in the test execution, as it will be an external command. We can either run it as a scaffolding task that's executed separately from the test execution, or within the test execution itself.

Anyway, I think that's probably the way to go :+1:

PR #1854, opened 2019-12-10
Fix/issue 1832

### LINKED ISSUE #1832: Add option to change Paper Size back to the label print screen
In version 1.22.0, the "redundant" Paper Size dropdown was removed from the label screen since it's configurable from the Settings screen.  This was not redundant.  When shipping domestically I print 4"x6" labels on a thermal printer, when shipping internationally I print on letter size adhesive paper.  It is very beneficial to be able to select the paper size (thus label type) when creating a label.  Returning to the settings page each time I want to print a label is very inconvenient.

roykho:
Fixes #1832 

* Add   - Paper size selection into purchase modal sidebar and reprint modal which was previously removed.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 1.22.2 - 2019-XX-XX =
 * Add   - Preselect rate when there is only one rate available for given shipping configuration.
+* Add   - Paper size selection into purchase modal sidebar and reprint modal which was previously removed.
 * Fix   - Packages weight total value formatting.
 * Fix   - Allow fulfillment flow redo.
 * Fix   - Use correct URL for variation products in Packages section.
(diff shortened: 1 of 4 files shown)


roykho:
I would say we could remove it because the selection is remembered for next time anyways so there is no need to have another setting even if it is just to set default.

cc @aleftick @laurendavissmith thoughts?

laurendavissmith:
lets leave it for now since we'll be taking a closer look at some UX updates next year

aleftick:
We don’t want to remove this paper size option. 

We removed it from this label buying flow in the last released and quickly learned that merchants relied on it. For example some merchants print domestic label on a label printer and international on regular 8x10 printer.
We’ll explore in the future if we can help merchants automate that selection.

roykho:
@aleftick just so we're on the same page, we were referring to the redundant one that is sitting in the "WCS->Settings". This PR is to add back in what we took out in last release.

PR #1859, opened 2019-12-19
Fix last box id for individually packed items.

budzanowski:
Don't save and don't provide "individual" as last box id.

I think that this actually fixes:
Fixes: https://github.com/Automattic/woocommerce-services/issues/1864

This solves this screen:
image

It happens when user has sent a package as individually packaged. Later moved to another order. that has different items.

### DIFF
--- a/classes/class-wc-rest-connect-shipping-label-controller.php
+++ b/classes/class-wc-rest-connect-shipping-label-controller.php
@@ -54,7 +54,7 @@ class WC_REST_Connect_Shipping_Label_Controller extends WC_REST_Connect_Base_Con
 			}
 		}
 
-		if ( ! empty( $last_box_id ) ) {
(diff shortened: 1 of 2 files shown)


classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -54,7 +54,7 @@ public function post( $request ) {
 			}
 		}
 
-		if ( ! empty( $last_box_id ) ) {
+		if ( ! empty( $last_box_id ) && $last_box_id !== "individual" ) {
```
bor0: I think we need to check for numeric here, to avoid bugs like this in the future.

bor0:
This PR seems like the best way forward for now. See discussion https://a8c.slack.com/archives/CAKN301C3/p1576832487012100

PR #1870, opened 2020-01-16
Add api heartbeat test.

budzanowski:
Implement a heartbeat test to offload the connect server.

### DIFF
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -295,6 +295,30 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 			return $this->request( 'GET', '/connection/test' );
 		}
 
+		/** Heartbet test
+		 * 
+		 * @return true|WP_Error
+		 */
+		public function is_alive() {
+			return $this->request( 'GET', '' );
+		}
+
+		/** Heartbet test with transient cache.
+		 * 
+		 * @return true|WP_Error
+		 */
+		public function is_alive_cached() {
+			$connect_server_is_alive_transient = get_transient( 'connect_server_is_alive_transient' );
+			if ( true === $connect_server_is_alive_transient ) {
+				return true;
+			}
+
+			$is_alive_request = $this->is_alive();
+			$new_is_alive = ! is_wp_error( $is_alive_request );
+			set_transient( 'connect_server_is_alive_transient', $new_is_alive, MINUTE_IN_SECONDS );
+			return $new_is_alive;
+		}
+
 		/**
 		 * Create a deferred Stripe Standard Account
 		 * @param $email string The user's email address
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -309,7 +309,7 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 		 */
 		public function is_alive_cached() {
 			$connect_server_is_alive_transient = get_transient( 'connect_server_is_alive_transient' );
-			if ( true === $connect_server_is_alive_transient ) {
+			if ( false !== $connect_server_is_alive_transient ) {
 				return true;
 			}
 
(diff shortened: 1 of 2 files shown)


woocommerce-services.php
```diff
@@ -1372,12 +1372,12 @@ function add_plugin_action_links( $links ) {
 		}
 
 		function enqueue_wc_connect_script( $root_view, $extra_args = array() ) {
-			$wcs_connection_schemas = $this->api_client->get_service_schemas();
+			$is_alive = $this->api_client->is_alive_cached();
```
superdav42: I think we actually want to use auth_test() here and hit the connection test so `wcs_server_connection` will be true if the server is up and the site is connected to jetpack. We should still make a cached version of `auth_test()` and use it but be sure to invalidate the cache when jetpack is connected or disconnected.

budzanowski: I think that this is already taken care of.
The script is enqueued only if Jetpack is connected or in dev mode. I think that this is enough.

budzanowski: https://github.com/Automattic/woocommerce-services/blob/develop/woocommerce-services.php#L437

superdav42: Me too

classes/class-wc-connect-api-client.php
```diff
@@ -295,6 +295,30 @@ public function auth_test() {
...
+			return $this->request( 'GET', '' );
+		}
+
+		/** Heartbet test with transient cache.
+		 * 
+		 * @return true|WP_Error
+		 */
+		public function is_alive_cached() {
+			$connect_server_is_alive_transient = get_transient( 'connect_server_is_alive_transient' );
+			if ( true === $connect_server_is_alive_transient ) {
```
c-shultz: This check is not working for me because get_transient is returning `"1"` for true. I think a `false !== $connect_server_is_alive_transient` should work because get_transient returns the bool false for expired transients.

budzanowski: Corrected.

classes/class-wc-connect-api-client.php
```diff
@@ -295,6 +295,30 @@ public function auth_test() {
...
+		 */
+		public function is_alive_cached() {
+			$connect_server_is_alive_transient = get_transient( 'connect_server_is_alive_transient' );
+			if ( false !== $connect_server_is_alive_transient ) {
+				return true;
+			}
+
+			$is_alive_request = $this->is_alive();
+			$new_is_alive = ! is_wp_error( $is_alive_request );
+			set_transient( 'connect_server_is_alive_transient', $new_is_alive, MINUTE_IN_SECONDS );
```
c-shultz: Wow, get_transient is annoying with booleans. My suggestion didn't fully fix it.  `true` is resulting in `"1"` and `false` is resulting in `""`. The following makes it work:
```suggestion
			if ( $new_is_alive ) {
				set_transient( 'connect_server_is_alive_transient', true, MINUTE_IN_SECONDS );
			}
```

classes/class-wc-connect-api-client.php
```diff
@@ -295,6 +295,32 @@ public function auth_test() {
 			return $this->request( 'GET', '/connection/test' );
 		}
 
+		/** Heartbet test.
```
laurendavissmith: I know this already got merged but can we fix these misspellings @budzanowski

classes/class-wc-connect-api-client.php
```diff
@@ -295,6 +295,32 @@ public function auth_test() {
...
+		 * 
+		 * @return true|WP_Error
+		 */
+		public function is_alive() {
+			return $this->request( 'GET', '' );
+		}
+
+		/** Heartbet test with transient cache.
```
laurendavissmith: I know this already got merged but can we fix these misspellings @budzanowski

classes/class-wc-connect-api-client.php
budzanowski: @laurendavissmith done.

classes/class-wc-connect-api-client.php
budzanowski: @laurendavissmith done.

Issue #1876, opened 2020-01-21
Cannot select package after removing only item from only package


c-shultz:
Cannot select package after removing only item from only package:

![2020-01-21 07 41 50](https://user-images.githubusercontent.com/6209518/72819176-a4852d00-3c21-11ea-984b-871f142d8cf2.gif)

The only way to resolve this is to refresh the order page.

harriswong:
There seems to be a warning in console when all items are removed.
```
backend.js:6 Warning: unstable_flushDiscreteUpdates: Cannot flush updates when React is already rendering.
    in PackageInfo (created by Localized(PackageInfo))
    in Localized(PackageInfo) (created by Connect(Localized(PackageInfo)))
    in Connect(Localized(PackageInfo)) (created by PackagesStep)
    in div (created by PackagesStep)
    in div (created by FoldableCard)
    in div (created by Card)
    in Card (created by CompactCard)
    in CompactCard (created by FoldableCard)
    in FoldableCard (created by Localized(FoldableCard))
    in Localized(FoldableCard) (created by StepContainer)
    in StepContainer (created by Localized(StepContainer))
    in Localized(StepContainer) (created by PackagesStep)
    in PackagesStep (created by Localized(PackagesStep))
    in Localized(PackagesStep) (created by Connect(Localized(PackagesStep)))
    in Connect(Localized(PackagesStep)) (created by LabelPurchaseModal)
    in div (created by LabelPurchaseModal)
    in div (created by LabelPurchaseModal)
    in div (created by LabelPurchaseModal)
    in div (created by DialogBase)
    in div (created by ModalPortal)
    in div (created by ModalPortal)
    in ModalPortal (created by Modal)
    in Modal (created by DialogBase)
    in DialogBase (created by Dialog)
    in MomentProvider (created by Connect(MomentProvider))
    in Connect(MomentProvider)
    in Provider
```

harriswong:
Currently, if we remove the items from the package, we can't add them back. The best we could do is display something like this:
![image](https://user-images.githubusercontent.com/572862/72945869-66c7f780-3d3a-11ea-87d3-59e388a96dc4.png)

The area under "Packages to be Shipped" will be empty, and "Items to fulfilled" would also be empty. The users won't be able to do anything beside new package. Creating new package simply adds a "Custom package" to the drop down, not creating a packaging in the list. 

I think this modal will be more useful if we can have some ways to display items that are not part of any package, and allow user to add items to new packages.

[...]

PR #1886, opened 2020-01-27
[WIP] Fix/issue 1876

harriswong:
## Work in progress, do not merge

This PR no longer remove the package item. Instead, when "remove from package" is clicked, it moves the item to "Unpackaged Items" so that it can move to another package(s).

Issue: https://github.com/Automattic/woocommerce-services/issues/1876

## TODO
- Currently in this PR, "Unpackaged Items" are still treated as a "package", so it will still have a label. We will have to add something like `if (package === 'unpack_item') then <ignore package>` so that label can ignore this package.


harriswong:
We decided to remove the "Remove from package" option for now. More details here: https://github.com/Automattic/woocommerce-services/issues/1876#issuecomment-580899363

Closing this in favour of https://github.com/Automattic/woocommerce-services/pull/1908.

PR #1890, opened 2020-01-28
decode html entities to avoid &#039; from appearing on the label

### LINKED ISSUE #1862: Unnecessary character escaping in shipping label origin
The apostrophe character (`'`) is being entity-encoded as `&#039;` in the top line ("name") of the shipping label's return address.

I believe this name is originally grabbed from the WP user name [[code](https://github.com/Automattic/woocommerce-services/blob/7be77b3200dba5093c12b2cf69a2d7abfc34bbd5/classes/class-wc-connect-service-settings-store.php#L116
)] but is getting escaped somewhere along the way (plugin, server, or label provider API).

Props @dphyled for raising the issue.

superdav42:
closes #1862

To test: 
add a ' to your Site Title in WP -> Settings -> General.
Delete any saved origin address in the option `woocommerce_connect_origin_address` 
Create a new label.
&#039; should not appear on the label.

### DIFF
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -112,7 +112,7 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 
 		public function get_origin_address() {
 			$wc_address_fields = array();
-			$wc_address_fields['company'] = get_bloginfo( 'name' );
+			$wc_address_fields['company'] = html_entity_decode( get_bloginfo( 'name' ), ENT_QUOTES ); // HTML entities may be saved in the option.
 			$wc_address_fields['name'] = wp_get_current_user()->display_name;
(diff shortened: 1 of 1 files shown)


superdav42:
> This is not working for me. It's still showing the HTML encoding on the label. I think this may be because I previously set my site name with a bunch of characters in it (`asdfjlk'asd#@398"""`) _before_ updating to this branch, and it's using the stored value (which has encoded html characters) on this line without doing any decoding:
> 
> https://github.com/Automattic/woocommerce-services/blob/e23fc1bc826cc56d5d9f8d5258d3f7c1476350d1/classes/class-wc-connect-service-settings-store.php#L139

Yes, You'll need to clear out the option or manually update the company name to clear that out. Although we could just html_entity_decode that field evertime to avoid making having to do that. I think I'll make that change.

PR #1895, opened 2020-01-31
Spike: Automating test data creation for E2E tests

### LINKED ISSUE #1840: Spike: Automating test data creation for E2E tests

### LINKED ISSUE #1896: Automate Test Data Creation using WooCommerce API
Based on @Ferdev's research (here: https://wcship.wordpress.com/2020/01/27/another-take-on-generating-test-data-for-e2e-tests/) we have an excellent path forward to automate test data creation using the WooCommerce API instead of having Puppeteer directly go through these steps.

Ferdev:
Closes #1896 and #1897.

Creates test data products and orders using WooCommerce REST API. It adds the helper function `withOrder` that passes the generate order to the provided callback.

The REST API client needs the two env vars: `WC_E2E_REST_API_CONSUMER_KEY` and `WC_E2E_REST_API_CONSUMER_SECRET`. [Here](http://woocommerce.github.io/woocommerce-rest-api-docs/?javascript#authentication) are detailed instructions to generate those keys in your local environment.

The only thing left is to generate this REST API keys and populate the env vars during the Travis setup phase. There isn't currently a way to do this programmatically (check slack conversation [here](https://a8c.slack.com/archives/C0E1AV8T0/p1579800658019200)), but it could be done using some php code to create a random key directly into the database. Here's the actual code WooCommerce uses to generate the keys: https://github.com/woocommerce/woocommerce/blob/7d39222334862cd058cd64097a51f0ba32200323/includes/class-wc-ajax.php#L1952


c-shultz:
I switched the issue this closes to #1896 instead of #1840.

Ferdev:
With the latest commit, the test data creation now works in Travis even though the rest of the test fails. That should be addressed in #1897

harriswong:
👀

.travis.yml
```diff
@@ -7,57 +7,66 @@ language: php
...
+     dist: trusty
      php: 7.2
      env: WP_VERSION=latest WP_MULTISITE=0 WC_VERSION=latest RUN_E2E=1
+     before_install:
+         # Enable user namespace cloning
+         - "sysctl kernel.unprivileged_userns_clone=1"
+           # Launch XVFB
+         - "export DISPLAY=:99.0"
+         - "sh -e /etc/init.d/xvfb start"
+         - sleep 3
```
harriswong: Instead of `trusty`, if we are using the `Xenial` distribution, we maybe able to skip launching XVFB this way https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-services. 

Could be something we can do if we switch to `Xenial` or `Bionic` in the future.

Ferdev: yeah, I'd love to do that! I didn't change it in this release because there were enough changes and I was afraid of breaking the other tests. But yeah, we should totally switch to a newer distro at some point.

Ferdev:
> Do we need all the `console.log`? I think they are there for debugging purposes when an error is thrown on Travis? If so, i think we should keep them.
Yeah, I'm good with leaving or deleting them. I left them because it's informative about what's going on in the Travis log, but we can remove them in the future if they introduce too much noise.

PR #1913, opened 2020-02-10
Pass new API version number so we can properly version the API

superdav42:
Required for https://github.com/Automattic/woocommerce-connect-server/pull/1348 and should be released much sooner than the server changes so ideally the majority of sites would be using V2 before we update the server.


superdav42:
Closing in favor of #1927

PR #1925, opened 2020-02-15
Delay the initial label status call by 1 seconds

### LINKED ISSUE #1900: Label purchase status polling happens too early.
When we start the label purchase process the WCS plugin asks the connect server to purchase the label. Then the WCS asks the server for the purchase status. The status asking is a process repeated every second until the server tells us that the label was purchased or there was an error. This is a fairly simple process. It has one simple flaw in my opinion. When we inspect the endpoints delay we can see that the majority ( or basically always ) the first time WCS asks about the purchase status it will receive a status that the purchase is ongoing. Billing daddy and EasyPost is not fast enough. In my opinion, we should delay the first status asks API call. We should save this way a lot of unnecessary calls. 

Alternatively, we can consider building a more sophisticated system using webhooks. Connect server would notify WCS when purchase process has ended. Polling the connect server would not be necessary. WCS frontend could ask WCS backend for the status.

harriswong:
Applying the first solution suggested in the issue https://github.com/Automattic/woocommerce-services/issues/1900. 

Since the request hardly completes in 1 second, the polling for the first couple of seconds aren't necessarily. Instead, delay the first call by a factor of 2 and then continue to poll per second. 

**Impact:**
Minimum label purchase time will change from `1` second to `2` seconds. This may delay some purchasing. Based on some data on the server side, the request seems to take between 1 to 2 seconds. 

Closes #1900

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
@@ -757,6 +757,10 @@ const getPDFFileName = ( orderId, isReprint = false ) => {
 
 // retireves the single label status, and retries up to 3 times on timeout
 const labelStatusTask = ( orderId, siteId, labelId, retryCount ) => {
+	let timeout = 1000;
+	if ( retryCount === 0) {
+		timeout = 3000;
+	}
 	return api
 		.get( siteId, api.url.labelStatus( orderId, labelId ) )
 		.then( statusResponse => statusResponse.label )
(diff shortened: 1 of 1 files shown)


harriswong:
For tests, we can use `rewire` to test `labelStatusTask` function and test only this method. Or, we can add tests to the `purchaseLabel` function to test the purchase label flow. 

I think having tests for the `purchaseLabel` flow would add confident since we don't currently have coverage on it. It would be outside the scope of this issue though.

harriswong:
Currently pipeline shows "passed" but there are some errors inside E2E testing. Merge
 https://github.com/Automattic/woocommerce-services/pull/1895 in once it's in master, and re-run  tests.

harriswong:
Looks like E2E tests failed, https://travis-ci.org/Automattic/woocommerce-services/builds/650678496?utm_medium=notification&utm_source=github_status. 

Check why chrome couldn't start: https://travis-ci.org/Automattic/woocommerce-services/jobs/650678507#L989

harriswong:
Merged develop, E2E test passed.

Issue #1909, opened 2020-02-07
Better differentiation of service options


c-shultz:
We have several closely related issues that I'm closing to be addressed altogether on this issue instead:
https://github.com/Automattic/woocommerce-services/issues/1819
https://github.com/Automattic/woocommerce-services/issues/1903
https://github.com/Automattic/woocommerce-services/issues/1885
https://github.com/Automattic/woocommerce-services/issues/1818
https://github.com/Automattic/woocommerce-services/issues/1865

@elizaan36 is working at looking at the overall design here

[This Spike](https://github.com/Automattic/woocommerce-services/issues/1884) was done to determine how we should store/transmit the needed data to address all the above issues.

https://github.com/Automattic/woocommerce-connect-server/issues/1347 is for the Connect Server implementation of this work. If this issue is worked on at the same time, we'll need to coordinate on what the API interface will be.

elizaan36:
@c-shultz @aleftick 
Here's a solution for better displaying the service options, shown for desktop and mobile. Probably not final but I'll post them here so we can discuss.

![Services display 2](https://user-images.githubusercontent.com/5121465/74184556-9ae15a80-4c0c-11ea-9e58-602df3a83b3d.png)

Quick walkthroughs:
**Desktop**
https://share.getcloudapp.com/kpuY98GE
- Top services to compare are shown in the abbreviated list of services. Thinking about the flow for a merchant, the top items to compare are probably whether tracking is included, and how much insurance is covered. We'd like to advertise the free pickup option, so that's listed next to those items if available, with a help icon for further information.
- Upon selection, the signature options are shown and the service area expands down
- When a signature option is selected, the pricing for all services in the list are updated to reflect that. When a merchant needs the signature requirement, they'll want to compare all pricing with that option selected. They will likely not de-select it (unless all of the options are too expensive) because that would most likely be an important requirement for them. 
- Just in case they want to de-select the signature option, it's shown as a checkbox. When no signature options are selected, that means "No signature required" for the package.
- Selecting a new service when a signature option is selected will keep it selected with pricing staying the same.

**Mobile**
https://share.getcloudapp.com/eDuxWXmG
- Mobile view lists out the services in a bulleted list rather than wrapping the line (maybe <400px?)

Also included a solution for better displaying the WCS savings. This would appear as a notice below the totals.

elizaan36:
[...]

PR #1927, opened 2020-02-19
Tweak/better differentiation service options 1909

Ferdev:
Implements the new design defined in #1909.

This is the list of tasks needed to implement the new design:

- [x] Adapt current unit test setup to support components testing.
- [x] Create unit test for the `ShippingRate` component.
- [x] Add test for the existing and the new uses cases.
- [x] Add the benefits list
- [x] Implement action to toggle the signature options section when a new rate is selected.
- [x] Transform the current signature combo box to a list of checkboxes.
- [x] Make the checkboxes behave as radio buttons.
- [x] Update the rate's amount when a signature option is selected.
- [x] Update the shipping label totals when a signature option is selected.


client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/shipping-rate.js
```diff
@@ -19,45 +21,79 @@ class ShippingRate extends Component {
...
+	renderServices( carrier_id, signatureOptions, includedServices ) {
+
+		const servicesToRender = [];
 
-		this.setState( { signatureOption: val } );
-		updateValue( service_id, val );
+		if ( includedServices.tracking ) {
+			switch ( carrier_id ) {
+				case 'usps':
+					// Ideally this would come from connect-server, but we have no info from EasyPost API
```
superdav42: I think we could get rid of this comment now. Tracking will be coming from the connect-server now and we'll only show this on services that support it. I guess we can keep the switch statement as it'll say what kind of tracking it uses. It seems a little redundant to say USPS tracking as USPS offers no other kind. I think we could just say "Includes tracking"  and not use a switch statement.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/shipping-rate.js
```diff
@@ -81,41 +118,31 @@ class ShippingRate extends Component {
...
-				// Refer to: https://www.easypost.com/docs/api/node#trackers, specifically
-				// `A Tracker is created automatically whenever you buy a Shipment through EasyPost`
-				details = translate( 'Includes USPS tracking' );
-				break;
-		}
+		const ratePlusSignatureCost = selectedSignature ? rate + selectedSignature.value : rate;
+
 		return(
-			<div className="rates-step__shipping-rate-container">
+			<div className="rates-step__shipping-rate-container" >
```
superdav42:
```suggestion
			<div className="rates-step__shipping-rate-container">
```

superdav42:
Also the tooltip after "Available for free pickup" is on the design but does not seem to be implemented. Was that intentional?

Ferdev:
This can be reviewed again, thanks @superdav42 !

Ferdev:
To test it with the latest API, grab this WC Connect Server PR https://github.com/Automattic/woocommerce-connect-server/pull/1353 too

classes/class-wc-connect-api-client.php
```diff
@@ -466,7 +466,7 @@ protected function request_headers() {
 			$lang = $locale_elements[ 0 ];
 			$headers[ 'Accept-Language' ] = $locale . ',' . $lang;
 			$headers[ 'Content-Type' ] = 'application/json; charset=utf-8';
-			$headers[ 'Accept' ] = 'application/vnd.woocommerce-connect.v1';
+			$headers[ 'Accept' ] = 'application/vnd.woocommerce-connect.v2';
```
superdav42: The v2 is kinda like a magic number. I think we should put it in a constant. This is how I had it in my PoC. You should be able to rebase from https://github.com/Automattic/woocommerce-services/pull/1913/files then it'll be more obvious how to update to a newer API later.

assets/stylesheets/style.scss
```diff
@@ -238,6 +238,9 @@
 @import 'components/popover/style';
 @import 'components/tooltip/style';
 
+// All WordPress Component styles
+@import '~@wordpress/components/build-style/style.css';
```
superdav42: Not sure why this change is showing here. It's already in develop, I added as part of my tooltip work, did you add it too? Probably you should merge develop and make sure this line is not duplicated.

Ferdev: Yeah, I rebased my branch off yours so the merge will be easier. It shouldn't cause any problems.

Ferdev:
Hey @superdav42 , please, give it another try. The tooltip not being displayed is actually caused by wc-admin. If you disable the plugin the notifications appear again. I've created this issue so we don't forget to fix it https://github.com/Automattic/woocommerce-services/issues/1951

PR #1931, opened 2020-02-25
Jetpack::is_staging_site() is deprectated in Jetpack 8.1

superdav42:
have to use \Automattic\Jetpack\Status::is_staging_site() when it exists

Test on Jetpack 7 and Jetpack 8.1+. No deprecation notices should show.


peterfabian:
FYI, this is throwing notices during the OBW on WC 4.0 pre-release, so we would very much like to have this released as soon as possible:

screenshot_2020-02-26_at_15 24 23

superdav42:
Switched base to merge into master as it will be a hotfix.

PR #1935, opened 2020-02-27
Remove tabs which cause so much problems

superdav42:
Let's see if it works!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,43 +7,43 @@ language: php
 matrix:
   include:
    - php: 5.3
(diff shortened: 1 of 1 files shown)


harriswong:
👍 :shipit:

harriswong:
Checked against https://config.travis-ci.com/explore
Just warning and infos
```
[warn] on jobs.include: deprecated key: "sudo" (The key `sudo` has no effect anymore.)

[warn] on root: deprecated key: "sudo" (The key `sudo` has no effect anymore.)

[info] on root: the key matrix is an alias for jobs, using jobs

[info] on root: missing os, using the default "linux"
```

PR #1936, opened 2020-02-27
Run JS unit tests in travis

### LINKED ISSUE #1921: Add Javascript tests to Travis
We have lots of Javascript tests we could add to our Travis matrix.

You can try them by running:

`npm run test-client`

superdav42:
Let's see if it works

Closes #1921

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,40 +7,40 @@ language: php
 jobs:
   include:
    - php: 5.3
-     env: WP_VERSION=4.5 WC_VERSION=3.2.6
+     env: WP_VERSION=4.5 WC_VERSION=3.2.6 PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1
      dist: precise
    - php: 5.4
      dist: trusty
-     env: WP_VERSION=4.6 WC_VERSION=3.1.2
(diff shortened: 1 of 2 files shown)


c-shultz:
My tests are all failing now 😱 . Do they work for you locally @superdav42? I wonder if there's a version difference causing. I'm running v6.9 of npm. Travis appears to be using 10, but I'm not sure if the npm version makes sense for just making the tests fail.

superdav42:
> My tests are all failing now  . Do they work for you locally @superdav42? I wonder if there's a version difference causing. I'm running v6.9 of npm. Travis appears to be using 10, but I'm not sure if the npm version makes sense for just making the tests fail.

Tests work for me locally. I doubt it's an npm problem. I'm sure it's related to timezones because travis will have a different TZ than us. I tried to fix it in https://github.com/Automattic/woocommerce-services/pull/1936/commits/08005028e248f2137bfa134445f573bc83226b63 but it didn't work. Not sure what else to try.

superdav42:
Had to use moment.tz to set default timezone. All tests passing now.

c-shultz:
Closes #1921

PR #1940, opened 2020-02-29
Expose wcs redux stores 

c-shultz:
Place redux stores in a global called wcsAppStores so the store can be interacted with by external dependencies.

This will allow us to open the shipping label modal by dispatching an action as was done in this PoC: https://gist.github.com/c-shultz/f15f124c45afdc4201ba033fef9c35ee#file-wcs-autoload-php-L94

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -125,4 +125,6 @@ Array.from( document.getElementsByClassName( 'wcc-root' ) ).forEach( ( container
 	);
 } );
 
-window.wcsAppStores = createdStores;
+window.wcsGetAppStore = function( storeKey ) {
+	return createdStores[ storeKey ];
+}
--- a/client/main.js
+++ b/client/main.js
@@ -124,3 +124,5 @@ Array.from( document.getElementsByClassName( 'wcc-root' ) ).forEach( ( container
 		container
 	);
 } );
+
+window.wcsAppStores = createdStores;


client/main.js
```diff
@@ -124,3 +124,5 @@ Array.from( document.getElementsByClassName( 'wcc-root' ) ).forEach( ( container
 		container
 	);
 } );
+
+window.wcsAppStores = createdStores;
```
superdav42: Seems like it would be better to introduce an interface to get the store instead allowing direct access to them.
something like
```js
window.wcsGetAppStore( storeKey ) {
    return createdStores[ storeKey ];
}
```

harriswong: Would object references here be a concern? I am guessing not since it should be the same store globally. We shouldn't need to clone it every time under the same `window` object.

c-shultz: @harriswong, I'm not totally clear on your question. Are you worried about wcsGetAppStore getting defined multiple times?

harriswong: I was thinking `createdStores ` is assigned to `window.wcsGetAppStore`. So anytime a component modify `createdStores`, the `window.wcsGetAppStore` updates. If component-x modifies `createdStores`, then component-y's `window.wcsGetAppStore` will get updated unexpectedly. But i think that's fine since the purpose of this function is to get the store based on a key.

PR #1955, opened 2020-03-16
Fix automated taxes for GB with WC 4.0

### LINKED ISSUE #1954: GB no longer included in EU causing automated taxes to be removed
**Affected customers**
19436393-hc

**The bug**
GB was removed from [the EU countries](https://github.com/woocommerce/woocommerce/blob/4.0.0/includes/class-wc-countries.php#L356-L365) in WC 4.0, causing automated taxes to be disabled for GB, according to [this check](https://github.com/Automattic/woocommerce-services/blob/1.22.4/classes/class-wc-connect-taxjar-integration.php#L212-L223). GB is still on the TaxJar supported list here: https://developers.taxjar.com/api/reference/#countries .

Set this as Critical since it was something that was working and now is not and can cost customers a significant amount of money if taxes are not collected properly.

superdav42:
closes #1954

Test:
Set country to United Kingdom
Install WC 4.0
Ensure WooCommerce->Settings->Taxes has the Automated taxes option and that it works.

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -204,22 +204,14 @@ class WC_Connect_TaxJar_Integration {
 	}
 
 	/**
-	 * TaxJar supports USA, Canada, Australia, and the European Union
+	 * TaxJar supports USA, Canada, Australia, and the European Union + Great Britain
 	 * See: https://developers.taxjar.com/api/reference/#countries
 	 *
 	 * @return array Countries supported by TaxJar.
 	 */
 	public function get_supported_countries() {
-		$supported_countries = array_merge(
-			array(
-				'US',
-				'CA',
-				'AU',
-			),
-			WC()->countries->get_european_union_countries()
-		);
-
-		return $supported_countries;
+		// Hard code list instead of using `WC()->countries->get_european_union_countries()` just in case anyone else decides to leave the EU.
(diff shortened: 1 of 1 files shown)


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -204,22 +204,14 @@ public function configure_tax_settings() {
...
-				'US',
-				'CA',
-				'AU',
-			),
-			WC()->countries->get_european_union_countries()
-		);
-
-		return $supported_countries;
+		// Hard code list instead of using `WC()->countries->get_european_union_countries()` just in case anyone else decides to leave the EU.
+		return array( 'US', 'CA', 'AU', 'AT', 'BE', 'BG', 'CY', 'CZ', 'DE', 'DK', 'EE', 'ES', 'FI', 'FR', 'GB', 'GR', 'HU', 'HR', 'IE', 'IT', 'LT', 'LU', 'LV', 'MT', 'NL', 'PL', 'PT', 'RO', 'SE', 'SI', 'SK' );
```
harriswong: Confirmed all countries with the list taxjar provides at https://developers.taxjar.com/api/reference/#countries.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -204,22 +204,14 @@ public function configure_tax_settings() {
...
 	 * @return array Countries supported by TaxJar.
 	 */
 	public function get_supported_countries() {
-		$supported_countries = array_merge(
-			array(
-				'US',
-				'CA',
-				'AU',
-			),
-			WC()->countries->get_european_union_countries()
```
harriswong: There is a `apply_filters( 'woocommerce_european_union_countries', $countries, $deprecated );` in core. I don't think it breaks anything if we no longer relies on it. Checking.

harriswong: I think this filter is here for 3rd party compatibility. The TaxJar implementation we have is kinda separated from this list. 
One thing I can think of is if this filter return a subset (ie. just 'US') instead of the full list, and WCS continue to return all taxes. 

Even then, I think that's fine.

PR #1977, opened 2020-03-30
Update order detail page if order is fulfilled

### LINKED ISSUE #1948: Update edit order page when modal marks order as complete
Follow up to https://github.com/Automattic/woocommerce-services/issues/1904
We need to update the order status on the page when the merchants checks the "mark order as complete option" after purchasing a label.
The status dropdown uses jQuery Select2 so it needs updated with this line:
```js
jQuery('#order_status').val('wc-completed').trigger('change');
```
The other part is the order notes which are more tricky. We'll need to create a new endpoint in WCS which renders all the notes the same way as wc core then use AJAX to update the note container.

Ferdev:
Closes #1948.

When the order should be marked as completed when a shipping label is bought, this PR updates the order detail section and the order notes, to reflect the changes performed by the label purchase action.

### DIFF
--- a/client/apps/shipping-label/view-wrapper-label.js
+++ b/client/apps/shipping-label/view-wrapper-label.js
@@ -27,6 +27,7 @@ import {
 import { getSelectedSiteId } from 'state/ui/selectors';
 import {
 	areLabelsFullyLoaded,
+	shouldUpdateOrderDetailPage,
 } from '../../extensions/woocommerce/woocommerce-services/state/shipping-label/selectors';
 import {
 	areLabelsEnabled,
@@ -162,6 +163,17 @@ export class ShippingLabelViewWrapper extends Component {
 		this.props.openTrackingFlow( orderId, siteId );
 	};
 
+	updateOrderDetailScreen = () => {
+		if ( window.jQuery ) {
+			window.jQuery.get( window.location.href, function( result ) {
+				const parsedResult = window.jQuery( result );
+				const updatedOrderDetails = parsedResult.find( '#order_data' );
+				const updatedOrderNotes = parsedResult.find( '#woocommerce-order-notes' );
+				window.jQuery( '#woocommerce-order-notes' ).html( updatedOrderNotes.html() );
+				window.jQuery( '#order_data' ).html( updatedOrderDetails.html() );
+			} );
+		}
+	}
 	render() {
 		const {
 			siteId,
@@ -172,6 +184,7 @@ export class ShippingLabelViewWrapper extends Component {
 			translate,
 			events,
 			moment,
+			updateOrderDetailPage,
 		} = this.props;
 
 		const shouldRenderButton = ! loaded || labelsEnabled;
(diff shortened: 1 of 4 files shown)


client/apps/shipping-label/view-wrapper-label.js
```diff
@@ -162,6 +163,17 @@ export class ShippingLabelViewWrapper extends Component {
...
 	};
 
+	updateOrderDetailScreen = () => {
+		if ( window.jQuery ) {
+			window.jQuery.get( window.location.href, function( result ) {
+				const parsedResult = window.jQuery( result );
+				const updatedOrderDetails = parsedResult.find( '#order_data' );
+				const updatedOrderNotes = parsedResult.find( '#woocommerce-order-notes' );
+				window.jQuery( '#woocommerce-order-notes' ).html( updatedOrderNotes.html() );
+				window.jQuery( '#order_data' ).html( updatedOrderDetails.html() );
```
Ferdev: Instead of creating a new API endpoint just to get the updated notes, we perform a GET request to the current page (order's detail page) to get its latest state, and just update the page sections we're interested in.

client/extensions/woocommerce/woocommerce-services/state/shipping-label/reducer.js
```diff
@@ -1346,8 +1346,14 @@ reducers[ WOOCOMMERCE_SERVICES_SHIPPING_LABEL_CLOSE_DETAILS_DIALOG ] = state =>
...
-reducers[ WOOCOMMERCE_ORDER_UPDATE_SUCCESS ] = () => {
-	return initializeLabelsState();
+reducers[ WOOCOMMERCE_ORDER_UPDATE_SUCCESS ] = ( { fulfillOrder }, { order: { status } } ) => {
+	const initialState = initializeLabelsState();
+
+	if ( fulfillOrder && "completed" === status ) {
+		initialState[ 'shouldUpdateOrderDetailPage' ] = true;
+	}
+
+	return initialState;
```
Ferdev: If we marked the option "mark order as complete" and upon order update success, we got the order state as "completed", we should update the order's detail page.

Not sure here if the state associated to the shipping label should hold this variable or if it should be move to another section of the state, though.

client/apps/shipping-label/view-wrapper-label.js
superdav42: Order notes works great but using the method for updating `#order_data` has problems. All the dynamic features of that card do not work. The order datepicker, customer auto completer and edit buttons on the address no longer function correctly when replacing the html like this. I suggest we stick with the line just to update the order status:
```
jQuery('#order_status').val('wc-completed').trigger('change');
```

client/extensions/woocommerce/woocommerce-services/state/shipping-label/reducer.js
```diff
@@ -1346,8 +1346,14 @@ reducers[ WOOCOMMERCE_SERVICES_SHIPPING_LABEL_CLOSE_DETAILS_DIALOG ] = state =>
 };
 
 // Reset the state when the order changes
-reducers[ WOOCOMMERCE_ORDER_UPDATE_SUCCESS ] = () => {
-	return initializeLabelsState();
+reducers[ WOOCOMMERCE_ORDER_UPDATE_SUCCESS ] = ( { fulfillOrder }, { order: { status } } ) => {
+	const initialState = initializeLabelsState();
+
+	if ( fulfillOrder && "completed" === status ) {
+		initialState[ 'shouldUpdateOrderDetailPage' ] = true;
```
superdav42: Compiler gives the warning:
```
  1353:17  error  ["shouldUpdateOrderDetailPage"] is better written in dot notation  dot-notation
```

Ferdev:
PR updated!

Issue #1916, opened 2020-02-12
Create E2E test for the settings screen


Ferdev:
We need to create a test for the settings screen in `/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings`

harriswong:
## Test can include

### Settings
- Toggling, save change
- Changing paper size, save change
- Email Receipts, save changes

### Adding/removing/choosing card
* _need to figure out how to mock the API for cards_
- Choose a different card 
- Add a new card 
- Remove card when there is 1 card
- Remove card when there is 2+ cards
- 0 card

### Packaging
- Add new package
- Edit package
- Init view when there is no package
- Delete package

PR #1978, opened 2020-03-30
Add e2e test for toggling shipping label

harriswong:
Closes https://github.com/Automattic/woocommerce-services/issues/1916

## Goal
Not to cover everything using E2E tests but to cover critical user flows that would affect WCS experience inside the settings page.

### Settings
- Toggling, save change
- Changing paper size, save change
- Email Receipts, check box should display recipient name and email based on their `account/settings` info,  `Email the label purchase receipts to harris (harris) at harris@localhost.com save changes`

### Choose a different card
- Add a new card
- Remove card when there is 1 card
- 0 card

### Packaging
- Add new custom package
- Add new service package
- Edit package
- Delete package

### DIFF
--- a/tests/e2e-tests/specs/admin/shipping-settings.test.js
+++ b/tests/e2e-tests/specs/admin/shipping-settings.test.js
@@ -5,27 +5,52 @@ import { clickReactButton } from '../../utils/index';
 import { StoreOwnerFlow } from "../../utils/flows";
 import { AccountWithNoCreditCard, AccountWithOneCreditCard, AccountWithTwoCreditCard } from "../../fixtures/account_settings";
 
-// describe( 'Saving shipping label settings', () => {
-// 	it( 'Can toggle shipping labels' , async () => {
-// 		let slug = 'woocommerce-services';
-// 		await StoreOwnerFlow.login();
-// 		await StoreOwnerFlow.openSettings('shipping', 'woocommerce-services-settings');
-//         await page.waitForSelector('.form-toggle__switch');
-//         await expect(page).toClick('.form-toggle__switch');
-//         await page.waitForSelector('.label-settings__cards-label form-label', {
-//             hidden: true
-//         });
-//         await expect( page ).toClick( '.button.is-primary', { text: 'Save changes' } );
-//         // Clicking save should persist the hidden state of the section.
-//         await page.waitForSelector('.label-settings__cards-label form-label', {
-//             hidden: true
-//         });
+// Click save and wait until it's saved.
+const saveAndWait = async () => {
+    await expect( page ).toClick( '.button.is-primary', { text: 'Save changes' } );
+    await page.waitForSelector('[class="button is-primary"]');
+};
 
-//         // Enable
-//         await expect(page).toClick('.form-toggle__switch');
-//         await expect( page ).toClick( '.button.is-primary', { text: 'Save changes' } );
-//     });
-// } );
+describe( 'Saving shipping label settings', () => {
+	it( 'Can toggle shipping labels' , async () => {
+		await StoreOwnerFlow.login();
+		await StoreOwnerFlow.openSettings('shipping', 'woocommerce-services-settings');
+        await page.waitForSelector('.form-toggle__switch');
+        await expect(page).toClick('.form-toggle__switch');
+        await page.waitForSelector('.card.label-settings__labels-container', {
+            visible: false
+        });
+        await expect( page ).toClick( '.button.is-primary', { text: 'Save changes' } );
+        // Clicking save should persist the visible state of the section.
+        await page.waitForSelector('.card.label-settings__labels-container', {
+            visible: false
+        });
+
+        // Enable
+        await expect(page).toClick('.form-toggle__switch');
+        await saveAndWait();
+    });
+
+    it ('Should be able to select a different paper size', async () => {
+        await page.select('select.form-select', 'label');
+        await saveAndWait();
+        // TODO: Refresh page to see if it is selected.
+        // await StoreOwnerFlow.openSettings('shipping', 'woocommerce-services-settings');
+
+        let paperSize = await page.$('select.form-select');
+        let selectedOption = await (await paperSize.getProperty('value')).jsonValue();
+        expect(selectedOption).toBe('label');
+
+        await page.select('select.form-select', 'legal');
+        await saveAndWait();
+        // TODO: Refresh page to see if it is selected.
+        // await StoreOwnerFlow.openSettings('shipping', 'woocommerce-services-settings');
+
+        paperSize = await page.$('select.form-select');
+        selectedOption = await (await paperSize.getProperty('value')).jsonValue();
+        expect(selectedOption).toBe('legal');
+    });
+} );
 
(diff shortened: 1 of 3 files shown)


harriswong:
All E2E tests passed locally 
image

tests/e2e-tests/specs/admin/shipping-settings.test.js
```diff
@@ -0,0 +1,348 @@
+/**
+ * Internal dependencies
+ */
+import { StoreOwnerFlow } from "../../utils/flows";
+import { AccountWithNoCreditCard, AccountWithOneCreditCard, AccountWithTwoCreditCard, AccountWithTwoCreditCardAndNoDefault } from "../../fixtures/account_settings";
```
harriswong, line 5: Create 4 mocked responses for `account/settings`.

tests/e2e-tests/specs/admin/shipping-settings.test.js
```diff
@@ -0,0 +1,348 @@
...
+    await page.waitForSelector('[class="button is-primary"]');
+};
+
+/**
+ * This function will wait for a button with any CSS selector + text value.
+ *
+ * @param {string} selector CSS selector
+ * @param {string} text The text value of the element we want to search for. ie. button's value, div's innertext.
+ */
+const waitForSelectorAndText = async (selector, text) => {
```
harriswong, line 19: We can't use `waitForSelector` with a `text` option. That's not available in the Puppeteer API. In order to do something similar to what `expect-puppeteer` does with the `toMatchElement` API (https://github.com/smooth-code/jest-puppeteer/tree/master/packages/expect-puppeteer#expectinstancetomatchelementselector-options), which support a `text` option, i created this function. 


await page.waitForSelector( '.address-step__suggestion-title', { text: 'Address entered' } );`

tests/e2e-tests/specs/admin/shipping-settings.test.js
```diff
@@ -0,0 +1,348 @@
...
+        await expect(page).toMatchElement('.notice.is-success .notice__text', { text: 'Your shipping settings have been saved.' });
+
+        paperSize = await page.$('select.form-select');
+        selectedOption = await (await paperSize.getProperty('value')).jsonValue();
+        expect(selectedOption).toBe('label');
+    });
+} );
+
+describe( 'Shipping label payment method', () => {
+    let response;
```
harriswong, line 69: I set this to global because `accountSettingsRequestListener` event handler needs to change it.

tests/e2e-tests/specs/admin/shipping-settings.test.js
```diff
@@ -0,0 +1,348 @@
...
+    };
+
+    const mockAccountSettingAPI = async (mockResponse) => {
+        response = mockResponse;
+        page.on('request', accountSettingsRequestListener);
+    };
+
+    afterEach(() => {
+        response = '';
+        page.removeListener('request', accountSettingsRequestListener);
```
harriswong, line 90: It is important to remove the `accountSettingsRequestListener` listener otherwise none of the API `request` will work afterwards.

tests/e2e-tests/specs/admin/shipping-settings.test.js
```diff
@@ -0,0 +1,348 @@
...
+        await expect( page ).toFill( '.packages__properties-group #name', packageName );
+        await expect( page ).toFill( '.form-text-input.form-dimensions-input__length', '5' );
+        await expect( page ).toFill( '.form-text-input.form-dimensions-input__width', '5' );
+        await expect( page ).toFill( '.form-text-input.form-dimensions-input__height', '5' );
+        await expect( page ).toFill( '.form-text-input-with-affixes #box_weight', '0.5' );
+        await expect( page ).toClick( '.button.form-button.is-primary', { text: 'Add package' } )
+
+        // Set this once globally, all following test will use this metric
+        const metricSystem = await page.$$('.form-text-input-with-affixes .form-text-input-with-affixes__suffix');
+        metricSystemValue = await (await metricSystem[0].getProperty('innerText')).jsonValue();
```
harriswong, line 252: The unit is either in `in` or `cm`. We need to get this from the system the tests are running on. 

We will find this unit the first time we run the tests, then save this in a variable and used it for all remaining tests.

tests/e2e-tests/specs/admin/shipping-settings.test.js
```diff
@@ -0,0 +1,348 @@
...
+        await saveAndWait();
+        await expect(page).toMatchElement('.notice.is-success .notice__text', { text: 'Your shipping settings have been saved.' });
+
+        paperSize = await page.$('select.form-select');
+        selectedOption = await (await paperSize.getProperty('value')).jsonValue();
+        expect(selectedOption).toBe('label');
+    });
+} );
+
+describe( 'Shipping label payment method', () => {
```
harriswong, line 68: For all test cases in this `describe`, intercept all API to `wp-json/wc/v1/connect/account/settings` and return a mocked response. 
Remove this interception after this test case.

tests/e2e-tests/specs/admin/shipping-settings.test.js
Ferdev: Love those helpers! I would refactor this and probably move them to a separate file, what do you think?

tests/e2e-tests/specs/admin/shipping-settings.test.js
```diff
@@ -0,0 +1,348 @@
...
+                body: JSON.stringify(response)
+            });
+        } else {
+            request.continue();
+        }
+    };
+
+    const mockAccountSettingAPI = async (mockResponse) => {
+        response = mockResponse;
+        page.on('request', accountSettingsRequestListener);
```
Ferdev, line 85: I wonder what happens if an unhandled exception is generated during the test run. We should probably catch all these exceptions and remove the active listener, otherwise we're gonna get weird errors in the rest of the test run.

Ferdev:
Everything looks great, just left a couple of comments!

tests/e2e-tests/specs/admin/shipping-settings.test.js
harriswong: I moved `waitForSelectorAndText` into `utils/index.

I didn't move `saveAndWait` since it is clicking a "Save changes" button. I think "Save changes" is more specific to just this account settings page so I didn't move it.

tests/e2e-tests/specs/admin/shipping-settings.test.js
harriswong: Good question! The `removeListener` is inside `afterEach` of this `describe`. It should run and remove listeners even if the `it()` test cases throw exceptions.

harriswong: Just tested by adding a `throw new Error` in one test case, and counted 7 times `afterEach` calls, same count when I remove the throw error. I think we should be okay!

Ferdev: awesome, thanks for checking this Harris! I wasn't aware the before/after hooks were called even after an unhandled error :+1:

PR #1988, opened 2020-04-02
Fix order status changing after purchasing label

### LINKED ISSUE #1987: Order status is changes when purchasing a label if it's anything other than completed
The new `Notify the customer with shipment details` option should only create a new order note but it also changes the order status. 
To reproduce Set an order's status to failed. Purchase as new label without changing the option.
Order's status will be changed to completed.

Checking and unchecking the option will have the correct behavior of only adding a new order note.
This is cause of how the initial state is setup.

Also there is another problem where the order notes will not be refreshed unless the order status is changed. We need to update our code so only the order notes are refreshed when only a new note is added.

superdav42:
Closes #1987

Test:
Change order status to 'failed'
Purchase a new label.
Order status should not be changed to 'completed'
Also notes should be updated without a refresh.

### DIFF
--- a/client/extensions/woocommerce/state/sites/orders/test/actions.js
+++ b/client/extensions/woocommerce/state/sites/orders/test/actions.js
@@ -145,7 +145,7 @@ describe( 'actions', () => {
 		test( 'should update order detail screen on order save', () => {
 			utils.updateOrderDetailScreen = spy();
 			saveOrderSuccess( siteId, 40, updatedOrder );
-			expect( utils.updateOrderDetailScreen ).to.have.been.called;
+			expect( utils.updateOrderDetailScreen ).to.have.been.called();
 		} );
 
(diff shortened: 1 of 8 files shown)


client/extensions/woocommerce/state/sites/orders/test/actions.js
```diff
@@ -137,6 +142,12 @@ describe( 'actions', () => {
 			} );
 		} );
 
+		test( 'should update order detail screen on order save', () => {
+			utils.updateOrderDetailScreen = spy();
+			saveOrderSuccess( siteId, 40, updatedOrder );
+			expect( utils.updateOrderDetailScreen ).to.have.been.called;
```
Ferdev, line 148: Missing parenthesis: `to.have.been.called()`

c-shultz:
@superdav42, were you wanting more reviews on this or are you ready to merge?

superdav42:
No, I just haven't yet

PR #1994, opened 2020-04-03
Optimize bundles

superdav42:
Before:
image

After:
image
Done:
Lazy load code via routes
Lazy load create shipping label modal
Extern all wp-* scripts that can be.

TODO:
Test in WP < 5 and older than 5.3. Possible mix bundling and extern to support WP < 5
Remove core-js by using wp-polyfill extern
Refactor ShippingLabelViewWrapper to be as light as possible on initial load and lazy load more things for the modal.


client/apps/shipping-label/view-wrapper-label.js
```diff
@@ -84,6 +86,7 @@ export class ShippingLabelViewWrapper extends Component {
 			// If there are no purchased labels, just show Create labels button
 			if ( ! activeLabels.length ) {
 				return (
+					<span>
```
Ferdev, line 89: why are these `span`s? Semantically, spans are not allowed to contain block elements. We should use a block element here, and if it needs to be displayed inline, we should adjust its styling.

client/apps/shipping-label/view-wrapper-label.js
```diff
@@ -93,6 +96,10 @@ export class ShippingLabelViewWrapper extends Component {
 					>
 						{ translate( 'Create shipping label' ) }
 					</Button>
+						<Suspense fallback={<div />}>
+							<LabelPurchaseModal orderId={ orderId } siteId={ siteId } />
+						</Suspense>
```
Ferdev, line 101: Indentation is a bit off.

client/apps/shipping-label/view-wrapper-label.js
```diff
@@ -134,6 +145,10 @@ export class ShippingLabelViewWrapper extends Component {
 					>
 						{ translate( 'Track Package', 'Track Packages', { count: activeLabels.length } ) }
 					</Button>
+					<Suspense fallback={<div />}>
+						<LabelPurchaseModal orderId={ orderId } siteId={ siteId } />
+						<TrackingModal orderId={ orderId } siteId={ siteId } />
+					</Suspense>
 				</span>
```
Ferdev, line 152: Same here with the `span`s. Block elements should be used instead.

superdav42: divs are better

PR #1995, opened 2020-04-03
Remove spinner from create shipping label button

Ferdev:
Removes the preloading step in the create shipping label component and provides it with all the that it needs from the PHP side.

### DIFF
--- a/classes/class-wc-connect-account-settings.php
+++ b/classes/class-wc-connect-account-settings.php
@@ -1,0 +1,58 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( class_exists( 'WC_Connect_Account_Settings' ) ) {
+	return;
+}
+
+class WC_Connect_Account_Settings {
+
+	/**
+	 * @var WC_Connect_Service_Settings_Store
+	 */
+	protected $settings_store;
+
+	public function __construct( 
+		WC_Connect_Service_Settings_Store $settings_store,
+		WC_Connect_Payment_Methods_Store $payment_methods_store
+	) {
+			$this->settings_store = $settings_store;
+			$this->payment_methods_store = $payment_methods_store;
+	}
+
+
+	public function get() {
+		$payment_methods_warning = false;
+		$payment_methods_success = $this->payment_methods_store->fetch_payment_methods_from_connect_server();
+
+		if ( ! $payment_methods_success ) {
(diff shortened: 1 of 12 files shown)


classes/class-wc-connect-shipping-label.php
```diff
@@ -39,11 +54,22 @@ class WC_Connect_Shipping_Label {
...
 			WC_Connect_API_Client $api_client,
 			WC_Connect_Service_Settings_Store $settings_store,
-			WC_Connect_Service_Schemas_Store $service_schemas_store
+			WC_Connect_Service_Schemas_Store $service_schemas_store,
+			WC_Connect_Payment_Methods_Store $payment_methods_store
 		) {
 			$this->api_client = $api_client;
 			$this->settings_store = $settings_store;
 			$this->service_schemas_store = $service_schemas_store;
+			$this->payment_methods_store = $payment_methods_store;
```
superdav42, line 63: This property appears to be unused by the class WC_Connect_Shipping_Label so it should be safe to remove.

classes/class-wc-connect-shipping-label.php
```diff
@@ -39,11 +54,22 @@ class WC_Connect_Shipping_Label {
...
+			$this->account_settings = new WC_Connect_Account_Settings(
+					$settings_store,
+					$payment_methods_store
+			);
+			$this->package_settings = new WC_Connect_Package_Settings(
+					$settings_store,
+					$service_schemas_store
+			);
+			$this->continents = new WC_Connect_Continents();
```
superdav42, line 72: Next hack week let's setup [PHP-DI](http://php-di.org/) so we don't have to do all this stuff.

classes/class-wc-connect-shipping-label.php
Ferdev: oh yes, you're right!

classes/class-wc-connect-account-settings.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( class_exists( 'WC_Connect_Account_Settings' ) ) {
+	return;
+}
```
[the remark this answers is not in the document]
Ferdev: I was wondering why these checks were all over the place. I'll remove them in the classes I've introduced in this PR.

classes/class-wc-connect-shipping-label.php
Ferdev: Yeah, let's do that. I'm a big fan of dependency injection.

Issue #1997, opened 2020-04-03
Update NodeJS to version 12.x (LTS)


harriswong:
Node 12.16.1 is now the LTS and `wp-calypso` has also moved to `12.13.1` (https://github.com/Automattic/wp-calypso/blob/master/package.json#L66).

- Update `.nvmrc`
- Update `.travis.yml` `install` section
- Update `README.md`

PR #1999, opened 2020-04-08
[PoC] Add support to nodejs 12+

harriswong:
# Do not merge, proof of concept.

Closes https://github.com/Automattic/woocommerce-services/issues/1997

**Summary**
Update WooCommerce Services to work with nodejs 12. 

**Problems:**
- WooCommerce has a submodule `wp-calypso` and it is using `df4ddfdc94438b6d31b686d6415e08a567570944 wp-calypso (@automattic/jetpack-blocks@13.1.0-890-gdf4ddfdc94)`. It doesn't compile well with nodejs 12 because of its node-sass version.
- Our `package.json` has a `prepare` script that goes into the `wp-calypso` submodule to run `npm ci`. The `prepare` scripts run before packed and pubilshed, so it rebuilds submodule per `npm i` inside WCS. This means that if we run `npm i` under WCS, it will automatically runs `npm ci` inside `wp-calypso`. Upgrading to nodejs 12 breaks the submodule.
- We don't want to update `wp-calypso` to the latest yet. Updating the submodule has risk of breaking WCS and there isn't a pressing need to update at the moment.

**Proposed change:**
- Remove `"prepare": "cd wp-calypso && npm ci && cd .."` in `package.json`. 
- Bump `"node-sass": "^4.13.1"` to allow any minor updates
- Generate shrinkwrap coorespondingly based on node version.
- Build `wp-calypso` submodule using nodejs 10. Build WCS using nodejs 12.

**Result:**
Pull this branch and then go into `wp-calypso` submodule, switch to nodejs 10, build.
```
cd wp-calypso
nvm use 10
npm ci
```

Then, go back to WCS root
```
cd ..
nvm use 12
npm ci
npm run update-deps
npm run start
```

Switching node version during builds creates a problem for travis CI and production release. 

**Conclusion:**
Travis CI and release scripts don't work well when part of the plugin relies on a different version of nodejs. Unless there is a huge benefit using nodejs 12 than 10. we should just use 1 version throughout the plugin. 

While there are some improvements in nodejs 12 (https://medium.com/@nodejs/introducing-node-js-12-76c41a1b3f3f), they are not updates that WCS require in order to operate. nodejs 10's end of life is in April 2021. We are also exploring the option to remove `wp-calypso` as a dependency. Perhaps then that would be a better time to upgrade to nodejs 12.


superdav42:
I think it may be possible to make it work by using npm to manage the calypso repo instead of using a submodule. npm supports installing a specific hash so if we use it to install calypso we could then override calypso's dependencies with npm-shrinkwrap. This should let everything work without much changes.

harriswong:
We actually had it before with a hash https://github.com/Automattic/woocommerce-services/pull/1601/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L93

I haven't dug into it too much as to why. I think it is related to some monorepo-related changes in Calypso which broke the hash dependency. `npm` seemed to break when resolving file: sub-dependencies

PR #2004, opened 2020-04-16
Fix svg images not showing on dev

superdav42:
we must set the publicPath is webpack config on dev so css urls are correct.

Also fixes a problem where CC images are not showing.

### DIFF
--- a/assets/stylesheets/_fix-image-paths.scss
+++ b/assets/stylesheets/_fix-image-paths.scss
@@ -4,16 +4,19 @@ input[type='checkbox'] {
 		content: url( '../../wp-calypso/public/images/checkbox-icons/checkmark-primary.svg' );
 	}
 }
-$payment-cc-brands: 'amex', 'diners', 'discover', 'jcb', 'mastercard', 'unionpay', 'visa';
-$payment-types: 'alipay', 'bancontact', 'giropay', 'eps', 'ideal', 'paypal', 'p24', 'brazil-tef',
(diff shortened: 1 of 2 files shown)


assets/stylesheets/_fix-image-paths.scss
```diff
@@ -4,16 +4,19 @@ input[type='checkbox'] {
...
 	}
 }
-$payment-cc-brands: 'amex', 'diners', 'discover', 'jcb', 'mastercard', 'unionpay', 'visa';
-$payment-types: 'alipay', 'bancontact', 'giropay', 'eps', 'ideal', 'paypal', 'p24', 'brazil-tef',
-'wechat', 'sofort';
-@each $brand in $payment-cc-brands {
-	&.is-#{$brand} {
-		background-image: url( '../../wp-calypso/public/images/upgrades/cc-#{$brand}.svg' );
+
+.payment-logo {
```
harriswong, line 8: How come we are adding another class?

harriswong: I think this fixes the "mastercard/visa card images not showing" problem.

harriswong:
👍 🚢

assets/stylesheets/_fix-image-paths.scss
superdav42: To fix the visa mastercard images. Basically it never worked the way it was.

PR #2008, opened 2020-04-22
Remove Stripe deferred account creation methods

chickenn00dle:
As of May 31, the API for stripe deferred account activation will no
longer be supported. As a result, the methods and class related to this
functionality are no longer needed.

See paJDYF-x4-p2

This removes the `WC_REST_Connect_Stripe_Account_Controller` class along
with some related methods in `WC_Connect_API_Client` and
`WC_Connect_Stripe`.

**Testing Instructions**

- With the changes implemented in https://github.com/woocommerce/woocommerce-admin/pull/4164 checked out, run through the Stripe connection flow in onboarding.
- Confirm there are no errors and that Stripe settings are displayed correctly after finishing the above.


c-shultz:
Is there any time sensitivity to this change, @chickenn00dle? Or it just won't be used by the WC OBW after May 31st?

harriswong:
👀

classes/class-wc-connect-stripe.php
```diff
@@ -63,43 +63,6 @@ public function get_oauth_url( $return_url = '' ) {
 			return $result->oauthUrl;
 		}
 
-		public function create_account( $email, $country ) {
-			$response = $this->api->create_stripe_account( $email, $country );
-			if ( is_wp_error( $response ) ) {
-				return $response;
-			}
-			return $this->save_stripe_keys( $response );
-		}
```
harriswong, line 72: Hi @chickenn00dle, 

Our `init_core_wizard_payments_config` rely on this `create_account` function https://github.com/Automattic/woocommerce-services/blob/develop/woocommerce-services.php#L623. Do we need to make change to how `init_core_wizard_payments_config` works?

classes/class-wc-rest-connect-stripe-account-controller.php
```diff
@@ -1,60 +0,0 @@
...
-if ( ! defined( 'ABSPATH' ) ) {
-	exit;
-}
-
-if ( class_exists( 'WC_REST_Connect_Stripe_Account_Controller' ) ) {
-	return;
-}
-
-class WC_REST_Connect_Stripe_Account_Controller extends WC_REST_Connect_Base_Controller {
-	protected $rest_base = 'connect/stripe/account';
```
harriswong, line 12: In our `client/extensions/woocommerce/state/sites/settings/stripe-connect-account/actions.js`, we makes call to the stripe end point https://github.com/Automattic/woocommerce-services/blob/develop/client/extensions/woocommerce/state/sites/settings/stripe-connect-account/actions.js#L99

Do we need to remove `createAccount` https://github.com/Automattic/woocommerce-services/blob/develop/client/extensions/woocommerce/state/sites/settings/stripe-connect-account/actions.js#L77 as well?

chickenn00dle:
> Is there any time sensitivity to this change, @chickenn00dle? Or it just won't be used by the WC OBW after May 31st?

This just won't be used by the OBW (Task List) after the changes are made in WC Admin. So I don't think there is any time sensitivity.

classes/class-wc-connect-stripe.php
chickenn00dle: Good catch!

It looks like this functionality becomes redundant once deferred account creation is removed.

From what I am seeing, `$user_elected_to_create_stripe_account` will always be false here:
https://github.com/Automattic/woocommerce-services/blob/bca5be594c939770c5d377ae9cbcd6c47110256d/woocommerce-services.php#L596

And all of the conditional blocks in `init_core_wizard_payments_config` are ignored.

I tested this by removing the function and the line where it is called and the Stripe onboarding flow worked with no issues:
https://github.com/Automattic/woocommerce-services/pull/2008/commits/079ce1cb72501eda4f569faf3a782e92f8d886c1

I may be missing something though (is this used in any other scenario?) as I'm not very familiar with WC Services, so it'd be great if you could confirm the above.

classes/class-wc-rest-connect-stripe-account-controller.php
chickenn00dle: Thanks for digging into this @harriswong!

I removed all action, reducer, and selector code related to Stripe account creation:
https://github.com/Automattic/woocommerce-services/pull/2008/commits/66472de0915afc978734cb262c8ef2c3b77a7934

I also made sure to clean up the tests for these as well (It looks like there is an unrelated failing test for Shipping Rate).

Truthfully, I'm not certain where to look for these changes to manually confirm them on my test site. But I did make sure to run through setting up Stripe via onboarding and verifying settings a few times.

It'd be great if you could let me know where to look for future reference.

Thank you!

harriswong: No problem! Thanks for looking into this! I am not too familiar with the Stripe codes inside WCS as well so I did a little digging. I ran a search on `create_account`. That's where I saw the use of `init_core_wizard_payments_config`. 

Then I saw that we are removing the controller `class-wc-rest-connect-stripe-account-controller.php` so the end point caught my eyes.

harriswong:
I tried to test this locally by forcing the onboarding page to show (trying to avoid building a new WP site locally), but I got a blank page instead. I will try again next week.

chickenn00dle:
Perhaps related to https://github.com/woocommerce/woocommerce-admin/pull/4169 ?

classes/class-wc-rest-connect-stripe-account-controller.php
chickenn00dle: Gotcha. Thanks for sharing!

chickenn00dle:
I should also note, just in case, that this PR shouldn't go to production until AFTER the related changes in wc-admin do. Otherwise, the onboarding steps for Stripe will break until then.

Perhaps it would be safest to aim to include this in a release that is after May 31.

harriswong:
Ran through onboarding okay with the old flow:
image

harriswong:
Just want to double check, do we need to also test the new onboarding flow with this change as well? 
image

chickenn00dle:
Hmm. It's my understanding that the old onboarding flow will be gone as of the next WC Admin release, so this change only applies to the new onboarding flow. @thenbrent could you confirm?

harriswong:
@chickenn00dle Thanks for the link to https://github.com/Automattic/woocommerce-services/pull/2008#issuecomment-619668383, I was able to test this locally by changing `homepage: false`in `development.json`.

I clicked "Set up":
image

Then I click "Connect"
image

Then I got `{"error":{"message":"No application matches the supplied client identifier"}}`. Is this expected?

Branch i am using:
WCS: `update/stripe_onboarding`
WooCommerce admin: I am using `master` (717be1ab) since https://github.com/woocommerce/woocommerce-admin/pull/4164 is merged

chickenn00dle:
Awesome! I'm glad you got the blank page issue resolved @harriswong!

As for the error you mentioned, I am not seeing the same problem on my end.

Using:
WCS: `update/stripe_onboarding`
WC Admin: `master`

And a fresh installation of WordPress, I get the following:

Screen Capture on 2020-04-30 at 11-31-51(1)

I tried a few things to replicate what you are seeing but had no luck.

I do know when I was working on [connect server changes involving Stripe Deferred Accounts](https://github.com/Automattic/woocommerce-connect-server/pull/1367), I ran into a similar issue.

Basically, if Stripe is in development mode, instead of returning `publishable_key` and `secret_key`, Stripe instead returns `test_publishable_key` and `test_secret_key`, which results in this stuck behavior at the connect button. 

Here is the code where WC Admin checks for these keys before determining whether to complete set up:
https://github.com/woocommerce/woocommerce-admin/blob/master/client/dashboard/task-list/tasks/payments/stripe.js#L43

Perhaps this is related?

harriswong:
@chickenn00dle Thanks for the notes. I was able to get the redirect after making some changes. Confirmed that the new onboarding flow works.

client/extensions/woocommerce/state/sites/settings/stripe-connect-account/actions.js
```diff
@@ -10,8 +10,6 @@
 import {
 	WOOCOMMERCE_SETTINGS_STRIPE_CONNECT_ACCOUNT_CLEAR_COMPLETED_NOTIFICATION,
 	WOOCOMMERCE_SETTINGS_STRIPE_CONNECT_ACCOUNT_CLEAR_ERROR,
-	WOOCOMMERCE_SETTINGS_STRIPE_CONNECT_ACCOUNT_CREATE,
-	WOOCOMMERCE_SETTINGS_STRIPE_CONNECT_ACCOUNT_CREATE_COMPLETE,
```
harriswong, line 14: We will probably need to remove these as well. https://github.com/Automattic/woocommerce-services/blob/develop/client/extensions/woocommerce/state/action-types.js#L209-L212

harriswong:
@chickenn00dle i want to mention that we have a Calypso submodule dependency https://github.com/Automattic/woocommerce-services/blob/develop/.gitmodules. This points to an older version of Calypso https://github.com/Automattic/wp-calypso/commit/df4ddfdc94438b6d31b686d6415e08a567570944. 

The changes you made inside `client/extension/woocommerce/*` will not reflect in the Calypso submodule. For example, the changes you made in `client/extensions/woocommerce/state/sites/settings/stripe-connect-account/actions.js` (https://github.com/Automattic/woocommerce-services/pull/2008/files#diff-1715c3bdb5a2aa3eecdb4852ab2c5389) will not reflect in
`woocommerce-services/wp-calypso/client/extensions/woocommerce/state/sites/settings/stripe-connect-account/test/actions.js`

I think it is okay to not update the submodule since we always attempted to load `client/*` first before we falls back to `wp-calypso/client`. https://github.com/Automattic/woocommerce-services/blob/develop/webpack.config.js#L234-L240

harriswong:
I only have one more comment https://github.com/Automattic/woocommerce-services/pull/2008#discussion_r418342886, otherwise I think this is good to go.

chickenn00dle:
Thanks @harriswong ! I've resolved that in https://github.com/Automattic/woocommerce-services/pull/2008/commits/a9afa49027004e8983d22d749b64675311b77d1d

harriswong:
Thanks for making all the changes, providing notes to other PRs, generating gifs. These made reviewing much easier!

Please feel free to merge.

PR #2018, opened 2020-04-24
Fix 404 of taxjar.js on new order page

superdav42:
We changed the name of javascript files to contain the version and this broke this js file which is loaded on the new order page.


harriswong:
👀

superdav42:
I'm not sure what it does either but at least now there is no errors

PR #2026, opened 2020-05-13
Add e2e test for labels refund

### LINKED ISSUE #2011: Test refunding a label
**Test the following use case:**
User has previously purchased a label but would like to refund.
image

**Test should include:**
- Verify "request refund" works if label has not been used.
- Verify "request refund" fails if label is already used. 
- Verify the user can purchase a label again after a successful refund.

Ferdev:
Fixes #2011

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "globby": "^6.1.0",
     "jest": "^24.9.0",
     "jest-puppeteer": "^4.4.0",
+    "moment": "^2.25.3",
     "moment-timezone-data-webpack-plugin": "^1.1.0",
     "nock": "10.0.6",
     "optimize-css-assets-webpack-plugin": "^5.0.3",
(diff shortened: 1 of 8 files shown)


superdav42:
:eyes:

package.json
```diff
@@ -46,6 +46,7 @@
     "globby": "^6.1.0",
     "jest": "^24.9.0",
     "jest-puppeteer": "^4.4.0",
+    "moment": "^2.25.3",
```
superdav42, line 49: I think this should be a dev dependency since it's just used be the e2e test.

Ferdev: yeah! It's in `devDependencies` :smiley:

c-shultz:
Is this ready to merge, @Ferdev ?

Ferdev:
Yeah it just seems the review is pending.

c-shultz:
@superdav42, did you have any follow up comments or is this good to go on your end?

PR #2027, opened 2020-05-18
Show which nexus automatted taxes will work with and link to doc

### LINKED ISSUE #2019: Make it clear automated taxes only work in one nexus
We need to add a bit of text to the automated taxes setting like so
Screen Shot 2020-04-28 at 11 24 59
This will help with merchant confusion as to why out of state taxes are always 0.

superdav42:
closes #2019

Looks like image

There may be an issue with adding links in the tooltips as it's impossible for me to hover over the tooltip and click on the link. I the problem may be limited to me and my browser/OS combination so I'd like to know if others can click on the link.

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -124,11 +124,16 @@ class WC_Connect_TaxJar_Integration {
 	public function add_tax_settings( $tax_settings ) {
 		$enabled = $this->is_enabled();
 
+		$store_settings = $this->get_store_settings();
+		$all_states = WC()->countries->get_states( $store_settings['country'] );
+		$full_state = $all_states[ $store_settings['state'] ];
+
 		$automated_taxes = array(
 			'title'    => __( 'Automated taxes', 'woocommerce-services' ),
 			'id'       => self::OPTION_NAME, // TODO: save in `wc_connect_options`?
-			'desc_tip' => __( 'Automate your sales tax calculations with WooCommerce Services, powered by Jetpack.', 'woocommerce-services' ),
-			'desc'     => $enabled ? '<p>' . __( 'Powered by WooCommerce Services ― Your tax rates and settings are automatically configured.', 'woocommerce-services' ) . '</p>' : '',
+			/* translators: 1: Full state name */
+			'desc_tip' => sprintf( __( 'Your tax rates and settings will be automatically configured for %1$s. <a href="https://docs.woocommerce.com/document/setting-up-taxes-in-woocommerce/#section-12">Learn more about setting up tax rates for additional nexuses</a>', 'woocommerce-services' ), $full_state ),
+			'desc'     => $enabled ? '<p>' . __( 'Powered by WooCommerce Services.', 'woocommerce-services' ) . '</p>' : '',
 			'default'  => 'no',
 			'type'     => 'select',
 			'class'    => 'wc-enhanced-select',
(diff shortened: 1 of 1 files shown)


superdav42:
So this will not actually work very well as it's impossible to click the link in the tooltip. I made a [PR to core](https://github.com/woocommerce/woocommerce/pull/26558) to allow this to work and also fixes [a long standing bug](https://github.com/woocommerce/woocommerce/issues/26555) that has been bugging me for a while. So it'll probably get merged into core but we'll need to do something about current and old versions. I'll see if I can backport the fix.

superdav42:
Added a [hack](https://github.com/Automattic/woocommerce-services/pull/2027/commits/0f54ae557fc99e51860a45850dc9324809697b81) to make the tooltips stay when you hover over them so you can click the link. It's a little ugly but I think it works and even if something goes sideways it'll just break the tooltips so that should be ok.

harriswong:
👀

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -67,6 +67,11 @@ public function init() {
 		// Add toggle for automated taxes to the core settings page
 		add_filter( 'woocommerce_tax_settings', array( $this, 'add_tax_settings' ) );
 
+		// Fix tooltip with link on older WC.
+		if ( version_compare( WOOCOMMERCE_VERSION, '4.3.0', '<' ) ) {
+			add_action( 'admin_enqueue_scripts', array( $this, 'fix_tooltip_keepalive'), 11 );
+		}
+
```
harriswong, line 74: `4.3.0` beta is out, would https://github.com/woocommerce/woocommerce/pull/26558 made it in `4.3.0` in time? Otherwise, we might need to bump this version up.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -154,6 +165,33 @@ public function add_tax_settings( $tax_settings ) {
...
+	/**
+	 * Hack to force keepAlive: true on tax setting tooltip.
+	 */
+	public function fix_tooltip_keepalive() {
+		global $pagenow;
+		if ( 'admin.php' !== $pagenow || ! isset( $_GET['page'] ) || 'wc-settings' !== $_GET['page'] || ! isset( $_GET['tab'] ) || 'tax' !== $_GET['tab'] || ! empty( $_GET['section'] ) ) {
+			return;
+		}
+
+		$tax_settings = $this->add_tax_settings( array() );
```
harriswong, line 177: If all we need is `desc_tip`, we could refactor the `desc_tip` out into its own function:
```
function getTaxRateTooltip() {
$store_settings = $this->get_store_settings();
	$all_states = WC()->countries->get_states( $store_settings['country'] );
	$full_state = $all_states[ $store_settings['state'] ];
	$descTip = sprintf( __( 'Your tax rates and settings will be automatically configured for %1$s. <a href="https://docs.woocommerce.com/document/setting-up-taxes-in-woocommerce/#section-12">Learn more about setting up tax rates for additional nexuses</a>', 'woocommerce-services' ), $full_state );
    return $descTip;
}
```

Then we don't need to worry about any side effects that may be caused by `add_tax_settings()` function. But I took a look at `add_tax_settings()`, looks like it's a function that transform input to certain output, it doesn't modify any global variables/states. I think this is fine as well.

superdav42:
Good call on the missing state. Some EU countries don't have a state that can be configured so in those cases we should just show the country name. I update it to show state, country is set or just country. Also bumped the version check to 4.4.0 hopefully https://github.com/woocommerce/woocommerce/pull/26558 will be merged by then.

PR #2043, opened 2020-06-10
Add carrier name to price summary and UPS carrier charge notice

### LINKED ISSUE #2037: Shipping summary for multiple carriers
Add support for multiple carriers in shipping summary area.

**UPS carrier account:**
When the UPS carrier account is selected, the line item will read “Package [x] - UPS” with a detail text just below that reads “Your UPS account will be charged”

**Handle multiple packages:**
A merchant may have split multiple items into two packages, then selected USPS for one package and UPS via carrier account for the other.
Here the line items would be listed separately with the carrier account showing as “Package [x] - UPS” and the other showing “Package [x] - USPS”. Since one of the packages is using a discounted USPS rate, the savings message is displayed. If the carrier account is selected for both packages, the messaging would be consistent for each line item.

Design screenshot:
image

Design P2: https://wcship.wordpress.com/2020/05/29/adding-a-carrier-account-in-wcs-i5-design/

superdav42:
closes #2037
Adds a notice in sidebar that UPS account will be charged. Also adds carrier name next too each item in the price summary. 

Extracts Carrier name from title which includes both the carrier and service separated by a dash. I considered passing the carrier name in a field by itself but this would require changes to the connect server and changing the API versioning so it seemed better just to parse the existing title field.

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
@@ -156,6 +156,8 @@ export const getTotalPriceBreakdown = ( state, orderId, siteId = getSelectedSite
 				retailRate: foundRate.retail_rate,
 				rateWithDiscount: rateTotal,
 				addons: [],
+				carrierId: foundRate.carrier_id,
+				carrierTitle: foundRate.title.split( '-' )[0],
 			}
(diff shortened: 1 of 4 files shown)


harriswong:
👀

harriswong:
Travis test failed complaining about old PHPUnit version. 👍  otherwise.

PR #2049, opened 2020-06-16
Add/carrier account settings 2028

### LINKED ISSUE #2028: Carrier Account Settings
For sites with the [UPS carrier account feature flag enabled](https://github.com/Automattic/woocommerce-connect-server/issues/1361), they should be able to see the following option to connect their account on the WCS settings screen:

image

Which takes them to this screen for entry:

image

After account info is entered, it will need to be sent to the Connect Server to be validated/registered with EasyPost/UPS. The Connect Server endpoint is [detailed on this issue](https://github.com/Automattic/woocommerce-connect-server/issues/1362).

With an account added, they will see the account number and be allowed to disconnect the account:
image

Some considerations:
- We'll need an E2E test for the new flow
- The Connect Server will respond with an error if the account info does not validate. We should display this info to the users so they know why the validation failed

Full design info: https://wcship.wordpress.com/2020/05/15/adding-a-carrier-account-in-wcs-i4-design/

Ferdev:
Closes #2028 

These two issues have not been implemented yet:

- #2050 - Breadcrumbs in the carrier account settings form.
- #2051 - State of the save button in the shipping settings screen.

I'm gonna move these two issues to separate GitHub issues because this PR is big enough already.

I've decided to display all errors inline except for the license agreement checkbox. Inline errors looked weird in the checkbox, so for this component, the errors follow the Figma design: are displayed in the top of the form.

Another thing pending is unit tests: I'll work on that on a separate PR and I'll merge back to it any modifications we make to this one.


superdav42:
:eyes:

c-shultz:
@harriswong, can you pick up the PR review on this once you're back from AFK?

harriswong:
👀

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/list-item.js
```diff
@@ -0,0 +1,166 @@
...
+	const showDisconnectDialogHandler = () => {
+		props.setVisibilityDisconnectCarrierDialog( siteId, data.carrierId, true );
+	}
+
+	const hideDisconnectDialogHandler = () => {
+		props.setVisibilityDisconnectCarrierDialog( siteId, data.carrierId, false );
+	}
+
+	const renderActions = () => {
+		const { credentials,  onConnect, onDisconnect } = data;
```
harriswong, line 84: My lint console complaint: `84:36  error  'onDisconnect' is assigned a value but never used  no-unused-vars` :D

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/settings.js
```diff
@@ -0,0 +1,354 @@
...
+						<div className="carrier-accounts__settings-header">
+							<h4 className="carrier-accounts__settings-subheader">{ translate( 'UPS account information' ) }</h4>
+							<Checkbox id={ 'license_agreement' } checked={ showUPSInvoiceFields } onChange={ updateShowUPSInvoiceFields  } />
+							<span>{ translate( 'I have been issued an invoice from UPS within the past 90 days' ) }</span>
+						</div>
+						{ showUPSInvoiceFields && upsInvoiceFields() }
+					</CompactCard>
+					<CompactCard className="carrier-accounts__settings-license-agreement">
+						<Checkbox id={ 'license_agreement' } checked={ !! getValue( 'license_agreement' ) } onChange={ updateValue( 'license_agreement' )  } />
+						<span>{ translate( 'I have read the {{a}}License Agreement{{/a}}', { components: { a: <a href="https://link.to.terms.com/" />  } } ) }</span>
```
harriswong, line 274: Do we have another issue for updating `https://link.to.terms.com`?

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/settings.js
```diff
@@ -0,0 +1,354 @@
...
+		return [ <Button compact primary scary onClick={ () => history.back() }>{ translate( 'Cancel' ) }</Button> ];
+	}
+
+	return (
+		<div className="carrier-accounts__settings-container">
+			{ displayErrors() }
+			<div className="carrier-accounts__settings">
+				<div className="carrier-accounts__settings-info">
+					<h4 className="carrier-accounts__settings-subheader">{ translate( 'Connect your UPS account' ) }</h4>
+					<p className="carrier-accounts__settings-subheader-description">{ translate( 'Set up your own UPS carrier account to compare rates and print labels from multiple carriers in WooCommerce Services. Learn more about adding {{a}}carrier accounts{{/a}}.', { components: { a: <a href="https://link.to.carrier.accounts.com/" />  } } ) }</p>
```
harriswong, line 133: Similar to the above, do we have an issue to update the `https://link.to.carrier.accounts.com` link?

harriswong:
"connect" button remained disabled when i filled out all information. It also flagged website URLas invalid.
image

harriswong:
@Ferdev Should we add a hint to date format? ie. 2020-01-02, 2020/01/02, etc? Or is every carrier different? I use dashes on mine. 

image

harriswong:
I managed to get the button to show by using `example.com` instead of `http://example.com`. And by filling out all fields. But when I click on "connect", nothing is happening. There is no XHR requests made and no console.log error.
Kapture 2020-07-08 at 11 22 05

harriswong:
When I clicked "cancel", a modal popped up. Then I wanted to click "yes, cancel this" but the only button was "cancel". So i thought the "cancel" here means to "cancel my cancel-request", the exact function as clicking "x".

image

@Ferdev @elizaan36 Would it be better if we change the wording to "Yes" or something like that?

client/extensions/woocommerce/app/settings/shipping/index.js
```diff
@@ -41,6 +41,7 @@ class Shipping extends Component {
 	};
 
 	render = () => {
+		alert( 'entra' )
```
harriswong, line 44: Do we need this? :D

client/extensions/woocommerce/woocommerce-services/lib/initialize-carrier-accounts-state.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+export default () => {
+	const requiredFields = [
+		'account_number',
+		'name',
+		'address',
+		'city',
+		'state',
+		'state',
```
harriswong, line 14: There are 2 states here. We should probably remove 1.

client/extensions/woocommerce/woocommerce-services/state/carrier-accounts/reducer.js
```diff
@@ -0,0 +1,145 @@
...
+
+	return newState;
+};
+
+reducers[ WOOCOMMERCE_SERVICES_CARRIER_ACCOUNTS_TOGGLE_SHOW_UPS_INVOICE_FIELDS ] = ( state, { carrier } ) => {
+	const settings = state[ carrier ].settings;
+
+	const newState = {
+		...state,
+		[ carrier ]: {
```
harriswong, line 85: If this reducer (and its action) is strictly for `UPS`, should we remove `carrier` from the parameter? and have it as 
```
UPS: {
  ...state,
  settings: {...}
}
```

Otherwise, we could have an action `WOOCOMMERCE_SERVICES_CARRIER_ACCOUNTS_TOGGLE_SHOW_UPS_INVOICE_FIELDS` for `fedex` and it will have `showUPSInvoiceFields` as a settings state.

client/extensions/woocommerce/woocommerce-services/lib/initialize-carrier-accounts-state.js
```diff
@@ -0,0 +1,41 @@
...
+	return {
+		UPS: {
+			settings: {
+				requiredFields,
+				values:{
+					country: 'US'
+				},
+				fieldErrors: {},
+				ignoreValidation: requiredFields.reduce( (accumulator, currentValue) => {   
+					accumulator[ currentValue ] = 'country' === currentValue ? false : true;
```
harriswong, line 34: Why do we ignore validations except `country`?

client/extensions/woocommerce/woocommerce-services/state/carrier-accounts/selectors.js
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	if ( values.email  && ! values.email.match( emailRegex ) ) {
+		fieldErrors.email = translate( "The email format is not valid" );
+	}
+	if ( values.company_website  && ! values.company_website.match( urlRegex ) ) {
+		fieldErrors.company_website = translate( "The company website format is not valid" );
+	}
+
+	if ( ignoreValidation ) {
```
harriswong, line 52: If we leave fields blank, `fieldErrors` will be an object with key = field_name, and value = `'This field is required'`.  But since all validation are ignored except country, the `'This field is required'` error will be removed. 

I think the `'This field is required'` should stay without being removed.

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/index.js
```diff
@@ -0,0 +1,107 @@
...
+			window.location.href = url.href;
+		}
+
+		const onDisconnect = () => {
+		}
+
+		const carriers = [
+			{ carrierId: 'UPS', name: 'UPS', onConnect },
+			{ carrierId: 'UPS', name: 'UPS', credentials: '989999847463', onDisconnect },
+		]
```
harriswong, line 66: Are we leaving the actual registration API calls out from this PR?

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/list-item.js
```diff
@@ -0,0 +1,166 @@
...
+			<span>{ credentials }</span>
+		</div>;
+	}
+
+	const showDisconnectDialogHandler = () => {
+		props.setVisibilityDisconnectCarrierDialog( siteId, data.carrierId, true );
+	}
+
+	const hideDisconnectDialogHandler = () => {
+		props.setVisibilityDisconnectCarrierDialog( siteId, data.carrierId, false );
```
harriswong, line 80: Is the show/hide for dialog a state only within `CarrierAccountListItem`? Curious why global state store (redux) is preferred over using local state?

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/settings.js
```diff
@@ -0,0 +1,354 @@
...
+	const submitCarrierSettingsHandler = () =>
+		props.submitCarrierSettings( siteId, carrier );
+	const showCancelDialogHandler = () => {
+		props.setVisibilityCancelConnectionDialog( siteId, carrier, true );
+	}
+	const hideCancelDialogHandler = () => {
+		props.setVisibilityCancelConnectionDialog( siteId, carrier, false );
+	}
+	const updateShowUPSInvoiceFields = () => {
+		props.toggleShowUPSInvoiceFields( siteId, carrier );
```
harriswong, line 68: Show/hide UPS forms seems to be inside only `CarrierAccountSettings` component, also curious why redux is preferred over local state? I always go with local state before putting things in redux, so I am actually interested to know if there are something I am not seeing.

i18n/languages/woocommerce-services.pot
```diff
@@ -4,1124 +4,13 @@ msgid ""
...
-"POT-Creation-Date: 2020-03-31T20:32:49.044Z\n"
+"POT-Creation-Date: 2020-06-15T16:43:59.511Z\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "PO-Revision-Date: 2014-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
 
-#: dist/woocommerce-services-1.23.0.js:107
```
harriswong, line 15: @Ferdev do you know why the i18n strings are all removed? Is this supposed to only show up in release zips but somehow they were added to `develop`?

elizaan36:
> When I clicked "cancel", a modal popped up. Then I wanted to click "yes, cancel this" but the only button was "cancel". So i thought the "cancel" here means to "cancel my cancel-request", the exact function as clicking "x".

> @Ferdev @elizaan36 Would it be better if we change the wording to "Yes" or something like that?

Following up here too after our call. We can use a primary action for "Ok" and a secondary action for "Cancel" in this modal.

> @Ferdev @elizaan36 Should we add a hint to date format? ie. 2020-01-02, 2020/01/02, etc? Or is every carrier different? I use dashes on mine.

Good call on this too @harriswong - we'll use the ghosted date format in the form field "YYYY-MM-DD".

Also, we decided to keep the form field errors as-is, but I'll have a think about what might make it easier to know which fields are required as a future enhancement.

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/settings.js
Ferdev: Apparently we don't need to include that checkbox anymore, so I'm gonna remove it. See: p1592578267297800-slack-fulfill

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/settings.js
Ferdev: I don't think so... I'll create one if we don't have it.

Ferdev:
> "connect" button remained disabled when i filled out all information. It also flagged website URLas invalid.

I'll check the error with the URL flagged as invalid.

Ferdev:
> I managed to get the button to show by using `example.com` instead of `http://example.com`. And by filling out all fields. But when I click on "connect", nothing is happening. There is no XHR requests made and no console.log error.
> Kapture 2020-07-08 at 11 22 05

Sorry, I thought I mentioned that in the PR's description. Yeah, that functionality is in another PR, #2063.

client/extensions/woocommerce/app/settings/shipping/index.js
Ferdev: Spanish lessons right there! :sweat_smile:

client/extensions/woocommerce/woocommerce-services/lib/initialize-carrier-accounts-state.js
Ferdev: Yep, already did that

client/extensions/woocommerce/woocommerce-services/state/carrier-accounts/reducer.js
Ferdev: Hmm, yeah your right, that action and variable names are not right. I'll change them to make them carrier agnostic. I wanted to make this code work for any carrier we may support in the future.

client/extensions/woocommerce/woocommerce-services/lib/initialize-carrier-accounts-state.js
Ferdev: This is only for the initial state. We're pre-setting the country to US, so we don't need to validate that field from the start. But if the field is updated by the user, `ignoreValidation.country` will be set to `false` again.

client/extensions/woocommerce/woocommerce-services/state/carrier-accounts/selectors.js
Ferdev: hey @harriswong why do you say validation is ignored except for the country? It's actually the other way around, at first validation is only ignored for country (because it's the only field starting with value)

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/index.js
Ferdev: Yeah, sorry I should've mentioned it in the PR's description.

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/list-item.js
Ferdev: this is probably a personal preference due to my Elm background. The global store is actually contextualized per view/component in the plugin itself: other views like the shipping label is using the global store to persist some views' state. I've had bad experiences in the past using the global redux store + components' state, and also I think the recommended path in the past in the React community was to only use the redux store if you're already using it. But, if you think we should store this state in the component, I'm totally fine with it too!

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/settings.js
Ferdev: See me answer above! :smile:

i18n/languages/woocommerce-services.pot
Ferdev: I'll investigate this!

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/settings.js
harriswong: If we know the link, we can just add it to this PR.

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/list-item.js
harriswong: Ah I see. Thanks!

client/extensions/woocommerce/woocommerce-services/state/carrier-accounts/selectors.js
harriswong: Because `ignoreValidation` is only `false` for `country`
image

harriswong: Oh I see what you mean by the "first validation". On page load, the `fieldErrors` are all `This field is required`:
image

But `ignoreValidations` values are all set to `true` except `country` so that all the `fieldsErrors` are removed on page load, such that there is no "This field is required" showing initially?

Shouldn't we show `This field is required` initially because they are required?

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/settings.js
Ferdev: It's probably this one? https://www.easypost.com/ups-guide Not 100% sure, @c-shultz do you know?

client/extensions/woocommerce/woocommerce-services/state/carrier-accounts/selectors.js
Ferdev: > Shouldn't we show This field is required initially because they are required?
@elizaan36 , correct me if I'm wrong, but I think the idea was to only display the required field error message after the user updates the field and leaves it blank. The optional fields have an "optional" keyword in their label, so the user is supposed to fill all the other ones. That's what I understood from the prototype.

elizaan36: > only display the required field error message after the user updates the field and leaves it blank.

That's correct, @Ferdev !

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/settings.js
c-shultz: I think it was going to be a doc page. I'll double-check on the issue: https://github.com/Automattic/woocommerce-services/issues/2065

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/index.js
harriswong: Preferably to leave the `credentials` blank than to have a value.

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/settings.js
c-shultz: @Ferdev 
Let's just remove that `Learn more...` sentence and the link, per [this comment](https://github.com/Automattic/woocommerce-services/issues/2065#issue-656506211).

We'll re-add it once we have a document created.

client/extensions/woocommerce/woocommerce-services/state/carrier-accounts/selectors.js
elizaan36: Noticed some things as I was testing the error state messaging that would help people complete the form easier, somewhat along the lines of what @harriswong suggested except only after the user has interacted with the field. We want to avoid the possibility of someone opening the form full of errors before they've started.

For the UPS account number error message (the first field), if the user enters less than 6 digits in the field the error message could say "Your UPS account number needs to be 6 digits in length" or something like that. Currently "This field is required" doesn't provide enough information in case they've entered something incorrectly.

For the Zip code and phone number fields, we could also give them a "The zip code format is not valid" error message for anything that's not 5 digits, and "The phone number format is not valid" for anything less than 10 digits.

Ferdev:
> Following up here too after our call. We can use a primary action for "Ok" and a secondary action for "Cancel" in this modal.

Does this look good, @elizaan36 ? 

iamge

elizaan36:
Yep, looks good just a few things.
- We should be consistent with the cancel action across the two modals. Here's the other one:
image

- The WordPress styles are sneaking into the modals and we should have a consistent experience throughout the flow (until we switch over more holistically.) Let's make the Cancel action for both modals the secondary button style from the Calypso components.
- The height of both modal windows could be reduced so the text is closer to the actions. Let's make `.carrier-accounts__settings-cancel-dialog .dialog__content` height: 200px. Does that sound right to you?

Ferdev:
Sounds good @elizaan36! [`94bcd74` (#2049)](https://github.com/Automattic/woocommerce-services/pull/2049/commits/94bcd743ef32cec410863199af12240482d48f6d) takes care of it.

@harriswong I think this PR is ready for a second review. Thanks a ton!

i18n/languages/woocommerce-services.pot
harriswong: I think `i18n/` is under `.gitignore`, and our `prerelease` npm script will regenerate them. I think these are okay to be removed in `develop`.

harriswong:
👀

i18n/languages/woocommerce-services.pot
Ferdev: Yeah, agree!

Ferdev:
>     * we should remove the `alert` if not needed

Sorry, I got confused with the other PR and forgot to remove it. Now it's fixed.
 
>     * verified the page shows
> 
>     * checked disconnect popup to match design
> 
>     * merge conflicts for language file :(

Fixed!

>     * clicking on `Disconnect` gave me the following error, i am assuming [this PR](https://github.com/Automattic/woocommerce-services/pull/2063) will fix it.

Yeah, you're right!

Issue #2044, opened 2020-06-12
Cannot refund orders


FreshPhil:
**Describe the problem**
On WCS version `1.23.1`, we have had multiple cases where administrators are not able to refund orders. Instead, the page will endlessly load.

In these cases, downgrading to `1.23.0` resolved the problem, so this may be linked to the changes made for #2039 as we've noticed this error in the browser console:

![](https://d.pr/i/OAv0cv+)
Image Link: https://d.pr/i/OAv0cv

**Steps to replicate**
- install and activate WooCommerce Services v `1.23.1`
- place an order with automated taxes enabled
- try to refund the order while WooCommerce Services is still active
- Note the order will not refund

Next
- Deactivate and remove WooCommerce Services
- Reinstall version `1.23.0`
- Try to refund and note that it goes through normally

FreshPhil:
The current workaround is for users to remove the WCS plugin and then reinstall using the previous version. Since this is an open repo, they would have access to doing that.

c-shultz:
Fixed in 1.23.2.

PR #2041, opened 2020-06-05
Fix tax recalculation on admin order screen missing street address.

c-shultz:
Fixes #2039 

To test:
- Follow replication steps on https://github.com/Automattic/woocommerce-services/issues/2039
   - With `22325 W 47th St, Shawnee, Kansas, US 66226` as the origin and destination address, the tax rate should be 9.6%
- Confirm that `woocommerce-services-admin-order-taxjar` javascript file loads on new order page and edit order page
- Confirm that `Recalculate` works correctly for new admin order and an existing admin order

PR #2053, opened 2020-06-16
Revert "Fix tax recalculation on admin order screen missing street address."

c-shultz:
Reverts Automattic/woocommerce-services#2041

Created a very critical bug that disabled order refunds: #2044. This is already reverted from `master` on #2047

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,8 +1,5 @@
 *** WooCommerce Services Changelog ***
 
-= 1.23.1 - 2020-xx-xx =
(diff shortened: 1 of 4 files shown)


harriswong:
Should we merge master into develop instead? So that the changelog is kept.

c-shultz:
Closing because @harriswong  has a better idea :)

c-shultz:
@harriswong, I merged master into develop as suggested. Thanks!

Issue #2074, opened 2020-07-17
Unit tests failing


c-shultz:
We have a number of JS unit tests failing in the latest `develop`. Running `npm run test-client`, I see:
<img width="863" alt="image" src="https://user-images.githubusercontent.com/6209518/87831278-f1943580-c837-11ea-8b71-21eb123c678d.png">

PR #2079, opened 2020-07-20
Improve service listing readability

elizaan36:
This is a style sheet-only fix to improve readability of the service listings in the shipping label modal, mainly tweaking the margins within the flex container. These changes will be especially noticable at smaller screen sizes. 
- Made the carrier logos consistent in size (they were decreasing in size with each row)
- Adjusted margins in the flex container for better readability (added some white space between the divs)
- Right aligned the detail text for time frame so it doesn't run into the services detail text

I tweaked a few things to get everything looking right so please have a look and let me know if this all makes sense!


harriswong:
👀

harriswong:
Unit test failed but is being addressed in https://github.com/Automattic/woocommerce-services/issues/2074. scss changes should not block this from merging. Merging.

PR #2084, opened 2020-07-21
Fix build warning due to missing package

### LINKED ISSUE #2071: Interpolate Element Warning on Build
I'm seeing this from the latest `develop` after running `npm run start`:
image

Ferdev:
`@wordpress/components` `v8.5.0` was missing the dependency
`@wordpress/element` `v2.10.0`

----

Closes #2071

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -22,6 +22,7 @@
       "dependencies": {
         "i18n-calypso": {
           "version": "file:wp-calypso/packages/i18n-calypso",
+          "bundled": true,
           "requires": {
             "commander": "^2.9.0",
             "debug": "^3.2.6",
@@ -38,13 +39,11 @@
           "dependencies": {
             "@babel/parser": {
               "version": "7.6.4",
-              "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.6.4.tgz",
-              "integrity": "sha512-D8RHPW5qd0Vbyo3qb+YjO5nvUVRTXFLQ/FsDxJU2Nqz4uB5EnUN0ZQSEYpvTIbRuttig1XbHWU5oMeQwQSAA+A=="
+              "bundled": true
             },
             "@mrmlnc/readdir-enhanced": {
               "version": "2.2.1",
-              "resolved": "https://registry.npmjs.org/@mrmlnc/readdir-enhanced/-/readdir-enhanced-2.2.1.tgz",
-              "integrity": "sha512-bPHp6Ji8b41szTOcaP63VlnbbO5Ny6dwAATtY6JTjh5N2OLrb5Qk/Th5cRkRQhkWCt+EJsYrNB0MiL+Gpn6e3g==",
+              "bundled": true,
               "requires": {
                 "call-me-maybe": "^1.0.1",
                 "glob-to-regexp": "^0.3.0"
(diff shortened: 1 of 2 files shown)


harriswong:
👀

harriswong:
Why would `@wordpress/element` not a dependency when installing `@wordpress/components`? I would imagine installing `@wordpress/components` would automatically install `@wordpress/element` if it is a dependency.

harriswong:
https://github.com/WordPress/gutenberg/blob/master/packages/components/package.json#L38
Looks like element is a dependency already. mm..

harriswong:
Adding the `2.10.0` dependency manually addressed the problem but I found this strange, because `npm` should figure out the dependency instead of us specifying. So I wanted to take deeper look. I switched back to `develop`, I removed `npm-shrinkwrap.json` and `package-lock.json`. I did a `rm -Rf node_modules`, then `npm i`. I just wanted to see what it installs.

This warning is thrown with our `8.5.0` @wordpress/components` dependency:
```
WARNING in ./node_modules/@wordpress/components/build-module/text-highlight/index.js 24:9-47
"export '__experimentalCreateInterpolateElement' was not found in '@wordpress/element'
 @ ./node_modules/@wordpress/components/build-module/index.js
```

In the `./node_modules/@wordpress/components/build-module/text-highlight/index.js`, I see this line `
import { __experimentalCreateInterpolateElement } from '@wordpress/element';`.

It is relying on `__experimentalCreateInterpolateElement` from the `@wordpress/element`. I went into my `node_modules/@wordpress/element` which exists, I check its version number inside `woocommerce-services/node_modules/@wordpress/element/package.json`. `"version": "2.16.0"`. Which is different than the `2.10.0` we wanted to add in the PR.

So I did a `grep` in `woocommerce-services/node_modules/@wordpress/element`, I see this
```
t' .
./CHANGELOG.md:-   Graduated `__experimentalCreateInterpolateElement` function to stable api: `createInterpolateElement` (see [20699](https://github.com/WordPress/gutenberg/pull/20699))
./CHANGELOG.md:-   Added `__experimentalCreateInterpolateElement` function (see [17376](https://github.com/WordPress/gutenberg/pull/17376))
```

I took a look at `CHANGELOG.md`, and I found this:
```
## 2.13.0 (2020-04-15)

### New Features

-   Include TypeScript type declarations ([#21248](https://github.com/WordPress/gutenberg/pull/21248))
-   Graduated `__experimentalCreateInterpolateElement` function to stable api: `createInterpolateElement` (see [20699](https://github.com/WordPress/gutenberg/pull/20699))

## 2.10.0 (2019-12-19)

### New Features

-   Added `__experimentalCreateInterpolateElement` function (see [17376](https://github.com/WordPress/gutenberg/pull/17376))
```

It looks like the function `__experimentalCreateInterpolateElement` is removed in version `2.13.0`. And since `npm install` installed `@wordpress/element` at `2.16.0`. Since our `8.5.0` `@wordpress/components/build-module/text-highlight/index.js` uses `__experimentalCreateInterpolateElement`, and we installed `2.16.0` @wordpress/elements`, the function is missing. 

**Why does `@wordpress/components` `8.5.0` not install @wordpress/elements` `2.10.0`?**
I went into `woocommerce-services/node_modules/@wordpress/components/package.json` to take a look. I found this line:
```
 "@wordpress/element": "^2.10.0",
```
Because of the caret `^`, it will allow installation from `2.10.0` to `3.0.0`. 

I took another look in the shrinkwrap file generated in this PR, as well as my `package-lock.json` generated during my `npm install` test, I see that `@wordpress/babel-preset-default` somehow requires a higher version of `@wordpress/element`
image

I think that's the reason why our `@wordpress/element` got bumped to `2.16.0`. And since `@wordpress/components` allows `^2.10.0`, it never flagged it as a problem even though `__experimentalCreateInterpolateElement` is deprecated after `2.13.0`, and components' `8.5.0` will break.

### Propose Solution
Instead of us installing `@wordpress/element` at `2.10.0`, which conflicts `"@wordpress/babel-preset-default"`'s ``@wordpress/element ^2.16.0` dependency, should we try upgrading `@wordpress/component` to a newer version that is at least 2.16.0? 

I checked `master` `wordpress/component` (v10.0.2)? https://github.com/WordPress/gutenberg/blob/master/packages/components/src/text-highlight/index.js, it no longer uses `__experimentalCreateInterpolateElement`. Maybe bumping the component version would be better option.

Ferdev:
OMG, that was an awesome post-mortem! Yeah, I exactly arrived at the same conclusion, but I didn't want to update the `@wordpress/components` package because that could lead to a lot of unexpected problems in the rest of the code (like, needing to bump the react version as well. I think it would be better to address that on a separate issue, or even, a spike. I'll add a note to myself to discuss it in the next weekly meetup.

harriswong:
Ok, I confirmed that this way we have a "global" `@wordpress/elements` at `2.10.0` but `@wordpress/babel-preset-default` still has its `2.16.0` installed.

```
node_modules/@wordpress/element/package.json // 2.10.0
node_modules/@wordpress/babel-preset-default/node_modules/@wordpress/element //v2.16.0
```

Adding `@wordpress/element v2.10.0` shouldn't break `@wordpress/babel-preset-default`. 👍

PR #2086, opened 2020-07-22
Load locale packages with lowercase file names

harriswong:
import `moment/locale/en-ca.js` instead of `moment/locale/en-CA.js` since moment locale files are in lowercase. https://github.com/moment/moment/tree/develop/locale

### How to test
1. Go to general settings and switch language to Spanish. https://localhost/wp-admin/options-general.php
2. Run `npm run i18n` to download translation files
3. Go to WCS order page and click refund popup modal
4. Verify the text are in Spanish.

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -29,7 +29,7 @@ __webpack_public_path__ = global.wcsPluginData.assetPath;
 // First we try language code with region if it's different then fall back to language code only.
 if ( window.i18nLocale && window.i18nLocale.localeSlug !== 'en' ) {
 	const localeSlugParts = window.i18nLocale.localeSlug.split('-');
-	let localeFileSlug = window.i18nLocale.localeSlug;
+	let localeFileSlug = window.i18nLocale.localeSlug.toLowerCase();
 	if ( localeSlugParts[0] === localeSlugParts[1] ) {
 		localeFileSlug = localeSlugParts[0];
 	}
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-refund-modal.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-refund-modal.js
@@ -9,7 +9,6 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { localize } from 'i18n-calypso';
 import formatCurrency from '@automattic/format-currency';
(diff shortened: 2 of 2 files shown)


harriswong:
Good catch! You are right, both `en-us` and `en` doesn't exist inside moment's locale folder. Import won't work.
```
.../wordpress/wp-content/plugins/woocommerce-services/node_modules/moment/locale (fix/2073-locale-missing-error)$ ls -al | grep en
-rw-r--r--    1 harriswong  staff  2625 26 Oct  1985 en-au.js
-rw-r--r--    1 harriswong  staff  2458 26 Oct  1985 en-ca.js
-rw-r--r--    1 harriswong  staff  2631 26 Oct  1985 en-gb.js
-rw-r--r--    1 harriswong  staff  2629 26 Oct  1985 en-ie.js
-rw-r--r--    1 harriswong  staff  2451 26 Oct  1985 en-il.js
-rw-r--r--    1 harriswong  staff  2625 26 Oct  1985 en-in.js
-rw-r--r--    1 harriswong  staff  2634 26 Oct  1985 en-nz.js
-rw-r--r--    1 harriswong  staff  2643 26 Oct  1985 en-sg.js
```

harriswong:
moment [by default](https://momentjs.com/docs/#/i18n/changing-locale/) includes English(US) so we can safely skip that if it's the system default.

harriswong:
@Ferdev please take a look again!

client/main.js
```diff
@@ -27,9 +27,9 @@ __webpack_public_path__ = global.wcsPluginData.assetPath;
 
 // We need to lazy load the moment locale files.
 // First we try language code with region if it's different then fall back to language code only.
-if ( window.i18nLocale && window.i18nLocale.localeSlug !== 'en' ) {
+if ( window.i18nLocale && window.i18nLocale.localeSlug !== 'en-US' && window.i18nLocale.localeSlug !== 'en' ) {
```
Ferdev, line 30: What about something like:

```if ( window.i18nLocale && ! [ 'en-US', 'en' ].includes( window.i18nLocale.localeSlug )``` 

?

IMHO is less verbose and easier to read.

harriswong: I like it!

harriswong: Updated!

PR #2094, opened 2020-07-23
Fix broken carrier accounts unit tests

### LINKED ISSUE #2074: Unit tests failing

Ferdev:
Closes #2074


harriswong:
👀

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/index.js
```diff
@@ -69,7 +66,7 @@ CarrierAccounts.propTypes = {
 	siteId: PropTypes.number.isRequired,
 	carriers: PropTypes.arrayOf(
 		PropTypes.shape( {
-			id: PropTypes.string.isRequired,
+			id: PropTypes.string,
```
harriswong, line 69: How come `id` is no longer required?

Ferdev: It's because the only property always consumed by the frontend is the carrier name, which is called `carrier` and it's the only one required right now.

Ferdev:
> I got the following when i run `npm run test`, I understand this is outside of fixing unit tests, we can create a separate issue if it's too much effort to address.

I'm gonna fix all the linting issues related to carrier accounts, just to keep all the changes in the current scope, but I'll leave the others for another issue.

Ferdev:
I've fixed all carrier account related linting issues that can be fixed. There're a few of them which are false positives due to a bug in eslint. I've created issue #2097 to fix all these problems.

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/index.js
harriswong: Oh I see. Thanks!

Issue #1973, opened 2020-03-26
npm run eslint throws error in latest develop


harriswong:
Run `npm run eslint` and fix all the errors. 

For example:
```
woocommerce-services/client/apps/shipping-label/test/view-wrapper-label.js
  16:8  error  'Button' is defined but never used  no-unused-vars

woocommerce-services/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/test/sidebar.js
   8:21  error  'mount' is defined but never used                             no-unused-vars
  79:20  error  'props' is assigned a value but never used                    no-unused-vars
  91:20  error  'props' is assigned a value but never used                    no-unused-vars
  92:9   error  'renderedCheckboxControl' is assigned a value but never used  no-unused-vars
```

I'm also seeing this in the `console.log`
![image](https://user-images.githubusercontent.com/572862/77691612-dc368b00-6f6a-11ea-8505-b195c9504251.png)

harriswong:
Block until `1.24.0` is released.

PR #2108, opened 2020-07-29
Fix eslint errors and install missing packages

harriswong:
Closes https://github.com/Automattic/woocommerce-services/issues/1973
Fix eslint errors so that pipeline can pass. Also added missing packages for tasks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,9 @@
     "enzyme-adapter-react-16": "1.12.1",
     "enzyme-to-json": "3.3.5",
     "file-loader": "^4.3.0",
+    "glob": "^7.1.6",
     "globby": "^6.1.0",
+    "i18n-calypso-cli": "^1.0.0",
     "jest": "^24.9.0",
     "jest-puppeteer": "^4.4.0",
     "moment": "^2.25.3",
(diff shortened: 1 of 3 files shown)


tasks/i18n-download.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console, import/no-nodejs-modules */
```
harriswong, line 1: ignoring these 2 rules the same way we ignore it in `tasks/i18njson.js`

package.json
```diff
@@ -53,12 +55,12 @@
...
     "postcss-safe-parser": "^4.0.2",
     "puppeteer": "^1.20.0",
+    "request": "^2.88.2",
     "shelljs": "^0.7.6",
     "sinon": "7.3.1",
     "sinon-chai": "3.3.0",
     "webpack-bundle-analyzer": "^3.6.1",
-    "xgettext-js": "^1.1.0",
-    "i18n-calypso-cli": "^1.0.0"
+    "xgettext-js": "^1.1.0"
```
harriswong, line 63: `glob` and `request` are missing as devDependency:
```
woocommerce-services/tasks/i18n-download.js
  3:17  error  'request' should be listed in the project's dependencies. Run 'npm i -S request' to add it  import/no-extraneous-dependencies

woocommerce-services/tasks/i18n.js
  3:14  error  'glob' should be listed in the project's dependencies. Run 'npm i -S glob' to add it  import/no-extraneous-dependencies
```

PR #2111, opened 2020-07-30
Update order detail note for UPS

### LINKED ISSUE #2089: Order notes and UPS
The order notes have the tracking number but don't mention it was shipped by UPS. 

Screen Shot 2020-07-22 at 10 29 11 PM

There is a similar issue on the email sent notifying that a note has added. Which I only start seeing 3 days ago, is this something new? This email goes out when the user sends an additional label ("Create new label"). Didn't we use to send the same email as when the order is marked completed when the first label is printed?  

cc: @elizaan36  

screencapture-mail-google-mail-u-1-2020-07-22-22_31_00

c-shultz:
Fixes #2089 

Purchase new label with this checked:
image

Check order note:
image

Also adds a pickup link for UPS:
image

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
@@ -788,35 +788,42 @@ const handlePrintFinished = ( orderId, siteId, dispatch, getState, hasError, lab
 	if ( shouldEmailDetails( getState(), orderId, siteId ) ) {
 		const trackingNumbers = labels.map( label => label.tracking );
 		const carrierId = first( labels ).carrier_id;
+		let carrierIdReadable = '';
 		let note = '';
-		if ( 'usps' === carrierId ) {
-			note = translate(
-				'Your order has been shipped with USPS. The tracking number is %(trackingNumbers)s.',
-				'Your order consisting of %(packageNum)d packages has been shipped with USPS. ' +
-					'The tracking numbers are %(trackingNumbers)s.',
-				{
-					args: {
-						packageNum: trackingNumbers.length,
-						trackingNumbers: trackingNumbers.join( ', ' ),
-					},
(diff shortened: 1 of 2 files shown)


harriswong:
👀

c-shultz:
🤦 Good catch

client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
```diff
@@ -788,35 +788,42 @@ const handlePrintFinished = ( orderId, siteId, dispatch, getState, hasError, lab
 	if ( shouldEmailDetails( getState(), orderId, siteId ) ) {
 		const trackingNumbers = labels.map( label => label.tracking );
 		const carrierId = first( labels ).carrier_id;
```
harriswong, line 790: Might need to figure out why we use `first(labels)`

c-shultz: Hmm, yeah, it's not handling this well:
image

c-shultz: This will be pretty unusual, so I'm comfortable releasing with this bug. I'll make an issue and we can prioritize it next week

harriswong: I wonder if labels are sorted in descending order based on last process/most updated datetime.

c-shultz: Issue created: https://github.com/Automattic/woocommerce-services/issues/2113

c-shultz: You can have multiple labels in an order, it assumes the carrier of the first label is the carrier for all labels.

Issue #2112, opened 2020-07-30
Fix widths/layout in Carrier card and Cancel modal


elizaan36:
I had a look at the tweaking we'd need to do in a few places to fix alignment/visibility of text and other elements.

1. **Settings > WooCommerce Services > UPS Disconnect**

In the `.carrier-accounts__settings-cancel-dialog`, we can remove the `width: 666px` and instead use a variable width to ensure visibility of text and exit icon on smaller screen sizes. Here's what it currently looks like where the text and close icon are hidden.

<img width="417" alt="Screen Shot 2020-07-30 at 12 11 29 PM" src="https://user-images.githubusercontent.com/5121465/88954654-53a16180-d260-11ea-8c5d-4adb59e61d9c.png">
<img width="417" alt="Screen Shot 2020-07-29 at 11 20 59 AM" src="https://user-images.githubusercontent.com/5121465/88954667-5734e880-d260-11ea-9b63-14b9288e1ffc.png">

Also in the modal, the header is slighly too high and not horizontally aligned with the exit icon. I played around the padding and found that this combination works to horizontally align the header with the exit icon: `padding: 22px 28px 18px 28px`
<img width="587" alt="image" src="https://user-images.githubusercontent.com/5121465/88955451-8ef06000-d261-11ea-925b-2de7b30915bc.png">

2. In the Carrier account card in **Settings > WooCommerce Services**, we can tweak the widths of the Name and Credentials elements to align with the UPS list data in the card. It may be a matter of reworking those divs to match the percentages for each row.

<img width="412" alt="Screen Shot 2020-07-29 at 11 24 54 AM" src="https://user-images.githubusercontent.com/5121465/88955083-0c67a080-d261-11ea-81af-9aef2c925cbc.png">

PR #2121, opened 2020-08-06
fix: carrier disconnect modal layout

frosso:
## Description
Changing the "Disconnect Carrier" modal layout to fix the mobile behavior

## Issue

Fixes https://github.com/Automattic/woocommerce-services/issues/2112 (ticket also references grid layout - done separately in https://github.com/Automattic/woocommerce-services/issues/2122)

## Steps to reproduce
- Ensure you already connected "UPS"
- Go to **Settings > WooCommerce Services**
- Click  "Disconnect" on "UPS"
- Go to mobile viewport
- Modal layout behavior is fixed:
  - Text is wrapping
  - Title is centered
Screenshot on 2020-08-06 at 15-11-37

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -14,6 +14,7 @@
 * Fix   - Fix localization issues
 * Tweak - Improve service listing readability
 * Add   - Support UPS as a carrier (beta)
+* Fix   - Carrier "disconnect modal" layout
 
 = 1.23.2 - 2020-06-12 =
 * Fix   - Refund not possible on order page.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Services Changelog ***
 
+= 1.24.1 - 2020-xx-xx =
+* Fix   - Carrier "disconnect modal" layout
+
 = 1.24.0 - 2020-xx-xx =
 * Fix   - PHP 7.4 notice for taxes at checkout.
 * Add   - Carrier logos next to rates.
@@ -14,7 +17,6 @@
 * Fix   - Fix localization issues
 * Tweak - Improve service listing readability
 * Add   - Support UPS as a carrier (beta)
-* Fix   - Carrier "disconnect modal" layout
 
 = 1.23.2 - 2020-06-12 =
 * Fix   - Refund not possible on order page.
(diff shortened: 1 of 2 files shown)


client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/dialog-styles.scss
```diff
@@ -1,8 +1,8 @@
 .carrier-accounts__settings-cancel-dialog {
 	.dialog__content {
 		padding: 0;
-		width: 666px;
-		height: 200px;
+		max-width: 666px;
```
frosso, line 4: The ticket stated:
> we can remove the `width: 666px` and instead use a variable width to ensure visibility of text and exit icon on smaller screen sizes

However By doing that, the modal would have expanded considerably. I changed to `max-width` so that it doesn't look _as_ wide on lg+ screens

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/dialog-styles.scss
```diff
@@ -1,8 +1,8 @@
...
+		min-height: 200px;
```
frosso, line 5: Same for the height - `min-height` ensures that the modal "expands" naturally

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/dialog-styles.scss
```diff
@@ -11,9 +11,9 @@
 
 			.carrier-accounts__settings-cancel-dialog-title {
 				display: block;
-				margin: 0;
-				padding: 18px 28px;
+				margin: auto 28px;
```
frosso, line 14: The title wasn't centered. Rather than hardcoding the top/bottom padding, I changed it to `margin-top`/`margin-bottom`: `auto`, so that it centers naturally.
If people still prefer `padding` to position the title, I could use flexbox instead.

changelog.txt
```diff
@@ -14,6 +14,7 @@
 * Fix   - Fix localization issues
 * Tweak - Improve service listing readability
 * Add   - Support UPS as a carrier (beta)
+* Fix   - Carrier "disconnect modal" layout
```
c-shultz, line 17: Thanks for getting a changelog note added! 1.24.0 is out already (the header date should be set now in latest `develop` now that we've merged back from `master). You can go ahead and start a `1.24.1` header and add your note there.

frosso: Thanks for catching that, it looks like I had to update my local `develop` 👍

Issue #2091, opened 2020-07-23
"Postal code" vs "ZIP code" vs "Postcode"


aleftick:
We use different terminology in different places (not just in WCShip but also in Core). It would be good to be consistent and have a global mindset. 

I did some research and it wasn't very conclusive; I haven't found a standard approach. However, the best I have seen has been "ZIP/Postal code".

<img width="1553" alt="Screen Shot 2020-07-22 at 11 07 13 PM" src="https://user-images.githubusercontent.com/2545592/88248751-3bce4a00-cc70-11ea-872d-03fcfc41053c.png">

c-shultz:
>However, the best I have seen has been "ZIP/Postal code".

Ok cool, let's go with that for both the strings there.

PR #2123, opened 2020-08-06
chore: ZIP/postcode/postal code messaging consistency

frosso:
## Description

Bringing more consistency on the "ZIP code"/"Postal Code" messaging

## Fixes

Fixes https://github.com/Automattic/woocommerce-services/issues/2091

## Steps to reproduce

Carrier connection:
Screenshot on 2020-08-06 at 16-04-45

Address selection (destination/source):
Annotation on 2020-08-06 at 16-31-52

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -315,15 +315,15 @@ class WC_Connect_TaxJar_Integration {
 		$invalid_postcode = false !== strpos( $formatted_message, 'isn\'t a valid postal code for' );
 		if ( ! is_admin() && ( $state_zip_mismatch || $invalid_postcode ) ) {
 			$fields = WC()->countries->get_address_fields();
-			$postcode_field_name = __( 'Postcode / ZIP', 'woocommerce-services' );
+			$postcode_field_name = __( 'ZIP / Postal code', 'woocommerce-services' );
 			if ( isset( $fields['billing_postcode'] ) && isset( $fields['billing_postcode']['label'] ) ) {
 				$postcode_field_name = $fields['billing_postcode']['label'];
 			}
 
 			if ( $state_zip_mismatch ) {
-				$message = sprintf( _x( '%s does not match the selected state.', '%s - Postcode/Zip checkout field label', 'woocommerce-services' ), $postcode_field_name );
+				$message = sprintf( _x( '%s does not match the selected state.', '%s - ZIP/Postal code checkout field label', 'woocommerce-services' ), $postcode_field_name );
 			} else {
-				$message = sprintf( _x( 'Invalid %s entered.', '%s - Postcode/Zip checkout field label', 'woocommerce-services' ), $postcode_field_name );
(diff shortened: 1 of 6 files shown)


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -315,15 +315,15 @@ public function _error( $message ) {
 		$invalid_postcode = false !== strpos( $formatted_message, 'isn\'t a valid postal code for' );
 		if ( ! is_admin() && ( $state_zip_mismatch || $invalid_postcode ) ) {
 			$fields = WC()->countries->get_address_fields();
-			$postcode_field_name = __( 'Postcode / ZIP', 'woocommerce-services' );
```
frosso, line 318: This I wasn't sure if I had to modify as well 😬  Should I revert the changes on this file?

bborman22: I think this makes sense to update as well. Only thing I noticed was this change has spaces around the slash while the other instances does not have spaces.

bborman22:
This looked good to me and I confirmed all the labels and messages were updated as expected to ZIP/Postal code. One thing I wasn't sure of is the translation files for WCS. I noticed there was a [generated translation files](https://github.com/Automattic/woocommerce-services/tree/develop/i18n) that only has the Postal code msgid, so that may need to be re-generated to pick up the ZIP/Postal code msgid. Maybe @harriswong or @Ferdev have more information on the translation set up for WCS and could weigh in on any translation impact of this update?

classes/class-wc-connect-taxjar-integration.php
frosso: Yeah, the existing change had spaces around the slash - do you think this should be converted to _not_ have spaces?

bborman22: Yeah I think for consistency it shouldn't have spaces. Given the overall goal of the update, I'd put more weight on matching the new messaging over matching any previous messaging.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -315,15 +315,15 @@ public function _error( $message ) {
...
+			$postcode_field_name = __( 'ZIP / Postal code', 'woocommerce-services' );
```
frosso, line 318:
```suggestion
			$postcode_field_name = __( 'ZIP/Postal code', 'woocommerce-services' );
```

classes/class-wc-connect-taxjar-integration.php
frosso: Sure thing, that makes sense 👍  Done in 401a709

harriswong:
> I noticed there was a generated translation files that only has the Postal code msgid, so that may need to be re-generated to pick up the ZIP/Postal code msgid. Maybe @harriswong or @Ferdev have more information on the translation set up for WCS and could weigh in on any translation impact of this update?

Good question. We usually update our translation files per release through a script. Recently, we added a new script to generate these files: https://github.com/Automattic/woocommerce-services/blob/develop/package.json#L13.

harriswong:
This process may change, but as far as I know, we don't need to regenerate language file per PR. Using the `__()` and `translate()` is good.

PR #2125, opened 2020-08-07
Added array to get shipping labels API to include product ids associated with a shipping label

### LINKED ISSUE #2102: Include product_id in the GET shipping labels API response
In the Woo app, we display the image, quantity, price and name of the products associated with a shipping label. In the API response, we only get the product names for each shipping label. While this is working fine for physical products, when it comes to variable products, the product name and SKU are the same, so we are unable to identify the correct variable product associated with that shipping label. 

We need to include the productID (or the variationID, for a variable product) in the API response, so that we can correctly identify the product. The ID would be unique enough so we won’t run into this issue.

anitaa1990:
Fixes #2102 by including a new array to the GET shipping labels API. This array includes `product_ids` associated with a shipping label. 

#### To test
- Create a shipping label for an order with single or multiple SIMPLE products.
  - Verify that the GET shipping labels API returns the `product_ids` that are associated with the shipping label in the `product_ids` array.
- Create a shipping label for an order with single or multiple VARIABLE products.
  - Verify that the GET shipping labels API returns the `variable_ids` that are associated with the shipping label in the `product_ids` array.
- Smoke testing the API to ensure that existing functionality is not broken :) 

Note: I didn't find any unit tests to test this specific scenario but it's possible I missed something! Do let me know if I need to add unit tests to the PR before review 😄

### DIFF
--- a/classes/class-wc-rest-connect-shipping-label-controller.php
+++ b/classes/class-wc-rest-connect-shipping-label-controller.php
@@ -106,8 +106,10 @@ class WC_REST_Connect_Shipping_Label_Controller extends WC_REST_Connect_Base_Con
 			}
 
 			$product_names = array();
+			$product_ids = array();
 			foreach ( $package[ 'products' ] as $product_id ) {
 				$product = wc_get_product( $product_id );
+				$product_ids[] = $product_id;
 
 				if ( $product ) {
 					$product_names[] = $product->get_title();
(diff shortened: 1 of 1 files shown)


anitaa1990:
Thank you for the review @frosso 🙇‍♀️ Merging the PR now

Issue #2113, opened 2020-07-30
Multi-carrier label orders create an incorrect note


c-shultz:
A label order with both USPS and UPS results in an incorrect order note that says both have been shipped with USPS:
<img width="380" alt="image" src="https://user-images.githubusercontent.com/6209518/88956604-73318d80-d252-11ea-81e5-544613247de8.png">

To reproduce:
Create label order with multiple items and packages:
<img width="1321" alt="image" src="https://user-images.githubusercontent.com/6209518/88956738-a4aa5900-d252-11ea-8df8-851929caec36.png">

Choose USPS for one package:
<img width="632" alt="image" src="https://user-images.githubusercontent.com/6209518/88956764-b3910b80-d252-11ea-84e3-1caf7484ad25.png">

And UPS for for another:
<img width="688" alt="image" src="https://user-images.githubusercontent.com/6209518/88956791-bc81dd00-d252-11ea-87ef-93b04b0bcea6.png">

Print label with this checked:
![image](https://user-images.githubusercontent.com/6209518/88956626-7b89c880-d252-11ea-8541-d1634dd8bdb6.png)

c-shultz:
@elizaan36, this is a bit of a corner case for a label order with multiple packages where one is USPS and one is UPS.

Since this will be a bit unusual, I'm thinking we should come back around and fix this in a patch.

What do you think on wording? Maybe --

>Your order consisting of 2 packages has been shipped. The tracking numbers are 9449000205309044702569 (USPS), 1Z1261E70332503088 (UPS).

elizaan36:
> Your order consisting of 2 packages has been shipped. The tracking numbers are 9449000205309044702569 (USPS), 1Z1261E70332503088 (UPS).

Sounds perfect👩🏻‍🍳

PR #2126, opened 2020-08-07
refactor: multi-carrier label orders note

frosso:
## Description

Changing the message on the created note when an order is shipped with multiple carriers.
This new messaging simplifies the logic a bit.

I didn't find existing unit tests for this logic in the codebase 👀 

Screen Shot 2020-08-07 at 1 57 00 PM

## Fixes

Fixes https://github.com/Automattic/woocommerce-services/issues/2113

## Steps to reproduce
- Create order with multiple items and packages
- Mark order as "Completed"
- Create the labels
- Choose USPS for one package and UPS for another
- Print label with this checked:
image

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
@@ -784,34 +784,34 @@ const handlePrintFinished = ( orderId, siteId, dispatch, getState, hasError, lab
 		return;
 	}
 
-	if (shouldEmailDetails(getState(), orderId, siteId)) {
+	if ( shouldEmailDetails( getState(), orderId, siteId ) ) {
 		dispatch(
-			createNote(siteId, orderId, {
+			createNote( siteId, orderId, {
 				note: translate(
 					'Your order has been shipped. The tracking number is %(trackingNumbers)s.',
 					'Your order consisting of %(packageCount)d packages has been shipped. The tracking numbers are %(trackingNumbers)s.',
 					{
 						args: {
 							packageCount: labels.length,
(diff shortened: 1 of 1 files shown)


harriswong:
👀

harriswong:
@elizaan36 We are modifying the message for 1 package to use the `(USPS)` format as well. Is this ok? Here is a screenshot of the original compares to the 2 new messages:
image

harriswong:
We don't have a test covering `handlePrintFinished` in our [actions.js test](https://github.com/Automattic/woocommerce-services/blob/develop/client/extensions/woocommerce/woocommerce-services/state/shipping-label/test/actions.js). Our coverage on this is actually quite low. It'll be a bit more work but it would be nice to add a test covering this function and the expected label text change.

elizaan36:
> We are modifying the message for 1 package to use the (USPS) format as well.

Hey @harriswong. Yeah this looks good! I like the consistent placement for the carrier in parentheses whether it's one or multiple.

harriswong:
> We don't have a test covering `handlePrintFinished` in our [actions.js test](https://github.com/Automattic/woocommerce-services/blob/develop/client/extensions/woocommerce/woocommerce-services/state/shipping-label/test/actions.js). Our coverage on this is actually quite low. It'll be a bit more work but it would be nice to add a test covering this function and the expected label text change.
> 
> **[edited]**
> @frosso and I had a brief chat on this `handlePrintFinished`. Since it's scoped private, it is a bit complicated to test. We could write a test to its caller but that looks like a scope creep. We will discuss this in team instead of blocking. If needed, we will open another issue for the test(s).

We talked about this at team chat today, we decided to create a follow up to cover the missing test: https://github.com/Automattic/woocommerce-services/issues/2143.

Issue #2116, opened 2020-08-05
Customer selected shipping does not show anymore. 


anant1811:
Customer selected shipping method and cost used to show in previous version of the extension, it doesn't show with 1.24.0.

Example, an order with version 1.23.0:
![https://d.pr/i/NuNN5z](https://d.pr/i/NuNN5z+) 
Full Size: https://d.pr/i/NuNN5z

Same order with version 1.24.0:
![https://d.pr/i/jipuRO](https://d.pr/i/jipuRO+) 
Full Size: https://d.pr/i/jipuRO

Affected ticket:
3191664-zen

fevered:
Had another report of this in 3201074-zen

tijosh:
Another report in 3194617-zen

dphyled:
Another report in 3191736-zen

johndcoy:
Another 10048733-hc

PR #2127, opened 2020-08-07
Transform order object to a presentable API object

harriswong:
Closes https://github.com/Automattic/woocommerce-services/issues/2116

### Description
WCS 1.24.0 [initializes order](https://github.com/Automattic/woocommerce-services/blob/d7f915ed13cb53dea0118325da0d912e93b8d501/classes/class-wc-connect-shipping-label.php#L420) as the initial states of the react app. 

Currently, `$order->get_data()` returns a PHP object of the order. This object is not suitable to output as `json_encode()`, some of its protected fields are not generated, resulting a `{}` at the end. 

We should transform this order object to something that can be represented as a JSON object. This can be done similar to [Woo core's API](https://github.com/woocommerce/woocommerce/blob/cbfe4bd59587aadb4b6c6e95a5787bebe3925310/includes/legacy/api/v3/class-wc-api-orders.php#L142-L320). 

### To test
1. Go to an order page, and click to the step where you can see rates.
2. Verify the "Customer paid..." box shows
image

### DIFF
--- a/classes/class-wc-connect-order-presenter.php
+++ b/classes/class-wc-connect-order-presenter.php
@@ -4,7 +4,7 @@ if ( ! class_exists( 'WC_Connect_Order_Presenter' ) ) {
 
 	class WC_Connect_Order_Presenter {
 
-		public function get_order_for_api( $order ) {
+		public function get_order_for_api( WC_Order $order ) {
 			$dp = 2; //decimal point defaults
 			$order_data = array(
 				'id'                        => $order->get_id(),
--- a/classes/class-wc-connect-order-presenter.php
+++ b/classes/class-wc-connect-order-presenter.php
@@ -4,6 +4,13 @@ if ( ! class_exists( 'WC_Connect_Order_Presenter' ) ) {
 
 	class WC_Connect_Order_Presenter {
 
+		/**
+		 * This function transform the WC_Order object to a representational JSON form for the react app.
+		 * This is based on WooCommerce v3's get_order API woocommerce/includes/legacy/api/v3/class-wc-api-orders.php
+		 *
+		 * @param WC_Order $order
+		 * @return array
+		 */
 		public function get_order_for_api( WC_Order $order ) {
 			$dp = 2; //decimal point defaults
 			$order_data = array(
--- a/classes/class-wc-connect-order-presenter.php
+++ b/classes/class-wc-connect-order-presenter.php
@@ -91,14 +91,6 @@ if ( ! class_exists( 'WC_Connect_Order_Presenter' ) ) {
 					'meta'         => array_values( $item_meta ),
 				);
 
-				if ( in_array( 'products', $expand ) && is_object( $product ) ) {
-					$_product_data = WC()->api->WC_API_Products->get_product( $product->get_id() );
-
-					if ( isset( $_product_data['product'] ) ) {
-						$line_item['product_data'] = $_product_data['product'];
-					}
-				}
-
 				$order_data['line_items'][] = $line_item;
 			}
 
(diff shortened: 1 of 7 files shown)


classes/class-wc-connect-shipping-label.php
```diff
@@ -411,13 +411,14 @@ protected function reducer_items_quantity( $sum, $item ) {
...
+			$connect_order_presenter = new WC_Connect_Order_Presenter();
+
 			$order = wc_get_order( $post );
 			$order_id = WC_Connect_Compatibility::instance()->get_order_id( $order );
 			$items = array_filter( $order->get_items(), array( $this, 'filter_items_needing_shipping' ) );
 			$items_count = array_reduce( $items, array( $this, 'reducer_items_quantity' ), 0 );
-
 			$payload = array(
-				'order'             => $order->get_data(),
+				'order'             => $connect_order_presenter->get_order_for_api( $order ),
```
harriswong, line 421: The data here should match the format in https://github.com/Automattic/woocommerce-services/blob/develop/tests/e2e-tests/fixtures/orders.json

classes/class-wc-connect-order-presenter.php
```diff
@@ -0,0 +1,153 @@
+<?php
+
+if ( ! class_exists( 'WC_Connect_Order_Presenter' ) ) {
+
+	class WC_Connect_Order_Presenter {
```
harriswong, line 5: I am opened to other names. I am not sure what is the best name for this. :D

tests/php/bootstrap.php
```diff
@@ -18,7 +18,16 @@
 	exit( 'Could not determine test root directory. Aborting.' );
 }
 
-$wp_tests_dir = getenv( 'WP_TESTS_DIR' ) ? getenv( 'WP_TESTS_DIR' ) : '/tmp/wordpress-tests-lib';
+$wp_tests_dir = getenv( 'WP_TESTS_DIR' );
+
+if ( ! $wp_tests_dir ) {
+	$wp_tests_dir = rtrim( sys_get_temp_dir(), '/\\' ) . '/wordpress-tests-lib';
```
harriswong, line 24: This is generated by the `wp` CLI. It also fixed the problem on my system where my "temp" folder isn't just `tmp/`, it was like `/private/var/folders/blah/blahblhablhablah/T/blahblahblah....`

tests/php/bootstrap.php
```diff
@@ -42,4 +42,4 @@ function _manually_load_plugin() {
 	define( 'WC_UNIT_TESTING', true );
 }
 
-require $wc_root . '/tests/legacy/bootstrap.php';
```
harriswong: older Woo core version hasn't changed the path yet.

classes/class-wc-connect-order-presenter.php
```diff
@@ -0,0 +1,153 @@
...
+
+		/**
+		 * This function transform the WC_Order object to a representational JSON form for the react app.
+		 * This is based on WooCommerce v3's get_order API woocommerce/includes/legacy/api/v3/class-wc-api-orders.php
+		 *
+		 * @param WC_Order $order
+		 * @return array
+		 */
+		public function get_order_for_api( WC_Order $order ) {
+			$dp = 2; //decimal point defaults
```
frosso, line 15: (minor / ⛏️ ) Instead of `$dp` maybe expand to `$decimal_points`? By expanding the acronym, then you can also save the comment that explains what `$dp` stands for 😄

harriswong: Updated!

harriswong:
Test isn't passing due to the pipeline running test against older version of woocommerce codes. Older woocommerce versions don't seem to support some functions used in this PR. 

Blocked by https://github.com/Automattic/woocommerce-services/issues/2131.

Issue #1971, opened 2020-03-26
Style tweaks 


harriswong:
Couple of style tweaks

1. Remove the gap left of the green checkbox
![image](https://user-images.githubusercontent.com/572862/77688787-3da82b00-6f66-11ea-9f0f-18eb9068f53e.png)

2. The "Notify the customer with shipment details" text should be vertically middle with the checkbox button
![image](https://user-images.githubusercontent.com/572862/77688820-4c8edd80-6f66-11ea-8a17-011160534791.png)

3. The button background colour should be pink instead of blue. Pink to match the "Create shipping label" background
![image](https://user-images.githubusercontent.com/572862/77688863-5ca6bd00-6f66-11ea-8af3-c7078c8e90ab.png)

bborman22:
I addressed number 1 in the #2130 PR and number 3 in the #2129 PR.

For number 2 though, I wasn't able to see the same vertical alignment issue. In fact in the screenshot provided it appears to be misaligned to the top, while what I see looked better, but if anything was misaligned to the bottom. 

![Notify-Customer-option](https://user-images.githubusercontent.com/68524302/89948244-c19c4000-dbf3-11ea-8f2a-677a77531dcb.png)

I had checked this on Chrome 84 and Firefox 79. I was wondering if this alignment issue is still present in the latest develop branch or if there is a particular browser or set up that got that vertical alignment?

harriswong:
@bborman22 I just checked my local, looks like the `Notify the customer with shipment details` text is aligned vertically with the checkbox. 

@elizaan36! I think I created this issue from one of our round tables demos. For the second comment in the issue regarding vertical aligned text, I just want to double check if this looks ok now?

elizaan36:
> regarding vertical aligned text, I just want to double check if this looks ok now?

hey, I just checked on my test site and it looks like it's been fixed. Thanks for remembering these tweaks @harriswong and @bborman22 for taking it on🎉

bborman22:
With #2130 and #2129 merged, this issue can be closed.

PR #2129, opened 2020-08-11
Fix button is-busy color to use pink

bborman22:
**Description**

Changes the color of the is-busy state on the primary buttons to use the appropriate brand color.

**Issue**

Fixes number 3 in #1971.

**Testing**

You should be able to test this fix on any of the primary buttons in the WCS area. The issue called out the button on the Buy Shipping Label button in particular as this button will have typically have the is-busy state the longest.

1. With WC store and WCS set up to print shipping labels, go to an order that is shippable.
2. Click on the Create shipping label button.
3. Provide any necessary information for addresses, packages, and shipping rates to get the Buy shipping label button active.
4. Click on the Buy shipping label button and while the purchase is processed, the button should have the pink colors with animated slashes signifying the is-busy state.

Alternatively, you can inspect one of the WCS primary buttons (any pink button) from the browser Dev Tools and manually add the is-busy class to the button to confirm it uses the brand pink instead of default blue.

Button-is-busy

### DIFF
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -45,14 +45,15 @@
 		//wp-admin styles override the disabled button with !important, so need to do the same here
 		
 		&.is-busy {
+			background-size: 120px 100% !important;
 			background-image: linear-gradient(
 				-45deg,
 				var( --color-accent ) 28%,
 				var( --color-accent-600 ) 28%,
 				var( --color-accent-600 ) 72%,
 				var( --color-accent ) 72%
-			);
-			border-color: var( --color-accent-dark );
+			) !important;
+			border-color: var( --color-accent-dark ) !important;
 		}
 
 		&.is-borderless {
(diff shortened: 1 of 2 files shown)


assets/stylesheets/style.scss
```diff
@@ -43,10 +43,16 @@
...
-			background-size: 120px 100% !important;
-			background-image: linear-gradient( -45deg, var( --color-primary ) 28%, var( --color-primary-400 ) 28%, var( --color-primary-400 ) 72%, var( --color-primary ) 72%) !important;
-			border-color: #0081a9 !important;
+			background-image: linear-gradient(
+				-45deg,
+				var( --color-accent ) 28%,
+				var( --color-accent-600 ) 28%,
+				var( --color-accent-600 ) 72%,
+				var( --color-accent ) 72%
+			);
```
frosso, line 54: The previous code had some `!important`, due to L45.

I checked out this branch, and with the changes as they are, the buttons look gray:




But as soon as the `!important` is re-introduced, then I see the pink color:


I tested this in Chrome & Firefox.

Should the `!important` be re-added?

bborman22: Thanks for this catch! I've added in the `!important`. I noticed the busy animation is broken by a disabled background attribute so created a new issue #2138 for that.

PR #2135, opened 2020-08-13
Remove the slice on the rates errors array

### LINKED ISSUE #2048: rate-step doesn't show errors unless there is more than 1 error in array
**Summary:**
When `packageErrors` is ["Validation failed: Weight must be greater than 0"]. The `.slice(1)` below would always remove the first element. 

https://github.com/Automattic/woocommerce-services/blob/8b58b63d55dd46dd22df7cf50adc0dae4a5a781b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js#L78

We should check why this `slice` logic is there in the first place before changing it.

**Expected behaviour**
If there is 1 error, we should still be able to display it.

**Actual behaviour**
No error will be displayed if there is <= 1 error.

bborman22:
**Description**

There was a JavaScript  array slice on the packageErrors array on the Label Purchase Modal, Rate Step that would remove the first error in the array. This was originally done to remove duplicate errors since errors would be duplicated for each carrier. With a fix for the duplicate errors coming, this slice is no longer needed.

**Steps to Reproduce**

1. Start with a WC store with WCS set up.
2. Go into a shippable order and click Create Shipping Label.
3. Go through the Address verification and package selection.
4. Will need to force an error from the rates request. Which can be challenging given the error handling from the previous steps. 
5. Once the duplicate errors issue is addressed the response will look like this:

Rates-Error-No-Dups

This will cause there to be no errors displayed as the single error is sliced out.

**Fixes**

Fixes #2048 

**Additional Notes**

This PR should be coordinated with the duplicate errors issue on the server. 87-GH-woocommerce-shipping-issues

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
@@ -75,7 +75,7 @@ export const ShippingRates = ( {
 						/>
 					} ) )
 				) }
-				{ packageErrors.slice( 1 ).map( ( error, index ) => {
+				{ packageErrors.map( ( error, index ) => {
 					// Print the rest of the errors (if any) below the dropdown
 					return <FieldError type="server-error" key={ index } text={ error } />;
 				} ) }
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
 = 1.24.1 - 2020-xx-xx =
 * Fix   - Services management CSS table layout
 * Fix   - Carrier "disconnect modal" layout
+* Fix   - Remove rates error array slice
 
 = 1.24.0 - 2020-07-30 =
 * Fix   - PHP 7.4 notice for taxes at checkout.


client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/list.js
```diff
@@ -51,7 +51,8 @@ export const ShippingRates = ( {
 		}
 		const selectedRate = selectedRates[ pckgId ] || '';
 		const packageRates = availableRates[ pckgId ].default.rates;
-		const packageErrors = errors[ pckgId ] || [];
+		// filter out duplicate error messages
+		const packageErrors = errors[ pckgId ].filter( ( item, index ) => errors[ pckgId ].indexOf( item ) === index ) || [];
```
frosso, line 55: Before it was checking that `errors[ pckgId ]` was always an array - now it's assuming it is.
Should this be instead:
```suggestion
		const packageErrors = ( errors[ pckgId ] || [] ).filter( ( item, index ) => errors[ pckgId ].indexOf( item ) === index );
```

bborman22: Great catch, thank you for that!

frosso: Thank _you_ still looking good after our discussion! 💯

Issue #2128, opened 2020-08-11
Discounted shipping label pop-up blocking access to edit order


AashikP:
*Customer Reports*

3168571-zen
3204865-zen
23345765-hc
3168571-zen

**Steps to replicate**

1. Activate WooCommerce Services on a store where it wasn't activated before
2. Attempt to open an order (Chrome) to find that you're not able to click anything on the screen:
    ![https://d.pr/i/H9SxWc+](https://d.pr/i/H9SxWc+)

It does work on some browsers for some users. For example, this is what I see on Firefox:

![https://d.pr/i/rnHFbP+](https://d.pr/i/rnHFbP+)

Dismissing the notice on firefox while logged in as the same user will ensure that the issue is fixed the next time I access the order edit page in Chrome.

Here are my browser details:

Chrome (bug): https://www.whatsmybrowser.org/b/41D1T
Firefox (works fine): https://www.whatsmybrowser.org/b/HI6DL

However, user in 3168571-zen reported that none of the browsers worked for them:

https://www.whatsmybrowser.org/b/ZK5RL
https://www.whatsmybrowser.org/b/5SEC9
https://www.whatsmybrowser.org/b/V6Y18

**Expected Behavior**

Either fix the problem with pop-up hiding other options on some browsers or make it a notification like other WordPress notifications and not a pop-up.

Marking as high priority because without reaching out to support or performing a complete conflict test, most users wouldn't know what exactly is happening.

Previous slack discussion: p1593417215220200-slack-fulfill-support

**Edit:** In my case, I could not replicate the behavior on my Macbook. The problem could only be replicated on iMac.

selectedselections:
reported in 3204865-zen as well 

screencast of issue:  https://www.loom.com/share/f7dd79eba3fa48d495105ffde8b240dc

RiaanKnoetze:
From the original bug report: 

> However, user in 3168571-zen reported that none of the browsers worked for them

User in 3204865-zen also reports that neither Firefox nor Chrome is working for them and shared the following screencast: https://www.loom.com/share/d52b86e11f124ff080dc9876c8b68a67

Considering that this breaks order management entirely, I'm bumping to critical.

harriswong:
Hi, 

I took a look at this and I could not reproduce the problem on both local environment and live environment, against both Chrome (latest 84) and FIrefox (latest 79).

I can see both showing:
[...]

PR #2139, opened 2020-08-13
fix(ShippingLabel): unsafe access to formSchema.custom

frosso:
## Description
Looks like there's an unsafe access to `packagesSettings.formSchema.custom.items`.
In fact, it can happen that the `custom` property is `null`, causing JS to break.

I think it would be nice if [client/admin-pointers.js](https://github.com/Automattic/woocommerce-services/blob/develop/client/admin-pointers.js) was _Reactified_ , so that it could be tied to the React lifecycle, maybe?

It happens to be `null` on a new installation, in this scenario:
- `packageSettings` is generated: https://github.com/Automattic/woocommerce-services/blob/develop/classes/class-wc-connect-shipping-label.php#L422
- `packageSettings.custom` is generated: https://github.com/Automattic/woocommerce-services/blob/develop/classes/class-wc-connect-package-settings.php#L29
- `packageSettings.custom` is `null` at https://github.com/Automattic/woocommerce-services/blob/develop/classes/class-wc-connect-service-schemas-store.php#L235 , when the `services` option is empty

~I also considered changing https://github.com/Automattic/woocommerce-services/blob/develop/client/admin-pointers.js#L42-L49 to ensure that `#wcs-pointer-page-dimmer` is added only when the pointer is created (and so that an `onClick` listener can be added to the overlay), but I'm afraid that the PayPal integration might not work correctly, otherwise.~
I also changed `client/admin-pointers.js` to ensure that `#wcs-pointer-page-dimmer` is added only when the pointer is created (and so that an `onClick` listener can be added to the overlay).

## Fixes
Fixes https://github.com/Automattic/woocommerce-services/issues/2128

## Steps to reproduce
- Create an order in WC
- Install WCS **for the first time** (or delete _all_ its options)
- Ensure Jetpack is enabled
- Go to an order page
- Issue should no longer be present

### DIFF
--- a/client/admin-pointers.js
+++ b/client/admin-pointers.js
@@ -37,15 +37,21 @@ import jQuery from 'jquery';
 
 				show_pointer( pointers, i + 1 );
 			},
+			open: function() {
+				if ( !pointer.dim ) {
+					return;
+				}
+
+				target.css( 'z-index', 9999 );
+				$( 'body' ).append( '<div id="wcs-pointer-page-dimmer" class="wcs-pointer-page-dimmer"></div>' );
+				$( '#wcs-pointer-page-dimmer' ).fadeIn( 500 ).one( 'click', function () {
+					target.pointer( 'close' );
+				} );
+			}
 		} );
 
 		function open() {
 			target.pointer( options ).pointer( 'open' );
-			if ( pointer.dim ) {
-				target.css( 'z-index', 9999 );
-				$( 'body' ).append( '<div id="wcs-pointer-page-dimmer" class="wcs-pointer-page-dimmer"></div>' );
-				$( '#wcs-pointer-page-dimmer' ).fadeIn( 500 );
-			}
 		}
 
 		/**
--- a/client/apps/shipping-label/index.js
+++ b/client/apps/shipping-label/index.js
@@ -36,7 +36,7 @@ export default ( { order, accountSettings, packagesSettings, shippingLabelData, 
 	const packages = packagesSettings.formData;
 	const dimensionUnit = packagesSettings.storeOptions.dimension_unit;
 	const weightUnit = packagesSettings.storeOptions.weight_unit;
-	const packageSchema = packagesSettings.formSchema.custom.items;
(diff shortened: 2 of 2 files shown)


client/apps/shipping-label/index.js
```diff
@@ -36,7 +36,7 @@ export default ( { order, accountSettings, packagesSettings, shippingLabelData,
 	const packages = packagesSettings.formData;
 	const dimensionUnit = packagesSettings.storeOptions.dimension_unit;
 	const weightUnit = packagesSettings.storeOptions.weight_unit;
-	const packageSchema = packagesSettings.formSchema.custom.items;
```
frosso, line 39: Before
Screen Shot 2020-08-13 at 4 28 33 PM

After:
Screen Shot 2020-08-13 at 4 29 55 PM
Screen Shot 2020-08-13 at 4 29 34 PM

client/admin-pointers.js
```diff
@@ -37,15 +37,21 @@ import jQuery from 'jquery';
...
 			},
-		} );
+			open: function() {
+				if ( !pointer.dim ) {
+					return;
+				}
 
-		function open() {
-			target.pointer( options ).pointer( 'open' );
-			if ( pointer.dim ) {
```
frosso, line 44: Moved this `if ( pointer.dim )` logic to the `onOpen` event on the pointer - this way the`wcs-pointer-page-dimmer` is added _only_ when the pointer has also been added.

This is safe to work with the PayPal integration - the PP integration does not use the dimmer

client/admin-pointers.js
```diff
@@ -37,15 +37,21 @@ import jQuery from 'jquery';
...
 				target.css( 'z-index', 9999 );
 				$( 'body' ).append( '<div id="wcs-pointer-page-dimmer" class="wcs-pointer-page-dimmer"></div>' );
-				$( '#wcs-pointer-page-dimmer' ).fadeIn( 500 );
+				$( '#wcs-pointer-page-dimmer' ).fadeIn( 500 ).one( 'click', function () {
```
frosso, line 47: Added an `onClick` event listener on the overlay, so that it can be dismissed by clicking on it

PR #2140, opened 2020-08-14
Displays breadcrumbs in carrier settings form

### LINKED ISSUE #2050: Add breadcrumbs navigation to carrier accounts settings screen
As specified in Figma designs in #2028, breadcrumbs navigation must be added to the top of the screen.

Screen implemented as part of the work done in this PR: https://github.com/Automattic/woocommerce-services/pull/2049

Ferdev:
----

Closes #2050

To test it, just go to the UPS settings form screen. The breadcrumbs should be there:

`/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings`

### DIFF
--- a/classes/class-wc-connect-settings-pages.php
+++ b/classes/class-wc-connect-settings-pages.php
@@ -89,9 +89,25 @@ if ( ! class_exists( 'WC_Connect_Settings_Pages' ) ) {
 				$extra_args['order_id'] = $_GET['from_order'];
 				$extra_args['order_href'] = get_edit_post_link( $_GET['from_order'] );
 			}
-			if ( isset( $_GET[ 'carrier' ] ) ) {
-				$extra_args['carrier'] = $_GET[ 'carrier' ];
+			if ( isset( $_GET['carrier'] ) ) {
+				$extra_args['carrier']    = $_GET['carrier'];
 				$extra_args['continents'] = $this->continents->get();
+				$translated_carrier_name  = null;
+				switch ( strtolower( $extra_args['carrier'] ) ) {
+					case 'ups':
+						$translated_carrier_name = __( 'UPS', 'woocommerce-services' );
+						break;
+					default:
+						break;
+				}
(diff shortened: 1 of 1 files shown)


classes/class-wc-connect-settings-pages.php
```diff
@@ -89,9 +89,25 @@ public function output_shipping_settings_screen() {
...
+						$translated_carrier_name = __( 'UPS', 'woocommerce-services' );
+						break;
+					default:
+						break;
+				}
+				if ( $translated_carrier_name ) {
+				?>
+					<h2>
+						<a href="<?php echo esc_url( admin_url( 'admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings' ) ); ?>"><?php esc_html_e( 'WooCommerce Services', 'woocommerce-services' ); ?></a> &gt;
+						<span class="wc-shipping-zone-name"><?php echo esc_html( $translated_carrier_name ); ?></span>
```
frosso, line 107: (minor) I don't see a style assigned to `wc-shipping-zone-name` - is that going to be done in the future? Or can it be removed?

Ferdev:
Hey @frosso sorry for getting back to you so late!

> Did you consider adding the breadcrumb to the view rendered by apps/shipping-settings, rather than adding the breadcrumb through JS? Was there a downside to that?

Actually, I just chose to do the same WC core was doing in the shipping settings page: I just copied the same breadcrumbs code they're using.

Yeah, as you have seen, there're several implementations for breadcrumbs in WordPress. I just decided to go with Core's approach.

classes/class-wc-connect-settings-pages.php
Ferdev: You're right, that class name shouldn't be there! I'll update the PR before merging it. Thanks for the review!! :bow:

frosso:
@Ferdev thanks for following up! 💯  Looks good!

PR #2156, opened 2020-08-21
Make preloading optional for wc-admin prompt compatibility

### LINKED ISSUE #2154: wc-admin order page prompt js error
1.24 added code splitting and pre-loading optimizations. Both of these changes introduced some incompatibilities with the wc-admin order page prompt:
image

There's a wc-admin bug (https://github.com/woocommerce/woocommerce-admin/issue/5024) that prevented the order page prompt from even installing WCS, but once a fix was created for this, the prompt still couldn't launch WCS. With the following wc-admin fix (https://github.com/woocommerce/woocommerce-admin/pull/5025) set up, the order page prompt will install and activate WCS 1.24.1, but loading it fails with the following errors:
image

To reproduce:
1. Store setup:
    - U.S. based store
    - Create shippable product (and order)
    - Install wc-admin branch https://github.com/woocommerce/woocommerce-admin/pull/5025
       - Note: wc-admin now requires `define( 'JETPACK_AUTOLOAD_DEV', true )` for loading everything in development mode
    - WCS **not** installed on site
1. Go to order page for shippable order
1. You should see the above prompt
1. Click "Create shipping label" and wait
1. See error on JS console

c-shultz:
Fixes #2154

This change makes it so the preloading of data is not required to open the label modal.

#1995 preloads data when the order page first loads (so additional requests are not needed), but requiring the preload breaks compatibility with the wc-admin order page prompt because it tries to open the label modal right after installing WCS.

With this WCS fix and the wc-admin fix( https://github.com/woocommerce/woocommerce-admin/pull/5025 ), the wc-admin order page prompt should be fully working again. It's a little challenging to test all of this together because the order page prompt will download the latest production WCS  from .org if none is installed. The fix here should only affect auto-loading on the order page, so we can still test it effectively by having WCS installed (**but not activated**) when the prompt is clicked.

To test:
1.  Install the wc-admin branch and follow the setup instructions to see the order page prompt to install WCS:
image
2. Check out this branch (`fix/optional-preloading`) and build it. **WCS should be installed but deactivated**
3. If WCS is not activated, the prompt to install should still be visible.
4. Click "Create shipping label" button and wait
5. Since WCS is already installed, wc-admin should activate, load, and finally launch the label purchase modal.

While testing, we should also check that normal WCShip functionality is unaffected and that the label purchase modal loads OK and labels can be purchased.

### DIFF
--- a/client/apps/shipping-label/index.js
+++ b/client/apps/shipping-label/index.js
@@ -31,9 +31,9 @@ import './style.scss';
 
 export default ( { order, accountSettings, packagesSettings, shippingLabelData, continents, context, items } ) => {
 	const orderId = order ? order.id : null;
-	let preLoadedState = false;
+	let isPreloaded = false;
 	if ( undefined !== accountSettings ) {
-		preLoadedState = true;
+		isPreloaded = true;
 	}
 
 	const addPreloadedState = function( initialState ) {
@@ -136,7 +136,7 @@ export default ( { order, accountSettings, packagesSettings, shippingLabelData, 
 					},
 				},
 			};
-			return preLoadedState ? addPreloadedState( initialState ) : initialState;
+			return isPreloaded ? addPreloadedState( initialState ) : initialState;
 		},
 
 		getStateForPersisting() {
--- a/client/apps/shipping-label/index.js
+++ b/client/apps/shipping-label/index.js
@@ -36,7 +36,7 @@ export default ( { order, accountSettings, packagesSettings, shippingLabelData, 
 	const packages = packagesSettings.formData;
 	const dimensionUnit = packagesSettings.storeOptions.dimension_unit;
 	const weightUnit = packagesSettings.storeOptions.weight_unit;
-	const packageSchema = packagesSettings.formSchema.custom.items;
+	const packageSchema = packagesSettings.formSchema.custom ? packagesSettings.formSchema.custom.items : undefined;
 	const predefinedSchema = packagesSettings.formSchema.predefined;
 
 	return {
--- a/client/apps/shipping-label/index.js
+++ b/client/apps/shipping-label/index.js
@@ -31,10 +31,7 @@ import './style.scss';
 
 export default ( { order, accountSettings, packagesSettings, shippingLabelData, continents, context, items } ) => {
 	const orderId = order ? order.id : null;
-	let isPreloaded = false;
-	if ( undefined !== accountSettings ) {
-		isPreloaded = true;
-	}
+	const isPreloaded = ( undefined !== accountSettings );
 
 	const addPreloadedState = function( initialState ) {
 		const { storeOptions, formMeta, userMeta, formData } = accountSettings;
--- a/client/apps/shipping-label/index.js
+++ b/client/apps/shipping-label/index.js
@@ -30,14 +30,64 @@ import initializeLabelsState from 'woocommerce/woocommerce-services/lib/initiali
 import './style.scss';
 
 export default ( { order, accountSettings, packagesSettings, shippingLabelData, continents, context, items } ) => {
-
 	const orderId = order ? order.id : null;
-	const { storeOptions, formMeta, userMeta, formData } = accountSettings;
-	const packages = packagesSettings.formData;
-	const dimensionUnit = packagesSettings.storeOptions.dimension_unit;
-	const weightUnit = packagesSettings.storeOptions.weight_unit;
-	const packageSchema = packagesSettings.formSchema.custom ? packagesSettings.formSchema.custom.items : undefined;
-	const predefinedSchema = packagesSettings.formSchema.predefined;
+	let preLoadedState = false;
+	if ( undefined !== accountSettings ) {
+		preLoadedState = true;
+	}
+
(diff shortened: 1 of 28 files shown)


harriswong:
👀

client/apps/shipping-label/index.js
```diff
@@ -30,14 +30,64 @@ import initializeLabelsState from 'woocommerce/woocommerce-services/lib/initiali
...
 export default ( { order, accountSettings, packagesSettings, shippingLabelData, continents, context, items } ) => {
-
 	const orderId = order ? order.id : null;
-	const { storeOptions, formMeta, userMeta, formData } = accountSettings;
-	const packages = packagesSettings.formData;
-	const dimensionUnit = packagesSettings.storeOptions.dimension_unit;
-	const weightUnit = packagesSettings.storeOptions.weight_unit;
-	const packageSchema = packagesSettings.formSchema.custom ? packagesSettings.formSchema.custom.items : undefined;
-	const predefinedSchema = packagesSettings.formSchema.predefined;
+	let preLoadedState = false;
```
frosso, line 34: (minor) simplify this to 
```suggestion
	const preLoadedState = undefined !== accountSettings;
```

Also, what does everyone think about prefixing booleans with a `has`/`is` prefix? 👀

client/apps/shipping-label/index.js
```diff
@@ -30,14 +30,64 @@ import initializeLabelsState from 'woocommerce/woocommerce-services/lib/initiali
...
+	}
+
+	const addPreloadedState = function( initialState ) {
+		const { storeOptions, formMeta, userMeta, formData } = accountSettings;
+		const packages = packagesSettings.formData;
+		const dimensionUnit = packagesSettings.storeOptions.dimension_unit;
+		const weightUnit = packagesSettings.storeOptions.weight_unit;
+		const packageSchema = packagesSettings.formSchema.custom ? packagesSettings.formSchema.custom.items : undefined;
+		const predefinedSchema = packagesSettings.formSchema.predefined;
+		initialState.extensions.woocommerce.woocommerceServices = {
```
frosso, line 46: It looks like `initialState` (which is an object passed by reference in the function's argument) is being altered directly (rather than either being initialized in the function's body or being cloned).

I haven't been able to see any direct side effects related to this decision, but I just wanted to point it out and double check if this could be causing any issues?

client/apps/shipping-label/index.js
c-shultz: Both good suggestions! I went with `isPreloaded`

client/apps/shipping-label/index.js
c-shultz: Since the variable is initialized in `getInitialState` and immediately modified, I don't personally see a risk here.

woocommerce-services.php
```diff
@@ -246,7 +246,11 @@ static function get_wcs_admin_script_url() {
 		 * @return string
 		 */
 		static function get_wcs_admin_style_url() {
-			return self::get_wc_connect_base_url() . 'woocommerce-services-' . self::get_wcs_version() . '.css';
+			if ( ! defined( 'WOOCOMMERCE_CONNECT_DEV_SERVER_URL' ) ) {
```
c-shultz, line 249: It turns out the `woocommerce-services-x.xx.x.css` is only used for the production build (since the code splitting and dynamic loading changes were added). This check of `WOOCOMMERCE_CONNECT_DEV_SERVER_URL` consolidates the check in one place so an empty path will be passed to the assets endpoint

client/main.js
```diff
@@ -146,11 +147,12 @@ Array.from( document.getElementsByClassName( 'wcc-root' ) ).forEach( ( container
 			</AppContainer>,
 			container
 		);
+		return createdStores[ routeClassName ];
 	});
 } );
 
-window.wcsGetAppStore = function( storeKey ) {
-	return createdStores[ storeKey ];
+window.wcsGetAppStoreAsync = function( storeKey ) {
```
c-shultz, line 154: The dynamic loading functionality that was added in `1.24.0` means there's no guarantee that `createdStores` will be ready after the main js file is loaded. Returning a promise instead allows dependent code to wait for everything to be resolved first (see: https://github.com/woocommerce/woocommerce-admin/pull/5025/files#diff-8f89acda6e8c126c62f0d46b8d47dd73R276-R278)

harriswong: Renaming this will need to bump our minimum WC requirement to 4.5+. Should we do the requirement bump in this PR or make another issue for 1.24.2 TODO?

c-shultz: I'm not sure if there's much we can do. WCS is still generally compatible with very old versions of WC core. Any version of WC with the prompt will always download the latest version of WCS and try to start it. `wcsGetAppStore` is broken as of `1.24.0`. I couldn't think of a way to make it backwards compatible `wcsGetAppStore` without trying to remove the code splitting.

harriswong: Good point on the prompt that always download the latest. Since this part of the code in WCS is only being used by wc-admin to get the latest WCS, I think this is fine as well.

c-shultz:
This why we should bump our minimum PHP version. 😄 The latest commit should clear it up 🤞

Issue #2155, opened 2020-08-21
Usage of deprecated `jQuery.load`


frosso:
## Description

Reported in the forums: https://wordpress.org/support/topic/deprecated-code-in-woocommerce-services/

> jQuery Migrate Helper — Warnings encountered

> This page generated the following warnings:

> https://bendablebody.com/wp-content/plugins/woocommerce-services/dist/woocommerce-services-admin-pointers-1.24.1.js:1:1065: jQuery.fn.load() is deprecated

> Please make sure you are using the latest version of all of your plugins, and your theme. If you are, you may want to ask the developers of the code mentioned in the warnings for an update.

See: https://github.com/Automattic/woocommerce-services/blob/develop/client/admin-pointers.js#L85

Let's replace the `.load(handler)` with the `.on('load', handler)`.

## Acceptance criteria
- The page no longer generates a warning

PR #2167, opened 2020-09-03
Change jquery.load to jquery.on('load')

harriswong:
Closes https://github.com/Automattic/woocommerce-services/issues/2155

### Description
Since jquery 1.8, `jquery.load` is deprecated. https://api.jquery.com/load-event/ Our jquery version is `1.12.4` so we need to fix that.

### How to test
1. In `wordpress` database, look for `meta_key like '%dismissed_wp_pointers%'` in the `wp_usermeta` table
2. Remove your entry
3. You should see the pointer to show again
image


frosso (approved):
ezpz 👍

Issue #2141, opened 2020-08-17
Packages directory filenames are too long


Stoyan0v:
Describe the bug
I'm one of the developers of Siteground Migrator plugin, and we have issues migrate sites which have WooCommerce Services Installed.

The problem is that the filenames generated in packages folder are too long, and we cannot create tar archive when we migrate sites to our servers.
Here is an example: `woocommerce-services/dist/chunks/vendors~plugin-status~print-test-label~settings~shipping-label~shipping-settings.3e5c5184b6d45dc8c6e9.min.js`

We are using Phar::buildFromDirectory, but we get the following error when we try to archive the a plugin directory:
`[message] => Uncaught PharException: tar-based phar "/home/customer/www/*/public_html/wp-content/uploads/1597612370-fb1d81a286496977c83270c37ed7f569ea1d43fe/plugins/woocommerce-services.tar" cannot be created, filename "dist/chunks/vendors~plugin-status~print-test-label~settings~shipping-label~shipping-settings.3e5c5184b6d45dc8c6e9.min.js" is too long for tar file format`

Is there any chance to change how filenames are generate, so they can become shorter?

We had similar issues with WooCommerce Admin. Please refer to :https://github.com/woocommerce/woocommerce-admin/issues/4512

c-shultz:
To add an additional note to this, we've seen some reports that the `~` in the paths causes issues with some server's security settings, so let's look into removing that when we address this.

PR #2171, opened 2020-09-09
Change webpack chunks from [name] to [chunkhash]

harriswong:
Closes https://github.com/Automattic/woocommerce-services/issues/2141

### Summary
Updated chunks from `[name]` to `[chunkhash]` to address reduce filename size and to remove `~` in chunk filenames. I removed `id`, `name` form the `chunkFileName` since we don't really need that anymore. I kept `chunkhash` for cache busting purposes. I also added `chunkFileName` to the `MiniCssExtractPlugin` so that the chunks are now named with `chunkhash`, and placed properly in the `chunks` folder.
Before:
image
After:
image

#### Local development
1. Pull this branch and run `npm run start`.
2. Open webpack dev server http://localhost:8085/chunks/ and you should see a bunch of `<hash>.css` and `<hash>.min.js`
3. Go to WooCommerce > Orders, and you should still be able to create a label.
4. You can also open up network tab and see all the chunks show up for `css` and `js`.

#### Production environment
1. Pull this branch and run `npm run build`
2. Go to `release/` folder and upload this `woocommerce-services.zip` to a production website (ie. JN).
3. You should be able to print a label without errors. Check WCS settings and the style, js should load fine (`wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings`)
4. You can also open up network tab and see all the chunks show up for `css` and `js`.

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -56,22 +56,22 @@ const getRouteClass = async ( className ) => {
 	let module = null;
 	switch ( className) {
 		case 'wc-connect-create-shipping-label':
-			module = await import(/* webpackChunkName: "shipping-label" */ './apps/shipping-label');
+			module = await import('./apps/shipping-label');
 			break;
 		case 'wc-connect-service-settings':
-			module = await import(/* webpackChunkName: "settings" */ './apps/settings');
+			module = await import('./apps/settings');
 			break;
 		case 'wc-connect-admin-status':
-			module = await import(/* webpackChunkName: "plugin-status" */ './apps/plugin-status');
+			module = await import('./apps/plugin-status');
 			break;
 		case 'wc-connect-shipping-settings':
-			module = await import(/* webpackChunkName: "shipping-settings" */ './apps/shipping-settings');
+			module = await import('./apps/shipping-settings');
 			break;
 		case 'wc-connect-admin-test-print':
-			module = await import(/* webpackChunkName: "print-test-label" */ './apps/print-test-label');
+			module = await import('./apps/print-test-label');
 			break;
 		case 'wc-connect-stripe-connect-account':
-			module = await import(/* webpackChunkName: "shipping-label" */ './apps/stripe-connect-account');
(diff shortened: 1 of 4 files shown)


client/main.js
```diff
@@ -56,22 +56,22 @@ const getRouteClass = async ( className ) => {
 	let module = null;
 	switch ( className) {
 		case 'wc-connect-create-shipping-label':
-			module = await import(/* webpackChunkName: "shipping-label" */ './apps/shipping-label');
```
harriswong, line 59: Removed Magic Comment `webpackChunkName` because we no longer want to bundle them into `[name]`.

frosso (approved):
Made a build with `npm run build`.
Tested on JN.

Tested label creation process, live rates, live rates configuration, status page.
Confirmed that the chunk names for WCS are now not including `~` and are shorter.

Looks good 👍  :shipit:

webpack.config.js
```diff
@@ -60,7 +60,7 @@ module.exports = {
 			{
 			path: path.join( __dirname, 'dist' ),
 			filename: '[name]-' + process.env.npm_package_version + '.js',
-			chunkFilename: 'chunks/[name].[chunkhash].min.js',
```
harriswong, line 63: @frosso @Ferdev @c-shultz @bborman22 I want to double check, is there a reason why we should use `'chunks/[id].[chunkhash].min.js'` or is `chunks/[chunkhash].min.js` okay?

frosso: I'm not sure - what would be the advantage of adding the `id`? 👀 
I think that just using the `chunkhash` would be fine for now. It's still going to change between builds/changes, so that the browser's cache is busted 👍

c-shultz:
Let's make sure we test the wc-admin install prompt banner too

harriswong:
Maybe good to test wc-admin's WCS banner as well.

harriswong:
We will test the banner in the release zip.

Issue #2165, opened 2020-09-03
Select all posts checkbox doesn't work when plugin is active


aheckler:
To reproduce:

1. Spin up a new test site.
2. Install WC and WCS, activate both.
3. Go to the Pages menu in wp-admin, or any other page that lists posts (of any type).
4. Click here: https://d.pr/i/EaRtXo/E1smvFSjcr

As you can see, the posts are not actually select. No JS errors in console.

Deactivating WC Services makes it work again.

EDIT: From this forums thread: https://wordpress.org/support/topic/issue-with-wp-admin-select-all-feature/?view=all

peterfabian:
We also got a report on the developer blog and I could reproduce with WP 5.5.1 + WC 4.4.1 + WCS 1.24.1

c-shultz:
Thanks for the report! I'll slate this for next week's sprint

harriswong:
I can reproduce this on the latest WP 5.5.1, WC 4.5.1, and WCS 1.24.2. To reproduce, create a clean site, download and install WCS and WC through store. Activate but do not go through onboarding or any setups.

harriswong:
Another note is deactivating either WC or WCS will allow the select-all checkbox to work again.

harriswong:
One thing that seems to be different is that when both WC and WCS are activated, it has different events listeners bound to the checkbox
When both WC and WCS are activated, there is 1 listener
![image](https://user-images.githubusercontent.com/572862/92805131-14ddeb80-f376-11ea-865b-2a539c5c85b8.png)

When only either WC/WCS is activated, there are 5 listeners, and the 4th one seems to be responsible for the check-all box
![image](https://user-images.githubusercontent.com/572862/92805286-37700480-f376-11ea-83f6-9ef5b5d6fd8d.png)

harriswong:
The problem seems to be coming from the `wc_connect_banner` script in [here](https://github.com/Automattic/woocommerce-services/blob/develop/classes/class-wc-connect-nux.php#L441).

This code is use for this banner:
![image](https://user-images.githubusercontent.com/572862/92815968-08f82680-f382-11ea-9828-8daa283070eb.png)

harriswong:
In source, we can see the following error:
```html
<script id='wc_connect_banner-js-extra'>
--
[...]

PR #2176, opened 2020-09-11
Remove wp-admin dependency from banner

harriswong:
Closes https://github.com/Automattic/woocommerce-services/issues/2165

### Summary
Remove the `'updates'` dependency since WP has it inside script-loader by default. 

### Description 
The select-all feature checkbox stopped working when WCS's `wc_connect_banner` has a `updates.min.js` dependency in it. When we add `'updates'` as dependency in `wp_register_script()`, it seems like it moved `<script src='https://localhost/wp-admin/js/updates.min.js?ver=5.5.1' id='updates-js'></script>` to the `<head>`. When this happens, for some reason, the [event handler](https://github.com/WordPress/WordPress/blob/master/wp-admin/js/common.js#L1170 ) from WP's `common.js` is removed, as shown in [here](https://github.com/Automattic/woocommerce-services/issues/2165#issuecomment-690726954). By removing `'updates'` dependency, the `updates.min.js` no longer sits in the `<head>` and the event handler is back.

### Do we need the `updates` dependency?
I don't think we do anymore. This was added in https://github.com/Automattic/woocommerce-services/commit/a20273aac03ca23bdc0334a058730d11831eddbe#diff-57ae8afb227e356d71d37582f678a831R710, which was part of the work from here https://github.com/Automattic/woocommerce-services/commit/62a24300c8270aa949ada912d8e2a5d441e5d32a#diff-57ae8afb227e356d71d37582f678a831R858-R864. The purpose was to make sure `wp.updates.installPlugin( { slug: 'jetpack' } )` runs.

Given that WP's `script-loader.php` includes `updates` [by default](https://github.com/WordPress/WordPress/blob/master/wp-includes/script-loader.php#L1264), I don't think we need to load it ourselves anymore. It is also worth mentioning that in the code above, we can see that `updates` depends on `common`. Perhaps it is a problem when we use `wp_register_script()` with a dependency of `updates` without `common`?

#### All "check-all" box should now work
1. Go to https://localhost/wp-admin/plugins.php
2. Click the "select-all" checkbox, all boxes should now show.
image

#### Regression, localhost, Jetpack still installs through banner
1. Pull this branch
2. Make sure Jetpack is deactivated/uninstalled. Then go to https://github.com/Automattic/woocommerce-services/blob/develop/classes/class-wc-connect-nux.php#L421-L426, add `$banner_to_display = 'before_jetpack_connection';` after `line 426` above to force the banner to show. 
Alternatively, you can disconnect Jetpack.
3. Click this
image
4. You should see Jetpack install and activate.

#### Regression, on a fresh live site (JN), Jetpack still installs through banner
1. Pull this branch, run `npm run build`
2. On your fresh live site, install latest WC
3. Upload `releases/woocommerce-services.zip` to your live's site plugin
4. You should see a similar banner, click Activate/Install/Connect
image
5. Jetpack should installed and you should see a redirection, followed by a setup completed message.
image
image

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -1232,7 +1232,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			}
 			wp_register_script( 'wc_services_admin_pointers', $this->wc_connect_base_url . 'woocommerce-services-admin-pointers-' . $plugin_version . '.js', array( 'wp-pointer', 'jquery' ), null );
 			wp_register_style( 'wc_connect_banner', $this->wc_connect_base_url . 'woocommerce-services-banner-' . $plugin_version . '.css', array(), null );
-			wp_register_script( 'wc_connect_banner', $this->wc_connect_base_url . 'woocommerce-services-banner-' . $plugin_version . '.js',  array( 'updates' ), null );
+			wp_register_script( 'wc_connect_banner', $this->wc_connect_base_url . 'woocommerce-services-banner-' . $plugin_version . '.js',  array(), null );
 
 			$i18n_json = $this->get_i18n_json();
 			/** @var array $i18nStrings defined in i18n/strings.php */


Ferdev (approved):
This is working for me nicely, both in local and a JN site!

PR #2179, opened 2020-09-15
chore: rename WCS to WC Shipping & Tax

frosso:
## Description
Renaming of user-facing `WooCommerce Service(s)` strings to `WooCommerce Shipping & Tax(es)`.

## Related Issue
Helps with https://github.com/Automattic/woocommerce-shipping-issues/issues/104


c-shultz:
@frosso, we'll go ahead and release this with all the other pending changes on the 13th. Is it ready for review now?

frosso:
@c-shultz I updated the description(s) based on the newer marketing documents.
I am still waiting for the updated assets for the .org entry.

Would you like me to open this for review, even though I'm still waiting on the new assets?
I can open a PR for the updated assets separately.

c-shultz:
@frosso, I would go ahead with opening it up for review. The .org assets can be separate if needed.

c-shultz (commented):
@frosso ,

Why did you go with "Shipping and Tax**es**" in some cases instead of just "Shipping and Tax"
My assumption is that it would just be "Tax" in all cases, but maybe I missed a conversation about this.

c-shultz (changes requested):
Everything looks good and tests well! My only comment is that I think we should just be consistent and use "Tax" instead of "Taxes"

frosso:
Chatted with @c-shultz offline.
The "Taxes" (instead of "Tax") came from the adjusted marketing material.

Confirmed to keep "Tax" for consistency, will change later if needed.

c-shultz (approved):
The changes look good!

I ran through:
- Settings
- Status screen
- Label purchase
- Live rates

Everything is working well and the string update is reflected correctly everywhere. 🚢

c-shultz:
>changelog entry    
I'm glad you at least remembered to check for that 💯

PR #2199, opened 2020-09-23
Avoid PHP notice by checking for WP_ERROR

superdav42:
Api client returns instance of WP_Error on failure so we should really check for that.
TEST:
Make the connect server return an error by stopping it or something.
See PHP notice undefined $carriers_response->carriers on /wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings
With this PR there will be no notice.


classes/class-wc-connect-settings-pages.php
```diff
@@ -81,7 +81,8 @@ public function output_shipping_settings_screen() {
 			}
 
 			$extra_args = array();
-			if( $carriers_response = $this->api_client->get_carrier_accounts() ) {
+			$carriers_response = $this->api_client->get_carrier_accounts();
+			if ( ! is_wp_error( $carriers_response ) && $carriers_response ) {
```
frosso, line 85: What do you think about flipping these 2 checks?
```suggestion
			if ( carriers_response && ! is_wp_error( $carriers_response ) ) {
```

That way we can ensure that it's truthy, before passing it to `is_wp_error`

frosso (approved):
Changes look good! 👍

classes/class-wc-connect-settings-pages.php
superdav42: I think the primary condition that this statement needs to check for is if it is a WP_Error so I'd like to keep that first to make the code more readable and obvious what it's doing. is_wp_error() will accept any type of var without causing any problems. 
I think it's impossible for $carriers_response to not be truthy but I kept the check just in case there's something I'm not aware of.

Issue #2210, opened 2020-10-06
Add husky pre-commit hooks to make sure the php files conform to WooCommerce plugin standards


harriswong:
### Summary
We would like to make sure our PHP code conform to WooCommerce plugin standards. Add `phpcs` to `husky` `pre-commit hook. Maybe we can do similar things like what WooCommerce does.
https://github.com/woocommerce/woocommerce/blob/master/phpcs.xml

Issue #2131, opened 2020-08-11
Update minimum requirements to match our support policy


harriswong:
### Description
Our support policy is based on https://docs.woocommerce.com/document/frequently-asked-questions/#section-62. We are supporting the following:

**WooCommerce**: latest - 2. (the latest version, plus 2 previous ones. For example, if current version is 4.3, then we should support 4.2 and 4.1)

**WordPress version**: latest - 2

**PHP version**: 7.0+

### Task
- Update travis.yml matrix for the above combinations.
- Update minimum versions in readme.txt and woocommerce-services.php plugin header
- Update minimum requirements in docs: https://docs.woocommerce.com/document/woocommerce-services/#section-1
- Update `phpcs.xml.dist` for the [minimum required WP version](https://github.com/Automattic/woocommerce-services/pull/2212/files#diff-05ae9cddcaec1e845771a7db224961439f83ef5939ec67d3a48744cb34d7e58bR13-R15) [Added Oct 19, 2020]

c-shultz:
I tweaked this to include bumping the overall minimum requirements beyond just the Travis config

harriswong:
https://github.com/Automattic/woocommerce-services/pull/2329 has updated this repo to use GitHub Actions. We are now testing against WP latest minus 2 and WC latest minus 2. Closing.

PR #2212, opened 2020-10-07
Add phpcs with WooCommerce-Core rules to pre-commit hook

harriswong:
### Description
Add phpcs with WooCommerce-Core rules to pre-commit hook. This only runs [WooCommerce-Core ](https://github.com/woocommerce/woocommerce-sniffs) on committed changes, not on all files. THis is similar to [how WooCommerce Admin is setup]( https://github.com/woocommerce/woocommerce-admin/blob/37f92a5cf142a8bd2b055e16b451307b2a2dd095/lint-staged.config.js#L8).

**Updated - Oct 16, 2020**
I [removed](https://github.com/Automattic/woocommerce-services/pull/2212/commits/c402222c1a5ccca3465fa3b0073a21859ee2fbf8) `phpcs` from pre-commit hooks. There were too many linting errors and it will slow down development. I have removed this for now until we addressed most of the linting errors. To run `phpcs`, you will have to do this manually with `composer run phpcs`.

### Related issue(s)
Closes https://github.com/Automattic/woocommerce-services/issues/2210

1. Pull this branch, run `composer install`. If you don't have `composer`, install it globally with [this instruction](https://getcomposer.org/doc/00-intro.md#installation-linux-unix-macos).
2. Make changes to any `.php` file, (ie. `woocommerce-services.php`)
3. Run test manually. `composer run phpcs`.
4. The terminal runs and you should see linting errors preventing commit to go through
image

### Checklist

- [ ] unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "analyze": "cross-env ANALYZE=true NODE_ENV=production DISABLE_FEATURES=wpcom-user-bootstrap CALYPSO_CLIENT=true webpack",
     "eslint": "eslint tasks/ client/",
     "eslint:fix": "eslint --fix tasks/ client/",
-    "test": "npm run test-client && npm run eslint",
+    "test": "npm run test-client",
     "test-client": "jest -c=tests/client/jest.config.js",
     "test-client:watch": "npm run -s test-client -- --watch",
     "prerelease": "npm run i18n && node tasks/i18n-download && npm run i18njson",
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,0 +1,55 @@
+<?xml version="1.0"?>
+<ruleset name="WordPress Coding Standards">
+	<!-- See https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml -->
+	<!-- See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress-Core/ruleset.xml -->
+
+	<description>WooCommerce dev PHP_CodeSniffer ruleset.</description>
+
+	<!-- Exclude paths -->
+	<exclude-pattern>*/node_modules/*</exclude-pattern>
+	<exclude-pattern>*/vendor/*</exclude-pattern>
+	<exclude-pattern>bin/*</exclude-pattern>
+
+	<!-- Configs -->
+	<config name="minimum_supported_wp_version" value="4.7" />
+	<config name="testVersion" value="5.6-"/>
+
+	<!-- Rules -->
+	<rule ref="WooCommerce-Core" />
+
+	<rule ref="WordPress.WP.I18n">
+		<properties>
+			<property name="text_domain" type="array" value="woocommerce-admin" />
+		</properties>
+	</rule>
+
+	<rule ref="PHPCompatibility">
+		<exclude-pattern>tests/</exclude-pattern>
+	</rule>
+
+	<rule ref="WordPress.Files.FileName.InvalidClassFileName">
+		<exclude-pattern>src/*</exclude-pattern>
+		<exclude-pattern>tests/*</exclude-pattern>
+		<exclude-pattern>woocommerce-admin.php</exclude-pattern>
+	</rule>
+
+	<rule ref="WordPress.Files.FileName.NotHyphenatedLowercase">
+		<exclude-pattern>src/*</exclude-pattern>
+	</rule>
+
+	<rule ref="Generic.Arrays.DisallowShortArraySyntax.Found">
+      <exclude-pattern>src/*</exclude-pattern>
+    </rule>
+
+	<rule ref="Generic.Commenting">
+		<exclude-pattern>tests/</exclude-pattern>
+	</rule>
+	<rule ref="Squiz.Commenting.FileComment.MissingPackageTag">
+		<exclude-pattern>src/</exclude-pattern>
+		<exclude-pattern>tests/</exclude-pattern>
+	</rule>
+	<rule ref="Squiz.Commenting.FileComment.Missing">
+		<exclude-pattern>src/</exclude-pattern>
+		<exclude-pattern>tests/</exclude-pattern>
+	</rule>
+</ruleset>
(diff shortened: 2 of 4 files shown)


phpcs.xml.dist
```diff
@@ -0,0 +1,55 @@
...
+			<property name="text_domain" type="array" value="woocommerce-admin" />
+		</properties>
+	</rule>
+
+	<rule ref="PHPCompatibility">
+		<exclude-pattern>tests/</exclude-pattern>
+	</rule>
+
+	<rule ref="WordPress.Files.FileName.InvalidClassFileName">
+		<exclude-pattern>src/*</exclude-pattern>
```
superdav42, line 31: Looks like some of these rules need to be update or removed because WCS does not have a src/ directory. 
I think the InvalidClassFileName rule doesn't need any exclusions except maybe the tests/ folder.
For the other rules I think src/ should be replaced with classes/

superdav42, line 31:
```suggestion
```

phpcs.xml.dist
```diff
@@ -0,0 +1,55 @@
...
+		<exclude-pattern>tests/*</exclude-pattern>
+		<exclude-pattern>woocommerce-admin.php</exclude-pattern>
```
superdav42, line 33:
```suggestion
		<exclude-pattern>woocommerce-services.php</exclude-pattern>
```

phpcs.xml.dist
```diff
@@ -0,0 +1,55 @@
...
+	</rule>
+
+	<rule ref="WordPress.Files.FileName.NotHyphenatedLowercase">
+		<exclude-pattern>src/*</exclude-pattern>
```
superdav42, line 37:
```suggestion
		<exclude-pattern>classes/*</exclude-pattern>
```

phpcs.xml.dist
```diff
@@ -0,0 +1,55 @@
...
+	</rule>
+
+	<rule ref="Generic.Arrays.DisallowShortArraySyntax.Found">
+      <exclude-pattern>src/*</exclude-pattern>
```
superdav42, line 41:
```suggestion
      <exclude-pattern>classes/*</exclude-pattern>
```

phpcs.xml.dist
```diff
@@ -0,0 +1,55 @@
...
+    </rule>
+
+	<rule ref="Generic.Commenting">
+		<exclude-pattern>tests/</exclude-pattern>
+	</rule>
+	<rule ref="Squiz.Commenting.FileComment.MissingPackageTag">
+		<exclude-pattern>src/</exclude-pattern>
```
superdav42, line 48:
```suggestion
		<exclude-pattern>classes/</exclude-pattern>
```

phpcs.xml.dist
```diff
@@ -0,0 +1,55 @@
...
+		<exclude-pattern>tests/</exclude-pattern>
+	</rule>
+	<rule ref="Squiz.Commenting.FileComment.Missing">
+		<exclude-pattern>src/</exclude-pattern>
```
superdav42, line 52:
```suggestion
		<exclude-pattern>classes/</exclude-pattern>
```

superdav42 (changes requested):
Some suggestions

phpcs.xml.dist
harriswong: Good point. Confirmed InvalidClassFileName doesn't apply to our `classes` folder since all ours have `class-` prefix
https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/Files/FileNameSniff.php#L190-L191

phpcs.xml.dist
harriswong: Fixed

harriswong:
Thanks for catching all those @superdav42 . I added a new commit to address all the comments. Please take another look.

phpcs.xml.dist
harriswong: > InvalidClassFileName

Ended up our test classes have `_` instead of `-`. We can probably rename all of them.

package.json
```diff
@@ -8,7 +8,7 @@
     "analyze": "cross-env ANALYZE=true NODE_ENV=production DISABLE_FEATURES=wpcom-user-bootstrap CALYPSO_CLIENT=true webpack",
     "eslint": "eslint tasks/ client/",
     "eslint:fix": "eslint --fix tasks/ client/",
-    "test": "npm run test-client && npm run eslint",
```
frosso, line 11: Should `npm run eslint` be added to the Husky `pre-commit`?

frosso (approved):
The instructions should maybe be updated, since `phpcs` is no longer part of the commit hooks :) 

Tested locally, looks good 👍

package.json
harriswong: Yep, this is already in `develop`. https://github.com/Automattic/woocommerce-services/blob/develop/lint-staged.config.js#L3

brezocordero:
Also tested locally and works great. 

One question, our configs are lower than core:
https://github.com/woocommerce/woocommerce/blob/e5f0ce0723d6335bbee913ac2f9ac49634f50760/phpcs.xml#L27

Any reason for that?

harriswong:
Good question @brezocordero! I was using [woocommerce-admin](https://github.com/woocommerce/woocommerce-admin/blob/e7f769efab64337cc04452b9836d08750ab68ee1/phpcs.xml.dist#L14-L15) as a reference point and I didn't look into this. You are right, our config is a few versions off compared to [woocommerce-core](https://github.com/woocommerce/woocommerce/blob/e5f0ce0723d6335bbee913ac2f9ac49634f50760/phpcs.xml#L28-L29). 

I agree. I think we should match the woocommere-core's policy.

Instead of bumping the version here, what do you think about us adding this to our [Update minimum requirements issue](https://github.com/Automattic/woocommerce-services/issues/2131)? I worry that If I bump the version here, there will be a scope creep.

brezocordero:
> Instead of bumping the version here, what do you think about us adding this to our Update minimum requirements issue? I worry that If I bump the version here, there will be a scope creep.

I like that, it is better to be consistent and make the change everywhere.

harriswong:
I added a note to https://github.com/Automattic/woocommerce-services/issues/2131. So that we remember to update the `phpcs.xml.dist` file.

harriswong:
@frosso I updated testing steps.

Issue #2157, opened 2020-08-24
Customs forms missing with UPS labels


elizaan36:
For international orders, the customs forms should print with UPS labels in the same way they do for USPS. I tested this on my site and the customs form appears for international shipments and I'm able to fill it out, but it doesn't print with the label.

USPS - with customs form attachment
<img width="1015" alt="Screen Shot 2020-08-24 at 9 23 47 AM" src="https://user-images.githubusercontent.com/5121465/91056779-0066c880-e5ec-11ea-8af5-51d586165636.png">

UPS - no customs form
<img width="1005" alt="Screen Shot 2020-08-24 at 9 41 14 AM" src="https://user-images.githubusercontent.com/5121465/91058516-2a20ef00-e5ee-11ea-834e-233761433009.png">

A UPS beta customer flagged this issue. Here's his message:
> Just a heads up, but for international orders, the UPS option doesn’t print all the required documentation like the USPS option does. Please take a look into that as it’s absolutely needed for someone like me when shipping. Thanks!

Ferdev:
I've performed several test to try to make UPS print the customs info in their labels. Since I couldn't make it work, and the docs didn't say anything about UPS needing special parameters to make it work, I contacted Easypost support and they replied this:

> The Commercial Invoice is provided as a courtesy, UPS labels do not include a print-out of the entered customs information for UPS as it does for the USPS. 

Apparently, this is something UPS don't provide, and their behavior is different from USPS labels. Some kind of customs info is provided with the UPS commercial invoice, but I'm not sure this would be useful for our customers.

Maybe this is something we could include in the future as added value for our customers.

elizaan36:
Hey @Ferdev that's interesting and it does make sense that it would be a different process to USPS. I suppose it's a separate form to the UPS shipping label and we'd have to set up as such.

Since the customs forms won't print for our merchants in the process of creating a UPS shipping label, we should remove the custom form step. I think that will be the biggest point of confusion, if people fill out the info in the customs form then it's not there when they print. 😆

elizaan36:
Just reached out to the original customer to shared the feedback and asked what his experience is printing on UPS.COM. Here's what he shared:

> When I create a label on UPS.com I am informed I also need to fill out a document that lists what products I am shipping and their costs. Attached is a recent shipment I made and what the forms look like.

Shared the labels separately in [this thread](https://a8c.slack.com/archives/CU6SYV31A/p1598992662033900).

[...]

PR #2213, opened 2020-10-07
Open customs forms in new browser tabs

Ferdev:
After a shipping label is purchased, if it needs a customs form to be printed, this patch creates an info
note for the user, and opens the customs form in a new browser tab.

### Test instructions.

1. Patch WooCommerce Connect Server to make it always include a commercial invoice:

``` javascript
1 file changed, 2 insertions(+), 1 deletion(-)
lib/shipping/easypost-client/index.js | 3 ++-

modified   lib/shipping/easypost-client/index.js
@@ -352,7 +352,8 @@ const buyShipment = ( shipment, rateToBuy ) => {
 const extractCommercialInvoiceFromShipment = ( shipment ) => {
 	if ( shipment.forms ) {
 		for ( const form of shipment.forms ) {
-			if ( FORM_TYPE_COMMERCIAL_INVOICE === form.form_type && ! form.submitted_electronically ) {
+			// if ( FORM_TYPE_COMMERCIAL_INVOICE === form.form_type && ! form.submitted_electronically ) {
+			if ( FORM_TYPE_COMMERCIAL_INVOICE === form.form_type ) {
 				return form.form_url;
 			}
 		}
```

2. Create an international order.
3. Purchase a new shipping label for the order created in step 2.
4. The customs form pdf should be open on a new tab and a notice should be displayed.

----

Related issue https://github.com/Automattic/woocommerce-services/issues/2157

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
@@ -920,6 +920,17 @@ function downloadAndPrint( orderId, siteId, dispatch, getState, labels ) {
 
 	let hasError = false;
 
+	const customsForms = labels.map( ( label ) => label.commercial_invoice_url ).filter( ( url ) => url );
+	if ( customsForms && 0 < customsForms.length) {
+		dispatch(
+			NoticeActions.infoNotice( translate( "Note: A customs form will open in a new tab and must be printed and included on this international shipment." ), {
+			} )
+		);
+		for (const customsFormsUrl of customsForms) {
+			window.open( customsFormsUrl );
+		}
+	}
+
 	api
 		.get( siteId, printUrl )
 		.then( fileData => {


c-shultz (approved):
This looks good to me. After purchasing the label, I see this notification:

image

And I get this in the new tab:

image

It's a little jarring as expected, but I think this balances our needs to find a quicker solution while making sure the very small percentage of users that need the form will see it. 

/cc @aleftick @elizaan36 

This took a little different direction than we had planned. It turns out, we don't know if the invoice printout is needed until _after_ the purchase, so showing a note in the modal before the purchase and print popup is not possible. This change shows the notification and opens the customs invoice in a new tab right after purchase.

c-shultz:
This is not a blocker for merging, but I'm just also noting here that the response for DHL purchases indicates that the form is always required. I see from some DHL literature that we need to be enrolled in Paperless Trade. I asked DHL about enabling this here:
https://docs.google.com/document/d/1P_-ENCEfSqajQXWKeOMMCpwI5wfcVrx6__rawGmuvtg/edit#bookmark=id.3qr7opwx5ihq

c-shultz:
I checked in with DHL. Our account was set up fine. It looks like there is an issue where test labels always indicate that the form is required. When testing real label creation with DHL, it does not display the form for destinations that don't require it.

PR #2220, opened 2020-10-13
Lookup carrier name for email tracking URL

### LINKED ISSUE #2218: Email to customer shows "DHLEXPRESS" instead of "DHL EXPRESS"
image

bborman22:
## Description

The email notification to customers with the shipment tracking information was displaying DHLEXPRESS as one word instead of DHL EXPRESS. This was due to the fact that the carrier name was generated by making the carrier ID uppercase, but the `dhlexpress` carrier ID is all one word for other compatibility reasons. I chose to do a lookup for the service `carrier_name` instead of hardcoding the name to limit the number of places we'd have the carrier name hardcoded. This comes with a slight performance hit, but allows for much easier updates/additions to carrier names.

### Related issue(s)

Fixes #2218 

1. Start with a WC and WCS site set up with DHL Express enabled.
2. Place an order that DHL Express would be able to fulfill (international).
3. Go to the order and purchase the shipping label. Be sure to check the "Notify the customer with shipment details" checkbox.
4. Check the email notification for the Provider name in the tracking section of the email.
5. Confirm the provider name is displayed as DHL Express.

NotificationEmail-Tracking

### Checklist

- [ x ] unit tests

- [ x ] `changelog.txt` entry added

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -1040,7 +1040,8 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			// Generate a table row for each label
 			foreach ( $labels as $label ) {
 				$carrier = $label['carrier_id'];
-				$carrier_label = strtoupper( $carrier );
+				$carrier_service = $this->get_service_schemas_store()->get_service_schema_by_id( $carrier );
+				$carrier_label = ( ! empty( $carrier_service->carrier_name ) ) ? $carrier_service->carrier_name : strtoupper( $carrier );
 				$tracking = $label['tracking'];
 				$error = array_key_exists( 'error', $label );
 				$refunded = array_key_exists( 'refund', $label );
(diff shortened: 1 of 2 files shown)


woocommerce-services.php
```diff
@@ -1040,7 +1040,8 @@ public function add_tracking_info_to_emails( $order, $sent_to_admin, $plain_text
 			// Generate a table row for each label
 			foreach ( $labels as $label ) {
 				$carrier = $label['carrier_id'];
-				$carrier_label = strtoupper( $carrier );
+				$carrier_service = $this->get_service_schemas_store()->get_service_schema_by_id( $carrier );
+				$carrier_label = ( ! empty( $carrier_service->carrier_name ) ) ? $carrier_service->carrier_name : strtoupper( $carrier );
```
frosso, line 1044: (minor) it could happen that `$carrier_service` is returned as `null`.
I think it would be safer if we checked for that case
```suggestion
				$carrier_label = ( ! $carrier_service || empty( $carrier_service->carrier_name ) ) ? strtoupper( $carrier ) : $carrier_service->carrier_name;
```

bborman22: Awesome, thanks for this! I had waffled a bit on this conditional, this was what I was looking for!

frosso: Thanks for considering it! 👍

PR #2223, opened 2020-10-15
Disable refund button for USPS letters.

### LINKED ISSUE #2120: Disable the refund button for USPS letters
Per USPS policy, letters are not eligible for refunds, so we should probably disable the refund button for letters to avoid false expectations to our customers.

I would ping @elizaan36 here because disabling the button just in this case may confuse the store owners. Maybe a message with the reasons why the button is disabled should be displayed instead to avoid confussions.

brezocordero:
## Description
USPS letters are not eligible for refunds as they are untracked.

This PR disables the `Request a Refund` button for USPS labels when the package has the value `is_letter = true` and shows a tooltip message `USPS letters are not eligible for refund`.

This PR adds `is_letter` to label metadata when a label is purchased. Labels purchased before this PR will not have that information and thus the button will not be disabled. As labels are only refundable for a certain amount of time, overtime the difference will disappear.

### Related issue(s)

closes #2120 

1. With WooCommerce Shipping and Tax ready to purchase labels, place an order on your store.
2. Edit the order, click on Purchase a Label, verify the addresses and select a package that is a USPS envelope.
USPSEnvelopes
3. Select one of USPS services and finish purchasing the label
USPSCarrier

4. On the order notes, click on the ellipsis for the label you just purchased and verify the "Request a Refund" button is disabled.
DisabledRefund

5. Hover over `Request a Refund`, verify there is a tooltip message: `USPS letters are not eligible for refund`
Tooltip

6. Purchase a label for another carrier, like DHL Express, using an envelope package. Verify the button is not disabled.
RequestRefund

### Checklist

- [x] unit tests

### DIFF
--- a/classes/class-wc-rest-connect-shipping-label-controller.php
+++ b/classes/class-wc-rest-connect-shipping-label-controller.php
@@ -106,6 +106,8 @@ class WC_REST_Connect_Shipping_Label_Controller extends WC_REST_Connect_Base_Con
 				$label_meta[ 'package_name' ] = __( 'Unknown package', 'woocommerce-services' );
 			}
 
+			$label_meta['is_letter'] = $package['is_letter'];
+
 			$product_names = array();
 			$product_ids = array();
 			foreach ( $package[ 'products' ] as $product_id ) {
--- a/classes/class-wc-rest-connect-shipping-label-controller.php
+++ b/classes/class-wc-rest-connect-shipping-label-controller.php
@@ -106,7 +106,7 @@ class WC_REST_Connect_Shipping_Label_Controller extends WC_REST_Connect_Base_Con
 				$label_meta[ 'package_name' ] = __( 'Unknown package', 'woocommerce-services' );
 			}
 
-			$label_meta['is_letter'] = $package['is_letter'];
+			$label_meta[ 'is_letter' ] = isset( $package[ 'is_letter' ] ) ? $package[ 'is_letter' ] : false;
 
 			$product_names = array();
 			$product_ids = array();
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
 = 1.25.1 - 2020-XX-XX =
 * Fix   - Update carrier name in tracking notification email
 * Add   - Add pre-commit and pre-push git hooks for linting and unit tests.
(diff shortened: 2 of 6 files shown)


classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -106,6 +106,8 @@ public function post( $request ) {
 				$label_meta[ 'package_name' ] = __( 'Unknown package', 'woocommerce-services' );
 			}
 
+			$label_meta['is_letter'] = $package['is_letter'];
```
bborman22, line 109: Two small changes here. One is just stylistic with spaces between bracketed strings and the other is to just assign a default value to avoid any undefined warnings etc. I went with null here based on our discussion, but don't think false would be bad here as it would be functionally the same and is nicer to be able to know it's always a boolean. This update should require the tests to be updated from undefined to null (or false if you wanted to go that route).

```suggestion
			$label_meta[ 'is_letter' ] = isset( $package[ 'is_letter' ] ) ? $package[ 'is_letter' ] : null;
```

bborman22 (changes requested):
This worked as expected for me. Just had the one note about setting a default value for `is_letter`. And I was wondering if it would be possible to put a test in:

client/extensions/woocommerce/woocommerce-services/views/shipping-label/test/label-item.js

for this? Overwrite the props with `isLetter: true, carrierId: 'usps'` and then look for the Tooltip or the class that disables the link? Otherwise, it all looked good and worked as expected!

classes/class-wc-rest-connect-shipping-label-controller.php
brezocordero: Done! I used `false` as default and updated the test with a mix of `true`, `false` and `undefined`.

brezocordero:
> And I was wondering if it would be possible to put a test in:
client/extensions/woocommerce/woocommerce-services/views/shipping-label/test/label-item.js

Thanks for pointing that file out @bborman22 ! I had overlooked it. I have added tests to it.

bborman22 (approved):
Confirmed refund isn't available for USPS isLetter packages and was still available for other methods. Tests all passed.

LGTM!

Issue #2055, opened 2020-06-17
Utilize E2E test setup from WC core


c-shultz:
WC core now provides a Docker-based E2E test environment. We should take advantage of this to provide reproducible environment for our E2E tests to run in. 

https://www.npmjs.com/package/@woocommerce/e2e-environment

c-shultz:
@Ferdev has a good start on this on this draft PR: https://github.com/Automattic/woocommerce-services/pull/2224

c-shultz:
I'll move this to the cooldown queue, to prioritize later.

PR #2224, opened 2020-10-16
Integrate `@woocommerce/e2e-environment` package for the e2e tests

Ferdev:
Closes: #2055 

This pull request refactors the e2e test suite and makes use of the new core package `@woocommerce/e2e-environment`.

This package provides a docker environment that allows us to standardize our test environment by using a docker container, which contains the required wordpress install our test suite needs to be able to run.

It also provides all the jest and puppeteer config our test suite needs so it can use the new docker container. 

Four new commands have been added to manage the e2e test suite:

- `npm run docker:up` and `npm run docker:down` Commands to start and stop the docker container.
- `npm run test:e2e` runs the test suite. It's needed to have the docker container running prior to execute this command.
- `npm run test:e2e-dev` runs the test suite on visible web browser (non-headless mode). It's needed to have the docker container running prior to execute this command.

### Pending work

Currently, the test suite doesn't work because the Jetpack connection that WooCommerce Shipping & Tax needs must be mocked. The best way to do this would be activating as a plugin the php file `tests/e2e/config/travis/wc-services-testing-helper.php`.

In order to do that, the docker container config files must be updated to activate this file.

### Changes in test fixtures helpers

The official `@woocommerce/woocommerce-rest-api` package has been replaced by `wpapi v 1.2.1` which, along with the plugin https://github.com/WP-API/Basic-Auth/archive/master.zip, allows to use basic auth to consume the WooCommerce REST API endpoints, instead of the more complex Oauth credentials.


Ferdev:
The `@woocommerce/e2e-environment` package has been upgraded to `v0.1.6` recently and among the changes, there's no need now to include the `docker-compose.yaml` file with the plugin. 

See https://github.com/woocommerce/woocommerce/tree/master/tests/e2e/env

tests/e2e/specs/admin/4-refund-label.test.js
```diff
@@ -53,10 +57,10 @@ describe( 'Refund shipping label', () => {
 	it( "doesn't work for expired labels", async () => {
 
 		await withExpiredShippingLabelAndOrder( async ( { order } ) => {
-			console.log( `# Opening order page for order #${order.id}` );
+			// console.log( `# Opening order page for order #${order.id}` );
```
c-shultz, line 60: Should we just delete these commented out lines (if they're really not needed)?

FangedParakeet:
### What do we have here...

These changes add an implementation for WooCommerce's end-to-end testing environment package within our plugin that allows us to run our own end-to-end tests inside a customisable Dockerized container. The container and the E2E tests run as separate scripts: the first script creates a Docker container running WordPress, WooCommerce, Jetpack, and WCS on an instance running on `http://localhost:8084`; the second script allows you to run the E2E tests within this pre-packaged environment. 

To this end, there are four new scripts added that will be described in more detail below. For the Docker container we have...
- `npm run docker:up`
- `npm run docker:down`

And for the tests we have...
- `npm run test:e2e`
- `npm run test:e2e-dev`

### Requirements

More comprehensive setup requirements can be found [here](https://github.com/woocommerce/woocommerce/tree/master/tests/e2e).

**Node, NPM, and NVM**
Obviously.

**Docker**
[Mac installation](https://docs.docker.com/docker-for-mac/install/)
[Windows installation](https://docs.docker.com/docker-for-windows/install/)

Also, here are my recommended Docker configuration settings.

```json
{
  "dns": [
    "YOUR.LOCAL.DNS.ADDRESS",
    "8.8.8.8"
  ],
  "features": {
    "buildkit": true
  },
  "experimental": true
}
```

**Jest**
The [e2e-environment package](https://www.npmjs.com/package/@woocommerce/e2e-environment) uses Jest to run its E2E tests. You will just have to run `npm install jest --global` once before you get started and you'll be fine.

### Testing steps

1. Pull this branch and run `npm install` followed by `npm install jest --global`, if you haven't done so already.
2. Our Docker container will contain a WordPress installation on a PHP image with MariaDB running as the local database. Before we set up the container, we can set [these three variables](https://github.com/woocommerce/woocommerce/tree/master/tests/e2e#how-to-run-tests-using-custom-wordpress-php-and-mariadb-versions) respectively to set your WordPress, PHP, and MariaDB versions:
 - `WP_VERSION`
 - `TRAVIS_PHP_VERSION`
 - `TRAVIS_MARIADB_VERSION`
 The default value for each is `latest`. I would recommend `WP_VERSION` >= `5.6`, because we will be installing the latest version of Jetpack on the instance and this is its WP minimum version requirement. I would also recommend `TRAVIS_MARIADB_VERSION` ~= `10.5.5`. The latest available version of MariaDB is currently `10.5.9` and when I was testing with this, for some reason WordPress was no longer able to communicate with the database.
 3. To start up the docker container, run `npm run docker:up`. This will create a production build of the plugin and then copy all the plugin files into the container. Once this is complete, you should see something like the following.
 ```bash
 Successfully built 08ea468dff6cd9dc5fd7aa6577029f49e00338f1ccaeac78b63ac0eca59cff2e
Creating woocommerce-services_php ... done
Creating woocommerce-services_db  ... done
Creating woocommerce-services_wordpress-www ... done
Creating woocommerce-services_wordpress-cli ... done
Docker exit code: 0
```
To view the Docker processes that have started up, run `docker ps`. You should see something similar to these three services.
```bash
CONTAINER ID   IMAGE                           COMMAND                  CREATED          STATUS          PORTS                  NAMES
03ac4e4a0253   e2e-environment_wordpress-cli   "entrypoint.sh"          15 seconds ago   Up 14 seconds                          woocommerce-services_wordpress-cli
d00bbea8c846   wordpress:5.6                   "docker-entrypoint.s…"   15 seconds ago   Up 15 seconds   0.0.0.0:8084->80/tcp   woocommerce-services_wordpress-www
6281f649e1d5   mariadb:10.5.5                  "docker-entrypoint.s…"   16 seconds ago   Up 15 seconds   3306/tcp               woocommerce-services_db
```
The first process there is a WordPress CLI instance that runs the initialization commands, so that the WordPress instance will be populated with everything we need. You can find these bash/CLI commands at `tests/e2e/docker/initialize.sh`. Moreover you can find the default user accounts and settings stored in `tests/e2e/config/default.json`. To follow these initialization commands as they run, run `docker logs -f` using the container ID for the `e2e-environment_wordpress-cli` image. So, for instance, to follow the initialization of the container above you would enter the following.
```bash
docker logs -f 03ac4e4a0253
```
Once the initialization is complete, hopefully without any errors, you should see something like the following the lines at the end.
```bash
Success: Created post 9.
Visit http://localhost:8084
```
You can now visit `http://localhost:8084` to explore our completely configured WordPress/WCS instance at last! Phew.

4. The E2E tests can either be run in headless mode by running `npm run test:e2e` or you can follow them in a spawning Chromium browser as they run by running `npm run test:e2e-dev`. You might notice that they don't run especially quickly: this is by design, to ensure tests don't timeout while waiting for buttons to appear. You can find this configuration in `tests/e2e/config/jest-puppeteer.config.js`. The key values here are `slowMo`, which sets the millisecond delay between test steps (the bigger this number, the slower the tests will run) and `runInBand`, which forces the tests to run sequentially instead of asynchronously. Either way, once the tests have finished running, you should see something like the below.
```bash
Testing URL: http://localhost:8084

 PASS  ../../../tests/e2e/specs/admin/1-activate-extension.test.js (9.681s)
  Store admin can login and make sure WooCommerce Shipping & Tax extension is activated
    ✓ Can activate WooCommerce Shipping & Tax extension if it is deactivated (7816ms)

 PASS  ../../../tests/e2e/specs/admin/2-shipping-settings.test.js (127.704s)
  Saving shipping label settings
    ✓ Can toggle shipping labels (11995ms)
    ✓ Should be able to select a different paper size (5429ms)
  Shipping label payment method
    ✓ should turn on request interception after all credit card test ran (521ms)
    ✓ should show "Add a credit card" button if Wordpress has no credit card (7630ms)
    ✓ should show "Choose a different card" button if Wordpress.com has a credit card (3247ms)
    ✓ should show 2 credit cards with the right information if Wordpress.com has 2 credit cards (6467ms)
    ✓ should have no default card checked if there are multiple cards in Wordpress.com (13288ms)
    ✓ Should show the correct email receipts message (841ms)
    ✓ should turn off request interception after all credit card test ran (519ms)
  Packaging
    ✓ Can add package (22828ms)
    ✓ Can edit package (23141ms)
    ✓ Can delete package (12096ms)
    ✓ Can select and add service package (10826ms)

 PASS  ../../../tests/e2e/specs/admin/3-create-shipping-label.test.js (43.338s)
  Create shipping label
    ✓ should create a new shipping label (31153ms)

 PASS  ../../../tests/e2e/specs/admin/4-refund-label.test.js (24.847s)
  Refund shipping label
    ✓ works if the label hasn't been used (14378ms)
    ✓ doesn't work for expired labels (9144ms)

Test Suites: 4 passed, 4 total
Tests:       17 passed, 17 total
Snapshots:   0 total
Time:        206.067s
Ran all test suites.
Jest exit code: 0
```
5. Once you're done playing around, you can wrap things up and bring the container down with `npm run docker:down`. That's it!

tests/e2e/specs/admin/4-refund-label.test.js
FangedParakeet: I am definitely not a fan of leaving in commented out code; however, these `console.log`s are pretty invaluable when it comes to debugging these tests and I left that in because doing this might not be immediately obvious to the next person who has to look at this and modify these tests (or at least it wasn't to me). Despite that defence, I will gladly remove them if you think so, because they are currently inactive.

c-shultz:
It all ran successfully for me! I just installed the latest version of Docker (in case that matters).

Setup:
`export WP_VERSION=5.6`
`export TRAVIS_MARIADB_VERSION=10.5.5`
`npm run docker:up`

And then ran it with `npm run test:e2e`
image


🎆

tests/e2e/specs/admin/4-refund-label.test.js
FangedParakeet: Actually, I changed my mind. Just removed these goofy comments. Good riddance!

c-shultz: Haha. Sounds good! If there was something really special worth keeping maybe a comment explaining why it was left there would be good.

c-shultz:
Since the E2E build is failing, I would expect the checks to say they've failed instead of all passing 
image

FangedParakeet:
```
ERROR: for wordpress-www  Cannot start service wordpress-www: driver failed programming external connectivity on endpoint woocommerce-services_wordpress-www (bedbbfbfa3bcd80046700dce2324b6df6a0325d28a7a8ce76cf842b73b031167): Error starting userland proxy: listen tcp4 0.0.0.0:8084: bind: address already in use
Encountered errors while bringing up the project.
Docker exit code: 1
```

It looks like Docker is exiting with a return value of `1`, but the GitHub Actions step is still returning `0` for some reason. I think I'll have to add another step in between to confirm that the container has been initialised before running the tests. In this build, if the container doesn't launch, the tests don't even run, so `npm run test:e2e` just exits normally. Hopefully, when the tests actually run, if there are failures, that will cause the step to exit with failure as well.

harriswong:
👀

changelog.txt
```diff
@@ -2,11 +2,12 @@
 
 = 1.25.10 - 2021-xx-xx =
 * Tweak - Removes deprecated Jetpack constant JETPACK_MASTER_USER
-* Fix	  - Ensure status page is displayed on new WC navigation menu.
+* Fix	- Ensure status page is displayed on new WC navigation menu.
```
harriswong, line 5: nice

tests/bin/get-base-url.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable no-console, import/no-nodejs-modules */
+const path = require( 'path' );
+const fs = require( 'fs' );
+
+const testConfigFilePath = path.resolve( __dirname, '../e2e/config/default.json' );
+const testConfigFile = fs.readFileSync( testConfigFilePath );
+const testConfig = JSON.parse( testConfigFile );
+
+console.log( testConfig.url.slice( 0, -1 ) );
```
harriswong, line 9: do we still need this log?

classes/class-wc-connect-account-settings.php
```diff
@@ -29,7 +29,7 @@ public function get() {
 		}
 
 		$master_user    = WC_Connect_Jetpack::get_master_user();
-		$connected_data = WC_Connect_Jetpack::get_connected_user_data( $master_user->ID );
+		$connected_data = $master_user ? WC_Connect_Jetpack::get_connected_user_data( $master_user->ID ) : false;
```
harriswong, line 32: Is this to handle jetpack not being connected? Since WCS requires jetpack, should we return false on the entire function? Or return payment information + blanking out user detail? I don't think there would be any payment methods if it's not connected.

classes/class-wc-connect-settings-pages.php
```diff
@@ -86,13 +86,13 @@ public function output_shipping_settings_screen() {
 			);
 
 			$carriers_response = $this->api_client->get_carrier_accounts();
-			if ( ! is_wp_error( $carriers_response ) && $carriers_response ) {
+			if ( ! is_wp_error( $carriers_response ) && (array) $carriers_response ) {
```
harriswong, line 89: Why do we need to cast the response object to an array for a boolean check here?

tests/bin/wait-for-e2e-build.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+# Max amount of time to wait for the Docker container to be built
+# Allowing 30 polling attempts, 10 seconds delay between each attempt
+MAX_ATTEMPTS=30
```
harriswong, line 4: Maximum 5 minutes timeout on the pipeline for this job seems reasonable.

tests/e2e/config/jest-puppeteer.config.js
```diff
@@ -0,0 +1,10 @@
+const { useE2EJestPuppeteerConfig } = require( '@woocommerce/e2e-environment' );
+
+const puppeteerConfig = useE2EJestPuppeteerConfig( {
+	launch: {
+		slowMo: 100,
```
harriswong, line 5: Do we still need `slowMo`?

tests/e2e/config/jest.config.js
```diff
@@ -15,18 +17,19 @@ module.exports = {
...
-	setupFilesAfterEnv: [
-		'<rootDir>/jest.setup.js',
-		'<rootDir>/jest.test.failure.js',
-		'expect-puppeteer',
-	],
+	// setupFilesAfterEnv: [
+	// 	'<rootDir>jest.setup.js',
+	// 	'<rootDir>jest.test.failure.js',
+	// 	'expect-puppeteer',
+	// ],
```
harriswong, line 31: Now that the test works, maybe we can remove these comments.

tests/e2e/utils/components.js
```diff
@@ -56,4 +56,28 @@ const loginAndConfirmExtensionActivation = async () => {
...
+			// Delete package
+			await page.waitForSelector( '.packages__add-edit-title.form-section-heading', { text: 'Edit package' } );
+			await expect( page ).toClick( '.button.packages__delete.is-scary.is-borderless', { text: 'Delete this package' } );
+		}
+		editButton = await page.$( '.button.is-compact', { text: 'Edit' } );
+	} while ( editButton );
+};
+
+// Click save and wait until it's saved.
+const saveAndWait = async () => {
```
harriswong, line 78: Let's use import `{ saveAndWait } from '../../utils/components';` for this function.

harriswong (commented):
I have a comment regarding the change on `classes/class-wc-connect-account-settings.php`. Otherwise looks good to me.

tests/bin/get-base-url.js
FangedParakeet: Yes, this is required, because this isn't actually a log, but an output that is read by the [bash script inside the `npm run docker:ping` command](https://github.com/Automattic/woocommerce-services/blob/dc6edb776bae802dfa56259d9a6cd1c2c12361c6/tests/bin/wait-for-e2e-build.sh#L11). This JS script reads the [JSON file used to initialise the WP Docker container](https://github.com/Automattic/woocommerce-services/blob/dc6edb776bae802dfa56259d9a6cd1c2c12361c6/tests/e2e/config/default.json#L2) and outputs the base URL, which should be http://localhost:8084. This value should be pretty constant, but I didn't want to hard-code it and needed to write this part in JS, because parsing JSON in bash is not a very fun time.

classes/class-wc-connect-account-settings.php
FangedParakeet: When I used [this file](https://github.com/Automattic/woocommerce-services/blob/dc6edb776bae802dfa56259d9a6cd1c2c12361c6/tests/e2e/config/travis/wc-services-testing-helper.php) to mock out Jetpack, I discovered that `WC_Connect_Jetpack::get_master_user()` would start returning either undefined or an empty array/object--I can't remember which--but basically this meant that any reference to its member variables (e.g. `$master_user->ID`, `$master_user->display_name`) threw a PHP error notice on the WCS settings page stating the index was undefined. I wasn't entirely sure of all the places where this function is used, so I just handled the error and returned empty strings for those variables, but @harriswong let me know if you think it would be better to just return `false` for the entire function here in that case.

classes/class-wc-connect-settings-pages.php
FangedParakeet: `$this->api_client->get_carrier_accounts()` returns a JSON decoded object representing the server response for this request. Presumably due to the mock, this was somehow returning an empty object, which meant that `( ! is_wp_error( $carriers_response ) && $carriers_response )` evaluated to `true`, because `$carriers_response` wasn't an `WP_Error` and it wasn't empty/falsey. Then the next line, `$carriers_response->carriers`, threw a PHP error notice for undefined index. The array cast here is for that scenario and prevents the error from occurring and consequently that notice from being shown.

tests/e2e/config/jest-puppeteer.config.js
FangedParakeet: Irritatingly, the `slowMo` value here is actually required. The `slowMo` basically sets a delay in between lines of execution while the E2E tests are running. Without the `slowMo` value in both headless and non-headless mode, the tests (particularly package tests) will fail periodically, because elements are not added to the DOM in time. In particular, regarding the package tests, a new package will be created and to confirm that it has been created correctly Jest will check the DOM for the new row containing the package before it has been added to the DOM and the test will fail. 

As an aside, this was particularly painful to debug and diagnose, because stepping through using the debugger would show no issues, but then the tests would fail from time to time when it came to running them at normal speed.  There may be room to decrease this number to make the tests run a bit quicker, but I found that keeping it 💯 hit the sweet spot, where the tests would consistently not fail randomly.

tests/e2e/config/jest.config.js
FangedParakeet: Done! Just removed all these redundant eyesores.

tests/e2e/config/jest-puppeteer.config.js
harriswong: Thanks for the explanation!

classes/class-wc-connect-settings-pages.php
harriswong: Instead of `(array) $carriers_response ) `, what do you think if we check the property directly instead? `empty($carriers_response->carriers)`, assuming `$carriers_response` is not `null`. If we want to check `null`, then we would need to do
both `empty($carriers_response)` ` empty($carriers_response->carriers)`

classes/class-wc-connect-account-settings.php
harriswong: Is there a way to get `WC_Connect_Jetpack::get_master_user()` to return a mock master user when we are testing? 

> let me know if you think it would be better to just return false for the entire function here in that case.

Here are my thoughts. This change seems to be here because we need it to get E2E test to work. I generally prefer not to modify source code because of test; if possible, I would prefer to modify/hack the tests instead. The change itself will not prevent an error to be thrown, would that change a 500 to a 200? Would this allow account settings to be displayed/returned when it shouldn't? If the latter are true, then maybe there are some risks exposing data. That said, WCS requires jetpack to connect; without Jetpack, this code should not run so the change here shouldn't affect anything.

If we can modify test code instead of this, it would be great; otherwise, I think this is okay.

classes/class-wc-connect-settings-pages.php
FangedParakeet: Updated this to remove that conniving cast and followed your suggestion!

classes/class-wc-connect-account-settings.php
FangedParakeet: > I generally prefer not to modify source code because of test; if possible, I would prefer to modify/hack the tests instead.

Totally agree with this sentiment. I had a bit of a play around today, but I managed to add a few lines to [`wc-services-testing-helper.php`](https://github.com/Automattic/woocommerce-services/blob/add/e2e-docker-container/tests/e2e/config/travis/wc-services-testing-helper.php#L27) to allow me to mock out these fields and revert this file back to the way it was before!

harriswong (approved):
Last 4 commits look good, double checked E2E tests passed! 👍
image
  
There is a conflict with changelog, once that's fixed, feel free to 🚢

PR #2229, opened 2020-10-22
Fallback to open PDF in new tab if Safari does not print natively

harriswong:
### Description
On some iPad devices, `navigator.userAgent` stopped including `iPad` in the string. This causes our logic to fallback on using `print()`, resurfacing [this problem](https://github.com/Automattic/woocommerce-services/issues/1166). 

### Related issue(s)
Closes https://github.com/Automattic/woocommerce-services/issues/1166

1. Use an iPad with Safari. Use the iPad (7th generation) if possible. 
image
1a. Alternatively, use browserstack and choose these iPad:
image
2. Go to your store and click re-print label in the "Shipment Tracking" block
image
3. The label should open in a new tab instead of closing the modal itself. 

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,6 +6,7 @@
 * Fix   - Update carrier name in tracking notification email
 * Add   - Add pre-commit and pre-push git hooks for linting and unit tests.
 * Add   - Disable refunds for USPS letters.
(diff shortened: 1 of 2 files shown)


frosso (approved):
Good catch! 👍

PR #2230, opened 2020-10-23
Add ZIP code validation to UPS signup form.

### LINKED ISSUE #2180: Limit UPS registration page's zip code to 5 digit only if address is US
### Summary
UPS registration fails when zip code is in the format of `12345-9876`. It works when it is just 5 digits, ie. `12345`

### Proposed change
If the address is in US, then we should accept only 5 digits zip code
image

tommyshellberg:
## Description
Provide some client-validation for the postal code field on the UPS signup form.

### Related issue(s)

Closes #2180 

GIF of it working in action:

Screen Capture on 2020-10-23 at 14-53-27

Verify also that it disables the Connect button when the ZIP/Postal code is incorrectly formatted(not a 5 digit number):
Screen Shot 2020-10-23 at 14 58 47

### Checklist

- [ ] unit tests

Not yet - but a full e2e test will be a great idea when we know how this form will look in its final form(it may be abstracted to allow for additional carriers).

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,9 +1,10 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
 = 1.25.1 - 2020-XX-XX =
-* Tweak - DHL refund days copy adjustment
+* Tweak - DHL refund days copy adjustment.
+* Tweak - Add ZIP code validation to UPS(beta) signup form.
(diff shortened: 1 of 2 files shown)


harriswong:
👀

harriswong (approved):
Tested the UPS registration form. I was able to register UPS with the valid information. 👍

PR #2238, opened 2020-10-28
Allows package dimension input to begin with zero, still validates positive values

### LINKED ISSUE #2232: Input validation does not allow dimensions starting with a zero
When adding a new package, the dimension inputs do not allow entry of values starting with zero (e.g. 0.1):
image

 The total package weight does the same thing:
image

It's good to have input validation that prevents the end value of being zero, but disallowing values that start with zero feels pretty awkward/buggy.

FangedParakeet:
## Description

When adding a new package, the inputs for package dimensions will not allow you to start entering a number beginning with zero (e.g. decimals beginning with "0."). This is because there are `onChange` validation hooks on the inputs that clear out the field if the current value is non-positive (0 or negative). In addition, when submitting the form, the three dimension values are validated together as a string (`"{length} x {width} x {height}"`) against the regex `^(\\d+|(?:\\d*\\.\\d+)) x (\\d+|(?:\\d*\\.\\d+)) x (\\d+|(?:\\d*\\.\\d+))$`. This regex will match any non-negative values _including_ zero, so the validation relies on the hooks on the input field to prevent zero from being entered.

My solution is to loosen the validation in the `onChange` hooks, preventing negative values, but allowing zero. To maintain the correct validation on submission, we need to now validate on submission that the field values are non-zero (and non-negative). The regex appears to originate from some external package schema and besides, it didn't seem entirely appropriate to rely on string validation for a numeric value, so I simply added in one more function (`validateDimensions`) to double-check and confirm this validation before submitting.

This also occurs on the input field for _Total Weight (with package)_ on the create shipping label dialog, but this is because the input validation clears the field if the value is not truthy only allowing positive and non-zero values.

### Related issue(s)
Closes #2232 

Go to **WooCommerce -> Settings -> Shipping -> WooCommerce Shipping**.
Click **Add package** under **Packaging**.
Enter decimals with a leading zero into the dimensions fields.
Click **Add package** to submit the dialog and check validations.

Add a package dialog

Go to **WooCommerce -> Orders**.
Select an order.
Click **Create Shipping Label**/**Create new label** button.
Edit **Total Weight (with package)** field.

Create shipping label dialog

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,6 +8,7 @@
 * Add   - Add pre-commit and pre-push git hooks for linting and unit tests.
 * Add   - Disable refunds for USPS letters.
 * Fix   - Issue with printing labels in some iOS devices through Safari.
+* Fix   - UI fix for input validation for package dimensions and weights.
 
 = 1.25.0 - 2020-10-13 =
 * Fix   - UPS connect redirect prompt
--- a/client/extensions/woocommerce/woocommerce-services/views/packages/input-filters.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/packages/input-filters.js
@@ -41,7 +41,8 @@ const checkDimension = dimension => {
 		return '';
 	}
 
-	if ( 0 >= dimension ) {
+	// Allow users to type `0`, but not negative numbers
+	if ( 0 > dimension ) {
 		return '';
 	}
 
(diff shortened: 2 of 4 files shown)


tommyshellberg (approved):
Looks good to me, nice first PR 👍

tommyshellberg:
P.S. @FangedParakeet:

sometimes GitHub automatically deletes the branch after merging but sometimes not so it's good to double-check and delete the branch once it's merged.

PR #2241, opened 2020-10-29
Remove second parameter from getenv for php5.3 compatability

### LINKED ISSUE #2025: Warning: getenv() expects exactly 1 parameter, 2 given
`Warning: getenv() expects exactly 1 parameter, 2 given in woocommerce-services/woocommerce-services.php on line 55`

This is caused by https://github.com/Automattic/woocommerce-services/blob/develop/woocommerce-services.php#L55 and the second parameter is only available to PHP `5.5.38, 5.6.24, 7.0.9`+. 

Since we support PHP 5.3 (https://github.com/Automattic/woocommerce-services/blob/develop/readme.txt#L5), we should handle without the second parameter.

Although we support PHP 5.3 (https://github.com/Automattic/woocommerce-services/blob/develop/readme.txt#L5), WooCommerce itself is going to be bumping PHP version to 7.0+ (https://github.com/woocommerce/woocommerce/blob/master/readme.txt#L6). Should we also bump ours?

FangedParakeet:
## Description

Our extension supports a minimum version of PHP 5.3; in versions of PHP less than 5.5 the function `getenv` only supports a single parameter: there is one instance of `getenv` with two parameters in our extension and this will cause a warning. 

`Warning: getenv() expects exactly 1 parameter, 2 given in woocommerce-services/woocommerce-services.php on line 55`

This fix removes that second parameter!

### Related issue(s)
Closes #2025 

As of WordPress 5.2, WordPress requires PHP 5.6 or greater, so you will need a version older than this to see this warning.

As of WooCommerce 3.8, WooCommerce also requires PHP 5.6 or greater, so you will need to use an older version of this extension.

As of Jetpack 8.4, Jetpack requires WordPress 5.3 or greater, so you will need to use an older version of this extension too.

If you can find the sweet spot for all of these deprecated tools to work together, you will be able to test to confirm that this warning no longer appears!

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,7 +11,6 @@
 * Fix   - Update carrier name in tracking notification email.
 * Add   - Add pre-commit and pre-push git hooks for linting and unit tests.
 * Add   - Disable refunds for USPS letters.
-* Fix   - Issue with printing labels in some iOS devices through Safari.
 
 = 1.25.0 - 2020-10-13 =
 * Fix   - UPS connect redirect prompt
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,6 +8,7 @@
 * Add   - Add pre-commit and pre-push git hooks for linting and unit tests.
 * Add   - Disable refunds for USPS letters.
 * Fix   - Issue with printing labels in some iOS devices through Safari.
+* Fix   - Prevents warning when using PHP 5.5 or lesser
 
 = 1.25.0 - 2020-10-13 =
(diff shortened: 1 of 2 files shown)


frosso (approved):
Looks straightforward 👍 

You'll want to resolve the merge conflicts and place the changelog entry in the correct version. Otherwise looks good! 💯

frosso:
I forgot to mention - weird that travis hasn't started yet, maybe it needs a little nudge

PR #2244, opened 2020-11-02
Adds more specific regex validation to UPS connect form

### LINKED ISSUE #2136: Connecting a UPS account, in-line form field errors
Noticed some things as I was testing the form field validations that would help people complete the form easier. The current problem is the message, "This field is required" doesn't provide enough information in case they've entered something incorrectly.

For the UPS account number error message (the first field), if the user enters less than 6 digits in the field the error message could say
> Your UPS account number needs to be 6 digits in length

For the Zip code fields, (since these are all US-based users) for anything that's not 5 digits we could also show the error message as:
> The zip code needs to be 5 digits in length 

For the phone number field, for anything less than 10 digits:
> The phone number needs to be 10 digits in length

Alternatively, if we can't get that specific we could use the same error state message as the current one for the website:
> The [form field] format is not valid

image

These would appear only after the user has interacted with the field.

FangedParakeet:
## Description

The current form field validations when a user is connecting their UPS account don't always provide enough information to inform a user on how to correct them: in a few cases, the only validation present was checking non-empty values for required fields and in others the error lacked specificity. 

This PR adds the following validations for fields in this form.

- **UPS account number**: `The UPS account number needs to be 6 digits in length`
- **Phone number**: `The phone number needs to be 10 digits in length`
- **ZIP/Postal Code** (US only): `The ZIP/Postal code needs to be 5 digits in length`
- **UPS Invoice Number**: If present, `The invoice number needs to be 9 or 13 digits in length`
- **UPS Invoice Date**: If present, `The date must be a valid date in the format YYYY-MM-DD`

### Related issue(s)
Closes #2136 

Enable UPS for your site, by adding a row to `wcc_site_flags` with `flag_name` as `can_use_ups` and your site's ID for `siteid`. Refresh the WCS plugin from **WooCommerce > Status > WooCommerce Shipping & Tax** to complete the enablement.
Enable UPS beta

Go to **WooCommerce > Settings > Shipping > WooCommerce Shipping** and from the Carrier Account section at the bottom of the page, select "Connect" next to UPS.
Connect UPS account

Enter values into the fields mentioned above and check the validation.
UPS account form field

### Checklist

- [ ] unit tests

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/carrier-accounts/selectors.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/carrier-accounts/selectors.js
@@ -20,7 +20,7 @@ export const getCarrierAccountsState = ( state, siteId = getSelectedSiteId( stat
 
 const emailRegex = /(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])/;
 const USPostalCodeRegex = /^\d{5}$/;
-const UPSAccountNumberRegex = /^\d{6}$/;
+const UPSAccountNumberRegex = /^[a-zA-Z0-9]{6}$/;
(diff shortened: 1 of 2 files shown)


client/extensions/woocommerce/woocommerce-services/state/carrier-accounts/selectors.js
```diff
@@ -39,17 +43,38 @@ export const getFormErrors = ( state, siteId, carrier ) => {
 		}
 	}
 
+	if ( values.account_number && ! values.account_number.match( UPSAccountNumberRegex ) ) {
+		fieldErrors.account_number = translate( 'The UPS account number needs to be 6 digits in length' );
```
frosso, line 47: The account number can be digits or letters, its length should be max 6 characters.
This is validating only on the digits, it should also include `[a-zA-Z]`

FangedParakeet: Got it, @frosso, I didn't realise this! I was following @elizaan36's note on the linked issue blindly, like a giant worm chasing Kevin Bacon off the edge of a cliff. I've just updated this PR with a new commit that corrects this. You should now be able to see the following error messages: 

Error state
_Error state_

Letters accepted
_Letters accepted_

```
const UPSAccountNumberRegex = /^[a-zA-Z0-9]{6}$/;
```
_New regex_

Also, I've today realised that my GitHub notifications are basically all set wrong, so sorry @frosso for taking so long to notice your feedback! 🤦

frosso: haha no worries! :) The new regex looks good!

frosso (approved):
Looks good to me! 👍

Issue #2236, opened 2020-10-27
Add WooCommerce.com connection information to Connect server requests


bborman22:
To allow the Connect server to check for any WooCommerce.com subscriptions, it will need some of the merchant site's WooCommerce.com connection information when available. The key pieces of information needed are:

- The `woocommerce_helper_data` option `access_token`.
- A signature for a `GET /subscriptions` request to WooCommerce.com. (The following [signature implementation](https://github.com/woocommerce/woocommerce/blob/788969d2bbfba8f7e177b83504e70e98bc1babdf/includes/admin/helper/class-wc-helper-api.php#L70-L114) here can be used as a guide or if possible used directly.)
- The `woocommerce_helper_data` option `site_id`.

A good spot to add this might be in the settings field of the request body. If there is no WooCommerce.com connection information, this settings field should be empty.

The idea is to have all the necessary information to make the same [get_subscriptions](https://github.com/woocommerce/woocommerce/blob/788969d2bbfba8f7e177b83504e70e98bc1babdf/includes/admin/helper/class-wc-helper.php#L1256-L1283) call from the Connect server that the merchant site makes. Then use the `woocommerce_helper_data` option `site_id` to confirm an [active subscription for that site](https://github.com/woocommerce/woocommerce/blob/788969d2bbfba8f7e177b83504e70e98bc1babdf/includes/admin/helper/class-wc-helper.php#L120) all from the Connect server.

bborman22:
Cleaning out some stragglers, marking as closed since all related PRs are merged and issue marked as waiting for demo.

PR #2245, opened 2020-11-04
Add/issue 2236

tommyshellberg:
## Description
Add additional WC Helper connection information globally to the body of PUT & POST requests so that WCS can use this information for its /subscriptions/ endpoint. 

### Related issue(s)

#2236 

Testing is easier with the following branch for WCS:

1620-gh-Automattic/woocommerce-connect-server

- With this branch active, you should be able to inspect any of the POST/PUT requests to the WCS server and ensure that the new fields are added(access_token and site_id) to the payload sent to WCS:

Screen Shot 2020-11-03 at 3 53 19 PM

Test that the signature is sent in the headers to WCS:
Screen Shot 2020-11-03 at 3 52 52 PM

You can also verify that the $data array passed into the signature generator contains the correct information:
Screen Shot 2020-11-03 at 4 14 10 PM

Since there are no consuming clients for this /subscriptions/ endpoint it's not super easy to test. There are a couple of things I did:

- Go into core [WC Helper code ](https://github.com/woocommerce/woocommerce/blob/788969d2bbfba8f7e177b83504e70e98bc1babdf/includes/admin/helper/class-wc-helper-api.php#L94)and log the generated signature when requesting subscriptions from WCCOM. This happens when you go to **WooCommerce->Extensions->WooCommerce.com Subscriptions**. You may have to click 'Update' to force a refresh and API call.

- With WS&T plugin active you basically get a call to the /services/ endpoint every time you open a settings page as admin. So you can make a call to the subscriptions getter method and pass the request body to make a call on every page load, here:

https://github.com/Automattic/woocommerce-services/blob/2b08d7cc61e071d383b0c511589cfec0f4e7a530/classes/class-wc-connect-api-client.php#L39

`$subs_response = $this->get_wccom_subscriptions( $this->request_body() );`

Hacky for sure but it's not clear how/where we will actually consume this endpoint yet.

### Checklist

- [ ] unit tests


classes/class-wc-connect-api-client.php
```diff
@@ -436,30 +445,42 @@ protected function request_body( $initial_body = array() ) {
...
 				'settings' => array(),
 			);
-			$body = array_merge( $default_body, $initial_body );
+			$body         = array_merge( $default_body, $initial_body );
+
+			$helper_auth_data = WC_Connect_Functions::get_wc_helper_auth_info();
+			$access_token = '';
+			$site_id      = '';
+
+			if ( ! is_wp_error( $helper_auth_data ) ) {
```
tommyshellberg, line 454: Because the get_wc_helper_auth_info() method can return a WP_Error I decided to set the token and site id to default empty string values and only grab their values from `$helper_auth_data` if the `get_wc_helper_auth_info()` method _doesn't_ return a WP_Error. If there's a cleaner way to do this I'm all ears. :)

classes/class-wc-connect-api-client.php
```diff
@@ -530,6 +554,21 @@ protected function authorization_header() {
 			return $authorization;
 		}
 
+		protected function request_signature_wccom( $token_secret ) {
+			$method = strtoupper( $_SERVER['REQUEST_METHOD'] );
+			$data = array(
+				'host'        => 'woocommerce.com', // host URL.
+				'request_uri' => '/wp-json/helper/1.0/subscriptions', // endpoint URL.
```
tommyshellberg, line 561: The request URI is hard-coded right now since we're only working with the /subscriptions/ endpoint, is there a way we can make this dynamic based on the endpoint being called? The trouble I had with that is there is no reference to an endpoint at this point.

classes/class-wc-connect-api-client.php
```diff
@@ -483,6 +504,9 @@ protected function request_headers() {
 			$headers[ 'Content-Type' ] = 'application/json; charset=utf-8';
 			$headers[ 'Accept' ] = 'application/vnd.woocommerce-connect.v' . static::API_VERSION;
 			$headers[ 'Authorization' ] = $authorization;
+
+			$wc_helper_auth_info = WC_Connect_Functions::get_wc_helper_auth_info();
+			$headers[ 'X-Woo-Signature' ] = $this->request_signature_wccom( $wc_helper_auth_info['access_token_secret'] );
```
bborman22, line 509: Should there be an is_wp_error check here as well?

classes/class-wc-connect-api-client.php
bborman22: I think having the URI hard-coded is fine for now, but was wondering if you had any thoughts about changing the parameters for this function to include the method, URI, secret, and body? A couple notes on this function to see any thoughts you'd have on them:

- In getting the method this way, wouldn't it use the method that the request it is currently processing was requested with? I'm thinking this signature should be generated without any consideration for what the request to the Connect server is and only focused on what the eventual request to WCCOM will be.
- So with that in mind, I was thinking the $data['body'] shouldn't be set based on the Connect server request body, but rather should be passed in to the function explicitly. For example, we might eventually make a POST request to WCCOM with a body of `{ product_id: 1234 }` or something like that, so we wouldn't want the rest of the Connect server request body included in the signature.
- I was wondering if the host could be pulled from this? https://github.com/woocommerce/woocommerce/blob/6e5cc0c2f2e2651b722f3806fa7c9e624f5b9204/includes/admin/helper/class-wc-helper-api.php#L32-L34 This would make it a little easier to test WCCOM in different environments.

classes/class-wc-connect-api-client.php
bborman22: This makes sense to me!

bborman22 (commented):
Just had a couple of notes mostly about the signature generation function, but I tested the token and signature getting passed along to the Connect server and the subscription request worked! As we discussed briefly, I'm still taking a look at the ability to test some of this, but wanted to get these notes in first.

classes/class-wc-connect-api-client.php
tommyshellberg: Good point. I think I avoided it at first because I was worried about preventing requests that did _not_ need the Helper auth info. But I don't see a case where we wouldn't want the auth info.

classes/class-wc-connect-api-client.php
tommyshellberg: Agreed on all of these points!

Now passing in the endpoint and method. 

>I'm thinking this signature should be generated without any consideration for what the request to the Connect server is and only focused on what the eventual request to WCCOM will be.

You're right; we care about eventually sending a GET request to WCCOM so let's pass 'GET' as the method rather than reading the request made to WCS since it's actually a POST. That was a bug for sure.

>So with that in mind, I was thinking the $data['body'] shouldn't be set based on the Connect server request body, but rather should be passed in to the function explicitly.

Also agreed, this is done now.

Also using WC_Helper_API::url( $endpoint ) to generate the request url. Should be much less brittle now. :)

classes/class-wc-connect-api-client.php
```diff
@@ -530,6 +557,22 @@ protected function authorization_header() {
...
+			$request_url = WC_Helper_API::url( $endpoint );
+
+			$data = array(
+				'host'        => parse_url( $request_url, PHP_URL_HOST ), // host URL.
+				'request_uri' => parse_url( $request_url, PHP_URL_PATH ), // endpoint URL.
+				'method'      => $method,
+			);
+
+			if ( in_array( $method, array( 'POST', 'PUT' ), true ) ) {
+				$data['body'] = $this->request_body();
```
bborman22, line 570: Is this still setting the body to the Connect server request body here? I was thinking we'll need to pass the WCCOM request body into this function, but wasn't sure if this was handling that or using the Connect server request body?

classes/class-wc-connect-api-client.php
```diff
@@ -475,6 +496,11 @@ protected function request_headers() {
 				return $authorization;
 			}
 
+			$wc_helper_auth_info = WC_Connect_Functions::get_wc_helper_auth_info();
+			if ( is_wp_error( $wc_helper_auth_info ) ) {
```
bborman22, line 500: I think your initial take on this was right and for now we'll still send requests even if the Helper auth is missing since we will only need it for certain features limited by a subscription. I'm thinking if the auth isn't found the rest of the plugin should still work, but any feature that requires a subscription shouldn't work. 

With that in mind, I was wondering if this check could just be placed near the end of this function and only include the `X-Woo-Signature` if `$wc_helper_auth_info` is not an error?

classes/class-wc-connect-functions.php
```diff
@@ -19,5 +19,25 @@ public static function should_send_cart_api_request() {
...
+		public static function get_wc_helper_auth_info() {
+			if ( class_exists( 'WC_Helper_Options' ) && is_callable( 'WC_Helper_Options::get' ) ) {
+				$helper_auth_data = WC_Helper_Options::get( 'auth' );
+			}
+
+			// It's possible for WC_Helper_Options::get() to return false, throw error if this is the case.
+			if ( ! $helper_auth_data ) {
+				return new WP_Error(
+					'missing_wccom_auth',
+					__( 'Unable to send request to WooCommerce Shipping & Tax server. WooCommerce Helper auth is missing', 'woocommerce-services' )
```
bborman22, line 37: If we are still sending requests even without the Helper auth, this error message could probably be trimmed down to just call out the missing Helper auth.

bborman22 (commented):
The signature function looks much better and sturdier! Just had that one question about the body still, but otherwise it looks good.

I also did some testing without a connection to WCCOM and realized it would prevent any authorized requests when trying to get the headers. I added some notes on the on the code, but I think we'll want to allow requests to go through even if the Helper auth isn't found. The only reason we'd need the Helper auth info on the Connect server is to check a subscription. My understanding is that the plugin should work at it's current capacity without any subscriptions and that subscriptions will only add new functionality.

classes/class-wc-connect-api-client.php
tommyshellberg: Yes, this is right - I should have trusted my gut. I added checks for both the signature and the access_token + site_id.

classes/class-wc-connect-functions.php
tommyshellberg: 👍

classes/class-wc-connect-api-client.php
tommyshellberg: I think what you're getting at here is that if we are only doing a GET request to /subscriptions/ on WCCOM it really shouldn't need to include $data['body'] since that won't exist for only getting subscriptions. I added this for a bit of future-proofing since we can safely assume we will be adding PUT/POST requests in the future to WCCOM. Premature? 🤔

bborman22: No, I think you're exactly right to anticipate future needs here. To give this a more concrete example, if for some reason we need to send an activation request to WCCOM, we'll need to pass along a body of: `{ 'product_key': 12345 }` like in here: https://github.com/woocommerce/woocommerce/blob/8db66f705f5906ff8e6f7f105c095fb71774dc40/includes/admin/helper/class-wc-helper.php#L1345-L1349

In this current implementation, I think the `$data['body']` would be set to our Connect server body, for example the settings payload, rather than just containing the product key to activate. The signature generation will only care about the product key body we'll be sending to WCCOM not the settings payload we're sending to Connect server.

That's why I was thinking we could pass in the body to the function for the most flexibility going forward and have it default to false or empty and skip setting `$data['body']` in those cases.

tommyshellberg: How about this implementation?

https://github.com/Automattic/woocommerce-services/pull/2245/commits/40eb8390575bf1a21f6193c54eb7c531278ec871

I've tested with the site connected and disconnected from WC.com as well as the /subscriptions/ endpoint and other areas of WCS(label printing, getting services) and it all continues to work without requiring a WC.com connection.

bborman22: Looks good to me!

bborman22 (approved):
Tested again and still all worked as expected! Only thing I had missed before was a changelog entry, but once that is in I think this is ready to merge!

PR #2259, opened 2020-11-18
Fix signature required checkbox not updating prices for label requests.

### LINKED ISSUE #2185: fix: label purchase change checkboxes and price
## Description

By checking the one of the checkboxes underneath a label ("Adult signature required", "Signature required") the price changes.
Selecting a checkbox, changing the shipping service and selecting the another service with the same service availability maintains the service as "checked". However, the price is not correctly reflecting the information.

## Acceptance criteria
- The price reflects the information based on the services being checked

tommyshellberg:
## Description
Ensure that the correct number of arguments are passed to the parent ShippingRates component when checking "require signature" option when requesting a shipping label.

### Related issue(s)

Closes #2185 

See the screencast in the original issue:

https://github.com/Automattic/woocommerce-services/issues/2185#issue-702784563

With this branch checked out you should see the prices updating when adding a signature:

Screen Capture on 2020-11-18 at 12-30-49

### Checklist

- [x] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
 = 1.25.3 - 2020-**-** =
 * Add   - Initial code for WooCommerce.com subscriptions API.
 * Fix   - UI fix for input validation for package dimensions and weights.
+* Fix   - When adding "signature required" to some packages, prices were not updating.
 
 = 1.25.2 - 2020-11-10 =
 * Tweak - Add ZIP code validation to UPS(beta) signup form.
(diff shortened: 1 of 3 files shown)


bborman22 (commented):
I found that when I switched services from a signature required to a new service and switched back, the sidebar did not maintain the signature required cost. I pushed up a commit to address this. If this looks good to you let me know and I'll approve the PR! 

One other thought I had while looking at this was wondering if maybe the signature required checkboxes should be "cleared" when changing services? I didn't look into what would be required for that, but thought I'd ask your opinion on that approach as well?

tommyshellberg:
> I found that when I switched services from a signature required to a new service and switched back, the sidebar did not maintain the signature required cost. I pushed up a commit to address this. If this looks good to you let me know and I'll approve the PR!

Thanks for catching this! The changes work and make sense to me. 🚢 

> One other thought I had while looking at this was wondering if maybe the signature required checkboxes should be "cleared" when changing services? I didn't look into what would be required for that, but thought I'd ask your opinion on that approach as well?

I think from a UX perspective it would be nice to actually keep the values when switching. I can imagine a shop owner adding extras to the available shipping services and switching between them to compare prices; in that case, they'd want to keep the selection.  🤷

bborman22 (approved):
Everything worked as expected with no issues!

PR #2262, opened 2020-11-19
Tweak/2250 add carrier description

### LINKED ISSUE #2250: Add a description paragraph to dynamic carrier registration form
Blocks by https://github.com/Automattic/woocommerce-services/issues/2234

### Summary
In #2234, we create a registration form dynamically for each carrier. We need to include a paragraph on the left for each of them:
image

### Proposed solution
1. Use a generic paragraph for all carriers. Find out what this paragraph will be. Use UPS as a reference. 
image

2. Each carrier will have its own unique message and its own ahref link. In this case, it may be better to retrieve the paragraph from the server the same way we retrieve the `carrier-types`. We will also need to create another ticket for the server side work.

brezocordero:
## Description
Add a carrier description to the generated carrier registration form.
We are starting with a default message but it is possible to use `carrierDesc` prop to pass a custom message.

### Related issue(s)

Closes #2250 

1. Go to https://localhost/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings&carrier=DhlExpressAccount

2. Check there is a default description for the carrier:
GenericCarrierMessage

3. It links to a documentation page that is not published yet, if you click the link you will get a 404.

New tests on: client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/test/dynamic-settings-form.js

### Checklist

- [x] unit tests


client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/dynamic-settings-form.js
```diff
@@ -170,6 +179,7 @@ const mapDispatchToProps = ( dispatch ) => ({
 DynamicCarrierAccountSettingsForm.propTypes = {
 	carrierType: PropTypes.string,
 	carrierName: PropTypes.string,
+	carrierDesc: PropTypes.string,
```
frosso, line 182: What are your thoughts on calling the prop `carrierDescription`? (without abbreviations)
There are some abbreviations that are common in codebases (`err`, `res`, `req`), but all the other props are not abbreviated

brezocordero: Done!

brezocordero:
Note: To test I commented on the server the condition that disables DHL carrier.

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/dynamic-settings-form.js
frosso, line 182: Looks like this is still calling it `carrierDesc` instead of `carrierDescription`

frosso (approved):
Looks good so far, thanks for the changes!
I'm guessing you'll also have to change `client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/dynamic-settings.js` in order to pass the `carrierDescription` prop?

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/dynamic-settings-form.js
brezocordero: Thanks for catching that! Done.

brezocordero:
> I'm guessing you'll also have to change client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/dynamic-settings.js in order to pass the carrierDescription prop?

For now we are only displaying a generic description, the specific carrier descriptions would be a future modification.
I didn't make any further changes in the plugin because we might have to change it again once we know where the specific description info is coming from (server or plugin). I can add the prop one level up if you think it will be beneficial.

frosso:
> For now we are only displaying a generic description, the specific carrier descriptions would be a future modification.
I didn't make any further changes in the plugin because we might have to change it again once we know where the specific description info is coming from (server or plugin). I can add the prop one level up if you think it will be beneficial.

I _believe_ we discussed on the team meeting on the week of Nov 5th that the description would come from the server, so that it could be updated regardless of the plugin's release cycle.
The ticket's proposed solution also states that it might be better to retrieve the paragraph from the server.

We can always iterate over it, tho!

brezocordero:
> I believe we discussed on the team meeting on the week of Nov 5th that the description would come from the server, 

Noted and added.

frosso (approved):
Thanks, still looks good! :)

PR #2287, opened 2020-12-21
Send WCCOM connection info in request header

FangedParakeet:
## Description

Subscription flags should be able to enable carriers in carrier account list. The items in the carrier account list will depend on current subscriptions, so the request to populate this list need to include WCCOM connection info in order to get subscription status on request.

New carrier account list

Currently, this WCCOM connection info is only sent in the body of `POST` requests. This PR adds this information to the header of all requests to the Connect Server, so that all request types will be able to send this information and have the Connect Server check existing subscription flags for a user's site. 

Our other options to resolve this would be to change the `GET /shipping/carriers` and `GET /shipping/carrier-types` requests to `POST` requests or to include the extra parameters as query parameters. Changing the requests to `POST` requests would require changes to the Connect Server API endpoints and could cause backwards compatability issues. I felt using the request header felt more appropriate than using query parameters, because these fields are fairly static and could be useful for many other requests/request types as well. Using the header also keeps the Connect Server integration relatively the same, as the server side code is already checking the request headers for the `X-Woo-Signature` to use to make the WCCOM request to fetch a user's WCCOM subscriptions. Additionally, when generating the request header values we are already making the request for the WCCOM connection information with `WC_Connect_Functions::get_wc_helper_auth_info()` to get the `X-Woo-Signature`, so we already have these additional fields unused without necessitating any further API calls. Since I was now able to remove these fields from POST requests, we only need to make this API call once on each request now.

### Related issue(s)

Closes Automattic/woocommerce-shipping-issues#147

1. Please pull, review, and approve the changes in this [PR for the Connect Server](https://github.com/Automattic/woocommerce-connect-server/pull/1673).
2. Add a free extension from WooCommerce.com to your site (e.g. WooCommerce Payments, Square for WooCommerce, etc.) and activate it on your site admin from **WooCommerce > Extensions > WooCommerce.com Subscriptions**.
3. Add a row in the `wcc_subscription_flags` table, using the `product_id` value for a subscription connected to your site, to enable a carrier (e.g. `can_use_dhlexpress_live_rates` to enable DHL Express). You can find the `product_id` value for subscriptions for your WooCommerce account by making a request to the Connect Server to `POST /subscriptions`, containing correct header values for `X-Woo-Signature`, `X-Woo-Access-Token`, and `X-Woo-Site-Id`. Ensure that the flag that you add is not also present in the `wcc_site_flags` table already.

wcc_subscription_flags
_Example row in `wcc_subscription_flags` to use Square extension to enable DHL Express_

4. You should now see the carrier you chose above as an available carrier to connect from the WooCommerce Shipping settings page. 

WooCommerce Shipping carriers
_Example DHL Express is now available to connect_

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,6 +7,7 @@
 * Add	- Allows registration of additional accounts.
 * Tweak - Carrier description on dynamic carrier registration form.
 * Fix   - Adjust documentation links.
+* Tweak - Adds WCCom access token and site ID to connect server request headers.
 
 = 1.25.3 - 2020-11-24 =
 * Add   - Initial code for WooCommerce.com subscriptions API.
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
 = 1.25.5 - 202x-xx-xx =
 * Fix	- Redux DevTools usage update.
 * Add	- Display subscriptions usage.
+* Tweak - Adds WCCom access token and site ID to connect server request headers.
 
 = 1.25.4 - 2020-12-08 =
 * Tweak - Remove Stripe connect functionality.
@@ -11,7 +12,6 @@
 * Add	- Allows registration of additional accounts.
 * Tweak - Carrier description on dynamic carrier registration form.
 * Fix   - Adjust documentation links.
-* Tweak - Adds WCCom access token and site ID to connect server request headers.
 
 = 1.25.3 - 2020-11-24 =
 * Add   - Initial code for WooCommerce.com subscriptions API.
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -11,6 +11,7 @@
 * Add	- Allows registration of additional accounts.
 * Tweak - Carrier description on dynamic carrier registration form.
 * Fix   - Adjust documentation links.
+* Tweak - Adds WCCom access token and site ID to connect server request headers.
```
c-shultz, line 14: Changelog update should go up in the 1.25.5 section

c-shultz (changes requested):
All good except the changelog section. Looks good to merge after that gets moved!

changelog.txt
FangedParakeet: Whoops, branch needed a belated update and I missed this one--just corrected the changelog entry now, cheers!

c-shultz (approved):
👍

c-shultz (approved):
Looks great! :shipit:

PR #2288, opened 2020-12-23
Added filter to allow third-party plugins check if we should skip calculating totals

seb86:
This is an enhancement to allow third-party plugins check if we should skip calculating totals. This PR is only part 1 of 2 PR's to allow calculating the totals via other conditions.


c-shultz (changes requested):
https://github.com/Automattic/woocommerce-services/pull/2289#pullrequestreview-616679007

c-shultz:
This has been quiet for a while. I'm going to close these, but please contact us again if you wish to revisit this! Thanks!

PR #2293, opened 2020-12-23
Removes DHL eCommerce International, replaces DHL eCommerce with DHL eCommerce Solutions (logos)

FangedParakeet:
## Description

This allows DHL merchant carrier accounts to be displayed with the same DHL logo that is used by DHL Express.

### Related issue(s)
Automattic/woocommerce-shipping-issues#125.

1. Pull the changes from Automattic/woocommerce-connect-server#1674 and follow the steps to enable DHL Express, DHL eCommerce Solutions, and DHL eCommerce Asia.
2. Visit **WooCommerce > Settings > Shipping > WooCommerceShipping**. You should now see the new DHL carrier accounts appear in the list of available Carrier Accounts, all using the same DHL logo as DHL Express.

DHL carrier account logos
_New DHL carrier accounts with correct logos_

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-rest-connect-shipping-carrier-controller.php
+++ b/classes/class-wc-rest-connect-shipping-carrier-controller.php
@@ -26,6 +26,7 @@ class WC_REST_Connect_Shipping_Carrier_Controller extends WC_REST_Connect_Base_C
 			return $error;
 		}
 
+		do_action( 'wc_connect_fetch_service_schemas' );
 		return $response;
 	}
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,12 +1,14 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
-= 1.25.5 - 202x-xx-xx =
+= 1.25.6 - 202x-xx-xx =
+* Fix 	- Refreshes shipping methods after registering or removing carrier accounts.
+
+= 1.25.5 - 2021-01-11 =
 * Fix	- Redux DevTools usage update.
 * Add	- Display subscriptions usage.
 * Add	- Subscription activation.
 * Add 	- Uses same DHL logo for all registered DHL accounts.
 * Tweak - Adds WCCom access token and site ID to connect server request headers.
-* Add 	- Refreshes shipping methods after registering new carrier account.
 
 = 1.25.4 - 2020-12-08 =
 * Tweak - Remove Stripe connect functionality.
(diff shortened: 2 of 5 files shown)


frosso (approved):
Looks straightforward 👍 
The `changelog.txt` entry seems to be "checked", but I'm not seeing changes on the file 👀

FangedParakeet:
@frosso, initially when I made these changes to use the same logos for DHL Express, DHL eCommerce, DHL eCommerce International, and DHL eCommerce Asia, I added the following line to the changelog.

> * Add 	- Uses same DHL logo for all registered DHL accounts.

Then pretty much exactly as that change got approved, I found out that EasyPost were actually removing DHL eCommerce and DHL eCommerce International to replace them with DHL eCommerce Solutions without explicitly warning us, prompting this quick update. This is basically additional work required for that last change (but I had to make a new PR, since the previous one had just been merged), so I thought I could avoid crowding the changelog entries with a redunandant row. So when I checked the `changelog.txt` entry, what I told you was true...from a certain point of view. But let me know if you think this little update necessitates an additional `Tweak` line and I'll slip it in there for good measure!

superdav42:
Ran into a bug https://github.com/Automattic/woocommerce-services/issues/2298 but not related to these changes.

classes/class-wc-rest-connect-shipping-carrier-controller.php
```diff
@@ -26,6 +26,7 @@ public function post( $request ) {
 			return $error;
 		}
 
+		do_action( 'wc_connect_fetch_service_schemas' );
```
FangedParakeet, line 29: @superdav42, I just added this line to refresh the shipping methods regardless of the value for `WOOCOMMERCE_CONNECT_FREQUENT_FETCH`. Seems a bit too easy: will this do the trick or would you recommend a better way to call `WC_Connect_Service_Schemas_Store::fetch_service_schemas_from_connect_server`?

superdav42: Using the action is perfect. You just need it add it to `WC_REST_Connect_Shipping_Carrier_Delete_Controller::delete()` as well.

superdav42 (changes requested):
Refresh carriers when a carrier is disconnected as well.

FangedParakeet:
Just added that to `WC_REST_Connect_Shipping_Carrier_Delete_Controller` as well and updated the changelog entry to reflect the recent release. Cheers, @superdav42!

superdav42 (approved):
Works for me now.

PR #2300, opened 2021-01-13
Remember post-print notification checkbox state

### LINKED ISSUE #2005: Option to deselect the `mark this order complete` checkbox by default
User in 2883635-zen is complaining about the **Mark this order as complete and notify the customer** checkbox showing now when buying a label:

Screenshot
Screenshot: https://d.pr/i/iUS6h3

They say this goes against their workflow and creates a lot of confusion with their customers if they forget to uncheck it.

I'm asking them for more details about what their workflow is to understand why this is a problem.

In the meantime, how complicated it would be to add an option in WooCommerce Services settings to leave this box unchecked by default?

waclawjacek:
## Description

Currently, the **Mark this order as complete and notify the customer** and **Notify the customer with shipment details** checkboxes in the **Create shipping label** modal are always set to a default state (unchecked). This is problematic for users who have workflows that do not match this behavior.

This PR introduces a new option `post_print_notification_settings` in `WC_Connect_Options` which stores the state of both checkboxes.

The settings can be manipulated via a REST controller which automatically happens when a label is printed (automatically or after **Print** is clicked) in the **Create shipping label** modal.

#### Relying on functions rather than state when sending

Previously, the state values `fulfillOrder` and `emailDetails` were set to a result calculated from checkbox state and order status:

```js
const onFulfillAndEmailOrderChange = ( value ) => {
	// Don't change order status if already finished.
	props.setFulfillOrderOption( orderId, siteId, value && ! isOrderFinished( order.status ) );
	// Email only if order is already complete.
	props.setEmailDetailsOption( orderId, siteId, value && isOrderFinished( order.status ) );
};
```

This PR repurposes these values to only contain the user preference of whether to send the notification if applicable.

Instead of relying on the state directly, the `shouldFulfillOrder()` and `shouldEmailDetails()` functions should be used instead.

#### Naming

The name is pretty long. :) I was also considering "Post-Print Actions" or just "Notification Settings" but for now settled for the most verbose "Post-Print Notification Settings".

#### Default state removal

I have removed the default state for the `emailDetails` and `fulfillOrder` fields from `/client/extensions/woocommerce/woocommerce-services/lib/initialize-labels-state/index.js`. I believe it was not used anywhere. If someone more familiar with the inner workings of the plugin could please let me know if this might be problematic, that would be great!

### Related issue(s)

Fixes #2005 

Markup on 2021-01-13 at 18:50:41

Markup on 2021-01-13 at 18:24:23

The screenshot shows an order that is not completed yet. The **Notify the customer with shipment details** checkbox is presented in the same modal for completed orders.

### Checklist

- [x] unit tests

### DIFF
--- a/classes/class-wc-connect-service-settings-store.php
+++ b/classes/class-wc-connect-service-settings-store.php
@@ -161,6 +161,67 @@ if ( ! class_exists( 'WC_Connect_Service_Settings_Store' ) ) {
 		}
 
 		/**
+		 * Gets post-label-printing notification settings.
+		 *
+		 * @return array Array of boolean values, indexed by setting names.
+		 */
+		public function get_post_print_notification_settings() {
+			$defaults = array(
+				'mark_order_complete_and_notify_customer' => false,
+				'notify_customer_with_shipment_details'   => false,
+			);
+
+			$settings = WC_Connect_Options::get_option( 'post_print_notification_settings', $defaults );
+
+			return array_merge( $defaults, $settings );
+		}
+
+		/**
+		 * Updates post-label-printing notification settings.
+		 *
+		 * @param string $name    Name of the setting to enable/disable.
+		 * @param bool   $enabled Whether the setting should be enabled.
+		 *
+		 * @return true|WP_Error WP_Error if an error occurred, `true` otherwise.
+		 */
+		public function set_post_print_notification_setting( $name, $enabled ) {
+			$allowed_names = array(
+				'mark_order_complete_and_notify_customer',
+				'notify_customer_with_shipment_details',
+			);
+
+			if ( ! in_array( $name, $allowed_names, true ) ) {
+				return new WP_Error(
+					'invalid_notification_setting_name',
+					__( 'Invalid notification setting name supplied.', 'woocommerce-services' )
+				);
+			}
+
+			$old_settings      = WC_Connect_Options::get_option( 'post_print_notification_settings' );
+			$settings          = $old_settings;
+			$settings[ $name ] = (bool) $enabled;
+
+			/*
+			 * WC_Connect_Options::update_option() returns `false` if the new value is the same as the old one.
+			 * As this is not an issue in this case, we return `true` here and leave the option unchanged.
+			 */
+			if ( $settings === $old_settings ) {
+				return true;
+			}
+
+			$result = WC_Connect_Options::update_option( 'post_print_notification_settings', $settings );
+
+			if ( ! $result ) {
+				return new WP_Error(
+					'save_failed',
+					__( 'Updating the option failed.', 'woocommerce-services' )
+				);
+			}
+
+			return $result;
+		}
+
+		/**
 		 * Attempts to recover faulty json string fields that might contain strings with unescaped quotes
 		 *
 		 * @param string $field_name
--- a/classes/class-wc-rest-connect-post-print-notification-settings-controller.php
+++ b/classes/class-wc-rest-connect-post-print-notification-settings-controller.php
@@ -1,0 +1,34 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit();
+}
+
+if ( class_exists( 'WC_REST_Connect_Post_Print_Notification_Settings_Controller' ) ) {
+	return;
+}
+
+class WC_REST_Connect_Post_Print_Notification_Settings_Controller extends WC_REST_Connect_Base_Controller {
+	protected $rest_base = 'connect/post-print-notification-settings';
+
+	public function post( $request ) {
+		$result = $this->settings_store->set_post_print_notification_setting( $request['name'], $request['enabled'] );
+
+		if ( is_wp_error( $result ) ) {
+			$error = new WP_Error(
+				'save_failed',
+				sprintf(
+					__( 'Unable to update notification setting. %s', 'woocommerce-services' ),
+					$result->get_error_message()
+				),
+				array( 'status' => 400 )
+			);
+
+			$this->logger->log( $error, __CLASS__ );
+
+			return $error;
+		}
+
+		return new WP_REST_Response( array( 'success' => true ), 200 );
+	}
+}
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/test/shipping-rate.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/test/shipping-rate.js
@@ -8,6 +8,7 @@ import { expect } from 'chai';
 import { configure, mount } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16'
 import { CheckboxControl, RadioControl } from '@wordpress/components';
+import { moment } from 'i18n-calypso';
 
 /**
  * Internal dependencies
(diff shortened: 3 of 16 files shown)


waclawjacek:
There is currently a delay between when the checkbox state is loaded as opposed to `paperSize` which is implemented in a similar way. This might have to be improved.

classes/class-wc-connect-service-settings-store.php
```diff
@@ -160,6 +160,67 @@ public function set_preferred_paper_size( $size ) {
...
+				);
+			}
+
+			$old_settings      = WC_Connect_Options::get_option( 'post_print_notification_settings' );
+			$settings          = $old_settings;
+			$settings[ $name ] = (bool) $enabled;
+
+			/*
+			 * WC_Connect_Options::update_option() returns `false` if the new value is the same as the old one.
+			 * As this is not an issue in this case, we return `true` here and leave the option unchanged.
```
harriswong, line 206: Nice comment!

classes/class-wc-rest-connect-post-print-notification-settings-controller.php
```diff
@@ -0,0 +1,34 @@
+<?php
```
harriswong, line 1: I see we use `classes/class-wc-rest-connect-shipping-label-preview-controller.php` (`connect/label/preview`) when we save the `paper_size` preference. I am a little surprised that I can't find an existing end point to update settings. So this new end point makes sense.

Were you also not able to find any existing end point to update settings?

classes/class-wc-connect-service-settings-store.php
```diff
@@ -160,6 +160,67 @@ public function set_preferred_paper_size( $size ) {
...
+		/**
+		 * Updates post-label-printing notification settings.
+		 *
+		 * @param string $name    Name of the setting to enable/disable.
+		 * @param bool   $enabled Whether the setting should be enabled.
+		 *
+		 * @return true|WP_Error WP_Error if an error occurred, `true` otherwise.
+		 */
+		public function set_post_print_notification_setting( $name, $enabled ) {
+			$allowed_names = array(
```
harriswong, line 188: We can't trust the user input and I like how you add the check here instead of the controller's level.

classes/class-wc-rest-connect-post-print-notification-settings-controller.php
```diff
@@ -0,0 +1,34 @@
...
+				),
+				array( 'status' => 400 )
+			);
+
+			$this->logger->log( $error, __CLASS__ );
+
+			return $error;
+		}
+
+		return new WP_REST_Response( array( 'success' => true ), 200 );
```
harriswong, line 32: Nice! There are a few different ways in our existing code to return a response/error. I am glad you picked `WP_REST_Response` :D 💯

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/test/shipping-rate.js
```diff
@@ -94,7 +95,9 @@ describe( 'ShippingRate', () => {
 		} );
 
 		it( 'renders the delivery date', () => {
-			expect( shippingRateWrapper ).to.contain( <div className="rates-step__shipping-rate-delivery-date">January 1</div> ); // eslint-disable-line
+			const expectedDate = moment( shippingRateWrapper.props().rateObject.delivery_date ).format( 'LL' ).split( ',' )[0];
```
harriswong, line 98: Is this the timezone issue? :D

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/sidebar.js
```diff
@@ -51,10 +49,13 @@ export const Sidebar = props => {
...
-		props.setFulfillOrderOption( orderId, siteId, value && ! isOrderFinished( order.status ) );
-		// Email only if order is already complete.
-		props.setEmailDetailsOption( orderId, siteId, value && isOrderFinished( order.status ) );
+		if ( isOrderFinished( order.status ) ) {
+			// Email only if order is already complete.
+			props.setEmailDetailsOption( orderId, siteId, value );
+		} else {
+			// Don't change order status if already finished.
+			props.setFulfillOrderOption( orderId, siteId, value );
+		}
```
harriswong, line 58: Instead of updating both all the time, now it updates either or based on `order.status`. I think this is fine since we have defaults and we don't need to update both every time.

harriswong (approved):
Tested locally. I checked "Mark this order as complete and notify the customer" and print a label. It refreshed with the "Notify the customer with shipment details" unchecked.

I opened up another order and "Mark this order as complete and notify the customer" is checked by default. 

Thanks for adding all the tests! LGTM! 👍

c-shultz:
Nice work pushing this forward and digging so deep into a new codebase, @waclawjacek 🥳

classes/class-wc-rest-connect-post-print-notification-settings-controller.php
waclawjacek: Yes!

I was considering adding the notification-settings-saving logic to an existing endpoint but:

1. Saving when shipping rates are requested or purchased is too early because the checkbox might be toggled afterward.
1. Saving when the label PDF is **requested** (this happens automatically after a request to purchase a label is made) was another option I considered but:
    1. the notifications are not yet sent at this point so again, this is too early;
    1. this is a `GET` request. The paper size setting is saved upon this request, but this makes more sense than saving notification settings here - the paper size has to be provided anyway so a label can be printed in the selected size.
1. Saving when the label PDF is **displayed** (either after "Print" is clicked or the PDF is opened automatically) *sometimes* triggers another request. This will be either a request to mark an order as completed, to send the email with details, but can also be no request at all. In such cases, a separate call would have to be made anyway.

Based on the above, I decided that placing the logic in a separate REST endpoint accessed after the "Print" button is clicked would be the cleanest solution.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/test/shipping-rate.js
waclawjacek: Yes! The issue was that the time provided in `createShippingRateWrapper()` (which is `2020-01-01T23:59:00.000Z`) can, in some timezones, not match the test's expectation of `January 1`. :)

classes/class-wc-rest-connect-post-print-notification-settings-controller.php
harriswong: This makes a lot of sense now. Thank you!

c-shultz:
Let's go ahead and hold off on this until we can give WCShip more focus, and do so more thorough testing.

waclawjacek:
The implementation here should be complete. I didn't merge this yet because I wasn't sure if the Redux state + respective React components are managed in a reliable and backward-compatible way, taking into account all the possible contexts the app can be initialized and run in.

If someone with more knowledge than me in this field could give it a look, that would be fantastic!

If #2367 is implemented, the approach taken here might need to change. The assumption in this PR was that things happen after printing. If they happen after purchase instead, maybe an additional REST endpoint will not be necessary, etc.

harriswong:
@waclawjacek Bump! I stumbled upon this when I am looking at cleaning up the Calypso stuff. I think we can take a look at this again.

samnajian:
Old and obsolete.

Issue #2285, opened 2020-12-15
Add transients to cache API live rates response


brezocordero:
Following the investigation for : #2263 
Add transients to store API [live rates response](https://github.com/Automattic/woocommerce-services/blob/eedd4af2d536e4fef2223879f24359e553f20bb9/classes/class-wc-connect-shipping-method.php#L347), to reduce the number of API requests. 
The reduction will be more visible on sites that do not have persistent object cache.

[Example](https://github.com/woocommerce/woocommerce-shipping-usps/blob/c6721a38486adddd858b8e163e3ad31bb70badbe/includes/class-wc-shipping-usps.php#L1908) of how it is done for other live rates shipping plugins.

c-shultz:
I'm prioritizing this now since we're further investing in a live rates offering, so we should look to make sure we're reducing our request rate where possible. 

I'm anticipating this is less than a week of effort, but if it starts to look larger, let's discuss to make sure it's still worth the investment.

frosso:
I believe I have a solution in https://github.com/Automattic/woocommerce-services/pull/2301 - spending some additional time to investigate if the number of requests can be limited even further

frosso:
Timeboxed it a bit - no luck in reducing further without major changes :( 
The PR can be reviewed, tho.

PR #2301, opened 2021-01-15
refactor: live rates response caching

frosso:
## Description

Changing from `wp_cache_get` to `get_transient`, so that the API response is set with transients (rather than WP cache).
If a store does not have the WP cache configured, they would not benefit from the caching of the API response.
By using the transients, we're leveraging `wp_options` instead, which is always configured.

I ensured that the transient key length is short enough ([must be 172 character or shorter](https://developer.wordpress.org/reference/functions/set_transient/#parameters)).
I also added 2 new debug messages, to help with the debugging:
- "Cache contains rates response"
- "Cache does not rates response"

I am aware that I am using the word "cache", but I didn't think that the word "transient" would be useful, in this case. After all, we're using the `transient`s as a caching mechanism.

### Related issue(s)

fixes https://github.com/Automattic/woocommerce-services/issues/2285

- Put your site in debug mode, by going to **WooCommerce > Status > WooCommerce Shipping & Tax > Debug**
- Put WooCommerce in debug mode in **WooCommerce > Settings > Shipping > Shipping options > Debug mode**
- Ensure your site has permissions to get live rates (set the `can_use_live_rates` flag on the server)
- Ensure your site is US-based
- Add and configure the USPS shipping method by WooCommerce Shipping in **WooCommerce > Settings > Shipping > Shipping Zones** in one of your shipping zones
- Live rates should still be working in your checkout
- With debug _disabled_, you should not see as many requests on the server

First request:
Screen Shot 2021-01-15 at 2 59 08 PM

Subsequent requests:
Screen Shot 2021-01-15 at 2 57 14 PM

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 1.25.6 - 202x-xx-xx =
 * Fix 	- Refreshes shipping methods after registering or removing carrier accounts.
+* Tweak	- Changed rates response caching method from cache to transient.
 
 = 1.25.5 - 2021-01-11 =
 * Fix	- Redux DevTools usage update.
--- a/classes/class-wc-connect-shipping-method.php
+++ b/classes/class-wc-connect-shipping-method.php
@@ -343,16 +343,17 @@ if ( ! class_exists( 'WC_Connect_Shipping_Method' ) ) {
 				'wcs_rates_%s',
 				md5( serialize( array( $services, $package, $custom_boxes, $predefined_boxes ) ) )
 			);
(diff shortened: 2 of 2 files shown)


frosso:
I believe I need some additional adjustments on the cache key for local development - removed the "needs review" label

bborman22 (approved):
This worked well for me. I confirmed that the server doesn't get hit when rate is cached. I updated the cart and/or address and confirmed it requested a new rate with either update. 

LGTM!

PR #2304, opened 2021-01-19
Error in checkout page first load if automated taxes has errors

Ferdev:
## Description
When the checkout page is loaded, the taxjar integration triggers automatically and, if there's an error in the process (like an invalid zip code from a previous request), the checkout page won't show.

Ideally, we should be hitting the Taxjar API only when the user triggers the checkout form validation, but not automatically in page load.

WORK IN PROGRESS

### Related issue(s)

#1576 

1. Add a product to the cart and go to the checkout page.
2. Enter an invalid zip code, like `XYZZY` and press the puchase button.
3. Reload the checkout page. The `cart-errors.php` will be displayed instead of the checkout page.

### Checklist

- [ ] unit tests


frosso:
Looks like this has been replaced in favor of https://github.com/Automattic/woocommerce-services/pull/2310

Closing

PR #2310, opened 2021-01-26
Remove error notices from checkout page load

### LINKED ISSUE #1576: "Invalid Zip Entered" Error when Automated Taxes are enabled.

bborman22:
## Description
As described in #1576 there was a situation where if the tax calculation failed while loading the checkout page, the checkout page would not display. This error primarily came up as an address, particularly zip code error. The only way to fix it previously was to modify the address some other way as you could no longer do it from the checkout page. There was a lot of digging on this issue to find the best approach and @Ferdev pinpointed an approach of not calculating tax on the checkout page load. I then picked up the issue and modified it slightly so that it may still calculate the tax on page load, but if it fails it doesn't add an error notice, which was breaking the checkout page and instead fail silently to give the user a chance to fix the error.

The taxes are actually calculated in two ways on the checkout page, once from the page load and one through AJAX (after updating the address or attempting to place the order). This PR checks to see if the request to calculate tax was from the checkout page and if it was AJAX or not. If it was not AJAX it will not add an error notice. I had previously experimented with adding a "notice" level notice instead of skipping it entirely, but found that these don't clear out as you update the page and would cause confusion.

Screen Shot 2021-01-26 at 4 01 04 PM

I feel it is safe to not display an error notice on the page load as the AJAX requests will catch and display the errors. If the user has AJAX disabled, other functionality for the checkout won't work well, but as a fallback there is still a check on valid address data when processing the checkout. 

### Related issue(s)

Fixes #1576 

1. Start with a WCS site with tax calculations enabled.
2. Go through a normal shopping experience up to the checkout page.
3. On the checkout page, enter a valid address with valid zip code.
4. Confirm that taxes calculated properly when the zip code was entered.
5. Change the zip code to an invalid zip code.
6. Confirm an error notice is displayed.
7. Refresh the checkout page.
8. No notice should be displayed at first while the subtotals are calculated. This happens quickly so may miss it while it's calculating subtotals.
9. Subtotals will be calculated and at that point the error notice should display.
10. Attempt to checkout with the error notice and receive more errors and are unable to checkout with invalid zip.

To test with AJAX disabled on the checkout page add the following hook to an mu-plugin or something similar:

```
add_action( 'wp_enqueue_scripts', function() {
	wp_dequeue_script( 'wc-checkout' );
} );
```

Attempt the same process as above, you won't get an error notice at first. Once clicking Place Order though, you should get a page refresh with the errors displayed.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.7 - 2021-xx-xx =
+* Fix	- Prevent error notices on checkout page load
+
 = 1.25.6 - 2021-01-26 =
 * Fix 	- Refreshes shipping methods after registering or removing carrier accounts.
 * Tweak	- Changed rates response caching method from cache to transient.
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -326,7 +326,8 @@ class WC_Connect_TaxJar_Integration {
 				$message = sprintf( _x( 'Invalid %s entered.', '%s - ZIP/Postal code checkout field label', 'woocommerce-services' ), $postcode_field_name );
 			}
 
-			if ( ! wc_has_notice( $message, 'error' ) ) {
+			// if on checkout page load (not ajax), don't set an error as it prevents checkout page from displaying
+			if ( ( is_cart() || ( is_checkout() && is_ajax() ) ) && ! wc_has_notice( $message, 'error' ) ) {
 				wc_add_notice( $message, 'error' );
 			}
 


frosso (approved):
Thanks for fixing this! LGTM! 🚢

JBSurfs:
I'm not as code savy as everyone here, but I see you all mentioning it is fixed.  I am still having this same problem, checking again this morning where the errors start if the zip code is wrong.  Am I looking to add the snippet of code above to go around this problem?  If so, to which file?  Thank you.

frosso:
@JBSurfs can you please confirm whether or not the version of WooCommerce Shipping & Tax used to test is at least `1.25.7`? You can find the version used in the *Plugins* section of your `wp-admin`.

If that's not the version used, you can download the newest one here: https://wordpress.org/plugins/woocommerce-services/
If the version used on your site is an older one, you can write a support request in the [community support forum](https://wordpress.org/support/plugin/woocommerce-services/) including some instructions and details about your blog.

Thanks!

JBSurfs:
Hi Frosso, yes, that is the version used, 1.25.7.  I am showing all plugins available as up to date on the dashboard also.  I have deactivated jetpack again.  

Just tried it out to see what happens and how.  So, on checkout, I can go straight to zipcode and type in 2 numbers and get a green check mark on the side and no error message.  If I change the state to North Carolina, which is the state I am collecting tax for, I immediately get the error message.  I can get the zip code changed, but it has to be a quick click in and type one number at a time until I get to a correct zip code.  So, 12 is in there, then click in the box, type 3, get the green check, error message, and cursor is out of the box. Click in the box, type 4, get the green check and error, cursor is out, click in box, type 5, then get green check and no error message.  The problem is most people are probably not going to know to keep clicking and typing so fast, plus I think it does not work that well on mobile.

frosso:
Hi @JBSurfs , it looks like the issue you described is related to element focus - the cursor is moved out of the input field.
I created a new ticket for it: https://github.com/Automattic/woocommerce-services/issues/2335

Thanks for pointing it out!

PR #2314, opened 2021-01-29
Highlight rate call usage in red and display tooltip when over limit

waclawjacek:
## Description

When the rate call usage is over limit, highlight the text in red and increase the font weight. Display a tooltip with the following message:

> Your store is over the limit for rate calls this month and your account will be charged for overages. To prevent future overage charges you can upgrade your plan.

Some things that I would like to point out in case they require further adjustment:

* The implemented view is not pixel-perfect but rather an approximation of the mockup.
* I have used the `--color-red-400` color rather than e.g. one of the `--color-error-*` colors.
* I have set the tooltip width to `200px` to be more like the design mockup as by default it is displayed as a single line.
  
  Relevant CSS:
  ```css
  .components-popover__content {
	  white-space: normal;
	  width: 200px;
  }
  ```
* The design mockup shows a cursor similar to when using `cursor: pointer`. I have left it in the default state as this is the default for the Tooltip component from `@wordpress/components` in hope this results in a more uniform experience.
* For the mobile view, I have placed the icon in a column and centered the text.

### Potential todos:

- [ ] On mobile devices, the tooltip is not displayed (tested on Safari and Firefox on iOS 14.4).

On desktop:

Desktop preview

On mobile:

Mobile preview

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 1.25.7 - 2021-xx-xx =
 * Fix	- Prevent error notices on checkout page load
+* Tweak	- Highlight rate call usage over limit on WooCommerce Shipping settings page.
 
 = 1.25.6 - 2021-01-26 =
 * Fix 	- Refreshes shipping methods after registering or removing carrier accounts.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,8 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
 = 1.25.7 - 2021-xx-xx =
-* Fix   - Prevent error notices on checkout page load.
-* Tweak	- Highlight rate call usage over limit on WooCommerce Shipping settings page.
+* Fix   - ∫Prevent error notices on checkout page load.
+* Tweak - Highlight rate call usage over limit on WooCommerce Shipping settings page.
 * Fix   - Connect carrier account link broken on subdirectory installs.
 
 = 1.25.6 - 2021-01-26 =
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
 = 1.25.7 - 2021-xx-xx =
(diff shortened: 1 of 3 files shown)


changelog.txt
```diff
@@ -2,6 +2,7 @@
 
 = 1.25.7 - 2021-xx-xx =
 * Fix	- Prevent error notices on checkout page load
+* Tweak	- Highlight rate call usage over limit on WooCommerce Shipping settings page.
```
brezocordero, line 5: Nitpicking on formatting: Align the messages on Fix and Tweak. And not from this PR but missing a period on Fix.

client/extensions/woocommerce/woocommerce-services/views/subscriptions-usage/style.scss
```diff
@@ -20,7 +20,43 @@
...
+				text-align: left;
+			}
+
+			& > .subscriptions-usage__list-item-usage-icon {
+				display: none;
+			}
+
+			&.is-over-limit {
+				& > .subscriptions-usage__list-item-usage-numbers {
+					color: var( --color-red-light );
```
brezocordero, line 41: On the PR there is a comment about not using `var( --color-error )` and using `--color-red` instead.
I think it is to be closer to the mockup but I am not sure. Looks ok though.

brezocordero (changes requested):
It is working well and looks like the mockup.

It would have been good to add a test here but not a requirement: https://github.com/Automattic/woocommerce-services/blob/develop/client/extensions/woocommerce/woocommerce-services/views/subscriptions-usage/test/index.js

Small nitpicking request on the changelog.txt format.

brezocordero (approved):
Remaining question about the color but looks consistent with carrier accounts and with the design.
🚢

brezocordero:
I am merging the PR because @waclawjacek is on leave. 
Thanks for the work @waclawjacek !

PR #2315, opened 2021-02-01
Use relative path so links work on subdirectory installs

### LINKED ISSUE #2298: Unable to connect a carrier account if site is installed in sub directory
This code at https://github.com/Automattic/woocommerce-services/blob/master/client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/list-item.js#L105:
```js
					<a
						href={
							`/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings&carrier=${data.type}`
						}
						// eslint-disable-next-line wpcalypso/jsx-classname-namespace
						className="button is-compact"
					>
```
will not work if WordPress is installed in a subdirectory like `http://localhost/wordpress/wp-admin/admin.php` clicking the link would go to `http://localhost/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings&carrier=carriername` which will most likely 404.

I was about to fix it by just removing `/wp-admin/` from the href but I was wondering if we should even use a full page load for this page. Should we use react to just render a new route?

superdav42:
## Description
Link to connect a carrier account will break if WP is installed in a subdirectory.

### Related issue(s)

closes #2298 

Install WordPress into a subdirectory.
Try to connect a carrier account.
Should not 404.

### Checklist

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.7 - 2021-x-x =
+* Fix   - Connect carrier account link broken on subdirectory installs
+
 = 1.25.6 - 2021-01-26 =
 * Fix 	- Refreshes shipping methods after registering or removing carrier accounts.
 * Tweak	- Changed rates response caching method from cache to transient.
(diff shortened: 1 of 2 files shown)


frosso (approved):
Tested locally, changes look good! 👍

PR #2317, opened 2021-02-03
Adds plugin hook to send information to track subscription events on checkout

FangedParakeet:
## Description

> For this approach:
>
> - We want to count every successful check out that used WCShip for live rates
> - We should ignore checkouts that did not use WCShip to get live rates
> - We should ignore orders that are manually created from wp-admin

This PR adds a checkout hook to send rates request related information to the Connect Server to track related subscription events. The function here sends both the cached rates request, if present, and the package and service data for all requested shipping methods, so that if the former is not present, we can repeat the shipping rates request on the Connect Server. 

The major caveat is that this request is blocking, but should be fired after the order is complete and hopefully, if using cached rates responses, should not block for too long.

### Related issue(s)
Automattic/woocommerce-shipping-issues#143

Testing instructions coming soon! WIP

### Checklist

- [ ] unit tests

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -1089,7 +1089,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 				);
 				$custom_boxes[]     = $service_schemas_store->get_packages();
 				$predefined_boxes[] = $service_schemas_store->get_predefined_packages_for_service( $service_schema->id );
-				$requested_rates[]  = ( false !== $cached_response_body ) ? $cached_response_body['rates'] : array();
+				$requested_rates[]  = $cached_response_body;
 			}
 
 			$api_client = $this->get_api_client();
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -1034,36 +1034,11 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		 * @param array $posted_data Post data for order.
 		 * @param array $order WC_Order object containing completed order details.
 		 */
-		public function track_completed_order( $order_id, $posted_data, $order ) {
-			// We need to recreate a temporary cart from the order
-			// so that we can create the same shipping packages from the cart
-			// to use to generate the correct cache key to return the cached
-			// shipping rates response.
+		public function track_completed_order( $order_id, $data, $order ) {
 			$logger = $this->get_logger();
 
-			$temp_cart   = new WC_Cart();
-			$order_items = $order->get_items();
-
-			foreach ( $order_items as $item ) {
-				$product = wc_get_product( $item->get_product() );
-
-				$product_id           = $product->get_id();
-				$quantity             = $item->get_quantity();
-				$variation_id         = 0;
-				$variation_attributes = array();
-
-				if ( 'variation' === $product->get_type() ) {
-					$variation_id         = $product_id;
-					$product_id           = $product->get_parent_id();
-					$variation_attributes = $product->get_variation_attributes();
-				}
-
-				$temp_cart->add_to_cart( $product_id, $quantity, $variation_id, $variation_attributes );
-			}
-
-			$packages = $temp_cart->get_shipping_packages();
-			$temp_cart->empty_cart();
-			$package = $packages[0];
+			$packages = WC()->cart->get_shipping_packages();
+			$package  = $packages[0];
 			foreach ( $package['contents'] as $item_key => $item_details ) {
 				ksort( $package['contents'][ $item_key ] );
 			}
@@ -1072,33 +1047,29 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			$shipping_zone    = WC_Shipping_Zones::get_zone_matching_package( $package );
 			$shipping_methods = $shipping_zone->get_shipping_methods( true );
 
-			// We will fill these with the required parameters
-			// should we need to repeat the shipping request for
-			// each service as a fallback.
-			$services         = array();
-			$custom_boxes     = array();
-			$predefined_boxes = array();
-			$requested_rates  = array();
-
+q			$tracking_request_body = array();
 			foreach ( $shipping_methods as $method ) {
 				try {
 					$handler = new WC_Connect_Shipping_Method( $method->get_instance_id() );
 
-					$service_settings       = $handler->get_service_settings();
-					$service_schema         = $handler->get_service_schema();
-					$service_settings_store = $handler->get_service_settings_store();
-					$cached_response_body   = $handler->get_cached_shipping_rates_response( $package );
-
-					$services[]         = array(
+					$service_settings     = $handler->get_service_settings();
+					$service_schema       = $handler->get_service_schema();
+					$cached_response_body = $handler->get_cached_shipping_rates_response( $package );
+					$services 			  = array(
 						array(
 							'id'               => $service_schema->id,
 							'instance'         => $method->get_instance_id(),
 							'service_settings' => $service_settings,
 						),
 					);
-					$custom_boxes[]     = $service_settings_store->get_packages();
-					$predefined_boxes[] = $service_settings_store->get_predefined_packages_for_service( $service_schema->id );
-					$requested_rates[]  = $cached_response_body;
+
+					if ( ! isset( $cached_response_body->rates ) ) {
+						continue;
+					}
+					$tracking_request_body[] = array(
+						'rates'    => $cached_response_body->rates,
+						'services' => $services,
+					);
 				} catch ( Exception $e ) {
 					$logger->error(
 						sprintf(
@@ -1112,7 +1083,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			}
 
 			$api_client = $this->get_api_client();
-			$response   = $api_client->track_subscription_event( $services, $package, $custom_boxes, $predefined_boxes, $requested_rates );
+			$response   = $api_client->track_subscription_event( $tracking_request_body );
 			if ( is_wp_error( $response ) ) {
 				if ( is_a( $logger, 'WC_Connect_Logger' ) ) {
 					$logger->error(
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -766,6 +766,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			add_filter( 'wc_connect_shipping_service_settings', array( $this, 'shipping_service_settings' ), 10, 3 );
 			add_action( 'woocommerce_email_after_order_table', array( $this, 'add_tracking_info_to_emails' ), 10, 3 );
 			add_filter( 'woocommerce_admin_reports', array( $this, 'reports_tabs' ) );
+			add_action( 'woocommerce_order_details_after_order_table', array( $this, 'track_completed_order' ), 10, 1 );
 
 			$tracks = $this->get_tracks();
 			$tracks->init();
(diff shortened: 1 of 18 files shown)


woocommerce-services.php
```diff
@@ -1025,6 +1026,104 @@ public function reports_tabs( $reports ) {
...
+		 * @param $order WC_Order object containing completed order details.
+		 */
+		public function track_completed_order( $order ) {
+			// We need to recreate a temporary cart from the order
+			// so that we can create the same shipping packages from the cart
+			// to use to generate the correct cache key to return the cached
+			// shipping rates response.
+			$logger = $this->get_logger();
+
+			$temp_cart   = new WC_Cart();
```
superdav42, line 1042: If we only need to determine if customer selected a WCS shipping method at checkout it is not necessary to create a temporary cart here. Or even recalculate the shipping and packages. This code should be enough:
```php
foreach( $order->get_shipping_methods() as $item_shipping ) {
					$method = WC_Shipping_Zones::get_shipping_method($item_shipping->get_instance_id());
					if ( $method instanceof  WC_Connect_Shipping_Method ) {
						// send event
					}
				};
```

woocommerce-services.php
```diff
@@ -1025,6 +1026,104 @@ public function reports_tabs( $reports ) {
...
+							$e->getMessage()
+						),
+						__FUNCTION__
+					);
+					continue;
+				}
+			}
+
+			$api_client = $this->get_api_client();
+			$response   = $api_client->track_subscription_event( $services, $package, $custom_boxes, $predefined_boxes, $requested_rates );
```
superdav42, line 1113: It seems unnecessary to send all this data to the connect server. We only need to know which services were used and even that doesn't seem necessary since they all have the same event type.

superdav42 (changes requested):
I also have a concern that this will be easy for any merchant with a bit of php experience to disable the usage tracking completely but the live rates would still work fine. I suppose we could monitor the tracks data and if a customer has a disproportionate number of rate request vs tracking event then we could investigate if they are doing something nefarious.

woocommerce-services.php
c-shultz: Limiting the amount data seems like a good goal. I would vote we include the services data though so we have some flexibility with what we track in the future without updating the client (there's been some discussion about tracking/limiting usage for certain carriers, but not others)

c-shultz:
>I also have a concern that this will be easy for any merchant with a bit of php experience to disable the usage tracking completely but the live rates would still work fine. I suppose we could monitor the tracks data and if a customer has a disproportionate number of rate request vs tracking event then we could investigate if they are doing something nefarious.

The approach we take is a balance between engineering time and the potential for abuse. As long we can monitor for suspicious behavior, we can take a more simple approach to start with and watch if the need comes up to invest in this further.

woocommerce-services.php
```diff
@@ -770,6 +770,7 @@ public function attach_hooks() {
 			add_filter( 'wc_connect_shipping_service_settings', array( $this, 'shipping_service_settings' ), 10, 3 );
 			add_action( 'woocommerce_email_after_order_table', array( $this, 'add_tracking_info_to_emails' ), 10, 3 );
 			add_filter( 'woocommerce_admin_reports', array( $this, 'reports_tabs' ) );
+			add_action( 'woocommerce_order_details_after_order_table', array( $this, 'track_completed_order' ), 10, 1 );
```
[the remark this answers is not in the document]
FangedParakeet: Ah, great--I definitely didn't write the book on hooks (in fact, I've barely gotten past the cover--struggling to find very helpful documentation for most of these hooks), so thanks for enlightening me here! 

I've just updated the hook and everything appears to be working the same, so happy days.

woocommerce-services.php
FangedParakeet: The reason why I created the temporary cart is to get the cached rates request after checkout from inside of the transient where we store it during checkout. We need to determine whether live rates associated with subscription product usage events were used; to do this, I send the live rates to the server or, if unavailable, I send the shipping packages, so that the server can repeat the request to get the rates.  However, if we can send the cached rates to the server, then there is no need to repeat the shipping rates request on the Connect Server and the request should be much quicker. 

The cache key to access the transient value for a user's checkout is computed from a hash of the shipping method's service settings and the package, to keep it unique enough to use for a customer's specific order. You can find that code [here](https://github.com/Automattic/woocommerce-services/blob/eadf71f025d53214d2628f2440f5f53d65e20f1c/classes/class-wc-connect-shipping-method.php#L439) and the relevant section is below.

```
$cache_key = sprintf(
	'wcs_rates_%s',
	md5( serialize( array( $services, $package, $custom_boxes, $predefined_boxes ) ) )
);
```

I need to recreate the shipping packages at checkout in order to create the same hash to use the same key to return the correct rates from the transient. However, since the original cart is no more, I make a new one. That's why I also needed to perform the below `ksort` on the package contents, to ensure that the package object is the same in order to create the same hash.

```
foreach ( $package['contents'] as $item_key => $item_details ) {
	ksort( $package['contents'][ $item_key ] );
}
```

Figuring this out and getting the right package formatted to create exactly the same hash for the key was probably the least fun part of all of this work, so if there is a better way to access this transient that I am unaware of, I am definitely open to any suggestions!

woocommerce-services.php
FangedParakeet: Right now we only have one event type for all live rates, but I imagine this will change pretty quickly, if we wanted to have separate subscription products for different carriers. All this logic for the associations between carriers, usage events, and subscription products is contained in the Connect Server, so my plan from the start has basically been to send the rates there and let the server figure it out.

The reason why I'm sending the additional information about the shipping packages has almost entirely been my idea and I'm not sure how necessary it really is. The additional package information is only used as a fallback if the cached rates are not present for some reason. In this case, the server just repeats the shipping rates request to get the rates and then continues as normal. The only ways I can think that this would happen is if a user crippled the entire checkout by disabling cookies (though, I think the transient might even survive this, since it is database backed) or if a user took more than two hours to complete their checkout and the transient expired (which I think might generally break the entire checkout as well).

This is basically a fallback for a (hopefully) thin edge case, so if there are enough voices telling me that it's not worth worrying about, I will be content to sleep easy and disregard it. However, it doesn't seem like there's too much extra cost having it there though, other than the size of the HTTP request package to the server, since I already need to generate the shipping packages to try to get the cached rates out of the transient, as described above somewhere.

FangedParakeet:
> I also have a concern that this will be easy for any merchant with a bit of php experience to disable the usage tracking completely but the live rates would still work fine.

This has been a great concern of mine, but I can't see any viable options available for us to secure our remote plugin once it's on someone else's site. I think the only defence we have are our eagle eyes and making sure we track and monitor events, which is included in the other PR via Tracks.

woocommerce-services.php
superdav42: I think the cart object will still exist after switching to the new hook. So you should be able to use `WC()->get _shipping_packages()` in this code here now instead of creating a new cart. If it doesn't work we could switch to using the `woocommerce_checkout_create_order` action which happens before the `woocommerce_checkout_order_processed`

woocommerce-services.php
superdav42: I can't imagine how it could ever happen. The cart must generate the shipping prices during checkout and in the same request we are triggering this tracking method so there is no time at all for it to expire. We probably don't even need to use a transient and generate the caching key. We could have the shipping method call a add_tracking_event_data() which would set a property on this class. Then during the checkout hook we just need to read the tracking events that need to be sent.

FangedParakeet: > We could have the shipping method call a add_tracking_event_data() which would set a property on this class.

I like this idea. Let me try this out tomorrow morning and I'll strip out the rates request fallback while I'm at it as well and report back.

FangedParakeet: @superdav42 I've just removed this extra information, so now I am only sending the cached rates and the shipping method services information with the request, removing all the extraneous shipping package information.

woocommerce-services.php
FangedParakeet: @superdav42 I've just removed the temporary cart and replaced it with the code you suggested here!

woocommerce-services.php
```diff
@@ -1025,6 +1026,77 @@ public function reports_tabs( $reports ) {
...
+							'instance'         => $method->get_instance_id(),
+							'service_settings' => $service_settings,
+						),
+					);
+
+					if ( ! isset( $cached_response_body->rates ) ) {
+						continue;
+					}
+					$tracking_request_body[] = array(
+						'rates'    => $cached_response_body->rates,
```
superdav42, line 1070: The only part of rates used by the connect server to track the event is rates->id. I'm pretty sure this is the same as $service_schema->id set above. We could eliminate even more code if we only send the services. Then update the code at
https://github.com/Automattic/woocommerce-connect-server/pull/1697/files#diff-112c1dace901da978e2fca31f58eee61289c177ac513d4222a46685261f9fdb4R45
to just look at the service.id instead of rate.id.
Then we don't need the new method get_cached_shipping_rates_response

woocommerce-services.php
```diff
@@ -1025,6 +1026,77 @@ public function reports_tabs( $reports ) {
...
+
+			$packages = WC()->cart->get_shipping_packages();
+			$package  = $packages[0];
+			foreach ( $package['contents'] as $item_key => $item_details ) {
+				ksort( $package['contents'][ $item_key ] );
+			}
+			$package['rates'] = array();
+
+			$shipping_zone    = WC_Shipping_Zones::get_zone_matching_package( $package );
+			$shipping_methods = $shipping_zone->get_shipping_methods( true );
```
superdav42, line 1048: We should be able simplify this code a little by using a core method.
```suggestion
			$shipping_methods = WC()->shipping()->load_shipping_methods( $package )
```

superdav42 (changes requested):
I think we could simplify it even more by not sending the rates object. Really we only want to track if a WCS shipping method was shown during checkout so service id should be enough.

woocommerce-services.php
FangedParakeet: The reason why I decided to send the rates response was because in some cases we make a rates request for a shipping method, but return no rates for that method. For example, when I was testing this, I had FedEx and USPS rates enabled for my shipping zone, but I only received rates for USPS. I counted this as only having USPS live rates involved in the checkout and excluded FedEx, because this seemed like it would make more sense from a store owner/customer's perspective. 

However, that's just me making more assumptions at my own peril. @c-shultz, when we are checking whether a carrier's live rates are involved in a checkout, should we only count the carrier rates that a customer sees at checkout or do we count the carriers that we make requests for? To put that more simply, should we still count a carrier as being involved in a checkout if we don't return any rates for it? 

If the answer to that is a resounding `YES`, then I will be able to shed some more code here and probably put the whole transient trauma to bed once and for all as well.

woocommerce-services.php
```diff
@@ -1025,6 +1026,77 @@ public function reports_tabs( $reports ) {
...
+		 *
+		 * @param int   $order_id int for completed order ID.
+		 * @param array $posted_data Post data for order.
+		 * @param array $order WC_Order object containing completed order details.
+		 */
+		public function track_completed_order( $order_id, $data, $order ) {
+			$logger = $this->get_logger();
+
+			$packages = WC()->cart->get_shipping_packages();
+			$package  = $packages[0];
```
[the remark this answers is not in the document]
FangedParakeet: Gotcha, that would explain why `$packages` is an array in the first place. Noted and updated!

woocommerce-services.php
```diff
@@ -1025,6 +1026,77 @@ public function reports_tabs( $reports ) {
...
+			}
+			$package['rates'] = array();
+
+			$shipping_zone    = WC_Shipping_Zones::get_zone_matching_package( $package );
+			$shipping_methods = $shipping_zone->get_shipping_methods( true );
+
+			$tracking_request_body = array();
+			foreach ( $shipping_methods as $method ) {
+				try {
+					$handler = new WC_Connect_Shipping_Method( $method->get_instance_id() );
```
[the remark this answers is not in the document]
FangedParakeet: Gotcha, updated this as well.

woocommerce-services.php
FangedParakeet: Nice one, just updated this with your suggestion as well.

woocommerce-services.php
superdav42: Even if we want to ensure rates for WCS are present we should be able to use `WC()->shipping()->calculate_shipping_for_package()` to get the rates and then extract the service id from the rate id. Calculating the rates again should use the session cache so it won't be very intense. Then we don't need the method `get_cached_shipping_rates_response`.

FangedParakeet: OK, cool. @superdav42 I've now just ripped out all the transient cache commotion, replaced those parts with the `WC()->shipping()->calculate_shipping_for_package()` core function, and changed the endpoint so that we only need to send services instead of rates. This thins out the code even more and reduces the messy reliance on generating that session cache key, which I'm happy with.

superdav42 (approved):
Re-tested and fully working. Everything looks good now. Thanks.

PR #2318, opened 2021-02-05
Add API support for creating packages with POST and updating packages with PUT

jaclync:
## Description

Endpoint proposal: p91TBi-4xT-p2

Currently, there is one endpoint `POST /connect/packages` to **update** all the packages. In Woo mobile, we'd like to **create** packages without modifying the existing packages - it prevents the client from accidentally deleting packages that it hasn't fetched since mobile might not be always in sync with the backend.

This PR added support for creating packages by updating the existing `POST /connect/packages` that updates all packages to `PUT /connect/packages` and implementing `POST /connect/packages` to create new packages. Below is a list of major changes:

#### API
- Added functions to create custom and predefined packages in `WC_Connect_Service_Settings_Store`
- In `WC_REST_Connect_Base_Controller`, added support to handle PUT requests for a controller
- In `WC_REST_Connect_Packages_Controller`:
  - Updated the existing POST handling to create packages
  - Updated the previous POST handling that updates packages to PUT

#### Client
- Added support for PUT requests
- Updated existing packages update request from `post` to `put`

### Testing steps

Please confidence check on the existing behavior to update packages in wp-admin:

#### WCS settings

- Go to the WCS settings under `WooCommerce --> Settings --> Shipping --> WooCommerce Settings` (`/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings`)
- In the Packaging section, tap "Add package"
- Add a custom package or select from any of the service packages
- Tap "Add package" to add the package(s)
- Tap "Save changes" at the bottom of the section --> the success banner "Your shipping settings have been saved." should appear, and the new package(s) should be shown after reloading the settings page

#### Order details > create shipping label

- Go to an order where a shipping label can be created
- Tap "Create shipping label" CTA
- Expand the Packaging tab, and tap "Add package" next to "Package details"
- Add a custom package or select from any of the service packages
- Tap "Add package" to add the package(s) --> the success banner "Your shipping packages have been saved.", and the created packages should appear in the "Package details" dropdown list

### Checklist

- [x] unit tests


harriswong:
👀

harriswong (approved):
Thanks @jaclync ! Sorry this took a while. I tested 
- Creating a package inside "Create shipping label" within an order
- Creating a package inside the admin link you posted
- Tested `PUT` against `http://localhost/wp-json/wc/v1/connect/packages` via postman 
- Tested `POST` against `http://localhost/wp-json/wc/v1/connect/packages` via postman. This will duplicate all packages if i post with existing package data. This is fine as long as the end point doesn't care about duplicates. I think that's fine. 
- phpunit tests look good and passed

LGTM 👍

harriswong:
@jaclync We recently deployed a hotfix and it seems to conflict with your changelog. Please update! Sorry about that.

classes/class-wc-connect-service-settings-store.php
```diff
@@ -491,6 +491,20 @@ public function get_packages() {
...
+		 * Extends the global list of packages with a list of new packages
+		 *
+		 * @param array new_packages - packages to extend
+		 */
+		public function create_packages( $new_packages ) {
+			if (is_null($new_packages)) {
+				return;
+			}
+			$packages = $this->get_packages();
+			$packages = array_merge($packages, $new_packages);
```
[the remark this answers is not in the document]
jaclync: Thanks for raising this potential duplication issue! I think it'd be better to throw an error in case of duplication, since it's a client side error. I added two commits https://github.com/Automattic/woocommerce-services/pull/2318/commits/901e5d7727acd8a772c14b07d31b59cd01e9ddf5 https://github.com/Automattic/woocommerce-services/pull/2318/commits/ce425b8ba72b7f3ee7d42d1f0ce6fc4a68166065 to handle the following two duplication scenarios:

- The custom/predefined packages in the request body do not have unique names
- Any of the custom/predefined packages have the same names of the existing packages

Sorry that this duplication validation logic gets a bit complicated, I tried adding a test case for each scenario and package type. If you think it's better to handle these duplication scenarios differently, please let me know and I can update the PR accordingly!

woocommerce-services.php
```diff
@@ -843,7 +843,7 @@ public function rest_api_init() {
 			$this->set_rest_self_help_controller( $rest_self_help_controller );
 			$rest_self_help_controller->register_routes();
 
-			require_once __DIR__ . 'classes/class-wc-rest-connect-service-data-refresh-controller.php' ) );
+			require_once __DIR__ . '/classes/class-wc-rest-connect-service-data-refresh-controller.php';
```
jaclync, line 846: After merging from `develop`, I updated this line for phpunit and the web client to run as before

jaclync, line 846: After merging from `develop`, I updated this line for phpunit and the web client to run as before

changelog.txt
```diff
@@ -3,6 +3,7 @@
 = 1.25.8 - 2021-xx-xx =
 * Tweak - Display better errors on checkout page when address fields are missing / invalid.
 * Fix   - UPS invoice number allows numbers and letters.
+* Add   - Update the existing endpoint `POST /connect/packages` to create shipping label packages, and add an endpoint `PUT /connect/packages` to update shipping label packages.
```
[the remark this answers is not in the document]
jaclync: Yea it'd be ideal if this breaking change can be in a new version so that it doesn't affect any users of this endpoint (Chris mentioned Calypso is one of them other than the WCS web client and it's being removed, but maybe there are external developers using this endpoint).

If API versioning is implemented soon (this or next release), we can certainly wait for the endpoint to be versioned 😄 For Woo mobile, we also have another new endpoint on shipping label creation eligibility https://github.com/Automattic/woocommerce-services/pull/2334 that we'd love to release in the near future that is higher priority than this package creation endpoint.

jaclync:
@harriswong Thanks for the review and catching the duplication issue! I updated the PR to handle duplicate packages and responded to your comments for another look. I'm open to any feedback 😄

woocommerce-services.php
harriswong: 😓 There was some merge conflict madness and this should be fixed with this PR! https://github.com/Automattic/woocommerce-services/pull/2350

changelog.txt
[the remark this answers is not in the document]
harriswong: We had a brief discussion on slack (p1614723041112800-slack-CU6SYV31A). We think the risk/impact is low and this is okay to go ahead

We talked about:

**Affecting areas**
- Only impact users that has custom code on the "shipping label creation UI" or the "admin create package UI". We don't have a metric on that but we don't think there would be a lot of users.
- Affect adding packages. 

**Impact**
- ~~They get duplicated packages~~ They won't be able to update packages.

**What user should do when upgrading**
Similar to how we upgrade npm packages., we expect users to look at the changelog when they upgrade.

classes/class-wc-connect-service-settings-store.php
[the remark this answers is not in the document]
harriswong: `$this->settings_store->get_predefined_packages();` and `$this->settings_store->get_packages();` returns an array without a `key`. Finding duplicates will involve looping through the list to map out the `key` we want. I think I would do more or less the same thing you did in the last 2 commits. This looks 👍  to me!

classes/class-wc-rest-connect-packages-controller.php
```diff
@@ -43,4 +43,93 @@ public function post( $request ) {
...
+					if (!empty($duplicate_package_names)) {
+						$duplicate_package_names_by_carrier[$carrier] = array_values($duplicate_package_names);
+					}
+				}
+			}
+
+			if (!empty($duplicate_package_names_by_carrier)) {
+				return new WP_Error(
+					400,
+					'duplicate_predefined_package_names_of_existing_packages',
```
harriswong, line 124: Is this supposed to be a string with underscore or is this supposed to be a translation? If it is a translation, we can use the `__()` function.

harriswong (commented):
Tested the last 2 commits with different payload involving:
**predefined**:
- duplicated value in `predefined` array
- a new carrier name, ie.`usps-3`
- rearranged `predefined` array indiced with the same value. Throw 400 as expected

**custom**
- tested duplicated packages
- tested new package 

*other payload**
- tested empty `{}` payload body

This looks good to me 👍 . Just 3 more things:
1. Please merge latest `develop` again, i hope it doesn't create too much conflicts for you 😓 
2. I have a question regarding the language string inside the 400 error. 
3. I noticed that my editor complain about some phpcs things. You can run `composer install` and then `vendor/bin/phpcs classes/class-wc-rest-connect-packages-controller.php` to see all the phpcs messages; or run `vendor/bin/phpcbf classes/class-wc-rest-connect-packages-controller.php` to help you auto-fix all/most of them.

classes/class-wc-rest-connect-packages-controller.php
jaclync: This is meant for a unique identifier of the error so that the client can handle the error based on the identifier - each client might show a different error string to the user, and a mobile user might be on a device of a different language from their site. If we want to add a translated string in the response in the future, we could probably do so in the error data (L137 below)?

jaclync: I'd also like to document the possible error cases including the scenario, error identifier, and error data format - do you know if there is a good place to do so?

woocommerce-services.php
jaclync: Cool I merged the latest `develop` to PR branch! I didn't get any merge conflicts 😄

changelog.txt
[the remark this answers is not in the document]
jaclync: Thank you for confirming this with the team! In case of any existing endpoint users, getting an error sounds better than getting duplicated packages 🙂

classes/class-wc-rest-connect-packages-controller.php
jaclync: Update: I noticed during Postman testing that the error response had status code 500 instead of 400 that I intended to use, and I understood more about using a localizable string in the message. In https://github.com/Automattic/woocommerce-services/pull/2318/commits/d05b7c7e2e3a1caa8109e67e29bc263db915267e, I replaced `WP_Error` (which I couldn't find a way to change the status code) to `WP_REST_Response` with 400 status code with `code`, `message`, and `data` in the response.

Please lemme know if there's a better way to return a 400 error!

jaclync:
Thanks a lot for the suggestions! I updated the PR, any other feedback is welcome 😃 

✅ Merged latest `develop` again in https://github.com/Automattic/woocommerce-services/pull/2318/commits/e13748c66622a526f3ebac829a0aafc9c8376e35
💬 Responded to your question regarding the language string inside the 400 errors https://github.com/Automattic/woocommerce-services/pull/2318#discussion_r587091101, and replaced `WP_Error` with `WP_REST_Response` to throw 400 errors with localizable strings in the message field in https://github.com/Automattic/woocommerce-services/pull/2318/commits/d05b7c7e2e3a1caa8109e67e29bc263db915267e
✅ Auto-fixed most of the phpcs issues with phpcbf in https://github.com/Automattic/woocommerce-services/pull/2318/commits/7d4b49b3d77688560fc49603e730c6b54ab65db2, and fixed some of the remaining issues by adding comments on the affected functions in https://github.com/Automattic/woocommerce-services/pull/2318/commits/83ff5670a80605cd404c38422b5e82105b1147a4

harriswong (approved):
I like the use of `WP_REST_Response` in the last commit. There was an issue I experienced where `'success'=>true` needs to be included (p1604349886080900-slack-CU6SYV31A) in the response. However I couldn't remember now exactly what the problem was. :D I think it only affects our `200` responses and it has something to do with how WCS makes request in our React app [here](https://github.com/Automattic/woocommerce-services/blob/6084e075f2b828a8fe7a45c4d4c02aeee76029b7/client/api/request.js#L41).

This PR already included `success` for 200 and the error will be thrown with `400` regardless, so this is fine. 

I tested both conditions for the errors and I can confirm that i am getting `400` back with the following:
```
"code": "duplicate_predefined_package_names_of_existing_packages",
    "message": "At least one of the new predefined packages has the same name as existing packages.",
    "data": {
        "package_names_by_carrier": {
```
```
{
    "code": "duplicate_custom_package_names_of_existing_packages",
    "message": "At least one of the new custom packages has the same name as existing packages.",
    "data": {
        "package_names": [
            "My box",
```

Great work figuring all these out @jaclync. LGTM! 👍

jaclync:
Thank you @harriswong for the thoughtful review! What's the merge process for WCShip - who does the merge, and is squash & merge the recommended way? There is one check on `continuous-integration/travis-ci` that is always pending without a way to trigger it (some open PRs also have this while some don't) 🤔

harriswong:
Yep! We do squash and merge for PRs. The travis-ci  was moved from .org to .com and it stopped running. We are in the process of migrating it to use GitHub Actions (https://github.com/Automattic/woocommerce-services/issues/2328). 

I can help you merge this in the mean time.

PR #2322, opened 2021-02-05
Run phpcbf to fix most code standards issues

superdav42:
## Description
All the phpcs notices were annoying me. Ran phpcbf over all the php files with:
```sh
vendor/bin/phpcbf classes/* tests/php/ woocommerce-services.php
```

Only about 1916 phpcs errors exists as reported by `vendor/bin/phpcs classes/* tests/php/ woocommerce-services.php | wc -l`
vs 4760 that used to be reported.

### Checklist

- [x] unit tests


harriswong (changes requested):
I kinda scroll through trusting `phpcbf` doing the work here. As long as tests passed, then this is good.

There are conflicts, i can 👍  after it's resolved.

harriswong (approved):
Tests passed.

PR #2325, opened 2021-02-10
Implement compatibility with WooCommerce Blocks Cart and Checkout.

budzanowski:
This PR enables WoooCommerce Shipping & Tax compatibility with WooCommerce Blocks: Cart and Checkout blocks.

This is only the tax side of the integration - as it looks like for Shipping there is nothing todo.

In general, the code supports the integration. There are two modifications that are needed to enable live tax rates. The code that blocks the functionality was originally added to limit the number of costly API requests. This PR adjusts the limiting code and tries to do it in a way that all of the unnecessary cases are still blocked.

Fixes: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3824
Fixes: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3836
Fixes: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3849
Fixes: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3838 (01c5ab8b01b69b4698e68c0675a6c84590c4d1c0) 

Project thread: pbIy4N-17p-p2


classes/class-wc-connect-functions.php
```diff
@@ -39,5 +41,44 @@ public static function get_wc_helper_auth_info() {
...
 			return $helper_auth_data;
 		}
+
+		/**
+		 * Check if we are currently in Rest API request for the wc/store/cart API call.
+		 *
+		 * @return bool
+		 */
+		public static function is_store_api_call() {
+			if ( ! defined( 'REST_REQUEST' ) && empty( $GLOBALS['wp']->query_vars['rest_route'] ) ) {
```
nerrad, line 51: would `wc()->is_rest_api_request()` help here?

budzanowski: In the scope of the whole line not really. It could replace the defined check but I don't see much benefit. Am I missing something?

nerrad: [wc()->is_rest_api_request()](https://github.com/woocommerce/woocommerce/blob/b7339a3114e3e20d2e2cc51e6ac0e9536be7c6af/includes/class-woocommerce.php#L296-L305) seems to be a more robust detection implementation? There's already an abstraction in Woo for detecting rest requests so I was just thinking it'd helpful to use it here and make the code a bit more concise. The only downside is that `is_rest_api_request` is filterable (not sure why really).

nerrad: Also and maybe related, it might be worthwhile double checking that implementation of `REST_REQUEST` constant use here is okay given when it is set in the request stack (more context can be found in [this trac ticket](https://core.trac.wordpress.org/ticket/42061)).

budzanowski: Adjusted. It should be OK to use `REST_REQUEST` for now, but the function should be bullet proof if anything changes in wcship.

superdav42 (approved):
LGTM

Issue #2328, opened 2021-02-18
Migrate travis-ci to GitHub Actions


harriswong:
### Description
We had some issues with our E2E tests and travis' availability (https://github.com/Automattic/woocommerce-services/issues/2305). Take this as an opportunity to explore migration to GitHub Actions.

### Requirements
- Support PHP 7+  (8 is blocked by https://github.com/Automattic/woocommerce-services/issues/2258)

### Output
- ~~Can we create a pipeline in similar fashion? Run tests, merge etc.~~
- ~~How do we support matrix~~
- ~~P2 on findings~~ ✅
- ~~POC PR~~
- Migrate to GitHub Actions and remove travis

harriswong:
### Update
- Following the same way we run our pipeline in [Travis](https://github.com/Automattic/woocommerce-services/blob/develop/.travis.yml), the current [PR](https://github.com/Automattic/woocommerce-services/pull/2329/files) will download WC `5.0.0`, WP `5.6.2`. Install and setup DB.
- This PR can run both JS unit tests, lints.

### TODO
When running phpunit via `./vendor/bin/phpunit -c phpunit.xml.dist`, there is an error `Could not determine test root directory. Aborting.`. This is thrown by [our bootstrap code](https://github.com/Automattic/woocommerce-services/blob/develop/tests/php/bootstrap.php#L20), suggesting that it is not finding the woocommerce core test utils. I'll look into this.

harriswong:
### Next step
- Get this to work under PHP 7.4
- Test WP and WC over a matrix

Issue #2258, opened 2020-11-17
PHPUnit tests with PHP 8


bborman22:
While testing WooCommerce Shipping & Tax with PHP 8 and WP 5.6 I ran into an issue with running PHPUnit with PHP 8. The summary is that WordPress only supports up to PHPUnit 7.x, but PHPUnit 7.x doesn't support PHP 8. You can see the [WP core issue](https://core.trac.wordpress.org/ticket/50902) and their solution was to essentially [override the problematic pieces](https://core.trac.wordpress.org/changeset/49037) of PHPUnit. And there is the proposed WooCommerce solution, mentioned at the bottom of that WP core issue, of [forking PHPUnit](https://github.com/woocommerce/phpunit/pull/1) and [using that fork for unit testing](https://github.com/woocommerce/woocommerce/pull/27844).

This issue is being created to make sure that once a solution is implemented in WP 5.6 and the latest WooCommerce, we confirm that the PHPUnit tests are working as expected on PHP 8.

dustinparker:
hey @bborman22,

I tested implementing the changes from [WP Core's Changeset 49037](https://core.trac.wordpress.org/changeset/49037) and running PHPUnit 7.5 with PHP 8.0 and all tests passed using WP 5.6 and WC 5.9.0.

bborman22:
That's great to hear! I'm a little bit fuzzy on all the details from this. Are those changes something we can commit in to the repo or could we get them from WC in some way? Would our update be similar to https://github.com/woocommerce/woocommerce/pull/27844/files? Thanks for looking in to this one!

dustinparker:
I just tested running the compatibility fork with PHP 8.0 [as instructed here](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/tests#running-tests-in-php-8) and it also works. That is actually the simpler way to do it.

dustinparker:
hey @bborman22, was there anything else related to this issue you'd like me to look into?

bborman22:
Thanks! No I don't think there is anything else to verify at this point since we're able to get tests in PHP 8 to work. @harriswong I see you had a linked issue around adding PHP versions to the GH actions, looks like that item is unblocked now and we could look in to [this solution](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/tests#running-tests-in-php-8) linked above. Seems like this could be a good cooldown / janitorial task at some point?

harriswong:
[...]

PR #2216, opened 2020-10-09
Add composer command to run phpunit on WCS

harriswong:
## Description
Add composer to run existing test suites inside WooCommerce Services. Update `phpunit.xml.dist` to use `legacy` WooCommerce Core test utils. 

### Notes
There is 1 failed test if we run against the latest WooCommerce Core. This will be fixed in another issue https://github.com/Automattic/woocommerce-services/issues/1631, where we will update our Travis pipeline to include a pipeline test to run against latest woo core.

### Related issue(s)
Closes https://github.com/Automattic/woocommerce-services/issues/2211

### Steps to reproduce & screenshots/GIFs
Do you already have WordPress PHPUnit database/test setup? If so, skip to step 7.

1. Install `composer` if you don't have it already. Install it [globally](https://getcomposer.org/doc/00-intro.md#globally) so you can use it in any repo.
2. Run `composer install`
3. Make sure you have `wp-cli` installed. If not, follow [this step](https://make.wordpress.org/cli/handbook/guides/installing/).
4. Now you need to setup a WordPress Test database. Follow instruction [here](https://make.wordpress.org/cli/handbook/misc/plugin-unit-tests/#running-tests-locally).
When running the `wp scaffold plugin-tests woocommerce-services` command, don’t overwrite existing files.
5. You will need a local copy of [WooCommerce core from GitHub](https://github.com/woocommerce/woocommerce). We need this because our tests use cores’ test utility.
6. Add an environmental variable to your `.bash_profile` file that points to your woocommerce core repo. Mine looks like:
`export WC_DEVELOP_DIR=/wordpress/wp-content/plugins/woocommerce-clone`. This allows WCS tests to look for WooCommerce core’s test utils.
7. Type `composer run test` and you should see
```shell
Installing...
Running as single site... To run multisite, use -c tests/phpunit/multisite.xml
Installing WooCommerce...
Not running ajax tests. To execute these, use --group ajax.
Not running ms-files tests. To execute these, use --group ms-files.
Not running external-http tests. To execute these, use --group external-http.
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.3.11
/woocommerce-services/phpunit.xml.dist

................................................................. 65 / 98 ( 66%)
.................F...............                                 98 / 98 (100%)

Time: 22.8 seconds, Memory: 62.50 MB

There was 1 failure:

1) WP_Test_WC_Connect_Tracks_No_Jetpack::test_no_jetpack
[...]

PR #2329, opened 2021-02-18
Add github action to run js and php test

harriswong:
## Description
Migrate to Github Actions from travis ci. Test should run under PHP 7.4, latest WordPress(Latest - 2), and WooCommerce(Latest - 2).

## Changes in this PR
- Run JS tests, linting in its own job
- Run PHP tests in its own job
- Remove Travis (`.travis.yml`)
- Rename all php `test_` to `test-` for our phpcs linting rules to pass
- Fix this broken test (https://github.com/Automattic/woocommerce-services/pull/2216#issuecomment-708064854) and make sure PHPUnit tests pass
```
1) WP_Test_WC_Connect_Tracks_No_Jetpack::test_no_jetpack
Expectation failed for method name is equal to 'log' when invoked 1 time(s)
Parameter 0 for invocation WC_Connect_Logger::log('Tracked the following event: ...ed_out', '', false) does not match expected value.
Failed asserting that 'Tracked the following event: woocommerceconnect_opted_out' contains "error".
```

### Related issue(s)
Closes https://github.com/Automattic/woocommerce-services/issues/2328

1. At the bottom of this page, make sure these action pass
image

### TODO
- ~~Run this against a matrix of PHP versions (7.3, 7.4, 8) etc~~ ✅ Using only 7.4 for now
- ~~Run this against WP matrix according to our [support policy](https://github.com/Automattic/woocommerce-services/issues/2131)~~ ✅
- ~~Run this against WC matrix according to our [support policy](https://github.com/Automattic/woocommerce-services/issues/2131)~~ ✅
- Would be nice to create our own WooCommerce GitHub Action so that it handles all the setup and it will support WP x WC test matrix
- Remove debugging `steps` in the GitHub actions if not necessarily.
- Can not test PHP 8, blocked by https://github.com/Automattic/woocommerce-services/issues/2258
- Remove Travis from running at the bottom of this page. 

### Checklist

- [x] unit tests


.github/workflows/main.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Checkout latest WCS
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Switch to PHP 7.4
+        run: |
+          sudo update-alternatives --set php /usr/bin/php7.4
+          sudo update-alternatives --set phar /usr/bin/phar7.4
+          sudo update-alternatives --set phar.phar /usr/bin/phar.phar7.4
+      - name: Verify PHP version
```
harriswong, line 50: I added some verifying steps to help debugging. These can be removed when we decided to use this in prod.

.github/workflows/main.yml
```diff
@@ -0,0 +1,73 @@
...
+          MYSQL_PASSWORD: wp_test
+          MYSQL_DATABASE: wordpress_default
+          MYSQL_ROOT_PASSWORD: root
+        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+    steps:
+      - name: Checkout latest WCS
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Switch to PHP 7.4
```
harriswong, line 45: Using GitHub's ubuntu images but test against PHP 7.4 only. We can't test against PHP 8 due to this issue https://github.com/Automattic/woocommerce-services/issues/2258

.github/workflows/main.yml
```diff
@@ -0,0 +1,73 @@
...
+          php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+          php -r "if (hash_file('sha384', 'composer-setup.php') === '756890a4488ce9024fc62c56153228907f1545c228516cbf63f885e036d37e9a59d27d63f46af1d4d07ee0f76181c7d3') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
+          php composer-setup.php
+          php -r "unlink('composer-setup.php');"
+          sudo mv composer.phar /usr/local/bin/composer
+          composer install
+      - name: Setup WP and WC
+        run: tests/bin/install-wc-tests.sh wordpress_test root root 127.0.0.1 5.6.2 release/5.0
+      - name: Run PHPUnit tests
+        run: ./vendor/bin/phpunit -c phpunit.xml.dist
```
harriswong, line 73: Using existing phpunit config that we use for travis

.github/workflows/main.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Install composer
+        run: |
+          php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+          php -r "if (hash_file('sha384', 'composer-setup.php') === '756890a4488ce9024fc62c56153228907f1545c228516cbf63f885e036d37e9a59d27d63f46af1d4d07ee0f76181c7d3') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
+          php composer-setup.php
+          php -r "unlink('composer-setup.php');"
+          sudo mv composer.phar /usr/local/bin/composer
+          composer install
+      - name: Setup WP and WC
+        run: tests/bin/install-wc-tests.sh wordpress_test root root 127.0.0.1 5.6.2 release/5.0
```
harriswong, line 71: Using the same setup script we use for Travis. It would be nice if we can make this into our own WooCommerce Action.

.github/workflows/main.yml
```diff
@@ -0,0 +1,73 @@
...
+          sudo update-alternatives --set phar /usr/bin/phar7.4
+          sudo update-alternatives --set phar.phar /usr/bin/phar.phar7.4
+      - name: Verify PHP version
+        run: |
+          lsb_release -a
+          sudo update-alternatives --config php
+          sudo update-alternatives --config phar
+          sudo update-alternatives --config phar.phar
+          php -v
+      - name: Verify MySQL connection
```
harriswong, line 57: I added this to check if the DB service was actually connected during debugging. We can remove this later as well.

.gitignore
```diff
@@ -20,7 +20,6 @@ release/
 !/logs/.htaccess
 
 /vendor/
-composer.lock
```
harriswong, line 23: We should always commit the lock files. Same reason why we should commit `package-lock.json`

phpunit.xml.dist
```diff
@@ -10,7 +10,7 @@
 	>
 	<testsuites>
 		<testsuite name="WooCommerce Services Test Suite">
-			<directory suffix=".php">./tests/</directory>
+			<directory prefix="test-" suffix=".php">./tests/php</directory>
```
harriswong, line 13: PHPUnit's XML config default on scanning files with `Test.php` suffix. suffix and prefix for our WP/WC file name rules are needed.

tests/php/test-woocommerce-connect-tracks.php
```diff
@@ -26,8 +26,7 @@ public function test_no_jetpack() {
 		$this->logger->expects( $this->once() )
 			->method( 'log' )
 			->with(
-				$this->stringContains( 'Error' ),
-				$this->anything()
+				$this->stringContains( 'Error. jetpack_tracks_record_event is not defined.' )
```
harriswong, line 29: `logger->log()` function only takes 1 parameter. The 2nd is not needed. Fixing this in this PR as well. This fix should resolve this issue https://github.com/Automattic/woocommerce-services/pull/2216#issuecomment-708064854.

harriswong:
@FangedParakeet For the E2E tests you are working on, I think you can just add a `e2e..yml` inside `.github/workflows/` and that should be it!

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,23 @@
+name: JS Tests
+
+on:
+  push:
+    branches: [ develop ]
+  pull_request:
+    branches: [ develop ]
```
frosso, line 7: This should maybe also include `master`, for the releases?
I understand that `master` and `develop` should contain similar stuff, but in the case of hotfixes they wouldn't

.github/workflows/php_tests.yml
```diff
@@ -0,0 +1,65 @@
...
+          submodules: recursive
+      - name: Switch to PHP 7.4
+        run: |
+          sudo update-alternatives --set php /usr/bin/php7.4
+          sudo update-alternatives --set phar /usr/bin/phar7.4
+          sudo update-alternatives --set phar.phar /usr/bin/phar.phar7.4
+      - name: Install composer
+        run: |
+          php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+          php -r "if (hash_file('sha384', 'composer-setup.php') === '756890a4488ce9024fc62c56153228907f1545c228516cbf63f885e036d37e9a59d27d63f46af1d4d07ee0f76181c7d3') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
```
frosso, line 57: Why include the hash check? When a new version of `composer-setup` is released, wouldn't this break?

.github/workflows/js_tests.yml
harriswong: Should i do all pull and pushes actually? `on: [push, pull_request]`

.github/workflows/php_tests.yml
harriswong: I copied the download script directly from https://getcomposer.org/download/. 

> When a new version of composer-setup is released, wouldn't this break?

This is a valid concern. Do we care about the file integrity/signing here? Or should we manually update it when it breaks?

harriswong: Maybe I should do this instread: https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md

.github/workflows/js_tests.yml
frosso: Sure - that would do it on all the branches, correct?
I think that makes sense 👍

.github/workflows/php_tests.yml
frosso: > This is a valid concern. Do we care about the file integrity/signing here? Or should we manually update it when it breaks?

I wouldn't be interested about the new version. I think I would find cumbersome to update it at every version, while nothing can get merged until a PR is opened to fix the hash in the CI configuration

.github/workflows/js_tests.yml
harriswong: Fixed.

.github/workflows/php_tests.yml
harriswong: Fixed. Now using a script to download latest composer and also check its signature.

tests/bin/install-wc-tests.sh
```diff
@@ -10,14 +10,19 @@ DB_USER=$2
 DB_PASS=$3
 DB_HOST=${4-localhost}
 WP_VERSION=${5-latest}
-WC_VERSION=${6-"2.6.14"}
+WC_VERSION=${6-"5.0.0"}
```
frosso, line 13: I got confused for a second here because I didn't see that the arguments were changing based on https://github.com/Automattic/woocommerce-services/pull/2329/files#diff-52b429ab0f697095c91efe50e63752acc671f157d2661ba5404da3bdeb497b0aR55
But I see now that these are default values :)

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,18 @@
...
+  js_build_and_test:
+    name: JS Estlin and Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '10'
```
frosso, line 15: Shouldn't this match the value in `.nvmrc` - `10.11.0`?

frosso (approved):
Thanks for the changes - I think this looks good AFAIK 👍

.github/workflows/js_tests.yml
harriswong: `lint-staged requires at least version 10.13.0 of Node, please upgrade` I think we need at least 10.13.0. I am going to bump up our `.nvmrc`

harriswong: Actually, this should probably be done with this issue: https://github.com/Automattic/woocommerce-services/issues/2082

I am going to address it there instead.

frosso: Oh yeah - always getting that error on the pre-commit hooks :)

PR #2330, opened 2021-02-19
Remove unused files

c-shultz:
## Description
This PR strips out a lot of unused js code. 

For some history on this: the WCS application code started out in the wp-calypso repo and was all copied into this repo. It looks like a lot of unnecessary stuff was copied over.

I ended up (temporarily) adding this plugin to Webpack to identify the unused files: https://www.npmjs.com/package/unused-files-webpack-plugin

I ended up using [BeyondCompare](https://www.scootersoftware.com/) to compare the result of `npm run build` before and after these changes. The resulting build files were totally identical except some of the translation strings were removed, e.g.
image

These strings all seem to have equivalents elsewhere in the strings file or appear to be totally unused in WCShip, so I think we're all good.

### Testing/checking
- General testing of WCShip (settings screen, label printing, etc.)
    - Check for new console errors
- Compare built files before/after these changes


brezocordero (approved):
559 files! This is great.

I have tested the following:
* Order screen
  * Create new label: Testing the different steps on the purchase form, errors on the form and selecting different options.
  * Options displayed on the ellipsis menu: Refund, Details and Print
  * Track packages
* Settings screen:
  * Shipping Labels Card: Toggle, credit card, paper size, checkbox
  * Packaging: Adding and removing
  * Carrier and Subscriptions: Check they display as expected
  * Saving settings
* Live rates: I don't think this would be affected but tested we still get live rates
* Status:  Refresh, toggles, dropdown.

It is looking good. 

It is a big change but I think we should merge into the feature branch so that we can continue testing as we make the other changes.

c-shultz:
Part of effort to clean up Calypso dependencies: https://github.com/Automattic/woocommerce-services/issues/2333

Issue #2333, opened 2021-02-20
Remove dependency on Calypso Repository Git Sub-module


c-shultz:
(todo) 😄

c-shultz:
Let's discuss this at the team meeting, but I'm tentatively thinking we should pause on merging this to `develop`. It would be a risky time to deploy this major of an update.

harriswong:
Added https://github.com/Automattic/woocommerce-services/issues/2435 to track issues related to releasing `calypso-cleanup` branch. This issue should remain open since we have not finished all things on step 2.

harriswong:
It seems like the all the Calypso's `<Form..>` components do not exist in WooCommerce Components nor WordPress components. We will have to copy all of them over.

harriswong:
> QueryStoredCards from 'components/data/query-stored-cards';

Removing this will be a bit tricky since it relies on a lot more than just Calypso components. We will have to copy/move these over as well:
```js
import { isFetchingStoredCards } from 'state/stored-cards/selectors';
import { fetchStoredCards } from 'state/stored-cards/actions';
```

 **update** 
I think we don't need this component anymore because we are fetching settings from the PHP side. https://github.com/Automattic/woocommerce-services/pull/2602

harriswong:
TODO: Before removing, maybe run `dependency-cruiser` after all these to make sure no dependency left.

harriswong:
I think tooltip is done. https://github.com/Automattic/woocommerce-services/pull/1930

harriswong:
We're pausing work on this.

PR #2336, opened 2021-02-22
Bump @wordpress version and use Modal component for shipping label purchase

c-shultz:
## Description
- Bumps @wordpress `components` and elements versions up. The updated styling of `components` v11 improved the look and feel of the Modal component. I ran into some dependency road blocks with using v12, so we can save that update for later.

- Replaces Calypo's `Dialog` component with @wordpress's `Modal` component for the shipping label purchase modal

This only replaces the component on the shipping label purchase modal. I'll continue the Dialog replacement efforts on future PRs. I decided to stop here because this was already pretty big (especially given the dependency updates).

The shipping summary has also been nested in a Card component because that worked best with the updated header styles.

There are numerous style updates to make this all work well.

### Related issue(s)

image

Testing:
- Test all JS since the `@wordpress` update could affect any part of the JS app
- Test shipping label purchase modal
    - Mobile vs. desktop
    - Styling

I took some liberties with the design. I think we should aim to use the replacement components as intended, so I prioritized trying to make things look nice instead of trying to totally replicate the old design/layout. Let me know if there's any big concerns, otherwise we can get some design feedback once more components have been replaced


assets/stylesheets/style.scss
```diff
@@ -250,6 +250,7 @@
 
 // Just the tooltip WordPress Component styles is all we use for now.
 @import "~@wordpress/base-styles/colors";
+@import "~@wordpress/base-styles/variables";
```
c-shultz, line 253: This has to be manually added here so @wordpress components can find it. I would have thought this would be handled automatically by npm/webpack, but this is also how wc-admin does things.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/index.js
```diff
@@ -39,44 +36,39 @@ const LabelPurchaseModal = props => {
...
-			additionalClassNames="woocommerce label-purchase-modal wcc-root"
-			isVisible={ props.showPurchaseDialog }
-			onClose={ onClose }
-		>
-			<div className="label-purchase-modal__content">
-				<div className="label-purchase-modal__header">
-					<FormSectionHeading>
-						{ translate( 'Create shipping label', 'Create shipping labels', { count: Object.keys( props.form.packages.selected ).length } ) }
-					</FormSectionHeading>
-					<Button className="label-purchase-modal__close-button" onClick={ onClose }>
```
c-shultz, line 52: Button is built into Modal component

package.json
```diff
@@ -27,10 +27,9 @@
   "license": "GPL-2.0",
   "description": "Connect allthethings",
   "devDependencies": {
-    "@slack/web-api": "^5.8.0",
+    "@slack/web-api": "^5.15.0",
```
c-shultz, line 30: Several version got bumped to the latest compatible version (by `npm update`). I figured this was a good time while we we're tearing everything up anyway 🤷

package.json
```diff
@@ -82,10 +81,11 @@
...
     "@babel/preset-react": "7.0.0",
     "@babel/runtime": "7.1.2",
-    "@hot-loader/react-dom": "^16.7.0",
-    "@wordpress/babel-preset-default": "^4.11.0",
-    "@wordpress/components": "8.5.0",
-    "@wordpress/element": "2.10.0",
+    "@hot-loader/react-dom": "^16.14.0",
+    "@wordpress/babel-preset-default": "^5.0.1",
+    "@wordpress/base-styles": "^3.3.0",
+    "@wordpress/components": "^12.0.7",
```
c-shultz, line 87: This is a major update of @wordpress/components from 8.5 to 12.0.7

npm-shrinkwrap.json
```diff
@@ -16143,15 +17636,38 @@
...
+          "version": "0.15.1",
+          "resolved": "https://registry.npmjs.org/reakit-utils/-/reakit-utils-0.15.1.tgz",
+          "integrity": "sha512-6cZgKGvOkAMQgkwU9jdYbHfkuIN1Pr+vwcB19plLvcTfVN0Or10JhIuj9X+JaPZyI7ydqTDFaKNdUcDP69o/+Q=="
+        },
+        "reakit-warning": {
+          "version": "0.6.1",
+          "resolved": "https://registry.npmjs.org/reakit-warning/-/reakit-warning-0.6.1.tgz",
+          "integrity": "sha512-poFUV0EyxB+CcV9uTNBAFmcgsnR2DzAbOTkld4Ul+QOKSeEHZB3b3+MoZQgcYHmbvG19Na1uWaM7ES+/Eyr8tQ==",
+          "requires": {
+            "reakit-utils": "^0.15.1"
```
c-shultz, line 17668: This was 0.14.4 by default but it was causing a conflict with the version required by @wordpress/components, so I manually bumped it to 0.15.1. It looks some level of instability is expected around the version updates with reakit-utils: "[This is experimental and may have breaking changes in minor versions.](https://www.npmjs.com/package/reakit-utils)"

package.json
c-shultz: I went back to 11.1.15 because the dependencies are messy on 12 and I can't fix it so far

bborman22 (changes requested):
Functionally everything worked as expected and visually it all matched up well. I only found two issues, one is a mobile visual issue and the other is a React warning. 

The mobile visual issue was with the label purchase modal width and padding. The modal didn't go full screen width and it had padding on the sides which squished everything together causing some overlap. 

Current WCS:
WCS-1 25 7

This branch:
replace-model-branch

The Reach warning I got was:

```
Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.
    in PackageInfo (created by Localized(PackageInfo))
```

I looked through the PackageInfo component and I'm pretty certain the issue is when a package is not selected for the order and it updates the props inside the render. Everything still worked as expected with the warning, but was wondering if there was a better way to handle that props update?

https://github.com/Automattic/woocommerce-services/blob/40ccf292aaa3669f52281b3c462ed6f6affba8ff/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/packages-step/package-info.js#L59

client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
```diff
@@ -271,6 +274,43 @@ const tryGetLabelRates = ( orderId, siteId, dispatch, getState ) => {
 		} );
 };
 
+export const setPackageType = ( orderId, siteId, packageId, boxTypeId ) => (
```
c-shultz, line 277: setPackageType is moved up here now so it's defined in time

c-shultz:
Good finds, @bborman22. I tweaked the styles for narrow screens. I think it's as good as the previous version at least. 

You found the right line that was triggering that warning. I moved that functionality into actions to trigger when the printing modal opens up.

bborman22 (commented):
Awesome, that warning is gone now! The modal is looking better, but I am still seeing a gap in the modal on my screen. 

modal-gap


I noticed it is coming from this line and the fact that the new modal doesn't have the `.dialog.card` classes to set width to 100% at that breakpoint. Not sure if we should add those classes or remove that selector and just apply it right to `.label-purchase-modal` which seemed to work?

https://github.com/Automattic/woocommerce-services/blob/466661036d3cf96216061f2049bb33b1ba54253d/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/style.scss#L7-L22

c-shultz:
Thanks for catching that. I'm so blind to these things sometimes on my own PR. 🙈 

The card style applies to the cards within the modal and that seems to be working OK. I put a 100% width style in for the modal on narrow screens. Much better now 😄

bborman22 (approved):
Looks good! I think this is good to go now!

PR #2337, opened 2021-02-22
Replace Calypso Card and CompactCard components with WordPress Card.

brezocordero:
## Description

Replace Calypso `Card` and `CompactCard` components with WordPress `Card` component.

`CompactCard` is a Calypso `Card` with an added class `.is-compact`. I have copied the stylesheet from Calypso into the plugin so that we can use the regular `Card` component with that added class. We have copied other styles from Calypso before, I have place the styles for the Card in the same folder.

I found a component stylesheet, `productSearch`, and readme that we were not using. I removed them together with the import.

One of the tests was failing because of the style changes. I have updated it and verified the functionality was correct.

### Related issue(s)

Related to #2333 

Verify the following screens look (and work) like before:

1) WooCommerce Shipping Settings. The different sections are cards.
2) WooCommerce Shipping Status. The different sections are cards.
3) Editing a WooCommerce Shipping method from a Shipping zone. The 'Save Settings' button is in a card.
4) Connect UPS account. The form sections are cards.
5) Connect DHL account. . The form sections are cards.

### Checklist

- [x] unit tests


c-shultz (approved):
The change is totally seamless as far as I could find. I think keeping the existing styles was a good call on this one. Code and test updates looks good! 🚢

PR #2339, opened 2021-02-24
refactor: status page refresh with ajax call

frosso:
## Description

Updating the status' view "refresh" button to make an AJAX call, rather than refreshing the whole page.
I added some tests for the page.

- Go to WooCommerce > Status > WooCommerce Shipping & Tax
- Clicking the "refresh" button should not reload the whole page
- You can even test the error status by blocking the call on the network panel

On success:

On error:

### Checklist

- [x] unit tests


client/apps/plugin-status/health.js
```diff
@@ -8,53 +8,27 @@ import { localize } from 'i18n-calypso';
 /**
  * Internal dependencies
  */
-import FormSettingExplanation from 'components/forms/form-setting-explanation';
 import Indicator from './indicator';
+import WooCommerceServicesIndicator from './woocommerce-services-indicator';
 import SettingsGroupCard from 'woocommerce/woocommerce-services/components/settings-group-card';
 
-const HealthView = ( { translate, moment, healthItems } ) => {
```
frosso, line 15: Tried to simplify this view by encapsulating most of the logic into a new `WooCommerceServicesIndicator` React component

client/apps/plugin-status/indicator.js
```diff
@@ -29,21 +29,11 @@ const Indicator = ( { title, subtitle, state, message, children } ) => {
 			icon = 'notice';
 	}
 
-	const renderSubTitle = () => {
```
frosso, line 32: No need to create a new function at each re-render just so that the function is called immediately

client/apps/plugin-status/state/actions.js
```diff
@@ -10,21 +10,9 @@ import { translate } from 'i18n-calypso';
 import * as api from 'api';
 // from Calypso
 import { successNotice, errorNotice } from 'state/notices/actions';
+import { PLUGIN_STATUS_DEBUG_TOGGLE, PLUGIN_STATUS_LOGGING_TOGGLE, SERVICE_DATA_REFRESH } from './action-types';
 
-export const PLUGIN_STATUS_DEBUG_TOGGLE = 'PLUGIN_STATUS_DEBUG_TOGGLE';
```
frosso, line 14: Having the action types inside the `actions` file is bad, because it forces the reducer to also import the actions (whereas the actions and the reducer should not be aware of each other)

Moved these to a new `action-types.js` file

client/apps/plugin-status/view.js
```diff
@@ -87,16 +81,17 @@ const StatusView = ( { actions, loggingEnabled, debugEnabled, translate } ) => {
...
+	isDebuggingEnabled: Boolean( state.status.debug_enabled ),
+} );
+
+const mapDispatchToProps = {
+	onLoggingToggle: toggleLogging,
+	onDebuggingToggle: toggleDebugging,
+};
+
 export default connect(
-	( state ) => ( {
```
frosso, line 91: Extracted to `const` functions, for consistency

client/apps/plugin-status/view.js
```diff
@@ -87,16 +81,17 @@ const StatusView = ( { actions, loggingEnabled, debugEnabled, translate } ) => {
...
-		loggingEnabled: Boolean( state.status.logging_enabled ),
-		debugEnabled: Boolean( state.status.debug_enabled ),
-	} ),
-	( dispatch ) => ( {
-		actions: bindActionCreators( {
```
frosso, line 96: This was creating a new `actions` object/reference at each re-render

client/apps/plugin-status/woocommerce-services-indicator.js
```diff
@@ -0,0 +1,70 @@
...
+
+const WooCommerceServicesIndicator = ({ translate, moment, status, onRefreshClick }) => {
+	const handleRefreshClick = useCallback((event) => {
+		event.preventDefault();
+		onRefreshClick();
+	}, [onRefreshClick]);
+
+	const currentTimestamp = Date.now() / 1000
+	let indicatorState, indicatorMessage
+	if (!status.has_service_schemas) {
```
frosso, line 23: Moved this logic from the PHP files

superdav42:
Unit testing are failing. Same tests fail for me on my machine.

frosso:
@superdav42 thanks! Looks like I didn't mock the date correctly, I'm investigating now

frosso:
@superdav42 I made the adjustments on the timezone, thanks for pointing it out! It's all ✅  now 👍

woocommerce-services.php
```diff
@@ -841,6 +848,12 @@ public function rest_api_init() {
 			$this->set_rest_self_help_controller( $rest_self_help_controller );
 			$rest_self_help_controller->register_routes();
 
+			require_once( plugin_basename( 'classes/class-wc-rest-connect-service-data-refresh-controller.php' ) );
```
superdav42, line 851:
```suggestion
			require_once __DIR__ . '/classes/class-wc-rest-connect-service-data-refresh-controller.php';
```
Improved CS pattern introduced with https://github.com/Automattic/woocommerce-services/pull/2322

classes/class-wc-rest-connect-service-data-refresh-controller.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( class_exists( 'WC_REST_Connect_Service_Data_Refresh_Controller' ) ) {
+	return;
+}
```
superdav42, line 9:
```suggestion
```
I know these checks are everywhere in other code but they are really unnecessary and new code should not include them. This file is included with require_once() so the class will never already exist. ABSPATH check is mostly used to prevent anyone from loading the file directly outside of WP context but if that happened nothing would happen because this file only declares a class.

woocommerce-services.php
```diff
@@ -379,6 +382,10 @@ public function set_rest_self_help_controller( WC_REST_Connect_Self_Help_Control
 			$this->rest_self_help_controller = $rest_self_help_controller;
 		}
 
+		public function set_rest_service_data_refresh_controller( WC_REST_Connect_Service_Data_Refresh_Controller $rest_service_data_refresh_controller ) {
```
superdav42, line 385: Not sure the point of storing a reference to WC_REST_Connect_Service_Data_Refresh_Controller is here. I know we do it with other classes but usually that is done for purposes of adding mocks for unit tests. Since we're not doing that lets just remove the rest_service_data_refresh_controller property and set_rest_service_data_refresh_controller method and stop calling it below.

superdav42 (changes requested):
Just a few small suggestions. The new component and js unit tests are :100:

classes/class-wc-rest-connect-service-data-refresh-controller.php
frosso: I agree. Although I'd like to clarify that the `class_exists( '...' )` check is made also to allow a rewrite of the class implementation from third party 👍 
I only added the checks for consistency with the existing code.

woocommerce-services.php
frosso: Yeah, all those `set_*` seemed useless to me too - again, I added it because I didn't want to create pushback, but it seemed to have created the opposite effect 😅  removed this instance in d3391a17

classes/class-wc-rest-connect-service-data-refresh-controller.php
superdav42: >  Although I'd like to clarify that the class_exists( '...' ) check is made also to allow a rewrite of the class implementation from third party

That sounds like something we definitely don't want to encourage or support. If there was a case were we did we'd want to make use of interfaces and inheritance instead of overwriting entire classes. 

> Should we remove the if(!defined(...)) from all the other controllers?

Eventually we probably don't want to do it with this PR but I'd recommended it as part of a whole project update PHP code to new features like class autoloading.

superdav42 (approved):
Looking great now!

woocommerce-services.php
```diff
@@ -845,6 +843,11 @@ public function rest_api_init() {
 			$this->set_rest_self_help_controller( $rest_self_help_controller );
 			$rest_self_help_controller->register_routes();
 
+			require_once __DIR__ . 'classes/class-wc-rest-connect-service-data-refresh-controller.php' ) );
```
c-shultz, line 846: @frosso, looks like a syntax error slipped in due to a merge. I think the trailing `) )` need to be removed.

frosso: Thanks! Approved https://github.com/Automattic/woocommerce-services/pull/2350

PR #2344, opened 2021-02-26
Replace Calypso Fieldset with internal Fieldset

bborman22:
## Description
Replace the Calypso `FormFieldset` component with an internal `FormFieldset` component. There was not a comparable WordPress component to replace with and the component itself is very simple so it was easy to move over to our internal components. This component is simply a wrapper `fieldset` HTML tag for form fields and is primarily used by other components (TextField, NumberField, WeightField, etc). 

I placed the replacement component within the top level `client/components` directory based on some recent discussions on how best to organize our internal components. I kept the naming the same to simplify the code changes required. Due to having the same name, the Webpack config `resolve.modules` array order is hitting this new internal component rather than the Calypso component.

### Related issue(s)

Related to #2333 

This component is used on most of the plugin's pages, but below are the main pages and areas to confirm:

1. The Status page, each item within the settings group cards is wrapped in a FormFieldset.
2. The Settings page, each item in the Shipping Labels card is wrapped in a FormFieldset.
3. Within Settings, the connect carrier forms and add package forms use FormFieldset. 
4. The Label Purchase modal, each form field in the modal is wrapped in a FormFieldset.

Since the component is identical to Calypso's, I confirmed it was using the new internal component by adding a class name to the component and confirm that class name was added in all the relevant places.

Some examples of the FormFieldset:

LabelPurchase-Fieldset
Settings-Fieldset
Status-Fieldset
 

### Checklist

- [ ] unit tests

### DIFF
--- a/client/components/forms/form-fieldset/index.js
+++ b/client/components/forms/form-fieldset/index.js
@@ -1,0 +1,14 @@
+/**
+ * External dependencies
+ */
+
+import React from 'react';
+import classnames from 'classnames';
+
+const FormFieldset = ( { className, children, ...otherProps } ) => (
+	<fieldset { ...otherProps } className={ classnames( className, 'form-fieldset' ) }>
+		{ children }
+	</fieldset>
+);
+
+export default FormFieldset;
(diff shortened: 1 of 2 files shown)


c-shultz (approved):
All the forms are working well and looking good. This looks like a good, simple way to approach this!

PR #2351, opened 2021-03-03
Replace FormTextInput in TextField component

bborman22:
## Description

As part of the goal to remove Calypso dependencies, this PR replaces the Calypso `FormTextInput` component in our `TextField` component with the WordPress `TextControl` component. There are still other uses of the `FormTextInput` component in other places that will be replaced in a separate PR to keep each PR more manageable and keep this one focused on the `TextField`. 

There were several differences between the components that required reworking a few areas of functionality. The two biggest differences were:
- The `TextControl` component comes with a built in label. This allowed us to remove the `FieldLabel` component from the `TextField`. It also required some additional styles to make sure the label display is carried over successfully.
- The `TextControl` component [does NOT provide the event](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/text-control/index.js), only the new value. This required reworking some event handlers to only expect a value. Some event handlers required an `id` for the change, so this was added as a parameter in addition to the value. 

One other quirk with the `TextControl` component is that if you pass in a HTML id, it will set it on the input, but will not properly set it on the for attribute of the label. Because of this, I opted to not use the HTML id and instead focus on the name attribute. This also allows the WordPress components to self manage the IDs as needed. Throughout my testing, things all seemed to work correctly still and all unit tests passed once the lookup was changed from id to name, but was something I wanted to call out for additional testing.

### Related issue(s)

Related to #2333

The TextField component is used in the following areas:
- Dynamic carrier form
- UPS settings form
- Label purchase modal
- Customs fields

### Checklist

- [x] unit tests

### DIFF
--- a/assets/stylesheets/_colors.scss
+++ b/assets/stylesheets/_colors.scss
@@ -1,4 +1,6 @@
 :root {
 	--color-link: #984A9C;
+	--color-error: #eb0001;
+	--color-error-rgb: 235, 0, 1;
 }
 
--- a/client/extensions/woocommerce/woocommerce-services/components/text-field/index.js
+++ b/client/extensions/woocommerce/woocommerce-services/components/text-field/index.js
@@ -5,13 +5,13 @@
  */
 import React, { useCallback } from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { TextControl } from '@wordpress/components';
 
 /**
  * Internal dependencies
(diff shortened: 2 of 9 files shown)


frosso (approved):
Tested with Jetpack 9.4.

Made a comparison with the current release at different viewports, functionality is the same 👍 
Also verified the input validation on account connection/label purchase.

PR #2354, opened 2021-03-05
Replace FormTextInputWithAffixes with WooCommerce Component

bborman22:
## Description

As part of the goal to remove Calypso dependencies, this PR replaces the Calypso `FormTextInputWithAffixes` component with an internal [copy of the WooCommerce `TextControlWithAffixes` component](https://github.com/woocommerce/woocommerce-admin/blob/main/packages/components/src/text-control-with-affixes/index.js). In addition to that, there were a few remaining `FormTextInput` components that needed to be replaced on the same pages. 

I had originally installed the `@woocommerce/components` package and used it directly, but when running a test build, it added significantly to the file size. Since this is currently just using one component, it seemed more in line with the goal of the Calypso cleanup project to just move that component over internally until we have more of a use for other components from the package. With that future possibility in mind, I replicated the component with very minor changes so that any possible future switch to the package would be seamless.

Since the `TextControlWithAffixes` component is just an implementation of the `TextControl` component, it has some of the same quirks mentioned in #2351, with the primary difference being that the `onChange` function only returns a value, not an event. This has been worked around in a few areas by providing a name/key with the value in instances where it was usually retrieved from the event.

### Related issue(s)

Related to #2333

This PR impacts the following areas:

- Package Create/Edit text inputs
- Shipping method service adjustments
- Shipping method fallback rate
- In the label purchase modal, the Customs item row weight field
- In the label purchase modal, the package weight field

Service-Adjustment
Create-Package
Customs-Weight

### Checklist

- [ ] unit tests


harriswong:
👀

harriswong (approved):
Tested the inputs in the description's screenshots. LGTM. 👍

PR #2358, opened 2021-03-08
Replace Calypso FieldLabel with internal FieldLabel

bborman22:
## Description

Replace the Calypso `FieldLabel` component with an internal `FieldLabel` component. There was not a comparable WordPress component to replace with and the component itself is very simple so it was easy to move over to our internal components. In WordPress components, the field label is often included with the field component, so there may be a few more of these labels removed going forward.

I placed the replacement component within the top level `client/components` directory based on some recent discussions on how best to organize our internal components. I kept the naming the same to simplify the code changes required. Due to having the same name, the Webpack config resolve.modules array order is hitting this new internal component rather than the Calypso component.

### Related issue(s)

Related to #2333 

This component is used in multiple areas, primarily with custom form fields (NumberInput, FormTextInputWithAffixes, FormToggle, FormCurrencyInput, etc). Some pages to review in particular:

- Create/Edit package
- Status logs
- Shipping method settings page
- Label purchase modal

Since the component is identical to Calypso's, I confirmed it was using the new internal component by adding a class name to the component and confirm that class name was added in all the relevant places.

### Checklist

- [ ] unit tests

### DIFF
--- a/client/components/forms/form-label/index.js
+++ b/client/components/forms/form-label/index.js
@@ -1,0 +1,48 @@
+/** @format */
+
+/**
+ * External dependencies
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import { localize } from 'i18n-calypso';
+import { isObject, omit } from 'lodash';
+
+const renderRequiredBadge = translate => (
+	<small className="form-label__required">{ translate( 'Required' ) }</small>
+);
+
+const renderOptionalBadge = translate => (
+	<small className="form-label__optional">{ translate( 'Optional' ) }</small>
+);
+
+const addKeys = elements =>
+	elements.map( ( elem, idx ) => ( isObject( elem ) ? { ...elem, key: idx } : elem ) );
+
+const FormLabel = ( { children, required, optional, translate, className, ...extraProps } ) => {
+	children = React.Children.toArray( children ) || [];
+	if ( required ) {
+		children.push( renderRequiredBadge( translate ) );
+	}
+
+	if ( optional ) {
+		children.push( renderOptionalBadge( translate ) );
+	}
+
+	return (
+		<label
+			{ ...omit( extraProps, 'moment', 'numberFormat' ) }
+			className={ classnames( className, 'form-label' ) }
+		>
+			{ children.length ? addKeys( children ) : null }
+		</label>
+	);
+};
+
+FormLabel.propTypes = {
+	required: PropTypes.bool,
+};
+
+export default localize( FormLabel );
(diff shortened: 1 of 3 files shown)


client/components/forms/form-label/index.js
```diff
@@ -0,0 +1,48 @@
...
+	if ( optional ) {
+		children.push( renderOptionalBadge( translate ) );
+	}
+
+	return (
+		<label
+			{ ...omit( extraProps, 'moment', 'numberFormat' ) }
+			className={ classnames( className, 'form-label' ) }
+		>
+			{ children.length ? addKeys( children ) : null }
```
frosso, line 39: I'd recommend doing this, which prevents from doing `addKeys` after cloning the elements, overwriting the `children` through `React.Children.toArray`, as well as the `lodash` dependency :

```
diff --git a/client/components/forms/form-label/index.js b/client/components/forms/form-label/index.js
index 81b6f425..58ff79e3 100644
--- a/client/components/forms/form-label/index.js
+++ b/client/components/forms/form-label/index.js
@@ -8,41 +8,24 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { localize } from 'i18n-calypso';
-import { isObject, omit } from 'lodash';
-
-const renderRequiredBadge = translate => (
-	<small className="form-label__required">{ translate( 'Required' ) }</small>
-);
-
-const renderOptionalBadge = translate => (
-	<small className="form-label__optional">{ translate( 'Optional' ) }</small>
-);
-
-const addKeys = elements =>
-	elements.map( ( elem, idx ) => ( isObject( elem ) ? { ...elem, key: idx } : elem ) );
-
-const FormLabel = ( { children, required, optional, translate, className, ...extraProps } ) => {
-	children = React.Children.toArray( children ) || [];
-	if ( required ) {
-		children.push( renderRequiredBadge( translate ) );
-	}
-
-	if ( optional ) {
-		children.push( renderOptionalBadge( translate ) );
-	}
 
+const FormLabel = ( { children, required, optional, translate, className, moment, numberFormat, ...extraProps } ) => {
 	return (
 		<label
-			{ ...omit( extraProps, 'moment', 'numberFormat' ) }
+			{ ...extraProps }
 			className={ classnames( className, 'form-label' ) }
 		>
-			{ children.length ? addKeys( children ) : null }
+			{children}
+			{required && ( <small className="form-label__required">{ translate( 'Required' ) }</small> ) }
+			{optional && ( <small className="form-label__optional">{ translate( 'Optional' ) }</small> )}
 		</label>
 	);
 };
 
 FormLabel.propTypes = {
 	required: PropTypes.bool,
+	optional: PropTypes.bool,
+	children: PropTypes.node,
 };
 
 export default localize( FormLabel );
```

(pardon the formatting)

frosso (approved):
Seems to be working well - a few suggestions on the component's code, which would allow to streamline its content/body

client/components/forms/form-label/index.js
bborman22: Wow that is a lot cleaner, thanks for taking a look at this!! I've pushed these changes up now. The only difference I saw was with the key on the children, but it still worked as expected.

PR #2359, opened 2021-03-08
Replaces deprecated Jetpack constant JETPACK_MASTER_USER

Ferdev:
## Description
Instead of using the deprecated `JETPACK_MASTER_USER` constant, the Jetpack class now uses proper method from the Jetpack manager to get the status of the connection
### Related issue(s)

Closes https://github.com/Automattic/woocommerce-shipping-issues/issues/185

1. Create a new test site with latest WooCommerce.
2. Install latest Jetpack version v9.6.
3. Install this plugin and connect the site to Jetpack. The connection should be made without incidents.

### Checklist

- [x] unit tests

### DIFF
--- a/classes/class-wc-connect-jetpack.php
+++ b/classes/class-wc-connect-jetpack.php
@@ -155,39 +155,43 @@ if ( ! class_exists( 'WC_Connect_Jetpack' ) ) {
 
 		/**
 		 * Determines if the current user is connected to Jetpack
+		 *
 		 * @return bool Whether or nor the current user is connected to Jetpack
 		 */
 		public static function is_current_user_connected() {
+			if ( class_exists( '\Automattic\Jetpack\Connection\Manager' ) && method_exists( '\Automattic\Jetpack\Connection\Manager', 'is_user_connected' ) ) {
+				$connection = new Manager();
+
+				return $connection->is_user_connected();
+			}
+
 			if ( defined( JETPACK_MASTER_USER ) ) {
-				$user_token = WC_Connect_Jetpack::get_master_user_access_token( JETPACK_MASTER_USER );
+				$user_token = self::get_master_user_access_token( JETPACK_MASTER_USER );
 
 				return ( isset( $user_token->external_user_id ) && get_current_user_id() === $user_token->external_user_id );
-			} else {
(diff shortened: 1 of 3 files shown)


classes/class-wc-connect-jetpack.php
```diff
@@ -152,5 +152,43 @@ public static function tracks_record_event( $user, $event_type, $data ) {
...
+		/**
+		 * Determines if the current user is connected to Jetpack
+		 * @return bool Whether or nor the current user is connected to Jetpack
+		 */
+		public static function is_current_user_connected() {
+			if ( defined( JETPACK_MASTER_USER ) ) {
+				$user_token = WC_Connect_Jetpack::get_master_user_access_token( JETPACK_MASTER_USER );
+
+				return ( isset( $user_token->external_user_id ) && get_current_user_id() === $user_token->external_user_id );
+			} else {
```
frosso, line 165: No need for `else` here, since in the previous `if` there's already an early return.

I think that for legibility we can just do:

```
if ( defined( JETPACK_MASTER_USER ) ) {
  ...
  return ...;
}

if ( class_exists( '\Automattic\Jetpack\Connection\Manager' ) && ... ) {
  ...
  return ...;
}

return false;
```

Also - should we put the newer code (`class_exists( '\Automattic\Jetpack\Connection\Manager' )`) above the check for the constant, so that the newer code is returned earlier?

Thoughts?

classes/class-wc-connect-jetpack.php
```diff
@@ -152,5 +152,43 @@ public static function tracks_record_event( $user, $event_type, $data ) {
...
+		/**
+		 * Determines if Jetpack is connected
+		 * @return bool Whether or nor Jetpack is connected
+		 */
+		public static function is_connected() {
+			if ( defined( JETPACK_MASTER_USER ) ) {
+				$user_token = WC_Connect_Jetpack::get_master_user_access_token( JETPACK_MASTER_USER );
+
+				return isset( $user_token->external_user_id );
+			} else {
```
frosso, line 184: Similar to above

frosso (approved):
Just a minor comment re code style

classes/class-wc-connect-jetpack.php
Ferdev: Sounds good @frosso! Those are some nice suggestions. I just pushed a new commit.

frosso (approved):
Thanks for considering the changes!

PR #2366, opened 2021-03-09
Only display shipping validation errors on cart or checkout page

### LINKED ISSUE #2362: fix: error messaging appears after add to cart
## Description

It looks like the error messaging introduced in earlier PRs appears immediately after adding a product to the cart.
Markup 2021-03-09 at 08 05 12

## Steps to reproduce
- Go to your site
- You can have shipping debug on/off, doesn't matter
- Make sure your cache is clear - as if you're a new visitor
- Add a product to the cart
- Messaging appears on product pages

Also reported here: https://wordpress.org/support/topic/recent-update-causes-error-messages/

bborman22:
## Description

After a recent update, users had reported getting an error when adding products to their cart. It appears these errors were coming from the shipping validation process and being added as notices to the page. This PR looks to address that issue by only adding the errors as notices when on the Cart or Checkout page. This seemed to be the least disruptive approach for addressing this issue.

AddToCart-ValidationError

### Related issue(s)

Fixes #2362 

1. With WCS&T set up on this branch and shipping zones/methods set up.
2. Use an Incognito window or clear cache to simulate a customer that does not have any shipping information entered/cached yet.
2. Visit a product's single page (I was unable to reproduce from a listing page).
3. Add the product to the Cart.
4. No error should be displayed.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -10,6 +10,7 @@
 * Add   - Run phpcbf as a pre-commit rule.
 * Fix   - Fix PHPUnit tests. Rename `test_` to `test-` to match our phpcs rules. Remove travis and move to github action.
 * Tweak - Updated .nvmrc to use 10.16.0
+* Fix   - Only display shipping validation errors on the cart or checkout pages.
 
 = 1.25.8 - 2021-03-02 =
 * Tweak - Add support for new Jetpack 9.5 data connection.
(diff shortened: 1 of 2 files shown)


superdav42 (approved):
Tested and working.
I'm going to review what core is doing here. I think it shouldn't try to calculate shipping unless it will show shipping information not only does it have potential for the problems here but it is often an expensive and long operation that should be avoided in the ajax event to add an item to the cart.

PR #2372, opened 2021-03-11
Detect checkout block REST API call correctly.

budzanowski:
The `wc/store/checkout` was blocked inside the shipping calculation. This PR unblocks it and allows the functions to fetch rates correctly. When the endpoint was blocked the rates were not registered and the cart shipping totals calculation was selecting the default shipping method.

### Related issue(s)
Found during 1.25.9 testing call.

From the testing post:
```
When using the following setup:

Cart page housing either the [woocommerce_cart] shortcode or Cart block,
Checkout page using the Checkout block.
Site has WCS&T enabled.
When picking one shipping option with a live rate on the Cart page, and then picking another shipping option with a live rate in the Checkout block, the order will have no shipping costs associated with it.

In addition to that, when I set up the following shipping methods:

Local pickup for $20.
Flat rate for $10.
Regardless of which of the two methods I selected on the Cart page, when I selected the other one on the Checkout page, the order had Local pickup associated with it anyway.
```

### DIFF
--- a/classes/class-wc-connect-functions.php
+++ b/classes/class-wc-connect-functions.php
@@ -43,7 +43,7 @@ if ( ! class_exists( 'WC_Connect_Functions' ) ) {
 		}
 
 		/**
-		 * Check if we are currently in Rest API request for the wc/store/cart API call.
+		 * Check if we are currently in Rest API request for the wc/store/cart or wc/store/checkout API call.
 		 *
 		 * @return bool
 		 */
@@ -51,8 +51,8 @@ if ( ! class_exists( 'WC_Connect_Functions' ) ) {
 			if ( ! WC()->is_rest_api_request() && empty( $GLOBALS['wp']->query_vars['rest_route'] ) ) {
 				return false;
 			}
-
-			return false !== strpos( $GLOBALS['wp']->query_vars['rest_route'], 'wc/store/cart' );
+			$rest_route = $GLOBALS['wp']->query_vars['rest_route'];
+			return ( false !== strpos( $rest_route, 'wc/store/cart' ) ) || ( false !== strpos( $rest_route, 'wc/store/checkout' ) );
 		}
 
 		/**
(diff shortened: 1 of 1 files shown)


classes/class-wc-connect-functions.php
```diff
@@ -43,16 +43,16 @@ public static function get_wc_helper_auth_info() {
...
 		 * @return bool
 		 */
 		public static function is_store_api_call() {
 			if ( ! WC()->is_rest_api_request() && empty( $GLOBALS['wp']->query_vars['rest_route'] ) ) {
 				return false;
 			}
-
-			return false !== strpos( $GLOBALS['wp']->query_vars['rest_route'], 'wc/store/cart' );
+			$rest_route = $GLOBALS['wp']->query_vars['rest_route'];
+			return ( false !== strpos( $rest_route, 'wc/store/cart' ) ) || ( false !== strpos( $rest_route, 'wc/store/checkout' ) );
```
waclawjacek, line 55: How about:

```suggestion
			return 1 === preg_match( '#wc/store/(cart|checkout)#', $rest_route );
```

or:

```suggestion
			return (
				false !== strpos( $rest_route, 'wc/store/cart' ) ||
				false !== strpos( $rest_route, 'wc/store/checkout' )
			);
```

In the latter, aside from the change in indentation, the brackets surrounding individual conditions were also removed.

waclawjacek (approved):
LGTM! :shipit: 

I have posted a suggestion regarding the condition if you'd like to adjust it.

classes/class-wc-connect-functions.php
budzanowski: Cool, merging the later.

Ferdev:
Since this branch was forked later than `release/1.25.9`, contains some commits not related to this fix that the release branch doesn't have. So I'm gonna go ahead and squash this into `develop` and then cherry-pick it in the release branch. cc/ @waclawjacek

PR #2374, opened 2021-03-12
Add DHL Live rates inbox notification

brezocordero:
## Description

This PR adds an inbox notification about DHL live rates for customers that can have DHL live rates.
( = when they have get_service_schema_by_id( 'dhlexpress' ); )

The note is only displayed once, goes away when they dismiss it or action it (Click on 'Go to Shipping Zones')

The note is deleted from the DB when the plugin is uninstalled.

### Related issue(s)
Closes: https://github.com/Automattic/woocommerce-shipping-issues/issues/189

1. With WooCommerce and WooCommerce Shipping and Tax active
2. Test on a store without DHL Express live rates available. Check there is no notice.
3. Activate the flag for the store. Refresh the status on the plugin Status page.
4. Check there is a notice like this:
DHLLiveRatesNotice
6. Dismiss the notice. Activate-deactivate the plugin, check the notice is not displayed again.

Note: If you want to display the note again, uninstall the plugin or remove it from the DB tables: `wp_wc_admin_notes `and `wp_wc_admin_note_actions`

### Checklist

- [ ] unit tests ( I haven't written new PHP tests but I made sure previous tests are still running)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** WooCommerce Shipping & Tax Changelog ***
 = 1.25.10 - 2021-xx-xx =
 * Tweak - Removes deprecated Jetpack constant JETPACK_MASTER_USER
+* Add   - WC Admin notice about DHL live rates.
 
 = 1.25.9 - 2021-xx-xx =
 * Tweak - Cleanup stripe functionality.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,9 +1,9 @@
 *** WooCommerce Shipping & Tax Changelog ***
 = 1.25.10 - 2021-xx-xx =
 * Tweak - Removes deprecated Jetpack constant JETPACK_MASTER_USER
-* Add   - WC Admin notice about DHL live rates.
 
 = 1.25.9 - 2021-xx-xx =
+* Add   - WC Admin notice about DHL live rates.
 * Tweak - Cleanup stripe functionality.
 * Tweak - Display better errors on checkout page when address fields are missing / invalid.
 * Tweak - Refresh on status page does not reload page.
--- a/classes/class-wc-connect-note-dhl-live-rates-available.php
+++ b/classes/class-wc-connect-note-dhl-live-rates-available.php
@@ -7,54 +7,51 @@
 
 defined( 'ABSPATH' ) || exit;
 
-if ( ! class_exists( 'WC_Connect_Note_Dhl_Live_Rates_Available' ) ) {
+/**
+ * Note to inform WooCommerce Shipping users with legacy live rates about DHL live rates.
+ */
+class WC_Connect_Note_DHL_Live_Rates_Available {
+
+	use Automattic\WooCommerce\Admin\Notes\NoteTraits;
 
 	/**
-	 * Note to inform WooCommerce Shipping users with legacy live rates about DHL live rates.
+	 * Name of the note for use in the database.
 	 */
-	class WC_Connect_Note_DHL_Live_Rates_Available {
+	const NOTE_NAME = 'wc-services-dhl-live-rates-available';
 
-		use Automattic\WooCommerce\Admin\Notes\NoteTraits;
+	/**
+	 * Maybe add note to inform WooCommerce Shipping users with legacy live rates about new DHL live rates.
+	 *
+	 * @param WC_Connect_Service_Schemas_Store $schemas   Store schemas.
+	 */
+	public static function init( WC_Connect_Service_Schemas_Store $schemas ) {
+		// If store has DHL Express live rates.
+		$has_dlh_express = $schemas->get_service_schema_by_id( 'dhlexpress' );
 
-		/**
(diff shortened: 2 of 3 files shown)


woocommerce-services.php
```diff
@@ -208,6 +207,17 @@ public static function plugin_deactivation() {
 
 		public static function plugin_uninstall() {
 			WC_Connect_Options::delete_all_options();
+			self::delete_notices();
```
brezocordero, line 210: I thought about cleaning the note also on plugin_deactivation but that means on upgrades, the note would be displayed again because [the plugin is deactivated](https://core.trac.wordpress.org/browser/tags/5.7/src/wp-admin/includes/class-plugin-upgrader.php#L512).

classes/class-wc-connect-note-dhl-live-rates-available.php
```diff
@@ -0,0 +1,60 @@
+<?php
+/**
+ * WooCommerce Shipping note: DHL live rates available.
+ *
+ * Only for legacy customers that had the feature available.
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+if ( ! class_exists( 'WC_Connect_Note_Dhl_Live_Rates_Available' ) ) {
```
superdav42, line 10:
```suggestion
```
I know these checks are added everywhere in the code but they are not really needed provided you include the file with `requrie_once`

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** WooCommerce Shipping & Tax Changelog ***
 = 1.25.10 - 2021-xx-xx =
 * Tweak - Removes deprecated Jetpack constant JETPACK_MASTER_USER
+* Add   - WC Admin notice about DHL live rates.
```
frosso, line 4: I think this needs to go in `1.25.9`

classes/class-wc-connect-note-dhl-live-rates-available.php
```diff
@@ -0,0 +1,60 @@
...
+
+			if ( ! ! $has_dlh_express ) {
+				self::possibly_add_note();
+			}
+		}
+
+		/**
+		 * Get the note.
+		 *
+		 * @return Note|null
```
frosso:
```suggestion
		 * @return Automattic\WooCommerce\Admin\Notes\Note
```

classes/class-wc-connect-note-dhl-live-rates-available.php
```diff
@@ -0,0 +1,60 @@
...
+defined( 'ABSPATH' ) || exit;
+
+if ( ! class_exists( 'WC_Connect_Note_Dhl_Live_Rates_Available' ) ) {
+
+	/**
+	 * Note to inform WooCommerce Shipping users with legacy live rates about DHL live rates.
+	 */
+	class WC_Connect_Note_DHL_Live_Rates_Available {
+
+		use Automattic\WooCommerce\Admin\Notes\NoteTraits;
```
frosso: It looks like this got introduced (with the `get_note` method) in this PR https://github.com/woocommerce/woocommerce-admin/pull/4340 in May 2020
So anybody testing with WC~< 4.2 would get a fatal error. We require at least WC 3.0 - should we add some checks for this? Or should we update our headers?

frosso (commented):
Tested locally - Working well!
TIL about the `Automattic\WooCommerce\Admin\Notes\NoteTraits` trait - it makes showing notices a breeze!

Question about compatibility, which I think might be important

classes/class-wc-connect-note-dhl-live-rates-available.php
brezocordero: That's a good point, bumping it only for a notice it is not really worth it.
I will add a check.

classes/class-wc-connect-note-dhl-live-rates-available.php
brezocordero: Removed.

changelog.txt
brezocordero: Changed.

classes/class-wc-connect-note-dhl-live-rates-available.php
brezocordero: Done.

classes/class-wc-connect-note-dhl-live-rates-available.php
brezocordero: Added a check for the trait.

frosso (approved):
The changelog in the GitHub PR view seems cached or something - but I checked the changes locally and it looks good 👍
(Minor) it looks like the returned value in `can_add_wc_admin_notice()` can be inlined with the `return`
(Minor) the check for the trait to exist should be sufficient - in which case would the trait exist if `WC()->is_wc_admin_active()` is false?

brezocordero:
> (Minor)

Details are important ❤. Done.

frosso (approved):
Thanks again for the changes!

Issue #2364, opened 2021-03-09
Automate GitHub Actions to use the latest WooCommerce versions to test


harriswong:
### Description
Our [support policy](https://github.com/Automattic/woocommerce-services/issues/2131) for WooCommerce Services & Tax is latest WordPress/WooCommerce minus 2 versions. 

Currently, we [hardcoded](https://github.com/Automattic/woocommerce-services/blob/68fd8318635db34524f8d242231ad04a836d5b7d/.github/workflows/php_tests.yml#L37-L38
) the following versions:
```
matrix:
        ...
        wc-version: [4.8.0, 4.9.2, 5.0.0]
```

### Output
This should be automated so we don't have to keep updating this field. Something like (pseudo):
```
matrix:
        ...
        wc-version: [latest - 0.2.0, latest - 0.1.0, latest]
```

### Issues
Wordpress part is tracked in another issue here: https://github.com/Automattic/woocommerce-services/issues/2383

harriswong:
Here is an idea to grab the latest WC releases based on their release tags:

#### latest_version.php
```php
<?php
function version_compare_reverse($i, $j) {
	 return version_compare($i, $j) * -1;
}

$versions = [];
$fp = fopen("php://stdin", 'r');
$contents = fgets($fp);
while ($contents !== false) {
	$contents = fgets($fp);
	if (!preg_match('/^\d+\.\d+\.\d+$/', $contents)) {
		continue;
	}
	$versions[] = $contents;
}

usort($versions, 'version_compare_reverse');
print_r($versions);

//TODO: take $argv to dump out the latest - 2 versions
```

#### example usage
```
cd woocommerce //assume this is where the git clone is
git tag | php version_latest //prints latest 
git tag | php version_latest 1 //prints latest -1 
git tag | php version_latest 2 //prints latest -2
```

PR #2379, opened 2021-03-15
Use a script to find latest 3 minor versions for WooCommerce for test

harriswong:
## Description
Automatically find the last 3 minor woocommerce versions to run our Github action matrix tests. This PR is inspired by https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-returning-a-json-object. 

### Changes
- Create another job to find the last 3 WC minor version numbers
- Added a php script to output the version numbers out to stdout, ie. `[4.9.2, 5.0.1, 5.1.0]`. 

### Related issue(s)
Closes: https://github.com/Automattic/woocommerce-services/issues/2364

1. Scroll to the bottom of the PR, check the actions ran and they passed
2. Click on the "Checks" tab for this PR, make sure it now has a job to find WooCommerce latest 3 minor versions, and it spawns a few other jobs to run the tests on. On the left side, you should see it testing the new WooCommerce 5.1
image

### How to test the script
We want to make sure the script is printing out the last 3 minor versions
1. Go to root of WCS folder
2. run `git ls-remote --tags --refs https://github.com/woocommerce/woocommerce.git | sed 's/.*refs\/tags\///' | php ./bin/wc-latest-version.php` on console
3. You should see `['4.9.2','5.0.0','5.1.0']` on console.

### Checklist

- [ ] unit tests

### DIFF
--- a/bin/wc-latest-version.php
+++ b/bin/wc-latest-version.php
@@ -27,11 +27,10 @@ function filter_valid_versions() {
  * @return 3 minor versions, in the order of oldest to newest.
  */
 function get_last_three_minor($versions) {
-	$count = 0;
 	$last_minor_version = '';
 	$results = [];
 	foreach ($versions as $index => $version_number) {
-		$current_minor_version = substr($version_number, 0, strrpos($version_number, '.') + 1);
+		$current_minor_version = substr($version_number, 0, strrpos($version_number, '.'));
 
 		if (count($results) >= 3) {
 			break;
--- a/bin/wc-latest-version.php
+++ b/bin/wc-latest-version.php
@@ -42,13 +42,12 @@ function get_last_three_minor($versions) {
 		}
 
 		$last_minor_version = $current_minor_version;
-		$results[] = $version_number;
+		$results[] = "'$version_number'";
 		$count++;
 	}
 
 	return array_reverse($results);
 }
-
 
 // Output 3 latest minor versions in comma separated values, wrapped in [] for github action matrix
 $versions = filter_valid_versions();
--- a/bin/wc-latest-version.php
+++ b/bin/wc-latest-version.php
@@ -53,6 +53,8 @@ function get_last_three_minor($versions) {
 // Output 3 latest minor versions in comma separated values, wrapped in [] for github action matrix
 echo "Processing...\n";
 $versions = filter_valid_versions();
+echo "Version are \n";
+print_r($versions);
 usort($versions, 'version_compare_reverse');
 $results = get_last_three_minor($versions);
 echo '::set-output name=wc-versions::[' . implode(",",$results) . ']' . "\n";
--- a/bin/wc-latest-version.php
+++ b/bin/wc-latest-version.php
@@ -1,0 +1,57 @@
+<?php
+function version_compare_reverse($i, $j) {
+	 return version_compare($i, $j) * -1;
+}
+
+/**
+ * Take input from stdin and adds all valid (ie. x.y.z) versinos to an array.
+ */
+function filter_valid_versions() {
+	$versions = [];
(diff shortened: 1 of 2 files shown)


bin/wc-latest-version.php
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/**
+ * Take an array of $versions and return the latest 3 minor versions.
+ *
+ * @param Array $versions The array need to be sorted first.
+ * @return 3 minor versions, in the order of oldest to newest.
+ */
+function get_last_three_minor($versions) {
+	$count = 0;
```
frosso, line 30: Looks like `$count` is no needed, since you're already doing `count($results)`, correct?

frosso (approved):
Looks like it's working well, just a minor comment on an unused variable

bin/wc-latest-version.php
harriswong: Good catch! Previously I was using `$count` before I refactored. Removing this.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/style.scss
```diff
@@ -66,6 +66,10 @@
 		// See: https://github.com/WordPress/gutenberg/pull/23706
 		border: 4px solid #d52c82;
 		box-sizing: border-box;
+
+		// Unset the translate() transform introduced in Gutenberg.
+		// See: https://github.com/WordPress/gutenberg/pull/27377
+		transform: none;
```
frosso, line 72: Looks like there are some unwanted changes after the force-push

harriswong: haha omg, you are so fast. yes. I need to rebase from the latest develop. D:

bin/wc-latest-version.php
```diff
@@ -27,11 +27,10 @@ function filter_valid_versions() {
  * @return 3 minor versions, in the order of oldest to newest.
  */
 function get_last_three_minor($versions) {
-	$count = 0;
 	$last_minor_version = '';
 	$results = [];
 	foreach ($versions as $index => $version_number) {
-		$current_minor_version = substr($version_number, 0, strrpos($version_number, '.') + 1);
+		$current_minor_version = substr($version_number, 0, strrpos($version_number, '.'));
```
harriswong, line 33: @frosso I removed the `$count` in this last commit. As well, I noticed I had a `+1` on the `substr`. If `$version_number` is `4.12.1`, then the `substr` would had changed it to `4.12.`. I removed the `+1` so that it is `4.12`. In terms of functionality, it doesn't make any differences, just better for my eyes. 

I also added a "How to test the script" step in the description.

frosso: Oh cool, thanks! Good either ways :D

bin/wc-latest-version.php
```diff
@@ -0,0 +1,55 @@
...
+			break;
+		}
+
+		if ($last_minor_version === $current_minor_version) {
+			continue;
+		}
+
+		$last_minor_version = $current_minor_version;
+		$results[] = "'$version_number'";
+		$count++;
```
frosso, line 45: Looks like this is missing

frosso (approved):
Still looks good! 🤩  - minor thing

bin/wc-latest-version.php
harriswong: NOOooooooo

Thank you.

harriswong: Fixed.

frosso (approved):
⚡

PR #2375, opened 2021-03-12
feat: add live rates carriers list section

frosso:
## Description

Adding a "Live rates at checkout" section on the settings page.
Some details are on the ticket.

Checked with UX - a few things that are ok for this first phase:
- the tooltip’s width cannot be modified and I also cannot add artificial line breaks - everything needs to go into one line: https://d.pr/i/I29c68
- I tried to keep the alignment of the columns consistent - but on small screens it becomes evident that we’re trying to squeeze a lot of content on a small column: https://d.pr/i/LwYhQB
  - on even smaller screens the whole app is a bit messed up

### Related issue(s)

Closes https://github.com/Automattic/woocommerce-shipping-issues/issues/188

### Steps to reproduce & screenshots/GIFs

When testing against production (when testing against local/staging you might need to remove `DhlExpressAccount` from the `enabled_carrier_accounts` setting on the connect server).

Brand new site, with no flags:
Screen Shot 2021-03-12 at 10 09 28 AM

With UPS beta:
Screen Shot 2021-03-12 at 10 10 02 AM

Only grandfathered live rates:
Screen Shot 2021-03-12 at 10 10 30 AM

UPS beta, grandfathered live rates, DHL Express live rates:
Screen Shot 2021-03-12 at 10 10 58 AM

### Checklist

- [x] unit tests

- [x] `changelog.txt` entry added

PR #2265, opened 2020-11-20
refactor(CarrierRegistration): hook conversion, fix bugs

frosso:
## Description

- Redirects to settings page on successful dynamic account registration
- Fixes the dynamic account registration fields
- Removes the `carrierAccounts` redux slice because it contained UPS-hardcoded data
- Dynamically sets the title of the account registration page
- Converts the UPS registration form to hooks, consistently with the Dynamic form
- Removes the hardcoded UPS registration data
- Fixes an issue where if the server returns DHL account data, the whole settings page crashes
- Introduces the `redux-mock-store` package, to test redux-connected components and redux actions
- Overall, while introducing additional features, this reduces the bundle size of the minified chunks by ~4KB

### Related issue(s)

Closes https://github.com/Automattic/woocommerce-services/issues/2253

### Steps to reproduce & screenshots/GIFs

- Ensure to be using the latest staging version of the server
- Go to `/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings`
- Connect/disconnect any account: https://d.pr/i/JEPLNk
- UPS registration form behavior does not change
- Feel free to try and connect a DHL Express account with the Automattic credentials on the secret store (unfortunately it doesn't look like there is any validation done server-side for account credentials)

### Checklist

- [x] unit tests

- [x] `changelog.txt` entry added

PR #2382, opened 2021-03-16
Fix PHP notice on carrier registration page

### LINKED ISSUE #2381: Unexpected PHP notice in carrier registration form
One of our recent fixes has introduced an error in our PHP code:

https://dl.dropboxusercontent.com/s/j4ltfpdk3ga3tdm/shot-2021-03-15-14-44-50.png

In order to reproduce it, just try to register a new carrier from the WooCommerce Shipping & Tax settings page.

waclawjacek:
## Description

The notice was caused by no array key existence check which is now performed by `empty()`. Also, an invalid key was used.

Replace `array_values()` with `current()` since only the first value is used.

It seems this issue was [caused by this change.](https://github.com/Automattic/woocommerce-services/pull/2375/files#diff-45ab2d6757efe6316bd243c6eb0cf9194e30ddd4c01b98de3ed0e4a9b0aec1a9)

For context on how this looked previously, [this is the PR where the `$carrier_information` variable was introduced](https://github.com/Automattic/woocommerce-services/pull/2265/files#diff-45ab2d6757efe6316bd243c6eb0cf9194e30ddd4c01b98de3ed0e4a9b0aec1a9).

### Related issue(s)

Fixes #2381.

Because of this bug, the following breadcrumbs were not displayed. They should be visible with the changes in this PR.

Markup on 2021-03-16 at 13:05:01

### Checklist

- [ ] unit tests

- [ ] `changelog.txt` entry added (I didn't add a changelog entry because this fixes an issue with the yet-unreleased feature #2375)

### DIFF
--- a/classes/class-wc-connect-settings-pages.php
+++ b/classes/class-wc-connect-settings-pages.php
@@ -106,10 +106,10 @@ if ( ! class_exists( 'WC_Connect_Settings_Pages' ) ) {
 				$extra_args['continents'] = $this->continents->get();
 
 				$carrier_information = array();
-				if ( $extra_args['carriers'] ) {
-					$carrier_information = array_values(
+				if ( ! empty( $extra_args['carrier_accounts'] ) ) {
+					$carrier_information = current(
 						array_filter(
-							$extra_args['carriers'],
+							$extra_args['carrier_accounts'],
 							function( $carrier ) {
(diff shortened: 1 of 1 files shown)


Ferdev (approved):
LGTM!!

frosso:
😬 thank you for fixing it!

PR #2384, opened 2021-03-16
chore: update carrier discount amounts

frosso:
## Description

As discussed in team meeting (and later on Slack `CKNC7KACA-p1615912171010900`) changing the copy on the discount amounts.

https://d.pr/i/R5Lumv

### Checklist

- [ ] unit tests

I didn't think a changelog entry is needed, since the whole paragraph hasn't been released yet and it's just a copy change


client/extensions/woocommerce/woocommerce-services/views/live-rates-carriers-list/carriers-list.js
```diff
@@ -32,14 +32,9 @@ const Actions = localize( ( { translate } ) => {
 const CarrierDiscount = localize( ( {
```
frosso, line 32: I wondered whether having a whole component/function for this is worth it at this point - but whatevs 🤷

Ferdev (approved):
:+1:

PR #2387, opened 2021-03-17
Check for older versions of WC Admin.

brezocordero:
## Description

Add a `class_exists` check for older versions of WC Admin.

To test, use a WC Version without WC Admin and install WC Admin 2.1.

### Checklist

- [ ] unit tests


c-shultz (approved):
This looks good! It works well for me with wc-admin 2.1 (no errors), and when I deactivated wc-admin and updated WooCommerce to 5.1, the notice showed up correctly.

PR #2391, opened 2021-03-18
Fix Shipping validation notice shown when no address entered.

### LINKED ISSUE #2362: fix: error messaging appears after add to cart

superdav42:
### Related issue(s)

Closes #2362 
Option 2 turned out to be much harder and not provide a good user experience so I'm going with option 1. Several reasons mostly it's tricky to actually detect if other shipping methods are loading and doing so would require loading all the shipping methods when they aren't actually used which would be unneeded load. Forcing the shipping address required option actually seemed to have a worse user experience than not using the option. With the option enabled and entering an address without zip results in no errors shown at all even with the code as is and no notice that the zip is required.

My conclusion is any further shipping address validation should be done in core.

Use a new session like incognito and add product to cart and visit cart page. No notices should be shown unless debug mode is enabled. 

### Checklist

### DIFF
--- a/.github/workflows/php_tests.yml
+++ b/.github/workflows/php_tests.yml
@@ -18,8 +18,20 @@ jobs:
           sudo update-alternatives --set phar.phar /usr/bin/phar.phar7.4
       - name: Lint
         run: find . -type f -name "*.php" -exec php -l {} \;
+  find_wc_latest_versions:
+    name: Find supported woocommerce versions
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.get_wc_versions_for_matrix.outputs.wc-versions }}
+    steps:
+      - name: Checkout latest WCS
+        uses: actions/checkout@v2
+      - name: Get the 3 latest WC minor versions
+        id: get_wc_versions_for_matrix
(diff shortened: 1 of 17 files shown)


Ferdev:
For a hotfix, last merge commit should be removed, we don't want to deploy everything we have right now at `develop`. So this commit should go away: [`da93990` (#2391)](https://github.com/Automattic/woocommerce-services/pull/2391/commits/da939905b6a8fa30dd616e1bddbb8a971050ae9a)

Ferdev:
Hope you don't mind, I've updated some files from `master` which seemed out of sync. I've also reverted the merge commit from `develop`. I'll start reviewing your changes from here!

superdav42:
> Hope you don't mind, I've updated some files from `master` which seemed out of sync. I've also reverted the merge commit from `develop`. I'll start reviewing your changes from here!

Rebasing might have worked better than reverting. Either way thanks for taking care of it.

Ferdev (approved):
Looks good to me!
In order to reproduce the error, I had to go first to a shipping zone and edit any of the shipping methods, then save the shipping zone again. Just leaving this note here for completeness.

PR #2393, opened 2021-03-19
Replace EllipsisMenu with WP component DropdownMenu

waclawjacek:
## To-dos

Although the implementation is done, this PR is posted as a draft as tests have to be updated before it is submitted for review.

## Description

Replace the `EllipsisMenu` component with `DropdownMenu` from `@wordpress/components`.

Since `EllipsisMenu` was only used in one place, instead of creating a new component wrapping `DropdownMenu` I opted for using `DropdownMenu` directly.

#### Keyboard navigation!

Thanks to using the component from `@wordpress/components`, the menu got full keyboard navigation support. Previously there was none.

#### Toggle animation is no longer there

In the current version of WCS&T (1.25.9), when the menu toggle is clicked, the ellipsis icon is animated: it rotates 90º with a slight bounce [(demo in Calypso docs)](https://wpcalypso.wordpress.com/devdocs/design/ellipsis-menu).

The version in this PR does not include that.

#### Tooltips replaced by `info` text

Although it might be possible to use tooltips on the menu items, I opted for DropdownMenu's supported feature of displaying a description text under the menu item label. This is very friendly for keyboard navigation: disabled menu items are skipped when using keyboard nav, therefore not allowing a "focus" state, which in turn leads to a lack of conditions under which the tooltip should be displayed.

It might be possible to work around this if necessary however the main reason why I wanted to point this out is that using the info text might cause some clutter when multiple items have it:

After - potential clutter

Another workaround might be to change the wording of these messages.

### Related issue(s)

Part of #2333.

#### Before

**Mouse pointer hovering over an active item**

Before

**Mouse pointer hovering over a disabled item**

Before - mouse hover over disabled item

#### After

**Mouse pointer hovering over an active item**

After

**Mouse pointer hovering over a disabled item**

After - mouse hover over disabled item

### Checklist

- [ ] unit tests

### DIFF
--- a/assets/stylesheets/style.scss
+++ b/assets/stylesheets/style.scss
@@ -43,7 +43,7 @@
 			text-shadow: none !important;
 		}
 		//wp-admin styles override the disabled button with !important, so need to do the same here
-		
+
 		&.is-busy {
 			background-size: 120px 100% !important;
 			background-image: linear-gradient(
@@ -311,16 +311,6 @@
 				text-shadow: 0 -1px 1px #006799, 1px 0 1px #006799, 0 1px 1px #006799, -1px 0 1px #006799;
 			}
 		}
-	}
-}
-
-// Adding here so it will not have the .wp-core-ui.wp-admin .wcc-root prefix
-.shipping-label__item-menu-reprint-expired.popover__menu-item {
-	cursor: pointer;
-	background: transparent;
-	color: $muriel-gray-100;
-	.gridicon {
-		color: $muriel-gray-100;
 	}
 }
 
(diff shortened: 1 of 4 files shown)


harriswong:
@waclawjacek This PR you have here will fix the `<PopoverMenuItem>` on https://github.com/Automattic/woocommerce-services/issues/2333 as well! 

The first 4 commits look good. I was going to copy the Calypso's `components/popover/menu-item` over but I see that you got the ellipsis and the dashicons figured out. 

Can we revive this?

waclawjacek:
Will do! Let me just dig my way out of the issues currently on my plate and I'll get back to this when I can. :)

harriswong:
image
If there is another order, the popup menu is hidden behind it. This blocks the user from clicking the custom form.

samnajian:
Closing as it's too old, please feel free to reopen if you disagree @waclawjacek.

PR #2394, opened 2021-03-19
Do not retry fetching services on auth failure

superdav42:
closes https://github.com/Automattic/woocommerce-shipping-issues/issues/192

## Description
Many extra requests are being sent to the connect server with changes in 1.25.9. Most of these are from stores which are not authenticated. This happens because a store with an invalid blog token we keep trying to fetch services even though it fails each time.
 
### Related issue(s)
https://github.com/Automattic/woocommerce-shipping-issues/issues/192

Ensure your site is connected to production connect server. 
Change your blog token to an invalid one. This code should do it
```php
add_action( 'jetpack_loaded', function () {
	Jetpack_Options::update_option( 'blog_token', 'PjbxIr&GJJ8MzH^Vfan4gjXoqciy8g*Y.3Ww95YzGFyW(aY*g!C&VszYI0q^K7SG6' );
} );
```
If you using master you will see a 401 request on each page load to `/services` with this branch it should only happen one time.

### DIFF
--- a/classes/class-wc-connect-service-schemas-store.php
+++ b/classes/class-wc-connect-service-schemas-store.php
@@ -26,9 +26,14 @@ if ( ! class_exists( 'WC_Connect_Service_Schemas_Store' ) ) {
 			$response_body = $this->api_client->get_service_schemas();
 
 			if ( is_wp_error( $response_body ) ) {
+				$error_data = $response_body->get_error_data();
+				if ( isset( $error_data['response_status_code'] ) ) {
+					$this->update_last_fetch_result_code( $error_data['response_status_code'] );
+				}
 				$this->logger->log( $response_body, __FUNCTION__ );
 				return false;
 			}
+			$this->update_last_fetch_result_code( '200' );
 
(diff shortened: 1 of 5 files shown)


classes/class-wc-connect-service-schemas-store.php
```diff
@@ -60,6 +65,17 @@ protected function update_last_fetch_timestamp() {
 			WC_Connect_Options::update_option( 'services_last_update', time() );
 		}
 
+		public function get_last_fetch_result_code() {
+			return WC_Connect_Options::get_option( 'services_last_result_code' );
+		}
```
Ferdev, line 70: Apparently this function is not being used, should we remove it?

superdav42: no, I missed committing a file that uses it

Ferdev (approved):
Gave an extensive test to this on a Jurassic Ninja site connected to `api.woocommerce.com` and `WOOCOMMERCE_CONNECT_FREQUENT_FETCH` set to `false` and I can confirm this PR fixes the reiterative 401 errors.

PR #2402, opened 2021-03-29
Catch exception if the "DHL live rates available" note cannot be created

### LINKED ISSUE #2400: "DHL live rates available" notice causes fatal error when WC Admin is disabled
## Description

This was first reported in p1616690071123300-slack-C7U3Y3VMY mentioning a report in 3853620-zd-woothemes.

It seems that on sites where WC Admin is disabled, wp-admin becomes inaccessible due to an error caused by the "DHL live rates available" notice:

```
PHP Fatal error:  Uncaught Exception: Invalid data store. in /var/www/html/wp-content/plugins/woocommerce/includes/class-wc-data-store.php:107
Stack trace:
#0 /var/www/html/wp-content/plugins/woocommerce/includes/class-wc-data-store.php(139): WC_Data_Store->__construct('admin')
#1 /var/www/html/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/Notes/Note.php(87): WC_Data_Store::load('admin-note')
#2 /var/www/html/wp-content/plugins/woocommerce-services/classes/class-wc-connect-note-dhl-live-rates-available.php(37): Automattic\WooCommerce\Admin\Notes\Note->__construct()
#3 /var/www/html/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/Notes/NoteTraits.php(74): WC_Connect_Note_DHL_Live_Rates_Available::get_note()
#4 /var/www/html/wp-content/plugins/woocommerce-services/classes/class-wc-connect-note-dhl-live-rates-available.php(27): WC_Connect_Note_DHL_Live_Rates_Available::possibly_add_note()
#5 /var/www/html/wp-content/plugins/woocommerce-services/woocommerce-services.php(746): WC_Connect_Note_DHL_Live_ in /var/www/html/wp-content/plugins/woocommerce/includes/class-wc-data-store.php on line 107
```

To be able to use wp-admin again, a site admin has to deactivate WCS&T or enable WC Admin.

## How to reproduce

WC Admin can be disabled using this filter:

```php
add_filter( 'woocommerce_admin_disabled', '__return_true' );
```

Tested on WP `5.7`, WC `5.1.0`, WCS&T `1.25.10`.

waclawjacek:
## Description

It is possible for WC Admin to be available but disabled if the `woocommerce_admin_disabled` filter is used. If this happens,
attempts to create a note will fail due to the "admin-note" data store being unavailable.

This commit adds a check if an exception is thrown on note creation and removal attempts.

### Some more details on the issue

I researched the way WC Admin is initialized to learn more about the context of how we are adding the note and here is what I learned:

We are adding the note during the `woocommerce_init` hook, which is called during the `init` hook, which is fired *after* `plugins_loaded` - meaning all plugins are loaded at this time. So far, so good.

The problem seems to be that when [the `woocommerce_admin_disabled` filter](https://github.com/woocommerce/woocommerce-admin/blob/0b68c1d7c0c5025789071022efa129126da1ded5/src/FeaturePlugin.php#L67) is set to `true`, even though WC Admin is loaded and all of its classes and traits are available (which is what we currently check for), it hasn't really registered its "admin-note" data store yet, hence the issue.

### Alternative approaches

We can generally solve this in a couple of different ways, checking one of the following:

1. The value of `apply_filters( 'woocommerce_admin_disabled' )`.
1. If the `admin-note` data store is available using `apply_filters( 'woocommerce_data_stores' )`.
1. If the attempt to add a note throws an exception.

I opted for the last approach as it does not require WCS&T to be aware of WC Admin's implementation details such as what exactly is disabled when `woocommerce_admin_disabled` is false, and data store names. We are already using the trait so checking if it throws an exception keeps WCS&T's dependencies to just this trait.

However, the *[Using the Admin Notes Inbox in WooCommerce](https://developer.woocommerce.com/2020/10/16/using-the-admin-notes-inbox-in-woocommerce/)* article and [these examples](https://woocommerce.github.io/woocommerce-admin/#/examples/activity-panel-inbox) interact with the data store directly by providing its name in calls to `WC_Data_Store::load()` so it might also be OK to replace the `try-catch` blocks with just one inside `can_add_wc_admin_notices()`.

### Related issue(s)

Fixes #2400.

**To reproduce the issue:**

1. Add the following code to the site:
    ```php
    add_filter( 'woocommerce_admin_disabled', '__return_true' );
    ```
1. The site throws a fatal error when trying to access wp-admin.

With the changes in this PR applied, a fatal error is not thrown.

**To verify no regressions happened:**

1. Set up a fresh site and install the plugin.
2. No notice is displayed.
3. Enable DHL live rates and refresh the schemas.
4. The notice appears. Click either of the buttons to dismiss it.
5. The notice is gone.
5. Reinstall the plugin and activate it with DHL live rates still enabled.
6. The notice appears immediately.

### Checklist

- [ ] unit tests


FangedParakeet (approved):
Code change here looks pretty simple and works as expected--the fatal error is avoided!

I did notice that, following the examples you linked, checking for the two classes `WC_Admin_Notes` and `WC_Data_Store` like below did also prevent the fatal error, without having to resort to using a try/catch block where we just ignore the exception instead of handling it. 

```php
if ( ! class_exists( 'WC_Admin_Notes' ) ) {
   return;
}

if ( ! class_exists( 'WC_Data_Store' ) ) {
   return;
}
// ...
``` 

I think checking for specific conditions is usually better form, but personally I am more inclined to more general solutions and in this case, where the notice seems like a small enough feature that could be affected by many other unexpected plugin interactions, I'm willing to allow you to push the global premiere of the try/catch block in our WCS plugin. 

@waclawjacek let me know your thoughts about the above. Sorry, I broke my local Docker (again), which delayed this review. Fix up that pesky changelog and I'll gladly approve. Nice one.

waclawjacek:
@FangedParakeet Thanks so much for reviewing the PR!

I agree that muting the exception is not the nicest approach.

I did a search for the issue in the WC Admin repo and noticed they ran into a similar issue and [fixed it in WC Admin 2.1.2](https://github.com/woocommerce/woocommerce-admin/pull/6532/files#diff-16eea59d38b2841021d4a1801a6f55f0ea9496d1da6c0dc1f6a15411509619acR153) by trying to load the data store and catching the exception if it does not exist for a single note. I have [posted an issue in the WC Admin repo](https://github.com/woocommerce/woocommerce-admin/issues/6703) suggesting doing this for all notes.

I have resolved the changelog conflicts and replaced the previous changes in this PR with one similar to what was implemented in WC Admin 2.1.2. If you feel like using either of the alternative approaches (below) would be better, please let me know and I will do that.

## Your proposed solution

Well, this is a bit of a tricky one and I missed it as well because the `WC_Admin_Notes` class actually is in a namespace. So if you check for this, like in the examples:

```php
if ( ! class_exists( 'WC_Admin_Notes' ) ) {
```

It is true that the exception is not thrown but this is because note creation is never attempted - this class does not actually exist. :) It is actually namespaced so the check should be:

```php
if ( ! class_exists( 'Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes' ) ) {
```

Using this code, the class always exists (just like the other classes WCS&T currently checks for) and the exception is thrown. This means that this approach unfortunately does not work.

## The alternative approaches

The alternative solutions that I mentioned would all change the `can_add_wc_admin_notice()` method and would work as follows:

**Approach #1 Checking the value of `apply_filters( 'woocommerce_admin_disabled' )`**

```php
public static function can_add_wc_admin_notice() {
  $classes_exist = trait_exists( 'Automattic\WooCommerce\Admin\Notes\NoteTraits' ) && class_exists( 'Automattic\WooCommerce\Admin\Notes\Note' );
  $is_wc_admin_enabled = ! apply_filters( 'woocommerce_admin_disabled', false );

  return $classes_exist && $is_wc_admin_enabled;
}
```

**Approach #2 Checking if the admin-note data store is available using `apply_filters( 'woocommerce_data_stores' )`**

```php
public static function can_add_wc_admin_notice() {
  $classes_exist = trait_exists( 'Automattic\WooCommerce\Admin\Notes\NoteTraits' ) && class_exists( 'Automattic\WooCommerce\Admin\Notes\Note' );
  $data_stores = apply_filters( 'woocommerce_data_stores', array() );

  return $classes_exist && in_array( 'admin-note', $data_stores, true );
}
```

FangedParakeet (approved):
Thanks for all the investigation and research you launched yourself into on the back of my one little comment! Appreciate the extensive discussion write-up as well and nice catch on the namespacing. 

In terms of the approach, I think this looks clean enough and if it's an approach that has already worked with roaring success in WC Admin, I see no reason to change a winning formula. Moreover I think moving the logic to `can_add_wc_admin_notice()` makes more sense and improves that function as well. 

Everything is still working as expected: no errors in sight.

Nice one, this one has my blessing for smooth sailing: ship it at will!

PR #2406, opened 2021-04-01
fix: UPS registration form submission on invalid data

frosso:
### Related issue(s)

I noticed that if the merchant tries to connect their UPS account (with the invoice information) and the submission doesn't go well, the button remains deactivated until all the fields are re-edited.
Fixing.

- Go register a UPS account
- Enter some jibberish (without checking the "I have been issued an invoice from UPS within the past 90 days" checkbox)
- Submission should fail
- Check the checkbox
- Enter additional jibberish
- Submit again
- Submission should fail, but the "connect" button should be clickable
- On the `develop` branch, the "Connect" button didn't become clickable until both "UPS invoice number" and "UPS invoice date" were edited

### Checklist

- [x] unit tests

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/ups-settings-form.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/ups-settings-form.js
@@ -191,12 +191,12 @@ const UpsSettingsForm = ({ translate, errorNotice, successNotice, countryNames, 
 					type: 'UpsAccount',
 				};
 
-				if( isInvoiceDetailsChecked ) {
-					delete formValues.invoice_number;
-					delete formValues.invoice_date;
-					delete formValues.invoice_amount;
-					delete formValues.invoice_currency;
-					delete formValues.invoice_control_id;
+				if( ! isInvoiceDetailsChecked ) {
+					delete requestBody.invoice_number;
+					delete requestBody.invoice_date;
+					delete requestBody.invoice_amount;
+					delete requestBody.invoice_currency;
+					delete requestBody.invoice_control_id;
 				}
 
 				const result = await api.post( siteId, api.url.shippingCarrier(), requestBody );
--- a/client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/ups-settings-form.js
+++ b/client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/ups-settings-form.js
@@ -389,12 +389,14 @@ const UpsSettingsForm = ({ translate, errorNotice, successNotice, countryNames, 
 								<div className="carrier-accounts__settings-two-columns">
 									<TextField
 										id="invoice_number"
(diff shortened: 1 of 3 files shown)


client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/ups-settings-form.js
```diff
@@ -191,12 +191,12 @@ const UpsSettingsForm = ({ translate, errorNotice, successNotice, countryNames,
 					type: 'UpsAccount',
 				};
 
-				if( isInvoiceDetailsChecked ) {
```
frosso, line 194: The form values were being deleted upon submission if the checkbox was checked.
We should _not_ send the values upon submission if the checkbox is _not_ checked.

harriswong:
👀

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/ups-settings-form.js
harriswong: It seems strange we did it this way. Do you know why we delete when it's check?

harriswong (approved):
Tested 
- Uncheck the "I have been issued an invoice from UPS within the past 90 days" checkbox, observe network request body to not have the invoice data
- Check the box and confirm data is passed in.
- Check the box, fill invoice data, uncheck box, click "connect". Confirmed no data passed through


👍 LGTM!

client/extensions/woocommerce/woocommerce-services/views/carrier-accounts/ups-settings-form.js
frosso: When the checkbox is checked and then unchecked, we don't clear the values for the invoice data in the `formValues` object.
So if the merchant checks the checkbox, then enters some values for the invoice details, then unchecks the checkbox, the "hidden" invoice values would otherwise be sent in the request.
But we also want to "keep" the values previously entered in mind if the merchant unchecks the checkbox by mistake.

Let me know if it makes sense, it's not too straightforward :D

harriswong: Ah I see, thanks. In this branch, if I enter check the box, enter invoice data, uncheck the box by mistake, check it again, all the invoice data is gone. This is still okay right?

frosso: Thanks for pointing it out - that's weird, it must have been missed. I fixed it in 50175438. I tried adding a test, but enzyme doesn't do a good job at representing the values of the field. RTL should be better, tho!

harriswong:
Tested again
- Check the box, fill invoice data, uncheck box, click "connect". Confirmed no data passed through
- Check the box, fill invoice data, uncheck box, check box again. All the data previously filled are there. Click "connect". Confirmed data passed through.

👍

PR #2410, opened 2021-04-10
Fix/2409 fix error on checkout without using wcs rates

### LINKED ISSUE #2409: Unnecessary track_completed_order errors
Noticed this issue after a user mentioned this error occurring frequently [here](https://wordpress.org/support/topic/error-track_completed_order/). The user noticed the following error occurring frequently in the logs for their site.

```
Unable to send subscription event: No shipping rate could be calculated. No items in the package are shippable. (track_completed_order)
```

The error looks scary and suggests that something is not working correctly. What is actually happening is that the client-side subscription handler for checkout events has not been able to find any WCS shipping methods involved in a user's checkout and throws an error, because it cannot send any subscription event. The client should not throw an error, as it is perfectly conceivable for a customer on a user's store to complete a checkout without using a WCS shipping method.

## Steps to reproduce issue

1. Ensure that you have checked out the latest version of the WCS plugin and the Connect Server. Also ensure that debug and logging are both enabled on your site. 
2. Create a shipping zone and enable any shipping method other than any WCS shipping methods on this zone. Do not enable any WCS shipping methods on this zone.
3. Add a product to your cart and complete a checkout using a destination shipping address within the above aforementioned zone.
4. Check the WCS debug log and react with flabbergasted horror when you discover the `Unable to send subscription event` error appearing upon completing your checkout.

FangedParakeet:
## Description

Noticed this issue after a user mentioned this error occurring frequently [here](https://wordpress.org/support/topic/error-track_completed_order/). The user noticed the following error occurring frequently in the logs for their site.

```
Unable to send subscription event: No shipping rate could be calculated. No items in the package are shippable. (track_completed_order)
```

The error looks scary and suggests that something is not working correctly. What is actually happening is that the client-side subscription handler for checkout events has not been able to find any WCS shipping methods involved in a user's checkout and throws an error, because it cannot send any subscription event. The client should not throw an error, as it is perfectly conceivable for a customer on a user's store to complete a checkout without using a WCS shipping method.

### Related issue(s)

Closes #2409 

1. Ensure that you have checked out the changes in this PR and the latest version of the Connect Server. Also ensure that debug and logging are both enabled on your site. 
2. Create a shipping zone and enable any shipping method other than any WCS shipping methods on this zone. Do not enable any WCS shipping methods on this zone.
3. Add a product to your cart and complete a checkout using a destination shipping address within the above aforementioned zone.
4. Check the WCS debug log and notice the absence of `Unable to send subscription event` errors.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 1.25.12 - 2021-xx-xx =
 * Fix   - Fix PHP 5.6 compatibility issue.
+* Fix   - Removes unnecessary subscription debug error logs.
 
 = 1.25.11 - 2021-04-06 =
 * Fix	- Ensure status page is displayed on new WC navigation menu.
@@ -12,7 +13,6 @@
 * Tweak	- Display spinner icon during service data refresh.
 * Add	- Adds Dockerized E2E tests with GitHub Action integration.
 * Fix   - Handle DHL live rates notice creation and deletion errors.
-* Fix   - Removes unnecessary subscription debug error logs.
 
 = 1.25.10 - 2021-03-24 =
 * Add   - Add an endpoint for shipping label creation eligibility and share code for store eligibility.
--- a/changelog.txt
+++ b/changelog.txt
@@ -9,6 +9,7 @@
 * Tweak	- Display spinner icon during service data refresh.
 * Add	- Adds Dockerized E2E tests with GitHub Action integration.
 * Fix   - Handle DHL live rates notice creation and deletion errors.
+* Fix   - Removes unnecessary subscription debug error logs.
 
 = 1.25.10 - 2021-03-24 =
 * Add   - Add an endpoint for shipping label creation eligibility and share code for store eligibility.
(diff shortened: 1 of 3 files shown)


frosso (approved):
Looks safe 👍  I wonder about the return type of the method now and whether it is consistent - but it doesn't seem to be used unless it's a `WP_Error`, so I think it's fine.

FangedParakeet:
@frosso, actually in hindsight, I agree with you and moreover don't like having a function return a value only to throw it away immediately. Like that nagging feeling of thinking you've left the oven on or forgotten to feed your cat or left your cat in the oven, the regret of this code snippet would only serve to prey on my mind, like a cat in an oven full of mice. 

Just made a quick update to move that logic out of the api client--and as a bonus, I even put the changelog entry in the right place all by myself! Cheers and hopefully we will both now sleep at ease tonight.

frosso:
Uuuh - love this version even better! 👏  Good idea!

PR #2411, opened 2021-04-10
Fix/194 update plugin author

FangedParakeet:
## Description

An HE [posed the question on Slack](p1617974817323500-slack-C7U3Y3VMY), "Why does it say 'by Automattic' rather than 'by WooCommerce' in the system report?"

I initially thought this was a little pedantic (and I probably still do), but checking through the list of Heisenberg/Extendables plugins, I learned that all of them with two exceptions--WooCommerce Shipping & Tax and WooCommerce Distance Rate Shipping--state that their author is WooCommerce, while the latter two state Automattic. 

I really hope this inconsistency hasn't been giving any of our users' sleepless nights, but since it's a one-line change, I created this issue, so that we could put the issue of who truly owns us to bed once and for all.

### Related issue(s)

Closes Automattic/woocommerce-shipping-issues#194

1. Print system status report and check the author name next to the plugin.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 1.25.12 - 2021-xx-xx =
 * Fix   - Fix PHP 5.6 compatibility issue.
+* Tweak - Update plugin author name.
 
 = 1.25.11 - 2021-04-06 =
 * Fix	- Ensure status page is displayed on new WC navigation menu.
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -12,6 +12,7 @@
 * Tweak	- Display spinner icon during service data refresh.
 * Add	- Adds Dockerized E2E tests with GitHub Action integration.
 * Fix   - Handle DHL live rates notice creation and deletion errors.
+* Tweak - Update plugin author name.
```
frosso, line 15: Does this need to go under `1.25.12`?

frosso (approved):
gg ez

changelog.txt
FangedParakeet: Gonna blame Git's auto-merge for my inability to edit a text file correctly, but just moved it up into place now.

frosso (approved):
Thanks! :)

PR #2412, opened 2021-04-16
chore(TaxJar): remove branching on tax calculation

frosso:
## Description

The `$wc_cart_object->remove_taxes();` branch is never going to be reached.
First of all, `wc_cart_object` is not a variable in the scope, so the whole application would break.
Second of all, it will never be reached because of the early return on this line: https://github.com/Automattic/woocommerce-services/blob/develop/classes/class-wc-connect-taxjar-integration.php#L870

### Checklist

- [ ] unit tests

I didn't add a changelog entry because this little change isn't worth it.

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -933,10 +933,7 @@ class WC_Connect_TaxJar_Integration {
 			}
 		}
 
-		// Remove taxes if they are set somehow and customer is exempt
-		if ( WC()->customer->is_vat_exempt() ) {
-			$wc_cart_object->remove_taxes();
-		} elseif ( $taxes['has_nexus'] ) {
+		if ( $taxes['has_nexus'] ) {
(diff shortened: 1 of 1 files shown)


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -933,10 +933,7 @@ public function calculate_tax( $options = array() ) {
 			}
 		}
 
-		// Remove taxes if they are set somehow and customer is exempt
-		if ( WC()->customer->is_vat_exempt() ) {
```
frosso, line 937: This line is never reached because of the early return at https://github.com/Automattic/woocommerce-services/blob/develop/classes/class-wc-connect-taxjar-integration.php#L870

harriswong:
👀

harriswong (approved):
LGTM 👍 

Nice clean up! The condition mentioned indeed exit early and this `if ( WC()->customer->is_vat_exempt() ) {` is not needed.

Issue #2226, opened 2020-10-21
Tax reports don't break down taxes with enough detail


FreshPhil:
This is a follow up to a Slack conversation: p1603061005032000-slack-CTYKKGC1Y

## Problem
When using Automated Taxes, the `WooCommerce > Reports > Taxes` report does not show a break down of the tax rates that were used/charged.

All transactions within a state are shown as a single rate. For example, all sales for a California based company would be filed under the `US-CA-CA TAX-1` tax.

This is because we use the same naming scheme for all taxes within a given state/province.

![standard_tax_rates_added_by_wcservices](https://user-images.githubusercontent.com/52214/106674772-cdb5f880-65ab-11eb-891a-99a432a6af18.png)

The problem with this is that it's a common practice to have different rates by `county` within the same state in the US. Users can't report on their taxes charged by county though, since all tax rates for their state are grouped under the same name.

If someone were manually entering in rates (not using WCTax), this wouldn't be an issue as each counties tax rate could be name for that county.

## Solution

TaxJar returns `county` in their request-response. If we append that to the end of the tax name, this would give users a way to see the different taxes charged in their tax report.

While this isn't a "bug". I think this is a very important quality of life improvement that we should prioritize as it severely limits the reporting effectiveness for users.

mouligreenlaw:
Closing the original ZD ticket and adding the number here so they will get notified hopefully:
3411927-zen

zwiesman31:
3653299-zen

zhollenbeak:
Another in 3892911-zen

zhollenbeak:
Another report of this problem in 29040201-zen

PR #1899, opened 2020-02-04
Ignore county when country is UK for automated tax

harriswong:
When country is UK, the shipping address allow county to be a text field. This text field does not respect a "state code" format (ie. 2 letters). This creates a problem with WooCommerce core's tax code. While the core team is working on the issue, we are going to patch this by disabling county in the tax field for automated tax. County is optional and this change will not break tax rate.

### Notes
County is optional field for UK. https://developers.taxjar.com/api/reference/#post-calculate-sales-tax-for-an-order
image
By removing the county field, WooCommerce core will insert this entry as `*` which can match any county.
image

### To test
1. Set store address to an UK address, https://localhost/wp-admin/admin.php?page=wc-settings&tab=general. We can use the ones from the issue, (e.g. 63 Elford Crescent, Plympton, Plymouth, UK, PL74BT)
2. Add an item, go to the checkout page. Use shipping addresses:
- 2 Chestnut Hall Avenue, Maghaberry, Moira, County Armagh, BT67 0GG
- 40 Whitehall Way, Rockingham, Rotherham, South yorkshire, S61 4HW
- 19 Carisbrooke Close, EASTBOURNE, East Sussex, BN23 8EQ
3. Refresh the checkout page a few times
4. Verify `VAT` shows up as 20% tax of the product cost
5. Go to https://localhost/wp-admin/admin.php?page=wc-settings&tab=tax&section=standard, there should only be 1 entry per address. 

Closes #1600

PR #2413, opened 2021-04-16
refactor: rename tax name for US

frosso:
## Description

Renames the "Tax name" column, so that people can more easily identify the tax rates by the counties for reports.
This change is only valid for the US (I didn't want to break it for other countries). Although I feel like it might be useful to extend this format for all countries 🤷 

### Related issue(s)

Fix: https://github.com/Automattic/woocommerce-services/issues/2226

- I recommend deleting all your existing tax rates before this - or maybe only keep the ones you want to see changing
- Change your store to be based in MN (just as an example for nexus)
- Place 2 orders on your store to both these addresses:
  - 1000 Nicollet Mall, Minneapolis, MN 55403
  - 2105 London Rd, Duluth, MN 55812
- Go to WooCommerce > Settings > Tax > Standard rates, you should see updated values: https://d.pr/i/aSdMzN
- Go to Analytics > Taxes, you should see updated values in the report (might take a bit after an order is placed due to internal caching)

Before: https://d.pr/i/psnvuh
After: https://d.pr/i/lg7S6a

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1048,15 +1048,13 @@ class WC_Connect_TaxJar_Integration {
 		// prevents from saving a "state" column value for GB
 		$to_state = 'GB' === $location['to_country'] ? '' : $location['to_state'];
 
-		// For the US, we're going to modify the name of the tax rate to simplify the reporting and distinguish between the tax rates at the counties level.
-		// I would love to do this for other locations, but it looks like that would create issues.
-		// For example, for the UK it would continuously rename the rate name with an updated `state` "piece", each time a request is made
-		$tax_rate_name = self::generate_tax_rate_name( $taxjar_response, $location['to_country'], $to_state );
-
 		$tax_rate = array(
 			'tax_rate_country'  => $location['to_country'],
 			'tax_rate_state'    => $to_state,
-			'tax_rate_name'     => sprintf( '%s Tax', $tax_rate_name ),
+			// For the US, we're going to modify the name of the tax rate to simplify the reporting and distinguish between the tax rates at the counties level.
+			// I would love to do this for other locations, but it looks like that would create issues.
+			// For example, for the UK it would continuously rename the rate name with an updated `state` "piece", each time a request is made
+			'tax_rate_name'     => sprintf( '%s Tax', self::generate_tax_rate_name( $taxjar_response, $location['to_country'], $to_state ) ),
 			'tax_rate_priority' => 1,
 			'tax_rate_compound' => false,
 			'tax_rate_shipping' => $freight_taxable,
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -50,6 +50,51 @@ class WC_Connect_TaxJar_Integration {
 		$this->cache_time = HOUR_IN_SECONDS;
 	}
 
+	/**
+	 * @param $taxjar_response
+	 * @param string          $to_country
+	 * @param string          $to_state
+	 *
+	 * @return string
+	 */
+	private static function generate_tax_rate_name( $taxjar_response, $to_country, $to_state ): mixed {
+		if ( 'US' !== $to_country ) {
+			return $to_state;
+		}
+
+		// for a list of possible attributes in the `jurisdictions` attribute, see:
+		// https://developers.taxjar.com/api/reference/#post-calculate-sales-tax-for-an-order
+		$jurisdiction_pieces = array_merge(
+			[
+				'city'    => '',
+				'county'  => '',
+				'state'   => $to_state,
+				'country' => $to_country,
+			],
+			(array) $taxjar_response->jurisdictions
+		);
+
+		// sometimes TaxJar returns a string with the value 'FALSE' for `state`.
+		if ( rest_is_boolean( $to_state ) ) {
+			$jurisdiction_pieces['state'] = '';
+		}
+
+		return join(
+			'-',
+			array_filter(
+				[
+					// the `$jurisdiction_pieces` is not really sorted
+					// so let's sort it with COUNTRY-STATE-COUNTY-CITY
+					// `array_filter` will take care of filtering out the "falsy" entries
+					$jurisdiction_pieces['country'],
+					$jurisdiction_pieces['state'],
+					$jurisdiction_pieces['county'],
+					$jurisdiction_pieces['city'],
+				]
+			)
+		);
+	}
+
 	public function init() {
 		// Only enable WCS TaxJar integration if the official TaxJar plugin isn't active.
 		if ( class_exists( 'WC_Taxjar' ) ) {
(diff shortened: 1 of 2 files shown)


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -993,13 +996,55 @@ public function calculate_tax( $options = array() ) {
 	 * Unchanged from the TaxJar plugin.
 	 * See: https://github.com/taxjar/taxjar-woocommerce-plugin/blob/9d8e725/includes/class-wc-taxjar-integration.php#L396
 	 *
-	 * @return void
+	 * @return int
 	 */
-	public function create_or_update_tax_rate( $location, $rate, $tax_class = '', $freight_taxable = 1 ) {
+	public function create_or_update_tax_rate( $taxjar_response, $location, $rate, $tax_class = '', $freight_taxable = 1 ) {
```
frosso, line 1043: Added the `$taxjar_response` argument

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -909,27 +909,32 @@ public function calculate_tax( $options = array() ) {
 
 		$response = $this->smartcalcs_cache_request( wp_json_encode( $body ) );
 
-		if ( isset( $response ) ) {
-			// Log the response
-			$this->_log( 'Received: ' . $response['body'] );
+		// if no response, no need to keep going - bail early
```
frosso, line 957: Refactored with an early return - there should be no difference compared to the previous behavior. I like to return early because then I don't need to continue reading the code with the previous conditions in mind.

harriswong:
👀

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -50,6 +50,51 @@ public function __construct(
...
 	}
 
+	/**
+	 * @param mixed  $taxjar_response
+	 * @param string $to_country
+	 * @param string $to_state
+	 *
+	 * @return string
+	 */
+	private static function generate_tax_rate_name( $taxjar_response, $to_country, $to_state ) {
```
harriswong, line 60: Why do we need `$to_country, $to_state` if `$taxjar_response` also contains these values?

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -993,13 +1041,20 @@ public function calculate_tax( $options = array() ) {
...
 	 * See: https://github.com/taxjar/taxjar-woocommerce-plugin/blob/9d8e725/includes/class-wc-taxjar-integration.php#L396
 	 *
-	 * @return void
+	 * @return int
 	 */
-	public function create_or_update_tax_rate( $location, $rate, $tax_class = '', $freight_taxable = 1 ) {
+	public function create_or_update_tax_rate( $taxjar_response, $location, $rate, $tax_class = '', $freight_taxable = 1 ) {
+		// all the states in GB have the same tax rate
+		// prevents from saving a "state" column value for GB
+		$to_state = 'GB' === $location['to_country'] ? '' : $location['to_state'];
```
harriswong, line 1049: The `GB` check was for https://github.com/Automattic/woocommerce-services/pull/1899. If we moved it, we should probably do some regression test on the issue.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -945,10 +995,6 @@ public function calculate_tax( $options = array() ) {
 				'to_city'    => $to_city,
 			);
 
-			if ( 'GB' === $to_country ) {
-				$location['to_state'] = '';
-			}
-
```
harriswong, line 948: The `GB` check was for https://github.com/Automattic/woocommerce-services/pull/1899. If we moved it, we should probably do some regression test on the issue.

classes/class-wc-connect-taxjar-integration.php
frosso: That's a good question - `to_state` might not be always present in the TaxJar response: https://d.pr/i/mYKd6V

I added both as a fallback scenario

classes/class-wc-connect-taxjar-integration.php
frosso: I moved it here to colocate colocate all the "state"/"country" checks: https://github.com/Automattic/woocommerce-services/blob/624b44e509c163715364e2bf2883f3e3f2fd787f/classes/class-wc-connect-taxjar-integration.php#L1049

I checked the scenarios in that PR as well, fell free to double check 👍

classes/class-wc-connect-taxjar-integration.php
harriswong: For the countries that don't have `$to_country, $to_state`, will they be removed form the `join`-`array_filter` logic below regardless?

frosso: Yeah, the logic below is just for the US. But it's worth mentioning that if `$taxjar_response->jurisdictions` doesn't include `state` or `country` for whatever reason, we're also using `$to_state` or `$to_country` as a fallback.

harriswong (approved):
👍 LGTM

Tested ship from and to a MN address. I can confirm the tax name being updated to include more parts. 
image

Do you have any concerns that these new tax names will not backward compatible with the existing reporting/table etc?

frosso:
@harriswong thanks for taking the time to review!

> Do you have any concerns that these new tax names will not be backward compatible with the existing reporting/table etc?

Sorry, can you please expand on that?
All the US tax names will be updated as new orders come in. So older for older orders placed before this update, merchants will still see rates with names like "US-MNTax-1" (until a new order comes in, that is).
So a mix of names could be displayed in the report 🤷‍♂️

harriswong:
From the screenshot in the [issue](https://github.com/Automattic/woocommerce-services/issues/2226), I was thinking if a user has existing records saying `CA Tax`. They upgrade to the newer WCS which will now generate suffix for the tax. They will have something like
```
CA Tax
CA Tax
CA-abc-Tax`
```

I was wondering if we need to worry how we want to deal with the existing records versus the new ones. And if we need to worry about the mixed entries in reporting.

frosso:
Thanks, I see what you mean.
Yeah, I thought about that. It would be great if the entries were all consistent.
We can't delete the older tax rate entries, because otherwise they wouldn't appear correctly in the report for the older orders.
One thing we could do is write some sort of migration script that looks at all the entries in the table and tries to rename them when the plugin update/installation occurs. But I thought that would lengthen the update/installation process, possibly going in timeout for the number of requests.
So instead we could schedule an action to run and run the migration (making all the requests to the connect server), but I thought that would be a bit overkill - also because we need to track whether we ever run the process and we need to keep maintaining that part of the code.
On top of that, we would need to "hack" the process with some sort of fake request with the current store's address and a fake destination address and a fake shipment (since we don't have an endpoint that just returns some rates).

So I was thinking that leaving the rest as-is was a better option, letting the entries update themselves as new orders come in.

What do you think?

harriswong:
Thanks! +1 on as-is.

PR #2416, opened 2021-04-22
fix: check jetpack constant by name

frosso:
## Description

`defined()` in PHP needs the constant's name as `string` for its argument

I tested this in JN by downloading the WCS plugin (without activating it), ssh-ing, patching the 2 files with these changes and then going through the WC onboarding.

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -297,7 +297,7 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 			}
 
 			// check if Jetpack is activated
-			if ( ! class_exists( '\Automattic\Jetpack\Connection\Manager' ) && ! class_exists( '\Automattic\Jetpack\Connection\Tokens' ) ) {
+			if ( ! is_plugin_active( 'jetpack/jetpack.php' ) ) {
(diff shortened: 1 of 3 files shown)


classes/class-wc-connect-nux.php
```diff
@@ -297,7 +297,7 @@ public function get_jetpack_install_status() {
 			}
 
 			// check if Jetpack is activated
-			if ( ! class_exists( '\Automattic\Jetpack\Connection\Manager' ) && ! class_exists( '\Automattic\Jetpack\Connection\Tokens' ) ) {
+			if ( ! is_plugin_active( 'jetpack/jetpack.php' ) ) {
```
frosso, line 300: The comment says `check if Jetpack is activated` - `is_plugin_active` seems to be a good candidate for that, otherwise the `\Automattic\Jetpack\Connection\Manager` class might have been loaded by another plugin (like WCPay)

harriswong:
👀

harriswong (approved):
LGTM 🚢 

- I ran `grep -Eri "defined\(\s*\w+" --include="*.php" .` and confirmed that there are no more string constants

Unit tests should had been able to catch these kinda thing :(

PR #2421, opened 2021-05-06
fix: encoding of html entities

### LINKED ISSUE #2420: "Create shipping label" button doesn't appear when order meta contains double quote
Discovered in: 3951688-zen

**Description:**
The "Create shipping label" button isn't displayed when custom order meta contains a double quote.

**Context:**

When there is custom order meta that contains a double quote then the attribute on .wcc-root contains malformed JSON so a [console error is thrown](https://d.pr/i/Tg7YA3) and rather than displaying the "Create Shipping label" button it reads "Section not loading? Visit the status page for troubleshooting steps." in the Shipping Label box.

_This was reported on a setup that was using the Gravity Forms extension but I assume it would happen no matter where the meta data came from_

frosso:
## Description

Plugin breaks on order page when HTML entities are not properly encoded.

### Related issue(s)

Fixes https://github.com/Automattic/woocommerce-services/issues/2420

- Go to order page
- Add item metadata with encoded entities (e.g.: `&quot;`): https://d.pr/i/HEXQQO
- Page breaks: https://d.pr/i/vXgwc5
- Checkout this branch
- Page does not break

### Checklist

- [ ] unit tests

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -1600,8 +1600,11 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 				)
 			);
 
+			$encoded_arguments = wp_json_encode( $extra_args );
+			$escaped_arguments = function_exists( 'wc_esc_json' ) ? wc_esc_json( $encoded_arguments ) : esc_attr( $encoded_arguments );
+
 			?>
-				<div class="wcc-root woocommerce <?php echo esc_attr( $root_view ); ?>" data-args="<?php echo wc_esc_json( wp_json_encode( $extra_args ) ); ?>">
+				<div class="wcc-root woocommerce <?php echo esc_attr( $root_view ); ?>" data-args="<?php echo $escaped_arguments; ?>">
 					<span class="form-troubles" style="opacity: 0">
 						<?php printf( __( 'Section not loading? Visit the <a href="%s">status page</a> for troubleshooting steps.', 'woocommerce-services' ), $debug_page_uri ); ?>
 					</span>
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -1600,11 +1600,8 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 				)
 			);
 
-			$encoded_arguments = wp_json_encode( $extra_args );
-			$escaped_arguments = function_exists( 'wc_esc_json' ) ? wc_esc_json( $encoded_arguments ) : esc_attr( $encoded_arguments );
-
 			?>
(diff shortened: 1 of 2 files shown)


woocommerce-services.php
```diff
@@ -1596,8 +1596,11 @@ function enqueue_wc_connect_script( $root_view, $extra_args = array() ) {
 				)
 			);
 
+			$encoded_arguments = wp_json_encode( $extra_args );
+			$escaped_arguments = function_exists( 'wc_esc_json' ) ? wc_esc_json( $encoded_arguments ) : esc_attr( $encoded_arguments );
```
frosso, line 1604: Unfortunately, `wc_esc_json` is only part of WC 3.5.5+: https://woocommerce.github.io/code-reference/namespaces/default.html#function_wc_esc_json
I am doing a fallback on the previous `esc_attr` function.

harriswong:
👀

woocommerce-services.php
harriswong: Our support policy is [L -2](https://github.com/Automattic/woocommerce-services/issues/2131). We don't need to support WC <= 3.5.5. We should use `wc_esc_json`.

harriswong (commented):
I was able to reproduce and see the error. I can confirm `wc_esc_json` will not add slash to `&qout;`. I added a comment.

woocommerce-services.php
frosso: Ah, I guess we can update this: https://github.com/Automattic/woocommerce-services/blob/trunk/woocommerce-services.php#L11

Did it in ef5d7f312b19901e98f93b279456041f9c1bcf2d , just to be safe 👍

harriswong: I think it would be good if we include a line in the changelog for the support version change.

harriswong: ^ And i see there is a commit at the end for it. 👍

harriswong (approved):
👍  Last 3 commits look good! 

I think bumping out WC requirements to 5 is good, but I am not sure if there are other processes we need to do when bumping WC/WP versions. I added a note for our team chat tomorrow.

frosso:
@harriswong thanks! I'll keep this open until we discuss tomorrow, is that ok?

harriswong:
Yea for sure! Thanks for taking a look at this!

woocommerce-services.php
frosso: Discussed in team meeting - we'll update the WC version in the header and keep the fallback.

harriswong (approved):
LGTM

Issue #2020, opened 2020-05-04
Incorrectly taxed on shipping


klishb:
I have a very long ticket with support on this (2865426).  They recommended I post here.

The Taxjar API returns:

“freight_taxable”: false,

This response is not used correctly by WooCommerce Services.  For example, in New York state shipping should only be taxed when taxable items are in the cart.  Even with no taxable items in the cart shipping is getting taxed.

klishb:
Since posting this I have learned more about the nature of this problem.  I believe the issue lies in the caching that WooCommerce Services does with the Taxjar responses.  For the reasons stated above this response cannot be cached. Whether or not the shipping should be taxed needs to be determined at the time of sale since it's based on whether or not there are items in the cart that are taxable.  That's just the rule in NY - not sure about other states.  WooCommerce Services needs to leave those calculations up to the TaxJar API since they know all the state rules.  Caching may be reasonable option for some states, but not all.

If you delete the tax tables that got cached then WooCommerce Services uses the API response on the next transaction which will be the correct shipping tax calculation for that particular sale.  Future sales may or may not be correct based on what response was cached and written to the tax tables for the shipping tax option.

superdav42:
@klishb Can you provide more specific details on how to reproduce the problem?

We use the line items in the cart as part of the cache key so if a customer adds/removes an item it will perform a new request for tax rates.

klishb:
@superdav42 very interesting to hear that.  I suspected that may have been the cause, but maybe not.  It's been a long time since I looked at this problem, but I just confirmed it's still a bug.  To reproduce this you could:

1. Create the following "additional tax classes" under WooCommerce tax options
Candy - 40010
Food - 40030
2. Create a product in WooCommerce and set it to taxable with a tax class of "Candy - 40010"
3. Create a product in WooCommerce and set it to taxable with a tax class of "Food - 40030"
3. Set your store's address to NY
4. Set your customer address to NY
5. Put your candy product in the cart.  Both the candy and the shipping should be taxed in the cart.
6. See API request and response to TaxJar API in the taxes log.  Notice TaxJar API responded saying to tax both product and shipping.  This response was used in the cart calculation.
8. Clear the cart
9. Put a food product in your cart.  Your food will not be taxed, but shipping will (this is the problem).
10. See API request and response to TaxJar API in the taxes log.  Notice TaxJar API responded saying "taxable_amount": 0.0 for both product and shipping.  This response was not used in the cart calculation for some reason.

[...]

PR #2424, opened 2021-05-13
Fix shipping tax when multiple tax classes are used

superdav42:
## Description
We store rates return from tax jar in the woocommerce tax tables. Sometimes non-taxable items will charge taxes on shipping because this table is not updated if the tax rate returned by taxjar is 0.0. Really we're backporting a bugfix from the taxjar plugin https://github.com/taxjar/taxjar-woocommerce-plugin/blob/master/includes/class-taxjar-tax-calculation.php#L393

### Related issue(s)
#2020 

Create a taxable and non-taxable product.
Set store location and customer location to NY or any state that requires collecting taxes on shipping charges.
Add taxable product to cart and observe taxes for product and shipping.
Remove taxable product
add non-taxable product to cart and verify that no taxes are charged.
Without this fix shipping taxes would still be charged.

### Checklist

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,6 +6,7 @@
 * Fix - Plugin deletion when WooCommerce core is not present.
 * Tweak - Rename automatic tax names for US.
 * Fix - Check JetPack constant defined by name.
+* Fix - Sometimes taxes charged on shipping when they should not.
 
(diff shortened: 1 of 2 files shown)


frosso (approved):
Changes are aligned with TJ plugin 👍

PR #2427, opened 2021-06-01
Safari native print does not display pdf blob. Falls back to new tab.

### LINKED ISSUE #2272: Blank shipping label generated on MacOS Safari (works on Chrome)
I purchased a USPS First Class International stamp (US to Russia) for an order on my store using Safari on MacOS Big Sur. The PDF sent to me was blank. Totally empty.

blank-on-safari

I opened Google Chrome on the same machine and was able to "reprint" and get a proper (non-empty) label.

harriswong:
## Description
Safari renders a blank print dialog if we [passed in blob URL to the iframe](https://github.com/Automattic/woocommerce-services/blob/5f103cf54f3c1e7e73b7c0b100aca3b41259c289/client/extensions/woocommerce/woocommerce-services/lib/utils/print-document.js#L68).This is a similar issue to the iPad's one reported [here](https://github.com/Automattic/woocommerce-services/pull/2229). 

Opening the blob URL (ie. `blob:http://localhost/123123-123123-123123-12313-123123123`) directly on a new tab renders the PDF okay. 

### Solution
Instead of opening the print dialog directly in Safari. Opens a new tab so the user can print from it. Similar solution described [here](https://github.com/Automattic/woocommerce-services/issues/1166). 

### Related issue(s)
Closes https://github.com/Automattic/woocommerce-services/issues/2272

1. Open Safari, open any existing order and click "Reprint shipping label". Confirms label opens in new tab and cmd-p shows the label in a print dialog.
2. Open Firefox, same result as Safari. The label should opens in a new tab.
3. Open Chrome, click "Reprint shipping label" and it should show up in a print dialog (not new tab).

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.14 - 2020-XX-XX =
+* Fix   - Issue with printing blank label in Safari.
+
 = 1.25.13 - 2021-05-20 =
 * Fix   - Prevent new sites from retrying failed connections.
 * Fix   - Data encoding when entities are part of order meta.
--- a/client/extensions/woocommerce/woocommerce-services/lib/utils/pdf-support.js
+++ b/client/extensions/woocommerce/woocommerce-services/lib/utils/pdf-support.js
@@ -31,15 +31,16 @@ export default memoize( () => {
 		return 'addon';
 	}
 
-	if ( navigator.mimeTypes[ 'application/pdf' ] ) {
-		return 'native';
-	}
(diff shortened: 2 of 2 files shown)


FangedParakeet (approved):
Tested before these changes to confirm that Safari wasn't displaying the labels and tested after to confirm that these changes have resolved the problem. The label now opens in a new tab in all its glory. Chrome and Firefox still working as expected. 

Tested Safari 14.1, Chrome 91.0.4472.77, and Firefox 88.0.1.

Changes LGTM and have my blessing.

PR #2434, opened 2021-06-11
fix: shipping metabox user permissions

### LINKED ISSUE #2433: Inconsistent account permissions to ship labels
## Description

Part of the background information can be found in [this forum thread](https://wordpress.org/support/topic/woocommerece-shipping-label-purchase-problem/) and in Slack p1623358576053300-slack-CKNC7KACA

WCS allows merchants with the `manage_woocommerce` permission to create/view shipping labels and create/view packages.
The `manage_woocommerce` permission comes with a bunch of stuff, including changing WooCommerce settings. The `edit_shop_orders` permission currently allows to view the order page and (consequently) to view the "Print shipping label" information.

A site merchant in the forums is using a plugin to change the permissions for some users on their site. These users have the `edit_shop_orders` permission (but not the `manage_woocommerce` one). The site merchant would like the users to be able to print shipping labels, but they don’t want to let users manage the WC settings. In the order page, the users can see the “create shipping label” modal and Shipment Tracking information, but when the order page loads (or when they open the label modal), they see some error messages (again, because they don’t have the manage_woocommerce permissions): 

Markup 2021-06-10 at 15 49 45

I can confirm that using the `manage_woocommerce` permission allows users to buy/refund shipping labels.

You can test/create user permissions with this plugin, if you'd like: https://wordpress.org/plugins/user-role-editor/

## Acceptance criteria

- Create a new permission check for a new `wcship_manage_labels` permission type, to allow printing/refund shipping labels
- Ensure that the "create shipping label" modal is displayed only to users with the `wcship_manage_labels` and the `manage_woocommerce` permission (it currently is displayed to all users that can see the order page)
- As an alternative (or addition) to the `wcship_manage_labels` permission type, we can even use a new filter to check permissions

frosso:
## Description

Little tiny itty bitty fix to check the user's permission to display the metabox to print shipping labels.

### Related issue(s)

closes https://github.com/Automattic/woocommerce-services/issues/2433

**Without the fix**
- Download a plugin to manage user capabilities, like https://wordpress.org/plugins/leira-roles/ or https://wordpress.org/plugins/user-role-editor/
- Create a user without the `manage_woocommerce` permission
- Go to an order page with shipping labels
- You get an error
Screen Shot 2021-06-11 at 12 53 27 PM

**With the fix**
- Download a plugin to manage user capabilities, like https://wordpress.org/plugins/leira-roles/ or https://wordpress.org/plugins/user-role-editor/
- Create a user without the `manage_woocommerce` permission
- Go to an order page with shipping labels
- No error is displayed
- No shipping label/shipment tracking metabox is displayed
Screen Shot 2021-06-11 at 12 57 47 PM

---

- Add a new `wcship_manage_labels` capability
- Assign it to the user
- For the same order, user can now print/refund labels and add custom packages

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-functions.php
+++ b/classes/class-wc-connect-functions.php
@@ -83,5 +83,17 @@ if ( ! class_exists( 'WC_Connect_Functions' ) ) {
 
 			return false;
 		}
+
+		/**
+		 * Checks whether the current user has permissions to manage shipping labels.
+		 *
+		 * @return boolean
+		 */
+		public static function user_can_manage_labels() {
+			/**
+			 * @since 1.25.14
+			 */
+			return apply_filters( 'wcship_user_can_manage_labels', current_user_can( 'manage_woocommerce' ) || current_user_can( 'wcship_manage_labels' ) );
+		}
 	}
 }
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,9 @@
 = 1.25.14 - 2020-XX-XX =
 * Fix   - Issue with printing blank label in Safari.
 * Fix   - Update DHL Express pickup link.
+* Fix   - Ensure shipping label metabox is displayed to users with the correct capabilities.
+* Add   - Added `wcship_user_can_manage_labels` filter to check permissions to print shipping labels.
+* Add   - Added `wcship_manage_labels` capability to check permissions to print shipping labels.
 
 = 1.25.13 - 2021-05-20 =
 * Fix   - Prevent new sites from retrying failed connections.
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -431,6 +431,12 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		}
 
 		private function calculate_should_show_meta_box() {
+			// not all users have the permission to manage shipping labels.
+			// if a request is made to the JS backend and the user doesn't have permission, an error would be displayed.
+			if ( ! WC_Connect_Functions::user_can_manage_labels() ) {
+				return false;
+			}
+
 			$order = wc_get_order();
 
 			if ( ! $order ) {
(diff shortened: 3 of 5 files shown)


classes/class-wc-connect-functions.php
```diff
@@ -83,5 +83,17 @@ public static function has_cart_or_checkout_block() {
...
+		/**
+		 * Checks whether the current user has permissions to manage shipping labels.
+		 *
+		 * @return boolean
+		 */
+		public static function user_can_manage_labels() {
+			/**
+			 * @since 1.25.14
+			 */
+			return apply_filters( 'wcship_user_can_manage_labels', current_user_can( 'manage_woocommerce' ) || current_user_can( 'wcship_manage_labels' ) );
```
harriswong, line 96: Curious on your thoughts regarding introducing a new permission versus re-using the existing `edit_shop_orders` (or combination of existing ones)? 

Is the `wcship_user_can_manage_labels` filter here simply to give other plugins a chance to manipulate WCS label permission check? Should we allow other plugins to affect how WCS check permissions? On one side, it enhances compatibility; on the other side, it seems to open up label permission control a bit much.

frosso: > Is the `wcship_user_can_manage_labels` filter here simply to give other plugins a chance to manipulate WCS label permission check?

That's correct, the intent is to make this flexible by either leveraging the new capability type _or_ by having a code snippet added by store owners (ideally) allow the change.

A lot of this discussion happened in this slack thread: p1623358576053300-slack-CKNC7KACA

We wouldn't want users with `edit_shop_orders` permission to print labels (since they could also change the plugin's configured packages via the shipping labels modal).
Another possibility is to have a wider refactor on the permission checks and the shipping modal to _not_ allow `edit_shop_orders` to change the plugin's configured packages 😬 

Let me know if you have other thoughts! :)

harriswong: Thanks for the slack link and I agree on the `edit_shop_orders` permission discussion. Unless we know exactly what permissions `edit_shop_orders` give, we should probably not use that.

harriswong (approved):
Tested by switching `manage_woocommerce` and adding `wcship_user_can_manage_labels`. LGTM 👍

PR #2439, opened 2021-06-30
refactor: remove jetpack dependency

### LINKED ISSUE #2654: Spike: Replace Jetpack requirement with Jetpack DNA
The goal is to replace the Jetpack plugin dependency with Jetpack Connection (a part of Jetpack DNA).

A POC was already created and tested under various scenarios by @frosso in PR #2439. The idea is to get this PR working and check what would be the level of effort to bring this task to completion.

This spike is a part of this P2 project: paJDVv-2g9-p2

## Spike outcomes

The list below was prepared by @laurajohnsonwoo.

1. What work can be repurposed?
2. What are the pros/cons/considerations of the approaches we can take? (And make a recommendation if you have one!)
3. Use Connection package in a way that is totally invisible to Woo users
5. Show My Jetpack UI in wp-admin
6. Define approach for implementation, including creating epic/issues to make it happen (we'll answer # 2 above before you get to #s 3-4)
8. Scope (LOE and timeline) implementation.
9. Coordinate with @adrianduffell and team to loosen the Jetpack dependency in the core onboarding task list.

frosso:
## Description

Replace the Jetpack plugin dependency with usage of the Jetpack Connection package.

- (optional) connect to your sandbox & apply D63626-code to see a modified jetpack DNA connection UI (rather than the default one) - this also ensures that the JP connection doesn't land on a 404 page upon successfull connection

Before:
Screen Shot 2021-07-01 at 4 10 19 PM

After:
Screen Shot 2021-07-01 at 4 41 09 PM

- Ensure you run `composer install`
- You can demo this locally or on JN by running `npm run build` and uploading the package to your JN installation
- Tested scenarios:
  - [x] Site with older version of Jetpack, then this is installed
  - [x] Site with older WCS & Jetpack connected, then upgrade to this => no need to connect
  - [x] Site with older WCS & Jetpack connected, then upgrade to this, then uninstall Jetpack => merchant is asked to connect site & plugin to Jetpack **PLEASE NOTE**: you'll need D63626-code
  - [x] Site with older WCS & Jetpack connected, then upgrade to this, then disconnect & uninstall Jetpack => merchant is asked to connect site to Jetpack **PLEASE NOTE**: you'll need D63626-code
  - [x] Brand new site with WCPay installed and connected, then merchant installs this => no need to connect to Jetpack
  - [x] Brand new site with this installed and connected, then WCPay is installed => merchant skips Jetpack connection, goes to Stripe onboarding
  - [x] Brand new site without Jetpack, then install this => merchant is asked to connect site to Jetpack
- Untested scenarios:
  - [ ] Brand new site, merchant installs WC and on onboarding also opt-in to install WCS&T
    - Should we modify WC onboarding to no longer require Jetpack?
      - The modification of the onboarding flows is happening in https://github.com/woocommerce/woocommerce/pull/39992 and is working when locally building both plugins and then installing on a test site. See: paJDVv-2JR-p2 
  - [ ] Atomic sites (shouldn't be an issue with the connection, the atomic site checks are just used to send data to the connect server)

### Checklist

- [x] unit tests


classes/class-wc-connect-api-client.php
```diff
@@ -466,7 +466,7 @@ protected function request_body( $initial_body = array() ) {
 					'dimension_unit'       => strtolower( get_option( 'woocommerce_dimension_unit' ) ),
 					'weight_unit'          => strtolower( get_option( 'woocommerce_weight_unit' ) ),
 					'wcs_version'          => WC_Connect_Loader::get_wcs_version(),
-					'jetpack_version'      => JETPACK__VERSION,
+					'jetpack_version'      => defined( 'JETPACK__VERSION' ) ? JETPACK__VERSION : 'embed',
```
frosso: It doesn't look like we're doing anything on the server with this string. However, the server is still requiring it in the payload.

classes/class-wc-connect-api-client.php
```diff
@@ -529,7 +529,7 @@ protected function authorization_header() {
 			}
 
 			list( $token_key, $token_secret ) = explode( '.', $token->secret );
-			$token_key                        = sprintf( '%s:%d:%d', $token_key, JETPACK__API_VERSION, $token->external_user_id );
+			$token_key                        = sprintf( '%s:%d:%d', $token_key, defined( 'JETPACK__API_VERSION' ) ? JETPACK__API_VERSION : 'embed', $token->external_user_id );
```
frosso: The server seems ok with this 🤷

classes/class-wc-connect-debug-tools.php
```diff
@@ -28,7 +28,7 @@ function woocommerce_debug_tools( $tools ) {
 		function test_connection() {
 			$test_request = $this->api_client->auth_test();
 			if ( $test_request && ! is_wp_error( $test_request ) && $test_request->authorized ) {
-				echo '<div class="updated inline"><p>' . __( 'Your site is succesfully communicating to the WooCommerce Shipping & Tax API.', 'woocommerce-services' ) . '</p></div>';
+				echo '<div class="updated inline"><p>' . __( 'Your site is successfully communicating to the WooCommerce Shipping & Tax API.', 'woocommerce-services' ) . '</p></div>';
```
frosso: typo

classes/class-wc-connect-help-view.php
```diff
@@ -102,14 +85,19 @@ protected function get_health_items() {
 					'state'   => 'warning',
 					'message' => __( 'This is a Jetpack staging site', 'woocommerce-services' ),
 				);
-			} else {
+			} elseif ( defined( 'JETPACK__VERSION' ) ) {
```
frosso: Does this really matter? IDK

woocommerce-services.php
```diff
@@ -708,7 +722,6 @@ public function load_dependencies() {
 			$shipping_label        = new WC_Connect_Shipping_Label( $api_client, $settings_store, $schemas_store, $payment_methods_store );
 			$nux                   = new WC_Connect_Nux( $tracks, $shipping_label );
 			$taxjar                = new WC_Connect_TaxJar_Integration( $api_client, $taxes_logger, $this->wc_connect_base_url );
-			$options               = new WC_Connect_Options();
```
frosso: There's no constructor in this class

woocommerce-services.php
```diff
@@ -34,30 +34,21 @@
...
 	define( 'WOOCOMMERCE_CONNECT_MINIMUM_WOOCOMMERCE_VERSION', '2.6' );
-	define( 'WOOCOMMERCE_CONNECT_MINIMUM_JETPACK_VERSION', '7.5' );
 	define( 'WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH', 32 );
 
 	if ( ! defined( 'WOOCOMMERCE_CONNECT_SERVER_API_VERSION ' ) ) {
 		define( 'WOOCOMMERCE_CONNECT_SERVER_API_VERSION', '5' );
 	}
 
-	// Check for CI environment variable to trigger test mode.
-	if ( false !== getenv( 'WOOCOMMERCE_SERVICES_CI_TEST_MODE' ) ) {
```
frosso, line 54: When `WOOCOMMERCE_SERVICES_CI_TEST_MODE` is defined and `!false`, the other constants are already defined as `true` - this simplifies things a bit

woocommerce-services.php
```diff
@@ -34,30 +34,21 @@
...
+require_once __DIR__ . '/vendor/autoload_packages.php';
+
 require_once __DIR__ . '/classes/class-wc-connect-extension-compatibility.php';
 require_once __DIR__ . '/classes/class-wc-connect-functions.php';
 require_once __DIR__ . '/classes/class-wc-connect-jetpack.php';
 require_once __DIR__ . '/classes/class-wc-connect-options.php';
 
 if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 	define( 'WOOCOMMERCE_CONNECT_MINIMUM_WOOCOMMERCE_VERSION', '2.6' );
-	define( 'WOOCOMMERCE_CONNECT_MINIMUM_JETPACK_VERSION', '7.5' );
```
frosso, line 46: Not needed anymore

tasks/release.js
```diff
@@ -15,6 +15,7 @@ const dirsToCopy = [
 	'dist',
 	'i18n',
 	'images',
+	'vendor',
```
frosso, line 18: Used to release - maybe we could just include the non-dev packages in the release?

composer.json
```diff
@@ -13,5 +13,10 @@
 		"phpcbf": [
 			"phpcbf -p"
 		]
+	},
+    "require": {
+		"automattic/jetpack-connection": "1.20.0",
```
frosso: These are not the latest versions.
The latest version of the JP packages has minimum compatibility with WP 5.5 - but we're still testing for 5.4

https://github.com/Automattic/jetpack/pull/17820

classes/class-wc-connect-help-view.php
frosso: I'm going to say no.

woocommerce-services.php
```diff
@@ -282,14 +276,9 @@ public static function get_wcs_admin_style_url() {
...
 
-		function wpcom_static_url( $file ) {
-			$i   = hexdec( substr( md5( $file ), -1 ) ) % 2;
-			$url = 'http://s' . $i . '.wp.com' . $file;
-			return set_url_scheme( $url );
-		}
-
 		public function __construct() {
 			$this->wc_connect_base_url = self::get_wc_connect_base_url();
+			add_action( 'plugins_loaded', array( $this, 'jetpack_on_plugins_loaded' ), 1 );
```
frosso, line 288: It's important that this code is executed with priority `1` - I banged my head for a while trying to figure out why it wasn't working before 😄

classes/class-wc-connect-tracks.php
```diff
@@ -66,10 +66,6 @@ public function saved_service_settings( $service_id ) {
 		}
 
 		public function record_user_event( $event_type, $data = array() ) {
-			if ( ! function_exists( 'jetpack_tracks_record_event' ) && ! class_exists( 'Automattic\\Jetpack\\Tracking' ) ) {
```
frosso, line 69: Not used anymore - `Automattic\Jetpack\Tracking` will always be defined

.github/workflows/e2e_tests.yml
```diff
@@ -13,9 +13,16 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
+      - uses: shivammathur/setup-php@v2
```
frosso, line 18: The plugin now has composer dependencies for releasing - we need them when executing the tests

classes/class-wc-connect-debug-tools.php
frosso: https://github.com/Automattic/woocommerce-services/pull/2444

classes/class-wc-connect-help-view.php
```diff
@@ -73,26 +73,8 @@ protected function get_health_items() {
 			// Only one of the following should present
 			// Check that Jetpack is active
 			// Check that Jetpack is connected
-			include_once ABSPATH . 'wp-admin/includes/plugin.php'; // required for is_plugin_active
 			$is_connected = WC_Connect_Jetpack::is_active() || WC_Connect_Jetpack::is_development_mode();
-			if ( ! is_plugin_active( 'jetpack/jetpack.php' ) ) {
```
frosso, line 87: No need for all this branching logic anymore - checking whether jetpack is installed or not won't matter, we are now using the composer packages directly.

classes/class-wc-connect-nux.php
```diff
@@ -677,61 +631,38 @@ class="woocommerce-services__connect-jetpack wcs-nux__notice-content-button butt
...
-				<?php
+			<?php
 			endif;
 		}
 
 		/**
-		 * Activates Jetpack after an ajax request
+		 * Connects the site to Jetpack.
 		 */
-		public function ajax_activate_jetpack() {
```
frosso, line 698: No need to activate jetpack anymore

classes/class-wc-connect-api-client.php
DanReyLop: I think it was only ever used for troubleshooting. If it's easy to get the version of the `jetpack-connection` package I would include it, otherwise `embed` is fine.

woocommerce-services.php
```diff
@@ -514,6 +500,33 @@ public function load_textdomain() {
 			load_plugin_textdomain( 'woocommerce-services', false, dirname( plugin_basename( __FILE__ ) ) . '/i18n/languages' );
 		}
 
+		public function jetpack_on_plugins_loaded() {
+			if ( false === WC_Connect_Jetpack::is_jetpack_version_supported() ) {
```
DanReyLop: I'm not sure that this is needed at all. Even if an old version of Jetpack is installed, the Autoloader will smartly load the newest version of `jetpack-connection` (the one bundled with WCS&T).

tasks/release.js
DanReyLop: Absolutely. The `dev` packages add **a lot** of bloat. [Here's how we do it](https://github.com/Automattic/woocommerce-payments/blob/develop/package.json#L22) in WCPay, we make sure that the task that's run when we're preparing the release executes `composer install --no-dev`.

classes/class-wc-connect-jetpack.php
```diff
@@ -142,14 +104,12 @@ public static function build_connect_url( $redirect_url ) {
 		 * @param
 		 */
 		public static function tracks_record_event( $user, $event_type, $data ) {
-			if ( version_compare( JETPACK__VERSION, '7.5', '<' ) ) {
-				if ( function_exists( 'jetpack_tracks_record_event' ) ) {
-					return jetpack_tracks_record_event( $user, $event_type, $data );
-				}
-			} elseif ( class_exists( 'Automattic\\Jetpack\\Tracking' ) ) {
+			if ( class_exists( 'Automattic\\Jetpack\\Tracking' ) ) {
```
DanReyLop: Nitpick: `jetpack-tracking` is a dependency of `jetpack-connection`, so it will always be available. However, since it's being used explicitly, it should be required explicitly too (both in `composer.json` and in the `Jetpack\Config` initialization).

woocommerce-services.php
frosso: Oh, interesting, thanks! I saw the same check in WCPay and when I tested this alongside the Jetpack plugin, I noticed that the JP plugin's classes were loaded. I assumed that the plugin's classes were taking precedence.
But your comment aligns with https://github.com/Automattic/jetpack/tree/master/projects/packages/autoloader - so that's great to know! :) 

> This is a custom autoloader generator that uses a classmap to always load the latest version of a class.

classes/class-wc-connect-jetpack.php
frosso: I agree, thanks for pointing it out! Done in 6f4d648f

woocommerce-services.php
DanReyLop: > I saw the same check in WCPay

The autoloader hasn't always worked flawlessly :)

I remember that, if the Jetpack version is REALLY old, it could take precedence over the Composer dependency. That mattered last year, but at this point if the merchant is using a 2+ year old version of Jetpack I'm ok with it crashing.

DanReyLop: Here's the explanation, it's your call if you think avoiding a crash on JP 8.1 is worth it: https://github.com/Automattic/woocommerce-payments/blob/16c5a7c9b713fc937c46955ca33afcc27c74fda2/woocommerce-payments.php#L102-L104

DanReyLop:
I see this is still marked as a "POC". Everything I reviewed looks ready to ship. Are there any blockers?

waclawjacek:
@frosso WDYT about asking Codeable to give this one last review as it's been a while since this was published, and shipping this?

frosso:
@waclawjacek I would love to, but I received a bit of pushback a while ago since it would have a bigger impact on Jetpack: paJDVv-19O-p2

In order to have Codeable review these changes, we would also need to commit & deploy D63626-code

waclawjacek:
Converting to draft so hopefully our bots don't continue pinging us about this one.

waclawjacek:
Nope, didn't do it. Looks like the bot relies on labels so changing the label.

woocommerce-services.php
```diff
@@ -514,6 +500,17 @@ public function load_textdomain() {
 			load_plugin_textdomain( 'woocommerce-services', false, dirname( plugin_basename( __FILE__ ) ) . '/i18n/languages' );
 		}
 
+		public function jetpack_on_plugins_loaded() {
+			$jetpack_config = new Automattic\Jetpack\Config();
+			$jetpack_config->ensure(
+				'connection',
+				array(
+					'slug' => WC_Connect_Jetpack::JETPACK_PLUGIN_SLUG,
+					'name' => __( 'WooCommerce Shipping & Tax', 'woocommerce-payments' ),
```
harriswong: `'woocommerce-payments'` should probably be `'woocommerce-services'`

frosso:
```suggestion
					'name' => __( 'WooCommerce Shipping & Tax', 'woocommerce-services' ),
```
ope

woocommerce-services.php
waclawjacek: Context: https://github.com/Automattic/woocommerce-payments/pull/637/files#diff-ad243b7f0e0ae2cc54963fec106a5c54ece9c5bb8b272acaeece1836d71418a2R36

classes/class-wc-connect-jetpack.php
```diff
@@ -1,30 +1,27 @@
...
-				return $connection->get_access_token( $user_id );
-			}
-
-			if ( class_exists( '\Automattic\Jetpack\Connection\Manager' ) && method_exists( '\Automattic\Jetpack\Connection\Manager', 'get_access_token' ) ) {
-				$connection = new Manager();
-
-				return $connection->get_access_token( $user_id );
+		public static function get_access_token( $user_id = false ) {
+			// Standalone Jetpack 9.5+ installation
+			if ( class_exists( '\Automattic\Jetpack\Connection\Tokens' ) ) {
```
sergeymitr: We might not need this.
Since you're requiring a fresh `jetpack-connection` package, `Connection\Tokens` will always be there.
Even if an old Jetpack installed alongside, Autoloader will ensure we're using the latest available Connection package.

That also makes excessive the call to Manager::get_access_token() below.

classes/class-wc-connect-jetpack.php
```diff
@@ -35,6 +32,7 @@ public static function get_master_user_access_token( $user_id ) {
 		public static function is_development_mode() {
 			if ( method_exists( '\\Automattic\\Jetpack\\Status', 'is_offline_mode' ) ) {
```
sergeymitr: This check seems to be excessive now as well, the `Status` package is either there or not.
And since we're loading the Connection package, it'll always be loaded as a dependency: https://github.com/Automattic/jetpack/blob/trunk/projects/packages/connection/composer.json#L11

However, it might make sense to load it directly in the plugin's `composer.json` for clarity and to prevent potential incompatibilities in the future.

classes/class-wc-connect-jetpack.php
```diff
@@ -47,14 +45,7 @@ public static function is_development_mode() {
...
 		public static function is_active() {
-			if ( defined( 'WOOCOMMERCE_SERVICES_LOCAL_TEST_MODE' ) && WOOCOMMERCE_SERVICES_LOCAL_TEST_MODE ) {
-				return true;
-			}
-			if ( method_exists( 'Jetpack', 'is_active' ) ) {
-				return Jetpack::is_active();
-			}
-
-			return false;
+			return ! empty ( self::get_access_token( true ) );
```
sergeymitr: The code only checks for connection owner's "user token". I'd recommend to also check for "blog token" in case of some connection issues that get the "blog token" removed, but leave "user tokens" intact.

```suggestion
			return self::get_connection_manager()->is_connected() && self::get_connection_manager()->has_connected_owner();
```

classes/class-wc-connect-jetpack.php
```diff
@@ -110,30 +86,16 @@ public static function get_connected_user_data( $user_id ) {
 		 * @return WP_User | false
 		 */
 		public static function get_master_user() {
-			include_once ABSPATH . 'wp-admin/includes/plugin.php';
 			if ( self::is_active() && method_exists( 'Jetpack_Options', 'get_option' ) ) {
 				$master_user_id = Jetpack_Options::get_option( 'master_user' );
+
 				return get_userdata( $master_user_id );
 			}
```
sergeymitr: I'd recommend calling `Manager::get_connection_owner()` instead of pulling the data from options directly.

classes/class-wc-connect-jetpack.php
```diff
@@ -159,40 +115,57 @@ public static function tracks_record_event( $user, $event_type, $data ) {
...
-			if ( class_exists( '\Automattic\Jetpack\Connection\Manager' ) && method_exists( '\Automattic\Jetpack\Connection\Manager', 'is_connected' ) ) {
-				$connection = new Manager();
-
-				return $connection->is_connected();
-			}
+			$manager = self::get_connection_manager();
 
-			if ( defined( 'JETPACK_MASTER_USER' ) ) {
-				$user_token = self::get_master_user_access_token( JETPACK_MASTER_USER );
+			return $manager->is_plugin_enabled() && self::is_active() && ! $manager->is_missing_connection_owner();
```
sergeymitr: `is_plugin_enabled()` is a rudimentary piece left from the "Soft Disconnect" times, deprecated now.
`is_missing_connection_owner()` is also excessive if you check for `has_connected_owner()` inside `is_active()`, as they both check for the following:
- connection owner is set
- they have a user token
- WP user account actually exists for their userID

So it loooks like `is_connected()` and `is_active()` might be serving the same purpose, and deprecating one of them wouldn't hurt.

classes/class-wc-connect-jetpack.php
```diff
@@ -159,40 +115,57 @@ public static function tracks_record_event( $user, $event_type, $data ) {
...
+		/**
+		 * Connects the site to Jetpack.
+		 * This code performs a redirection, so anything executed after it will be ignored.
+		 *
+		 * @param $redirect_url
+		 */
+		public static function connect_site( $redirect_url ) {
+			// Mark the plugin as enabled in case it had been soft-disconnected.
+			$jetpack_connection_manager = self::get_connection_manager();
+			$jetpack_connection_manager->enable_plugin();
```
sergeymitr: `enable_plugin()` is also deprecated.

We used those `enable_plugin()`, `is_plugin_enabled()`, and `disable_plugin()` methods to "softly disconnect" plugins, meaning plugins could "disconnect" from WPCOM and leave the connection intact for all other connected plugins.
We no longer do that, as individual plugins cannot disconnect on their own (except for Jetpack), and "My Jetpack" is the main connection management place.
Here's some context: p9dueE-4n9-p2

classes/class-wc-connect-jetpack.php
```diff
@@ -159,40 +115,57 @@ public static function tracks_record_event( $user, $event_type, $data ) {
...
+			// Register the site to wp.com.
+			// standalone JP 9.2+ uses `is_connected`
+			$is_registered = false;
+			if ( method_exists( $jetpack_connection_manager, 'is_connected' ) ) {
+				$is_registered = $jetpack_connection_manager->is_connected();
+			} else {
+				$is_registered = $jetpack_connection_manager->is_registered();
+			}
+
+			if ( ! $is_registered ) {
```
sergeymitr: Since we "autoload" the latest Connection package, there's no need to double check if methods exist:
```suggestion
			if ( ! $jetpack_connection_manager->is_connected() ) {
```

classes/class-wc-connect-jetpack.php
```diff
@@ -159,40 +115,57 @@ public static function tracks_record_event( $user, $event_type, $data ) {
...
+				$result = $jetpack_connection_manager->register();
```
sergeymitr: `register()` would work fine, but `try_registration()` sets the "Agreed to TOS" flag, runs some hooks, and does some tracking, so I'd recommend using that instead.
```suggestion
				$result = $jetpack_connection_manager->try_registration();
```

classes/class-wc-connect-nux.php
```diff
@@ -290,28 +279,14 @@ public static function get_banner_type_to_display( $status = array() ) {
...
-			if ( 0 !== validate_plugin( 'jetpack/jetpack.php' ) ) {
-				return self::JETPACK_NOT_INSTALLED;
-			}
-
-			// check if Jetpack is activated
-			if ( ! is_plugin_active( 'jetpack/jetpack.php' ) ) {
-				return self::JETPACK_INSTALLED_NOT_ACTIVATED;
-			}
-
 			if ( defined( 'JETPACK_DEV_DEBUG' ) && true === JETPACK_DEV_DEBUG ) {
```
sergeymitr: There are some cases where we switch to Offline Mode without the constant set, so it might make sense to use `Status::is_offline_mode()` instead.

composer.json
```diff
@@ -14,9 +14,15 @@
 			"phpcbf -p"
 		]
 	},
+    "require": {
+		"automattic/jetpack-connection": "^1.51.9",
```
sergeymitr: Connection and Autoloader packages got new versions released recently, we could update them here.

sergeymitr (changes requested):
Impressive work, just a few minor details I noticed.

classes/class-wc-connect-api-client.php
waclawjacek: Yep, just checked and we are not using it anywhere [aside from Tracks calls](https://github.com/Automattic/woocommerce-connect-server/blob/32608bf9fffe39abaaa40826d3d06c537948c086/lib/analytics/tracks.js#L73). We [require it to be a string](https://github.com/Automattic/woocommerce-connect-server/blob/32608bf9fffe39abaaa40826d3d06c537948c086/lib/settings/validator.js#L18) **if** it is present.

I've changed this from `embed` to `embed-{jetpack_connection_package_version}`.

classes/class-wc-connect-api-client.php
waclawjacek: The format is `%s:%d:%d` so the second and third placeholders should be numbers (i.e. can't be "embed"). Testing on my local, it looks like the code below returns `1` anyway so I suggest we hardcode it to that, and in the future use Jetpack's header building function instead of our own, if possible.

```php
Constants::get_constant( 'JETPACK__API_VERSION' )
```

classes/class-wc-connect-jetpack.php
waclawjacek: Great ideas!

classes/class-wc-connect-help-view.php
```diff
@@ -78,49 +78,28 @@ protected function get_health_items() {
...
-				$health_item = array(
-					'state'   => 'error',
-					'message' => sprintf(
-						__( 'Please install and activate the Jetpack plugin, version %s or higher', 'woocommerce-services' ),
-						WOOCOMMERCE_CONNECT_MINIMUM_JETPACK_VERSION
-					),
-				);
-			} elseif ( version_compare( JETPACK__VERSION, WOOCOMMERCE_CONNECT_MINIMUM_JETPACK_VERSION, '<' ) ) {
+			if ( WC_Connect_Jetpack::is_offline_mode() ) {
```
waclawjacek, line 81: Adding offline mode as a separate possible status because this mode can be enabled when a dev mode is on (using `JETPACK_DEV_DEBUG` or `WP_LOCAL_DEV`) but also when simply running the site on a local TLD like `.test` or `.local`.

tasks/release.js
waclawjacek: Done in 573a36379a0555908681a70320d946373835b59d

classes/class-wc-connect-nux.php
```diff
@@ -261,12 +232,9 @@ public static function get_banner_type_to_display( $status = array() ) {
 				This is an existing user. Do nothing.
 			*/
 			switch ( $status['jetpack_connection_status'] ) {
-				case self::JETPACK_NOT_INSTALLED:
-				case self::JETPACK_INSTALLED_NOT_ACTIVATED:
-				case self::JETPACK_ACTIVATED_NOT_CONNECTED:
+				case self::JETPACK_NOT_CONNECTED:
 					return 'before_jetpack_connection';
 				case self::JETPACK_CONNECTED:
-				case self::JETPACK_DEV:
```
waclawjacek, line 269: I don't think we should present either the `after_jetpack_connection` or the `tos_only_banner` on local or dev sites (i.e. in offline mode) so I am removing this case.

classes/class-wc-connect-nux.php
```diff
@@ -216,34 +215,6 @@ public function register_new_carrier_dhl_pointer( $pointers ) {
...
-
-			if (
-				( self::JETPACK_NOT_INSTALLED === $jetpack_status ) ||
-				( self::JETPACK_INSTALLED_NOT_ACTIVATED === $jetpack_status )
-			) {
-				return false;
-			}
-
-			// Developer case
-			if ( self::JETPACK_DEV === $jetpack_status ) {
```
waclawjacek, line 238: Since the `JETPACK_NOT_INSTALLED` and `JETPACK_INSTALLED_NOT_ACTIVATED` statuses are no longer possible as Jetpack Connection is bundled with WCS&T, the only options left would be `JETPACK_DEV` or the result of `is_current_user_connected()`.

As we are using Jetpack Status' `is_offline_mode()` now to determine if a site should be in offline (a.k.a. dev) mode, but the offline mode can also be enabled by simply running the site locally, I am removing this method and instead inlining the `is_current_user_connected()` call in `set_up_nux_notices()` which is the only place it is used.

This is because I want to avoid presenting folks working on the site in local environments with the option to accept the TOS on behalf of the connection owner.

waclawjacek:
I have finished implementing my changes to this PR. @Automattic/escargot It's now ready for review.

waclawjacek:
@sergeymitr In the past you've shared some great insights into how to use Jetpack Connection. Could you please check if things look good?

changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.4.0 - 2023-xx-xx =
+* Add - Ability to connect to the server infrastructure without the Jetpack plugin.
```
waclawjacek:
```suggestion
* Add - Ability to connect to WordPress.com without the Jetpack plugin.
```

classes/class-wc-connect-api-client-live.php
```diff
@@ -17,20 +17,6 @@ class WC_Connect_API_Client_Live extends WC_Connect_API_Client {
 		protected function request( $method, $path, $body = array() ) {
 
 			// TODO - incorporate caching for repeated identical requests
-			if ( ! class_exists( '\Automattic\Jetpack\Connection\Manager' ) && ! class_exists( '\Automattic\Jetpack\Connection\Tokens' ) ) {
```
waclawjacek, line 20: These classes are always present because we bundle Jetpack Connection with WCS&T now.

classes/class-wc-connect-nux.php
```diff
@@ -216,34 +215,6 @@ public function register_new_carrier_dhl_pointer( $pointers ) {
...
 
-		/**
-		 * Check that the current user is the owner of the Jetpack connection
-		 * - Only that person can accept the TOS
-		 *
-		 * @uses self::get_jetpack_install_status()
-		 *
-		 * @return bool
-		 */
-		public function can_accept_tos() {
```
waclawjacek, line 227: Since the logic now boils down to "is offline mode or is current user connected", I am inlining this logic in `set_up_nux_notices()` instead.

client/banner.js
```diff
@@ -12,78 +9,8 @@ import jQuery from 'jquery';
...
 jQuery( document ).ready( ( $ ) => {
-	$( '.woocommerce-services__connect-jetpack' ).one( 'click', function( event ) {
-		event.preventDefault();
+	$( '.woocommerce-services__connect-jetpack' ).one( 'click', function() {
 		const btn = $( this );
 		btn.addClass( 'disabled' );
-
-		installStep()
-			.then( activateStep )
-			.then( connectStep )
```
harriswong, line 22: We don't need any of these steps on the banner anymore?

classes/class-wc-connect-tracks.php
```diff
@@ -100,11 +86,11 @@ public function record_user_event( $event_type, $data = array() ) {
...
 			$data['_via_ip']         = isset( $_SERVER['REMOTE_ADDR'] ) ? wc_clean( wp_unslash( $_SERVER['REMOTE_ADDR'] ) ) : '';
 			$data['_lg']             = isset( $_SERVER['HTTP_ACCEPT_LANGUAGE'] ) ? wc_clean( wp_unslash( $_SERVER['HTTP_ACCEPT_LANGUAGE'] ) ) : '';
-			$data['blog_url']        = $site_url;
+			$data['blog_url']        = get_option( 'siteurl' );
 			$data['blog_id']         = $jetpack_blog_id;
-			$data['wcs_version']     = $wcs_version;
-			$data['jetpack_version'] = $jp_version;
-			$data['is_atomic']       = $is_atomic;
+			$data['wcs_version']     = WC_Connect_Loader::get_wcs_version();
+			$data['jetpack_version'] = 'embed-' . WC_Connect_Jetpack::get_jetpack_connection_package_version();
```
harriswong, line 92: What does the `embed-` prefix do?

harriswong (commented):
I have some questions but LGTM otherwise.

I tested buying and printing a USPS label, works.

I tested 2 nux flows by hacking around this condition:
```

			$banner_to_display = self::get_banner_type_to_display(
				array(
					'jetpack_connection_status'       => $jetpack_install_status,
					'tos_accepted'                    => WC_Connect_Options::get_option( 'tos_accepted' ),
					'can_accept_tos'                  => WC_Connect_Jetpack::is_current_user_connection_owner() || WC_Connect_Jetpack::is_offline_mode(),
					'should_display_after_cxn_banner' => WC_Connect_Options::get_option( self::SHOULD_SHOW_AFTER_CXN_BANNER ),
				)
			);
```

The ToS one looks good and I traced the code a bit, looks good to me. 
image

The second nux flow gave me an error:
image
image

client/banner.js
waclawjacek: Nope! In the past the flow was AJAX-driven and led through the installation, activation, and connection via Jetpack. Now that we no longer depend on Jetpack, all we do is a single redirect.

classes/class-wc-connect-tracks.php
waclawjacek: This is just something to distinguish values from back when this used to be the version of Jetpack-the-plugin in use and now that it is the version of the embedded Jetpack Connection package. The prefix itself doesn't do anything.

waclawjacek:
@harriswong Thank you so much for the review! It's hard to say why exactly you got the error. Is there any chance of getting a list of steps to reproduce to achieve that result? It might be something not caused by this PR.

harriswong:
I think it could be me setting the following:
```
array(
					'jetpack_connection_status'       => 'not-connected',
					'tos_accepted'                    => false, //WC_Connect_Options::get_option( 'tos_accepted' ),
					'can_accept_tos'                  => false, //WC_Connect_Jetpack::is_current_user_connection_owner() || WC_Connect_Jetpack::is_offline_mode(),
					'should_display_after_cxn_banner' => WC_Connect_Options::get_option( self::SHOULD_SHOW_AFTER_CXN_BANNER ),
				)
```

We can upload this plugin to a brand new site, and if that is fine, then we should be good! Have you tested that already? If so, I think we are good.

harriswong:
👋 @waclawjacek For sift, I added these 2 lines:
https://github.com/Automattic/woocommerce-services/blob/trunk/woocommerce-services.php#L1605
https://github.com/Automattic/woocommerce-services/blob/trunk/woocommerce-services.php#L1610

I think we will have to rename these. My recent changelog also conflicted with your PR. Let me know if you want me to fix these! Thanks.

waclawjacek:
> 👋 @waclawjacek For sift, I added these 2 lines: [`trunk`/woocommerce-services.php#L1605](https://github.com/Automattic/woocommerce-services/blob/trunk/woocommerce-services.php?rgh-link-date=2023-10-24T16%3A58%3A54Z#L1605) [`trunk`/woocommerce-services.php#L1610](https://github.com/Automattic/woocommerce-services/blob/trunk/woocommerce-services.php?rgh-link-date=2023-10-24T16%3A58%3A54Z#L1610)
> 
> I think we will have to rename these. My recent changelog also conflicted with your PR. Let me know if you want me to fix these! Thanks.

Thank you! Great catch. I've just fixed these.

> I think it could be me setting the following:
> 
> ```
> array(
> 					'jetpack_connection_status'       => 'not-connected',
> 					'tos_accepted'                    => false, //WC_Connect_Options::get_option( 'tos_accepted' ),
> 					'can_accept_tos'                  => false, //WC_Connect_Jetpack::is_current_user_connection_owner() || WC_Connect_Jetpack::is_offline_mode(),
> 					'should_display_after_cxn_banner' => WC_Connect_Options::get_option( self::SHOULD_SHOW_AFTER_CXN_BANNER ),
> 				)
> ```

I don't think that's it. Here's the line of code reporting that error that I found in OpenGrok: fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Szh%2Qcyhtvaf%2Swrgcnpx%2Spynff.wrgcnpx%2Qfreire%2Qirefvba.cuc%3Se%3Qp86809os%26zb%3Q43279%26sv%3Q1292%231292-og I've just tried hardcoding the same arguments that you tried locally and I was able to connect my site (with the exception of having to remove this later to get the NUX banner to disappear :)). Could you please retry this on a fresh site?

> We can upload this plugin to a brand new site, and if that is fine, then we should be good! Have you tested that already? If so, I think we are good.

Yep! In so many test cases. 😄 Also just tested now. I'm attaching a build of the latest commit in this PR if you'd like to test on JN: 

[woocommerce-services-702256c7de757357fd36b93c040406a383502e69.zip](https://github.com/Automattic/woocommerce-services/files/13165493/woocommerce-services-702256c7de757357fd36b93c040406a383502e69.zip)

harriswong:
Tested on JN and there is no error, woot!

I have one question. After clicking the image
image

It bounces me to Jetpack page, is this expected? I was thinking the authentication should just connects the site but not bring to this page.
image

waclawjacek:
Thanks for checking! Did you apply the sandbox patch mentioned in the original post? Without it, you go through the old auth flows instead.

waclawjacek:
We've just merged the WPCOM changes so you can now test without sandboxing. Could you please give it a try?

harriswong:
Hi @waclawjacek, I gave this another test 
🧪 With staging
✅ print USPS label + refund using staging, reprinting
✅ international shipping DHL + refund. Reprinting + reprinting custom form 
✅ First onboarding nux flow
✅ Saving settings, credit card updates
✅ Refreshing connect server statuses

🧪 Without staging, using a new build with the latest commit(1d90a663ea7014f73c4f40e6c3439fdd91492a3c)
✅ Installing WCS&T 
✅ Activating and the NUX onboarding
✅ Tested creating an order, all the way to printing a label but without actually buying it with a real credit card.

🧪 I didn't test a **fresh** local dev setup env. 

🧪 Tested locally
✅ Pulled this branch, `npm run start` still works with Jetpack installed.

PR #2370, opened 2021-03-10
refactor: calypso form-select

frosso:
## Description

Replacing the `form-select` component from Calypso.
I copied the styles and the body from Calypso, making it a functional component.
Unfortunately, the `@wordpress/components` component doesn't allow `optgroup` children.
I have a PR open for that: https://github.com/WordPress/gutenberg/pull/29540
Moreover, the Calypso styles are targeting _all_ `select` elements, causing some visual artifacts when using the `SelectControl` from `@wordpress/components`: https://d.pr/i/iYhj0G
In the meantime, the team agreed to just copy the styles/component to get rid of the Calypso submodule.

Original: https://github.com/Automattic/wp-calypso/blob/df4ddfdc94438b6d31b686d6415e08a567570944/client/components/forms/form-select/index.jsx

You might think the Calypso styles are all in here: https://github.com/Automattic/wp-calypso/blob/df4ddfdc94438b6d31b686d6415e08a567570944/client/components/forms/form-select/style.scss
But you'd be wrong, there's more from here: https://github.com/Automattic/wp-calypso/blob/df4ddfdc94438b6d31b686d6415e08a567570944/assets/stylesheets/shared/_forms.scss

### Related issue(s)

Part of https://github.com/Automattic/woocommerce-services/issues/2333

### Steps to reproduce & screenshots/GIFs

This component is used in many places.
A few examples:
- Account connection page: https://d.pr/i/9stEFr
- WCS settings: https://d.pr/i/2lvK6c
- Label creation: https://d.pr/i/2jzaxG | https://d.pr/i/MbjoRm | https://d.pr/i/PE1Aq2 | https://d.pr/i/Uy3eai
- Settings page (currently broken, but https://github.com/Automattic/woocommerce-services/pull/2369 fixes it)

### Checklist

- [ ] unit tests

- [ ] `changelog.txt` entry added

PR #2441, opened 2021-06-30
Move FoldableCard component out from calypso submodule

harriswong:
## Description
Part of effort from https://github.com/Automattic/woocommerce-services/issues/2333, this PR moves `FoldableCard` out from the calypso's submodule. This PR includes the following changes:
- Added a new `client/components/foldable-card, a copy from the calypso submodule
- Added a new `client/components/screen-reader-text, a copy from the calypso submodule

### Related issue(s)

1. Make sure "Foldable Card" works. There are two types, compact, and not compact. 
2. To test, go to an order and create an label, the "step" should open/close properly: 
image
3. Go to WooCommerce Services settings (`wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings`), click "Add package" under "Packaging", these card should open/close properly:
image

### Checklist

- [ ] unit tests


client/components/foldable-card/index.js
```diff
@@ -0,0 +1,161 @@
...
+ import PropTypes from 'prop-types';
+ import React, { Component } from 'react';
+ import classNames from 'classnames';
+ import { localize } from 'i18n-calypso';
+ import { noop } from 'lodash';
+
+ /**
+  * Internal Dependencies
+  */
+ import { Card } from '@wordpress/components';
```
harriswong: The current submodule's code uses `CompactCard` and `Card`, I changed this to use WordPress' components instead. 

The current submodule: https://github.com/Automattic/wp-calypso/blob/df4ddfdc94438b6d31b686d6415e08a567570944/client/components/foldable-card/index.jsx#L16-L17

client/components/foldable-card/index.js
```diff
@@ -0,0 +1,161 @@
...
+             const iconSize = 24;
+             const screenReaderText = this.props.screenReaderText || this.props.translate( 'More' );
+             return (
+                 <button
+                     disabled={ this.props.disabled }
+                     type="button"
+                     className="foldable-card__action foldable-card__expand"
+                     onClick={ clickAction }
+                 >
+                     <ScreenReaderText>{ screenReaderText }</ScreenReaderText>
```
harriswong: I debated whether or not to copy over Calypso's `<ScreenReaderText`> component. I ended up copying it because:
- it has less change compared to the existing submodule
- other Calypso components may also need it
- we will clean up this entire `client/components` folder so then it's not introducing long term tech debt.

brezocordero: `<ScreenReaderText>` has a different name on the WordPress components, it is called [VisuallyHidden](<ScreenReaderText>). 
This is [the issue where the component was added to Gutenberg and why the name was changed for alignment with other libraries](https://github.com/WordPress/gutenberg/pull/18022).

I am ok with copying it over, it is small and [the latest version on Calypso](https://github.com/Automattic/wp-calypso/tree/trunk/packages/components/src/screen-reader-text) is the same as our copy.

We can also create a separate issue to replace `<ScreenReaderText>`

client/components/foldable-card/index.js
brezocordero: 👍

client/components/foldable-card/index.js
```diff
@@ -0,0 +1,161 @@
...
+                     { summary }
+                     { expandedSummary }
+                     { this.renderActionButton() }
+                 </span>
+             </div>
+         );
+     }
+
+     render() {
+         const itemSiteClasses = classNames( 'foldable-card', this.props.className, {
```
brezocordero:
```suggestion
         const itemSiteClasses = classNames( 'foldable-card', 'card', this.props.className, {
```
The styles are a bit different when using the WordPress card. 
NewFoldable
CurrentFoldable


[On this PR](https://github.com/Automattic/woocommerce-services/pull/2337), we copied over the Calypso CSS for Card and then pass it to the WordPress Card. A bit redundant but it doesn't change the display.

client/components/foldable-card/style.scss
```diff
@@ -0,0 +1,183 @@
+// Multisite
+.foldable-card.card {
+	@include clear-fix;
+	position: relative;
+	transition: margin 0.15s linear;
+	padding: 0;
+
```
brezocordero, line 8:
```suggestion
border-radius: 0;
```
By default, the WordPress card is rounded and it looks wavy when we have multiple cards. 

RoundedCorners

The newest version allows you to disable it with [isRounded={ false }](https://github.com/WordPress/gutenberg/blob/fe77cd3c7f9df1a5ac0a273097f7eb05282e638e/packages/components/src/card/card/component.js#L33) but that is not available yet.
Please take a second look in case adding the CSS line changes something else.

client/extensions/woocommerce/app/order/order-activity-log/day.js
```diff
@@ -9,7 +9,7 @@ import React, { Component } from 'react';
 /**
  * Internal dependencies
  */
-import FoldableCard from 'components/foldable-card';
+import FoldableCard from 'wcs-client/components/foldable-card';
```
brezocordero, line 12: Where can I test this?

brezocordero (changes requested):
@harriswong works pretty well!
A couple of small comments to make the UI similar to what we have now.

client/extensions/woocommerce/app/order/order-activity-log/day.js
harriswong: This was introduced in https://github.com/Automattic/woocommerce-services/pull/2370#discussion_r591760370. If we want to test whether or not webpack found the right file (`client/components/foldable-card/index.js`), we could add a random class name in one of the divs and see if it renders.

client/components/foldable-card/index.js
harriswong: If there is a WordPress component for this, maybe we should switch over to it. If the goal is to remove Calypso components completely, then I think it is beneficial to have as little calypso component dependency as possible.

client/extensions/woocommerce/app/order/order-activity-log/day.js
brezocordero: Thanks! Sorry my question was not clear. Where in the UI do I find the order-activity-log/day.js to test the changes?

client/components/foldable-card/index.js
harriswong: Thanks! Updated.

client/components/foldable-card/style.scss
harriswong: Added the `radius:0 change`.

client/components/foldable-card/index.js
harriswong: Update: Removed ScreenReaderText component and use VisuallyHidden component instead.

client/extensions/woocommerce/app/order/order-activity-log/day.js
harriswong: I believe it is the notes on the order page itself: 
image

harriswong:
@brezocordero I updated the PR with the following changes, please take another look again.
- Removed round corners for the cards
- Removed `ScreenReaderText` 
- Use `VisuallyHidden` component instead.

brezocordero (approved):
Thanks for making all the changes!
LGTM!
:shipit:

PR #2361, opened 2021-03-08
Replace/calypso buttons

brezocordero:
## Description

Replace Calypso Button component with WordPress Button component.
The styles from Calypso have been copied over, to maintain the same design.

Part of #2333

The following buttons have been replaced:

**WooCommerce Shipping Settings screen:**
Choose a different card link and loading button:
ChooseCard
Packages: Add Package, Remove and Edit Buttons:
Packages
Carriers: Connect and Disconnect:
Carriers
Subscriptions: Activate and Manage.
Save Settings.
SubsAndSave

Carriers Forms:
UPS:
UPSConnectCancel
UPSCancelConnection
DHL Form Connect and Cancel.
DHLConnectCancel

**Order Screen:**
Create Label (Pink), Create Label and Track Package (Grey)
CreateLabel

CreateAndTrack

Purchase Label Form
Add packages, select Package:
AddItemsSelectPackage

Purchase Button
BuyLabel

Note: `Purchase` button becomes `Add Credit Card` button when there is no payment selected. 

**Shipping method Screen**
ShippingMethodSave

### Checklist

- [x] unit tests

- [ ] `changelog.txt` entry added

PR #2443, opened 2021-07-01
Move Calypso's FormButton out into our own component folder

harriswong:
## Description
Similar to https://github.com/Automattic/woocommerce-services/pull/2344, There is no comparable WordPress nor WooCommerce `FormButton`. I moved the Calypso submodule's `FormButton` over.

I searched the repo for `<FormButton` and there are 3 props we used that could be different than `@wordpress/components`: `isPrimary`, `borderless`, disabled`.

I made some changes to the `<FormButton>` component after copying it over from our Calypso submodule. I  have tested these on the purchasing steps modal. 

### Related issue(s)

1. Pull this branch, run `npm run start` to recompile the assets
2.  Go to an order, click "Create new label" 
3. Make sure the "Verify address" button works. 

### Checklist

- [ ] unit tests


harriswong:
Inside the `FormButton` component, I'm looking into updating `import Button from 'components/button';` to `import { Button } from '@wordpress/components';`. This would reduce the dependency of the Calypso's button component. 

 Moving this back to "in progress".

client/components/forms/form-button/index.js
```diff
@@ -0,0 +1,48 @@
...
+	getDefaultButtonAction = () => {
+		return this.props.isSubmitting
+			? this.props.translate( 'Saving…' )
+			: this.props.translate( 'Save Settings' );
+	};
+
+	render() {
+		const { children, className, isPrimary, ...props } = this.props,
+			buttonClasses = classNames( className, 'form-button', 'button', {
+				'is-borderless': !! this.props.borderless
```
harriswong: Added handling for `borderless` prop that we use in a few cases. Using `is-borderless` similar to https://github.com/Automattic/woocommerce-services/pull/2361/files#diff-dc83733ab281e3cf1cb35835dbedfc2abdd540bf59d78e0f4e4e55863c3edab4R229.

client/components/forms/form-button/style.scss
```diff
@@ -0,0 +1,8 @@
+.form-button {
+	float: right;
+	margin-left: 10px;
+
+	&:first-child {
+		margin-left: 0;
```
harriswong, line 6: Added this CSS to remove margin for first button if they are side by side.

waclawjacek:
LGTM! Except for this tiny quirk when reloading the WCS&T settings page:

https://user-images.githubusercontent.com/1759681/219782061-ed4d0835-5615-425b-9685-535b274d8ae3.mp4

waclawjacek (changes requested):
LGTM except for a tiny thing mentioned above.

harriswong:
While looking into the CSS issue with the button, I concluded that we don't need a clickable button for placeholders. Reasons are as follow:

1. During rendering, users shouldn't be allowed to click on "Save settings"
2. For the purpose of rendering placeholders, we don't really need to show the button.
3. The "Save settings" placeholder button is in the wrong place. It is not at the bottom of the screen. It's confusing to have this fake button there.

I have removed it. @waclawjacek What do you think?
image

waclawjacek (approved):
:shipit: :shipit: :shipit:

PR #2445, opened 2021-07-02
deps: upgrade husky

frosso:
## Description

I upgraded the `husky` package.
During the upgrade, Husky moved the `npm run test` step to the `pre-commit` hook. I agree with that.

I didn't add a changelog entry because it's just dependencies.

### Checklist

- [ ] unit tests


package.json
```diff
@@ -155,11 +155,5 @@
   },
   "config": {
     "wp_org_slug": "woocommerce-services"
-  },
-  "husky": {
```
frosso, line 164: No longer necessary. See https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v7

harriswong:
👀

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx --no-install lint-staged && npm test
```
harriswong: +1 on lint per-commit, but I personally prefer running test per push.

I use commit to commit a "work in progress" history; I use push to indicate "this is ready". If a PR is "work in progress", it may fail certain tests in some commits. These intermediate commits may not require fixing the tests yet because the work is not finished.

Lint per commit has its value because we can omit the "fix lint" commits. Also, fixing lints is a much easier effort than fixing tests. 

I think having a per-commit rule for `npm run test` will slow down development time, defeating the purpose of "commit often". 

I would like to hear what others think as well.

harriswong (commented):
Tested `npm i` on this branch. Working as expected. 👍 

Just one comment on running test per commit. Would like to hear what others think as well.

.husky/pre-commit
frosso: Thanks for the input - I also try to commit often, but I do those intermediate commits with the `-n`/`--no-verify` flag.

What would your thoughts be on pushing without running the tests?
I was going through `woocommerce/woocommerce-payments`, `woocommerce/woocommerce`, `woocommerce/woocommerce-admin` and `Automattic/jetpack`, `WordPress/gutenberg`, they all don't run the tests on pre-commit nor on pre-push.
I would be kinda sad maybe to not have the tests run, but since they are already running on CI, that would make things faster 🤷

frosso:
@harriswong thanks!
I was just going to reply 😄 

I rebased this branch with the changes on `trunk` and ran your commands on my machine, they seem to be working ok now 🤷

.husky/pre-commit
harriswong: I personally think that WCS&T should continue to run test per-push. I think we should make sure tests pass before we push, so automating this makes sense.

frosso: Changed in 92139cc8cb2eaf40e03eed83166a564cb01bd22b 👍

.husky/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx --no-install lint-staged
```
harriswong: Not doing `npm run test` in pre-push?

frosso: OMG sorry about that! 🤦 (tested locally with `git checkout -b test && echo 'asd' >> README.md && git commit -am "test" && git push -u origin test && git checkout trunk && git push origin :test`)
```suggestion
npm test
```

harriswong: Haha np at all!

harriswong (approved):
👍 
Tested with
```
nvm i
npm i
npm run prepare
```

Then, i created a new branch off this. Made some commits, verified `git commit` runs through the following:
```
✔ Preparing...
✔ Running tasks...
✔ Applying modifications...
✔ Cleaning up...
Aborting commit due to empty commit message.
```

I didn't test `git push`.

PR #2447, opened 2021-07-02
feat: add dev docker image

frosso:
## Description

Adding a docker image to ease the local development process.
webpack is also in a container - which means that if a new file is added or the webpack configuration changes, you might need to restart the container to load the changes

- Stop your existing containers
- Run `npm run up`
- Wait for everything to build, there shouldn't be errors 🤞 
- Forward your site to something like Jurassic Tube or ngrok
- Connect your site to Jetpack
- Access again at http://localhost:8050/wp-admin/plugins.php
- Profit!

### Checklist

- [ ] unit tests

Nothing that should be visible to the merchants, so I didn't add one.


bin/docker-setup.sh
```diff
@@ -0,0 +1,140 @@
...
+
+echo "Importing some sample data..."
+cli wp import wp-content/plugins/woocommerce/sample-data/sample_products.xml --authors=skip
+
+echo "Copying the WooCommerce Services helper plugin..."
+mkdir -p docker/wordpress/wp-content/mu-plugins
+cp docker/mu-dev.php docker/wordpress/wp-content/mu-plugins/
+cp docker/mu-wccon-staging.php docker/wordpress/wp-content/mu-plugins/
+
+echo "Activating the WooCommerce Services plugin..."
```
frosso, line 137: (idea): maybe we can also figure out a way to auto-setup the shipping zones for WCS&T 🤷

FangedParakeet:
I had a little play around with this and for the most part everything worked pretty well!

I started off with a completely fresh clone of the repo and ran through the entire setup from zero to hero. The first time I started up the containers, I got a bunch of unhappy errors, which were a bit weird.

Unhappy errors
_Sad_

But then I literally just turned everything off and then back on again (restarted the containers) and it worked perfectly fine. I didn't mind having to set up the shipping zones myself, but I do have one minor complaint: when we upload the sample products into our store, none of them have any shipping dimensions or weight, so we're unable to get any rates for them at checkout. If there's a way to update these products from the CLI after we've uploaded them or if we could include our own `sample_products.xml` file, that would be a nice little improvement here. 

Also, I mucked around with the Jurassic Tube connection as well. Unfortunately, using `jetpack` as an environment variable doesn't magically make everything work together. However, I spun up a branch [here](https://github.com/Automattic/woocommerce-services/tree/fangedparakeet/docker-jurassic-tube) where I just used `wcpay` instead...and everything appeared to work out fine! A better solution would probably to just add `wcship` as an option in that famous JTube switch case, but I think we can reuse the JTube setup instructions/environment, just like a couple other plugins appear to be doing. I was able to make the Jetpack connection and continue on contentedly with my affairs using this solution.

Lastly, I can see the E2E tests are failing, but I think @waclawjacek already ran into this issue and it looks like all we need to do is update the `@woocommerce/e2e-environment` package apparently and everything should be right as rain.

package.json
```diff
@@ -17,9 +17,11 @@
     "i18njson": "node tasks/i18njson",
     "update-deps": "rm npm-shrinkwrap.json && rm -rf node_modules && npm install && npm shrinkwrap",
     "build": "echo yes | npm run release && cp release/woocommerce-services.zip ./",
-    "docker:up": "npm run dist && npx wc-e2e docker:up",
-    "docker:down": "npx wc-e2e docker:down",
-    "docker:ping": "bash ./tests/bin/wait-for-e2e-build.sh",
+	"up": "docker-compose up --build --force-recreate -d && ./bin/docker-setup.sh",
+	"docker:down": "docker-compose down",
```
FangedParakeet: Just for consistency's sake, could we either use `npm run up` and `npm run down` or `npm run docker:up` and `npm run docker:down`? Already had to curse my terminal window a few times after seeing the `npm ERR! missing script` error flash across my screen because I picked the wrong one.

Personally, I prefer the shorter `npm run up/down`, because it's, well, shorter. But also, consistent with other containerised Woo plugins.

frosso: You're absolutely right, let's do the shorter version `up`/`down` 👍 
```suggestion
	"down": "docker-compose down",
```

frosso:
@FangedParakeet thanks so much for taking the time to test this out! 🙌 

It's indeed pretty weird that the plugins were disabled 👀 but I'm glad it ended up working ok!

> when we upload the sample products into our store, none of them have any shipping dimensions or weight, so we're unable to get any rates for them at checkout

Yeah, that would be a nice next step - I could do some modifications for that!
What dimensions do you think we can use? For most of my products I just do `10x10x10` and `0.5` weight 😄 
One option could be to duplicate the XML from core, and embed it in our plugin with the shipping dimensions. I could generate the XML by entering the dimensions manually and then exporting again.
The other option could be to create a script to set the dimensions, taking into account variable and virtual products. That would maybe take a bit more time, tho.
I'm kinda leaning towards the first option (embedding the XML to import within the plugin), then we could create the script when we have some more time.

Thoughts?

> Also, I mucked around with the Jurassic Tube connection as well. Unfortunately, using `jetpack` as an environment variable doesn't magically make everything work together. However, I spun up a branch [here](https://github.com/Automattic/woocommerce-services/tree/fangedparakeet/docker-jurassic-tube) where I just used `wcpay` instead...and everything appeared to work out fine! A better solution would probably to just add `wcship` as an option in that famous JTube switch case, but I think we can reuse the JTube setup instructions/environment, just like a couple other plugins appear to be doing. I was able to make the Jetpack connection and continue on contentedly with my affairs using this solution.

That's great! I don't know what the difference is with the `wcpay` configuration, but if it works, it works!
I think the branch linked maybe doesn't have all the updates pushed with the JT setup - at this moment it looks like it has the changes from this branch 👀

FangedParakeet:
> What dimensions do you think we can use? For most of my products I just do 10x10x10 and 0.5 weight 😄

Yeah, I just put 5 for everything, because it's at the centre of my numpad, so I can just mash my keyboard without thinking. 😄 Your dimensions sounds probably a bit more reasonable.

>  One option could be to duplicate the XML from core, and embed it in our plugin with the shipping dimensions. I could generate the XML by entering the dimensions manually and then exporting again.
The other option could be to create a script to set the dimensions, taking into account variable and virtual products. That would maybe take a bit more time, tho.

I think the first option sounds pretty good to me. The second option sounds like a little bit of overkill. The purpose of this container should just to be able to boot things up quick, hop in, get some quick and dirty shipping rates, and spam some labels without having to worry about the details. I don't think worrying about that other stuff is so important. 

>  I think the branch linked maybe doesn't have all the updates pushed with the JT setup - at this moment it looks like it has the changes from this branch 👀

Lol, oops, I pushed the branch without committing anything. Welcome to Clown Town, population: this guy. [Here's a PR to this branch](https://github.com/Automattic/woocommerce-services/pull/2451), so you can see the few tiny things I had to add in and my one little quibble with it.

frosso:
@FangedParakeet I got some interesting findings!
The sample data provided in the XML doesn't have shipping dimensions.
But next to the XML, there's also a CSV with sample data! And the CSV _does_ have shipping dimensions!

I took some time to copy the shipping dimensions in the CSV to the XML file and I opened a PR within WC - this way we should be good for the next WC release, without having to create additional import scripts.

Here's the changes: https://github.com/woocommerce/woocommerce/pull/30243

FangedParakeet (approved):
Fateful findings indeed. Nice one looking into that one. Think I probably would've taken one look at all the noise in that XML file and said, "No thanks." 😄 

I tossed you an approval there, but I'm not sure if we need someone from Woo to give us the green light as well--seems like an inoffensive enough addition regardless. I'll toss you one here as well, because this is working pretty consistently for me and I'm loving it.

PR #2451, opened 2021-07-08
Adds WCPay Jurassic Tube setup and implementation to Docker feat

FangedParakeet:
## Description

Friendship ended with ngrok. Now Jurassic Tube is my best friend. 

1. With Docker containers up and running, run `npm run tube:setup` once to download the installer files and configure Jurassic Tube.
2. To open up a JT tunnel, run `npm run tube:start`. Go to the endpoint that you created in the previous step and behold the glory.
3. To close your tunnel once you're done fooling around, run `npm run tube:stop`.

### Checklist

- [ ] unit tests


bin/jurassic-tube-setup.sh
```diff
@@ -0,0 +1,48 @@
...
+    mkdir "${PWD}/docker/bin/jt"
+fi
+
+echo "Downloading the latest version of the installer script..."
+echo
+
+# Download the installer (if it's not already present):
+if [ ! -f "${PWD}/docker/bin/jt/installer.sh" ]; then
+    # Download the installer script:
+    curl "https://jurassic.tube/get-installer.php?env=wcpay" -o ${PWD}/docker/bin/jt/installer.sh && chmod +x ${PWD}/docker/bin/jt/installer.sh
```
FangedParakeet, line 21: So the only foible I have with this, is over here I'm using `env=wcpay` to get the setup installer for Jurassic Tube, where really we should probably be adding `wcship` as an additional case to the switch over [here](https://github.com/Automattic/jurassic-tube/blob/master/get-installer.php#L7) and then passing `env=wcship` instead. However, while looking a bit neater, ultimately this would make everything work just the same.

frosso (approved):
> Friendship ended with ngrok. Now Jurassic Tube is my best friend.

lol love it!

Tested this locally, works like a charm!

harriswong:
What's this magic I see here?! ❤️

PR #2448, opened 2021-07-03
Replace Calypso FormCheckbox with CheckboxControl.

brezocordero:
## Description

Replace Calypso FormCheckbox with [CheckboxControl](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/checkbox-control/index.js).

### Related issue(s)

#2333 

### Steps to reproduce & screenshots/GIFs

* Go to Settings > Shipping > WooCommerce Shipping 
* Click on Choose a different credit card, verify that that the look and feel and behavior of the checkbox is unchanged.
* Verify that the checkbox for Email receipts looks and works as before.
CheckboxLabelPayment

* Click on Add Package, Service Package, verify that the look and feel and behavior of checkboxes is unchanged.
CheckboxesAddPackage

* Place an order with multiple packages. 
* Edit the order, create a label, on packaging, click on 'Move' on one of the packages, then add to a new package.
* Click on 'Add items', verify that that the look and feel and behavior of the checkbox is unchanged.
CheckboxAddToPackage

### Checklist

- [ ] unit tests

- [ ] `changelog.txt` entry added

PR #2458, opened 2021-07-13
Update changelog to include Calypso component replacement

harriswong:
## Description
Update changelog to include changes from https://github.com/Automattic/woocommerce-services/pull/2448

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.17 - xxxx-xx-xx =
+* Tweak - Replace Calypso FormCheckbox with CheckboxControl.
+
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.17 - xxxx-xx-xx =
```
frosso, line 3: TIL we don't need the year

frosso (approved):
Looks good, tested with `woorelease simulate`

changelog.txt
harriswong: Just double checked, `xxxx` should work for year as well!

PR #2459, opened 2021-07-21
Do not fail on `npm run update-deps` if npm-shrinkwrap.json is missing

waclawjacek:
## Description

This is a tiny change that will allow the `npm run update-deps` script to proceed if `npm-shrinkwrap.json` is missing.

Previously if you ran `npm run update-deps` and something failed, you could end up with a missing `npm-shrinkwrap.json`. If you then retried `npm run update-deps`, the script would fail because `rm npm-shrinkwrap.json && ...` would fail.

Replacing `rm` with `rm -f` resolves this.


frosso (approved):
https://man7.org/linux/man-pages/man1/rm.1.html#OPTIONS 👍

PR #2466, opened 2021-08-04
Fix fatal error when viewing a draft order

### LINKED ISSUE #2423: Fatal error when viewing an auto draft order
A merchant had run into an issue when trying to view/edit an order that was just in the auto-draft status. The WCS plugin assumes that an order has been created when viewing and throws a fatal error when trying to access the created date. 

**To Reproduce the Error**
- Go to WooCommerce > Orders and click "Add order"
- Add a product to the order
- Click "Reclaculate"
- At this point, copy the order number on this page
- Then use the URL to go to that order page, such as /wp-admin/post.php?post=1234&action=edit
- Note the fatal error in the Shipment Tracking meta box
- Disable the WooCommerce Shipping & Tax plugin on the site
- Refresh that order page, note you can see the start of an order as expected

**Error**
`Fatal error: Uncaught Error: Call to a member function getTimestamp() on null in /srv/www/woocommerce-services-site/public_html/wp-content/plugins/woocommerce-services/classes/class-wc-connect-order-presenter.php:20`

https://github.com/Automattic/woocommerce-services/blob/4efc568d76630858dfc9f3c28a51fe3f6ca6be04/classes/class-wc-connect-order-presenter.php#L20

This line will need to be updated to match the `update_at` and `completed_at` format.

brettshumaker:
## Description
When viewing an order with a `post_status` of `draft`, the `date_created` order meta hasn't been added yet, and this fatal error occurs:

```
Fatal error: Uncaught Error: Call to a member function getTimestamp() on null in /srv/www/woocommerce-services-site/public_html/wp-content/plugins/woocommerce-services/classes/class-wc-connect-order-presenter.php:20
```

`$order->get_date_created()` returns `null` in this case and the fatal error is triggered when `WC_Connect_Order_Presenter->get_order_for_api()` tries to call `getTimestamp()` on `null`.

This PR checks that `$order->get_date_created()` is a truthy value before attempting to call `getTimestamp()` on the returned value. I've also added unit tests for the three meta values that are expected to be `WC_DateTime` objects to ensure that if they're returned as `null`, no fatal error occurs.

### Related issue(s)

Fixes #2423

- Go to WooCommerce > Orders and click "Add order"
- Add a product to the order
- Click "Reclaculate"
- At this point, copy the order number on this page
- Then use the URL to go to that order page, such as /wp-admin/post.php?post=1234&action=edit
- Note the fatal error in the Shipment Tracking meta box
- Disable the WooCommerce Shipping & Tax plugin on the site
- Refresh that order page, note you can see the start of an order as expected

### Checklist

- [x] unit tests


tests/php/test-class-wc-connect-order-presenter.php
```diff
@@ -8,6 +8,25 @@ public static function setupBeforeClass() {
 		WC_Connect_Compatibility::set_version( '3.0.0' );
 	}
 
+	public function test_get_order_for_api_null_values() {
+		// Setup order
+		$order = WC_Helper_Order::create_order();
+		$order->save();
```
waclawjacek, line 14: `WC_Helper_Order::create_order()` is already calling `save()` on the created order object, but it seems saving is not necessary at all here as we're just using a `WC_Order` object and not retrieving it from the DB.

waclawjacek (approved):
Fantastic PR! :tada: Very clean code both for the change itself and the test.

:shipit:

PR #2204, opened 2020-09-28
Add link to print customs info in the label tracking metabox

Ferdev:
Whenever a shipment needs to print a customs form to include it in their international shipments, this PR allows WooCommerce Shipping to present a link in the ellipsis menu in the tracking metabox that allows to print it.

It will only appear for carriers that can't receive the customs information electronically.

### Testing instructions

Since currently all our carriers support eletronically submitted customs forms, WooCommerce Connect Server won't include the `commercial_invoice_url` in its response for any of our current supported carriers. An easy way to force it could be filling the `commercial_invoice_url` field in the `wcc_shipping_label` table of your connect server database. Any public PDF url will work in this case.

After that, just open the order containing the modified shipping label, and click "Print customs form" in that label's ellipsis menu in the tracking metabox:

image

Closes #2191

PR #2468, opened 2021-08-10
Fix PHP Notice: Undefined property: stdClass::$commercial_invoice_url

### LINKED ISSUE #2455: PHP Notice:  Undefined property: stdClass::$commercial_invoice_url
Observed in the PHP logs when purchasing a US Priority Mail label (origin: Washington State, destination: Washington State) using 1.25.13. Label purchase seemed unaffected.

```
[09-Jul-2021 17:43:45 UTC] PHP Notice:  Undefined property: stdClass::$commercial_invoice_url in /home/xxxxxx/staging.marsiswaiting.com/wp-content/plugins/woocommerce-services/classes/class-wc-rest-connect-shipping-label-controller.php on line 118
```

brettshumaker:
## Description
Currently, `commercial_invoice_url` is allowed to be `null` in the return from the connect server and a PHP notice (PHP Notice: Undefined property: stdClass::$commercial_invoice_url) will be shown in this instance. This PR handles the `null` instance and sets the `commercial_invoice_url` to be an empty string.

### Related issue(s)
Fixes #2455 

I'm not sure which circumstances could lead to the commercial invoice URL being `null`, but it's certainly allowed in the connect server return so we should properly handle that case here.

### Checklist

- [ ] unit tests


harriswong:
👀

classes/class-wc-rest-connect-shipping-label-controller.php
```diff
@@ -115,7 +115,7 @@ public function post( $request ) {
 				'carrier_id'             => $label_data->label->carrier_id,
 				'service_name'           => $service_names[ $index ],
 				'status'                 => $label_data->label->status,
-				'commercial_invoice_url' => $label_data->label->commercial_invoice_url,
+				'commercial_invoice_url' => isset( $label_data->label->commercial_invoice_url ) ? $label_data->label->commercial_invoice_url : '',
```
harriswong, line 118: ~If `$label_data->label->commercial_invoice_url` was set explicitly to `null`, then `isset` won't catch it. If we want to be more defensive on it, we could use `isset( $label_data->label->commercial_invoice_url ) ? ... ` instead.~

**updated
Tested with 
```
$label_data = new stdClass();
$label_data->label = new stdClass();
$label_data->label->commercial_invoice_url = null;

$output = isset( $label_data->label->commercial_invoice_url ) ? $label_data->label->commercial_invoice_url : '';

var_dump($output);
//output ''
```
👍

harriswong:
> I'm not sure which circumstances could lead to the commercial invoice URL being null

We introduced API v4 to address the [custom issue](https://github.com/Automattic/woocommerce-services/pull/2204/files#diff-4cb5834c68b324f4fbecc437b25f62021e55ecc28db9adb93053eca0391f5cc5R51). Anything before v4 will be missing `commercial_invoice_url` in the response body.

harriswong (approved):
👍  LGTM

PR #2470, opened 2021-08-11
Add validation for destination phone number for label printing.

### LINKED ISSUE #2276: Destination Address Phone Number is required for DHL Express and UPS International Orders
We had been getting a few reports of merchants unable to purchase DHL Express labels and receiving a generic error message. Upon digging deeper it appears that the Destination Address Phone number is a required field for these label purchases. With some testing I was able to confirm that this error is triggered when purchasing an international label through DHL Express or UPS, but is not triggered on USPS international labels.

Since the majority of the services require this field, I was thinking it would be best to require the field for all international orders. I was going to call out three things that would be needed to address this issue:

1. Update the documentation to mention that Destination Address Phone Number is required for International orders as this would require the merchant to update their site settings.
2. Display a more useful error message to the merchant, though this may need to come from the Connect server.
3. Make the field required for international orders similar to what was done for the From Address.

For 3 I was thinking making a [similar check to destination as origin](https://github.com/Automattic/woocommerce-services/blob/be6d1fcb0396143ba0dfc5c7546ad82035d611f6/client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js#L487-L490) if a customs form is required?

**Steps to Reproduce**
1. Place an international order that would be eligible for DHL Express.
2. Go to create a shipping label for the order.
3. Remove the Destination Address phone number (if one was provided).
4. Attempt to purchase the shipping label.
5. Get generic error message about a problem purchasing the shipping label.

Screen Shot 2020-12-04 at 2 17 54 PM

tommyshellberg:
### Related issue(s)

Closes #2276 

This PR will require a destination phone address for all international shipments. To determine this, I used the return of `isCustomsFormRequired()`, I hope that is not too strong of an assumption since it's currently used when getting address errors.

I added some tests and refactored a method to allow for validating origin and destination numbers separately(and makes it more future-proof in case we need to validate other stuff).

Screen Shot 2021-08-11 at 1 36 13 PM

Screen Shot 2021-08-11 at 1 36 04 PM

### Checklist

- Test with US origin + destination, shouldn't need a phone for either address.
- Test with US origin + int'l destination, should need a destination phone and origin phone.

- [x] unit tests


client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
```diff
@@ -218,10 +218,10 @@ export const isCustomsFormRequired = createSelector(
  * @param {Object}  appState            Local Redux state.
  * @param {Object}  addressData         Address to check, including normalization state and values.
  * @param {number}  siteId              The ID of the current site ID.
- * @param {boolean} shouldValidatePhone An indiator whether phone validation is required.
+ * @param {Object}  fieldsToValidate    Specify which form fields to validate.
  * @return {Object}                     A hash of errors with field names as keys.
  */
-const getRawAddressErrors = ( appState, addressData, siteId, shouldValidatePhone ) => {
+ export const getRawAddressErrors = ( appState, addressData, siteId, fieldsToValidate ) => {
```
harriswong, line 224: When I set a break point on the calls to `getRawAddressErrors()`, it seems like `fieldsToValidate` is a boolean than an object.

harriswong (changes requested):
I checkout an item, created a new order with (use a different shipping address so i can skip phone number). I opened the order and manually removed "phone number" from the billing information. 

I clicked "Create shipping label" and typing in or removing phone numbers don't give me any error message, is there something I am missing?
image

tommyshellberg:
> I checkout an item, created a new order with (use a different shipping address so i can skip phone number). I opened the order and manually removed "phone number" from the billing information.

Well, that's embarrassing. This PR was missing a commit from my local repo 🤦. Should be working now and tests are passing again.

client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
```diff
@@ -480,9 +486,11 @@ export const getFormErrors = createSelector(
 		const destinationCountryCode = form.destination.values.country;
 		const destinationCountryName = getCountryName( state, destinationCountryCode, siteId );
 		const shouldValidateOriginPhone = isCustomsFormRequired( state, orderId, siteId );
+		// If it's an international order, validate the destination phone as well.
```
harriswong, line 489: One note here for `isCustomsFormRequired()` function. There is a special case that custom is required for military addresses. In this case, phone number will now also be required. I guess that's okay?

client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
```diff
@@ -264,11 +265,16 @@ const getRawAddressErrors = ( appState, addressData, siteId, shouldValidatePhone
 			);
 		}
 	}
+	if ( true === destinationPhone && !phone ) {
+		errors.phone = translate(
```
harriswong, line 269: For custom purposes, in line 255, it would go validate the phone number to match a 10 digit phone number format in origin. Since WCS&T supports US store only. The 10 digit rule is fine. 

In destination, we don't do any phone number validation (ie. i can put a single digit and it would still be okay). But I don't know if we should add international phone number validation here with this PR. Maybe something to add in the future?

client/extensions/woocommerce/woocommerce-services/state/shipping-label/test/selectors.js
```diff
@@ -580,3 +580,62 @@ describe( 'Shipping label selectors', () => {
...
+		expect( errors ).to.be.an( 'Object' ).and.to.not.have.property( 'phone' );
+	} );
+
+	it ( 'with no origin phone entered and origin phone required, return error', () => {
+		const fieldsToValidate = { originPhone: true };
+		const errors = getRawAddressErrors( appState, addressData, siteId, fieldsToValidate );
+		expect( errors ).to.be.an( 'Object' ).and.to.have.property( 'phone' );
+	} );
+
+	it ( 'with no origin phone entered and origin phone not required, do not return error', () => {
```
harriswong, line 619: Is this the same test cases as line 608?

harriswong: Oh i see why the codes are the same. Because the `addressData` could be from origin or destination, so the `phone: ''` could be empty and `fieldsToValidate = {};`.

harriswong (approved):
👍  LGTM

- [x] Tested a new order with international addresses
- [x] Tested a new order with US->US addresses
- [x] Tested existing order with international addresses
- [x] Tested existing order with US->US addresses

client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
tommyshellberg: I avoided phone number validation here because of the plethora of different phone formats and didn't want to potentially annoy someone with an imperfect regex validation. 🤷

client/extensions/woocommerce/woocommerce-services/state/shipping-label/test/selectors.js
tommyshellberg: Yeah, they are technically the same but I thought I would make each case combination explicit.

client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
tommyshellberg: I think the upside of re-using the function outweighs the downside. We could keep an eye out for complaints about those specific addresses(it's a tiny tiny percentage of shipments).

PR #2479, opened 2021-10-04
Fix country drop down showing currency name

### LINKED ISSUE #2477: Currency Displaying in Country Dropdown Shipping
I've deactivated all other plugins, yet I am having issues competing orders because the dropdown is displaying currencies rather than countries.
sp

bborman22:
## Description
It was discovered that the shipping label purchase modal was showing a currency name instead of the country name when using the latest WooCommerce version. This was the result of the changes in https://github.com/woocommerce/woocommerce/pull/30216 where some locale info keys were added. The culprit was the addition of the name key to the locale info which was then overwriting the country name key when we did an array merge. 

I went with what seemed like the cleanest fix by just changing the order of the arrays in the merge so that the country information will always override the locale (essentially currency) information in case they share a key. This allowed the country name to remain in the name key and shows the proper country name in the drop down.

### Related issue(s)
Fixes #2477 

1. Setup local site with latest version of WCShip and WC 5.7.1.
2. Create a new order and attempt to purchase a label for the order.
3. Note that the country fields inside the Origin address and Destination address both contain the correct country names now.
4. Set Destination address to an international address. Note that Origin country field on customs form is filled in with the correct country name.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.19 - 2021-xx-xx =
+* Fix   - Country drop down list no longer showing currency name.
+
 = 1.25.18 - 2021-08-16 =
 * Add   - Added "Automated Taxes" health item on status page.
 * Fix   - Show error when missing required destination phone for international shipments.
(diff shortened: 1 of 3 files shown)


frosso (approved):
Tested locally, looks like it's fixed 👍 
I'm amazed and afraid at the same time by the simplicity of the solution 😆

PR #2482, opened 2021-10-05
Bump phpunit version down to match WC

harriswong:
### Description
```
PHP Fatal error:  Uncaught Error: Call to private method PHPUnit\Util\Test::parseAnnotations() from context 'PHPUnit\Framework\TestSuite' in /home/runner/work/woocommerce-services/woocommerce-services/vendor/phpunit/phpunit/src/Framework/TestSuite.php:305
```
There was an error in the PHPUnit test step within GitHub Actions. This seems to be caused by a conflict between WCS using PHPUnit 7.* and Woo using PHPUnit 6.5.14:
```
// This is the WCS dependency
Install Composer...
- Downloading phpunit/phpunit (7.5.20)
```
```
// This is the WC dependency
Setup WP and WC
 - Downloading phpunit/phpunit (6.5.14)
```

Given that WCS tests [relies on WC's codes](https://github.com/Automattic/woocommerce-services/blob/e36bf5a36e6d9c8218196ef87c430ff38e4be5b9/tests/php/bootstrap.php#L51). The different version of PHPUnit could be a problem. 

This PR proposes downgrading WCS' PHPUnit to 6.5.14.

1. Run `./vendor/bin/phpunit -c phpunit.xml.dist` locally and tests should run. Or make sure the pipeline passes with this PR.

### Checklist

- [x] unit tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
     "require-dev": {
         "woocommerce/woocommerce-sniffs": "^0.1.0",
-        "phpunit/phpunit": "^7"
+        "phpunit/phpunit": "6.5.14"
     },
     "scripts": {
 		"test": [
--- a/composer.lock
+++ b/composer.lock
@@ -4,27 +4,27 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


frosso (approved):
If GH Actions is happy I'm happy

harriswong:
If @frosso is happy then I'm happy.

PR #2484, opened 2021-10-06
feat: add notice about tax nexus in settings

frosso:
## Description
I'd like to elevate the notice about the tax nexus in the settings.
It seems that it might not be clear to merchants that taxes are calculated for a single nexus.

With the setting disabled:
Screen Shot 2021-10-06 at 6 11 41 PM

With the setting enabled:
Screen Shot 2021-10-06 at 6 10 52 PM

- Go to http://localhost:8050/wp-admin/admin.php?page=wc-settings&tab=tax
- The plugin now has the notice about the tax nexus elevated next to the setting.

The link goes to the WC docs, but I'm not sure where the anchor went.
The information is already presented in the tooltip, I'm just "elevating" it so that it becomes more visible.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.19 - 2021-xx-xx =
+* Add - Notice about tax nexus in settings.
+
 = 1.25.18 - 2021-08-16 =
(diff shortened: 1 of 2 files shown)


bborman22 (approved):
Worked for me and makes the distinction much more clear now.

LGTM!

PR #2491, opened 2021-10-28
Fix/issue 2235 - Hide Shipment Tracking and Shipping Label metabox if the setting is disabled

### LINKED ISSUE #2235: Shipment tracking meta box shows up even when shipping label options disabled
A user in 3441996-zen noticed that with the extension Shipment Tracking enabled, the shipment tracking meta box shows up twice. I can reproduce this even with the shipping label function of WooCommerce Shipping and Tax disabled:

double shipment tracking

@c-shultz via A8C Slack says: "It looks like we may not be respecting the label setting on the order page.
There's a toggle at `wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings` that looks like it should allow the order page label options to be hidden."

image (5)

"But turning this off is not hiding the label printing meta boxes."

Although this is largely cosmetic rather than functional, it does raise questions and doubt in the mind of users so would be good to include in an upcoming release.

iyut:
## Description

Hide **"Shipment Tracking"** and **"Shipping Label"** metabox on order page when the Shipping Label setting is disabled.

### Related issue(s)

Fixes #2235 .

1. Disable the "Shipping Label" from `/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings`.
screenshot-fritflow com-2021 10 28-17_26_52

2. Edit one of your order from **WooCommerce >> Orders**.
3. **Before the fixes**, "Shipping Label" and "Shipment Tracking" meta box will always be displayed.
screenshot-fritflow com-2021 10 28-17_47_09

4. **After the fixes**, the "Shipping Label" and "Shipment Tracking" meta box will be disappeared.
screenshot-fritflow com-2021 10 28-17_48_54

### Checklist

- [ ] unit tests


dustinparker (approved):
Looks good!

PR #2496, opened 2021-11-04
Fix/issue 2490 - Add 3 digits currency code for non USD order in shipping label price

### LINKED ISSUE #2490: Compatibility with customer facing currencies other than USD
Merchants are able to add currencies to their store through various extensions, including the Multi-Currency feature within WooCommerce Payments. This allows the store currency to remain USD, but they are able to accept payments in other currencies, such as CAD and EUR. 

I tested WooCommerce Shipping and Tax with WCPay Multi-Currency and found:
* Legacy shipping rates in cart/checkout convert correctly
* Taxes convert correctly
* Amounts show in order correctly
* When creating a shipping label:
  * The amount the customer paid is in USD, not the currency of the order.
  * The amount the merchant will need to pay for the label is in USD, as it should be, but not explicit format. Meaning USD is not shown after the amount.

The amount the customer has paid is hardcoded to USD here:
https://github.com/Automattic/woocommerce-services/blob/1.25.19/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/index.js#L120

The amount the merchant needs to pay is also hardcoded to USD, which is fine since it should always be USD, however, it is not in explicit format.
https://github.com/Automattic/woocommerce-services/blob/1.25.19/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/rates-step/shipping-rate.js#L160

It appears it is possible to pass additional values to the `formatCurrency` function that is used for both instances to allow for explicit formatting in different currencies. 
https://github.com/Automattic/wp-calypso/blob/trunk/packages/format-currency/src/index.js#L7-L38

I believe a fix may be to:
* Always use the order currency for the amount the customer has paid.
* Add a check and if the order currency is not USD, then display both the customer cost and label cost in explicit format.
  * FWIW, [WooCommerce core has added a lot for referencing currency formatting.](https://github.com/woocommerce/woocommerce/blob/5.8.0/i18n/currency-info.php)

Markup 2021-10-22 at 09 01 32

iyut:
## Description
Get the currency information from the order meta and add a conditional to display 3 digits currency code in shipping label price for non USD currency.

### Related issue(s)

Fixes #2490

1. Install WooCommerce Payments and enable the multi-currency feature from this [page](http://localhost:8050/wp-admin/admin.php?page=wc-settings&tab=checkout&section=woocommerce_payments).
screenshot-localhost_8050-2021 11 04-15_54_19
2. Go to [WooCommerce >> Settings >> Multi-Currency](http://localhost:8050/wp-admin/admin.php?page=wc-settings&tab=wcpay_multi_currency) and set the settings like the screenshot below :
screenshot-localhost_8050-2021 11 04-15_57_01
3. Order any item using EUR currency.
4. Attempt to create the shipping label from order edit page.
5. Notice the 3 digits currency code is added in Shipping label price.
screenshot-localhost_8050-2021 11 04-16_00_14

### Checklist

- [] unit tests

- [x] changelog.txt entry added

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.20 - 2021-xx-xx =
+* Fix - Added 3 digits currency code on shipping label price for non USD.
+
 = 1.25.19 - 2021-10-14 =
 * Add - Notice about tax nexus in settings.
 * Fix - Country drop down list no longer showing currency name.
--- a/client/extensions/woocommerce/lib/order-values/index.js
+++ b/client/extensions/woocommerce/lib/order-values/index.js
@@ -104,6 +104,6 @@ export function getOrderTotalTax( order ) {
  * @param {Object} order An order as returned from API
  * @return {String} Currency value
  */
- export function getOrderCurrency( order ) {
+export function getOrderCurrency( order ) {
 	return get( order, 'currency', false );
 }
--- a/client/extensions/woocommerce/lib/order-values/index.js
+++ b/client/extensions/woocommerce/lib/order-values/index.js
@@ -97,3 +97,13 @@ export function getOrderTotalTax( order ) {
 	const fees = getOrderFeeTotalTax( order );
 	return subtotal + shipping + fees;
 }
+
+/**
+ * Get the currency value in order
+ *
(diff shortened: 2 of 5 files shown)


dustinparker (approved):
Everything works as expected. Looks good!

PR #2507, opened 2021-11-22
Fix/issue 2480 - Use 'native' pdf support feature for Firefox version 94 or later.

### LINKED ISSUE #2480: Firefox blob scheme marked as "Connection not secure"
A user brought it to our attention that in Firefox, the shipping labels are opened in a new tab using the blob scheme to display the label. In doing so the page is marked with a Connect not secure warning on the URL. While there isn't any security concerns because of how the blob scheme works, it could alleviate any concerns a merchant might have when they see this warning if we're able to avoid it?

**Steps to Reproduce**
1. Start with a store with an order with a shipping label created.
2. Using Firefox, log in to the store as the merchant and go to the relevant order.
3. On the sidebar click on the label's menu and select reprint label.
4. In the modal click Print
5. This should bring up a new tab using the blob scheme.
6. The Firefox URL bar will be marked as Connection not secure.

FF-Blob

iyut:
## Description
Originally, the plugin need to use blob in Firefox because there is some concern with the browser's PDF reader reliability based on [this comment](https://github.com/Automattic/woocommerce-services/blob/c39f7c7cbb506830beb399d0e326f19d30620100/client/extensions/woocommerce/woocommerce-services/lib/utils/pdf-support.js#L29-L30). **It seems Firefox is not stable enough to do direct printing for PDF file because of this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=911444).**

However, I did some testing by changing this [code](https://github.com/Automattic/woocommerce-services/blob/c39f7c7cbb506830beb399d0e326f19d30620100/client/extensions/woocommerce/woocommerce-services/lib/utils/pdf-support.js#L31) into `return 'native';` and try to print the label using **Firefox (version 94)** native PDF reader.
**The result** : I can print the PDF label without any issue.
**Screenshot** : https://paste.pics/8e10a4a6fc880f66c221bb775473c96a

This PR is to change the PDF support for Firefox ver. 94 to use `native` instead of `addon`.
### Related issue(s)

Fixes #2480 

1. Start with a store with an order with a shipping label created.
2. Using Firefox, log in to the store as the merchant and go to the relevant order.
3. On the sidebar click on the label's menu and select reprint label.
4. In the modal click Print.
5. This should bring up a printing box like in this [screenshot](https://paste.pics/8e10a4a6fc880f66c221bb775473c96a).

### Another scenario that needs to be tested : 

1. Try to test this PR in different OS : MacOS, Windows, Linux.
2. Try to test this PR in later version of Firefox.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.21 - 2021-xx-xx =
+* Fix - Use 'native' pdf support feature for Firefox version 94 or later.
+
 = 1.25.20 - 2021-11-15 =
 * Fix - Hide "Shipping Label" and "Shipment Tracking" metabox when the label setting is disabled.
 * Fix - Wrap TaxJar API zipcodes with wc_normalize_postcode() before inserting into the database.
--- a/client/extensions/woocommerce/woocommerce-services/lib/utils/pdf-support.js
+++ b/client/extensions/woocommerce/woocommerce-services/lib/utils/pdf-support.js
@@ -26,9 +26,14 @@ export default memoize( () => {
 	}
 
 	if ( includes( navigator.userAgent, 'Firefox' ) ) {
-		// Firefox has a long-lived bug (https://bugzilla.mozilla.org/show_bug.cgi?id=911444),
-		// it's not reliable to consider its PDF reader "native"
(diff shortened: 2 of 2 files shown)


dustinparker (approved):
Everything works as expected. Nice work!

PR #2512, opened 2021-12-06
Fix/issue 2511

### LINKED ISSUE #2511: Invalid address - "Name" is required even when it is present
## Description

When validating the origin address, the message "We were unable to automatically verify the address — Name is required." could appear even when the "name" is already filled.
Screen Shot 2021-12-02 at 2 47 54 PM

It looks like in https://github.com/Automattic/woocommerce-services/blob/a0b068d3124dc661c48cbdaf29e0f2f7ad8fc125/classes/class-wc-rest-connect-address-normalization-controller.php#L16-L25 we are stripping out the `name` field from the validation - maybe that's related?

## Steps to reproduce
- Go to the order's page
- Click "Create new label" to open the modal
- In the "Origin" address, enter ` ` (space) on the "name" field
- Click "Verify address"
- Click "Edit address" again, to edit the address
- Fill in the details of the "Name" field with "Test Tester"
- Click "Verify address" again
- The message "We were unable to automatically verify the address — Name is required." appears, even tho the "name" field is populated

dustinparker:
## Description

We needed to add logic to ensure the connect server only validates the address name field if using a version of WCS newer than 1.25.20.
### Related issue(s)

Fixes #2511 

1. Make sure you are using WCC branch fix/wcs-issue-2511 and that [this constant](https://github.com/Automattic/woocommerce-services/blob/a0b068d3124dc661c48cbdaf29e0f2f7ad8fc125/woocommerce-services.php#L48) is set to 5.
2. Create an order and create a label for that order.
3. Add a blank space to the origin/destination name field and click verify address.
4. The address should be verified automatically if the address is valid.
5. Change [this constant](https://github.com/Automattic/woocommerce-services/blob/a0b068d3124dc661c48cbdaf29e0f2f7ad8fc125/woocommerce-services.php#L48) to 6.
6. Change WCS to version 1.25.21 (I just hardcoded the passed version [here](https://github.com/Automattic/woocommerce-connect-server/blob/8e5aebed5932f8d09aa6a1dadc276690a066b620/lib/shipping/handler.js#L198))
7. Now try to create a shipping label for the order.
8. Add a blank space to the origin/destination name field and click verify address.
9. An error should be returned.
10. Add a name to the origin/destination name field and click verify address.
11. The address should be verified automatically if the address is valid.

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-rest-connect-address-normalization-controller.php
+++ b/classes/class-wc-rest-connect-address-normalization-controller.php
@@ -19,6 +19,16 @@ class WC_REST_Connect_Address_Normalization_Controller extends WC_REST_Connect_B
 
 		unset( $address['company'], $address['phone'] );
 
+		/**
+		 * We only want to validate the name field if the API
+		 * version is greater than 5. If it's 5 or less
+		 * don't validate the name field.
+		 */
+		if ( WOOCOMMERCE_CONNECT_SERVER_API_VERSION <= 5 ) {
+			$name = $address['name'];
+			unset( $address['name'] );
+		}
+
 		$body     = array(
 			'destination' => $address,
 		);
@@ -40,6 +50,15 @@ class WC_REST_Connect_Address_Normalization_Controller extends WC_REST_Connect_B
 				'success'      => true,
 				'field_errors' => $response->field_errors,
 			);
+		}
+
+		/**
+		 * We only want to validate the name field if the API
+		 * version is greater than 5. If it's 5 or less
+		 * don't validate the name field.
+		 */
+		if ( WOOCOMMERCE_CONNECT_SERVER_API_VERSION <= 5 ) {
+			$response->normalized->name = $name;
 		}
 
 		$response->normalized->company = $company;
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,7 @@
 * Fix - Only call WC Subscriptions API when "access_token_secret" value is saved in database.
 * Fix - Add name field to fields sent for EasyPost API address verification.
 * Fix - Display company name under origin and destination address when create shipping label.
+* Fix - Add WCC version logic to address name validation functionality.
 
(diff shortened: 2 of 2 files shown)


iyut (approved):
I have tested this code as in description. It works separating the checks for WCC server api version 5 and 6. LGTM!

classes/class-wc-rest-connect-address-normalization-controller.php
```diff
@@ -19,6 +19,16 @@ public function post( $request ) {
...
 
+		/**
+		 * We only want to validate the name field if the API
+		 * version is greater than 5. If it's 5 or less
+		 * don't validate the name field.
+		 */
+		if ( WOOCOMMERCE_CONNECT_SERVER_API_VERSION <= 5 ) {
+			$name = $address['name'];
+			unset( $address['name'] );
+		}
```
waclawjacek, line 30: I'm fairly certain these version checks aren't necessary, as per our chat.

```suggestion
		$name = $address['name'];
		unset( $address['name'] );
```

classes/class-wc-rest-connect-address-normalization-controller.php
```diff
@@ -42,6 +52,15 @@ public function post( $request ) {
...
 		}
 
+		/**
+		 * We only want to validate the name field if the API
+		 * version is greater than 5. If it's 5 or less
+		 * don't validate the name field.
+		 */
+		if ( WOOCOMMERCE_CONNECT_SERVER_API_VERSION <= 5 ) {
+			$response->normalized->name = $name;
+		}
```
waclawjacek, line 62:
```suggestion
		$response->normalized->name = $name;
```

waclawjacek (changes requested):
I'm adding suggestions for removing the API version checks as per our chat. I'm fairly certain they aren't necessary but please verify that before including these changes.

PR #2515, opened 2021-12-14
Change API version back to 5.

### LINKED ISSUE #2511: Invalid address - "Name" is required even when it is present

dustinparker:
## Description

We needed to add logic to ensure the connect server only validates the address name field if using a version of WCS newer than 1.25.20.
### Related issue(s)

Fixes #2511 

1. Make sure you are using WCC branch fix/wcs-issue-2511.
2. Change WCS to version 1.25.21 (I just hardcoded the passed version [here](https://github.com/Automattic/woocommerce-connect-server/blob/8e5aebed5932f8d09aa6a1dadc276690a066b620/lib/shipping/handler.js#L198))
3. Now try to create a shipping label for the order.
4. Add a blank space to the origin/destination name field and click verify address.
5. An error should be returned.
6. Add a name to the origin/destination name field and click verify address.
7. The address should be verified automatically if the address is valid.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,6 +6,7 @@
 * Fix - Add name field to fields sent for EasyPost API address verification.
 * Fix - Display company name under origin and destination address when create shipping label.
 * Fix - Don't override general "Enable Tax" setting with WC Services Automated Taxes setting.
+* Fix - Change WOOCOMMERCE_CONNECT_SERVER_API_VERSION back to 5.
 
 = 1.25.20 - 2021-11-15 =
 * Fix - Hide "Shipping Label" and "Shipment Tracking" metabox when the label setting is disabled.
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,7 +6,6 @@
 * Fix - Add name field to fields sent for EasyPost API address verification.
 * Fix - Display company name under origin and destination address when create shipping label.
 * Fix - Don't override general "Enable Tax" setting with WC Services Automated Taxes setting.
-* Fix - Change WOOCOMMERCE_CONNECT_SERVER_API_VERSION back to 5.
 
 = 1.25.20 - 2021-11-15 =
 * Fix - Hide "Shipping Label" and "Shipment Tracking" metabox when the label setting is disabled.
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -44,7 +44,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 	define( 'WOOCOMMERCE_CONNECT_MINIMUM_JETPACK_VERSION', '7.5' );
 	define( 'WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH', 32 );
 
-	if ( ! defined( 'WOOCOMMERCE_CONNECT_SERVER_API_VERSION ' ) ) {
+	if ( ! defined( 'WOOCOMMERCE_CONNECT_SERVER_API_VERSION' ) ) {
 		define( 'WOOCOMMERCE_CONNECT_SERVER_API_VERSION', '5' );
 	}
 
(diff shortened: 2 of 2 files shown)


iyut:
Hi @dustinparker , 

I have tried to test this. It does block the user when the name field in destination address is empty. But I'm not sure if the error that I get is the correct one. 

Here is what I did : 
1. change the WCS version in woocommerce-services.php : ` * Version: 1.25.21`
2. change the version in package.json file : 
```
{
  "name": "woocommerce-services",
  "version": "1.25.21",
...
...
}
```
3. Create an order.
4. Try to create shipping label.
5. let the name field empty by using [space].
6. This is the error that I get : 
screenshot-localhost_8050-2021 12 15-13_45_19

I'm not sure whether this is the correct error message that I should get or not. If it's the correct error, then I will approve this PR.

woocommerce-services.php
```diff
@@ -45,7 +45,7 @@
 	define( 'WOOCOMMERCE_CONNECT_MAX_JSON_DECODE_DEPTH', 32 );
 
 	if ( ! defined( 'WOOCOMMERCE_CONNECT_SERVER_API_VERSION ' ) ) {
```
waclawjacek: This is entirely unrelated to your PR but it looks like a trailing space snuck into this line. Could you please remove it?

```suggestion
	if ( ! defined( 'WOOCOMMERCE_CONNECT_SERVER_API_VERSION' ) ) {
```

changelog.txt
```diff
@@ -6,6 +6,7 @@
 * Fix - Add name field to fields sent for EasyPost API address verification.
 * Fix - Display company name under origin and destination address when create shipping label.
 * Fix - Don't override general "Enable Tax" setting with WC Services Automated Taxes setting.
+* Fix - Change WOOCOMMERCE_CONNECT_SERVER_API_VERSION back to 5.
```
waclawjacek: The const is [still set to `5` in the previously released `1.25.20`](https://github.com/Automattic/woocommerce-services/blob/1.25.20/woocommerce-services.php#L48) so from a release-to-release standpoint, nothing's changed. Let's remove this entry.

```suggestion
```

waclawjacek (changes requested):
Unfortunately something isn't working on my end. It looks to be related to this commit that was introduced on `trunk` a while back: 01c3f2ea7d0800ec2f4388e353755aa69a5e7ab1

This is the error I get when providing a non-empty name:
Markup on 2021-12-15 at 12:49:35
Markup on 2021-12-15 at 12:50:14

An unit test should catch this. 🤔 Could you please add one that will test this and a valid case?

dustinparker:
> Unfortunately something isn't working on my end. It looks to be related to this commit that was introduced on `trunk` a while back: [01c3f2e](https://github.com/Automattic/woocommerce-services/commit/01c3f2ea7d0800ec2f4388e353755aa69a5e7ab1)
> 
> This is the error I get when providing a non-empty name: Markup on 2021-12-15 at 12:49:35 Markup on 2021-12-15 at 12:50:14
> 
> An unit test should catch this. thinking Could you please add one that will test this and a valid case?

hey @waclawjacek, if we release the Connect Server change that adds name to the Joi schema, this will actually no longer be an issue. We'd just need to make sure the WCC change is released before this WCS change. It seems like a unit test wouldn't be necessary in that case. Let me know what you think. Thanks!

waclawjacek (approved):
> We'd just need to make sure the WCC change is released before this WCS change.

Ah, I see! Thanks! LGTM! :shipit: 

Could you put a note in the Fulfill Extension Releases spreadsheet that the Connect Server should be deployed before deploying the WCS&T extension?

PR #2516, opened 2022-01-11
Fix/issue 2514 - TaxJar does not get the tax if the cart has non-taxable item.

### LINKED ISSUE #2514: Mixing $0 items that are only taxed for shipping with products that aren't taxed at all removes all taxes from cart
Say that there are two products.  One is taxed only on shipping and one is not taxed at all.  When both are added to the same cart and shipping is selected, no tax will be applied.

**Steps to Reproduce**
* Create a product for $0 and set it to be taxed only on shipping.
* Create a second product for an amount like $10 and set its taxes so it isn't taxed.
* Add the $0 to the cart and enter an address that should be charged taxes.
* Add the second product to the cart and notice that the taxes are now 0.

**Cart with product that only charges tax on shipping**
only-shipping-cart

no-tax-in-cart

iyut:
## Description

Remove the line item that has product tax code : 99999 when getting a tax value from TaxJar. Line item with product tax code is blocking the plugin to get a tax information from TaxJar.

### Related issue(s)
Fixes #2514 

1. Create a product and set it to be taxed only on shipping.
2. Create a second product for any amount like $10 and set its taxes so it isn't taxed.
3. Add the non-tax product to the cart and enter an address that should be charged taxes.
4. Add the second product to the cart and notice that the taxes are now not 0.
screenshot-localhost_8050-2022 01 11-19_29_21
screenshot-localhost_8050-2022 01 11-19_29_54
screenshot-localhost_8050-2022 01 11-19_30_39

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,12 +1,14 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.22 - 2022-xx-xx =
+* Fix - TaxJar does not get the tax if the cart has non-taxable item.
+
 = 1.25.21 - 2022-01-26 =
 * Fix - Use 'native' pdf support feature for Firefox version 94 or later.
 * Fix - Only call WC Subscriptions API when "access_token_secret" value is saved in database.
 * Fix - Add name field to fields sent for EasyPost API address verification.
 * Fix - Display company name under origin and destination address when create shipping label.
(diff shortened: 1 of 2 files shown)


dustinparker (approved):
Looks good!

PR #2534, opened 2022-03-10
Fix/issue 2528

### LINKED ISSUE #2528: Auto-trigger of print dialog is an issue
The most recent version of WooCommerce plugin (6.2.0) introduced an annoying glitch while purchasing and printing shipping labels. It looks to me like the devs were trying to improve a feature but missed the mark. I'd like to provide some end-user feedback on how this change effects my UX in the browser of my choice. Please see screen recording at [https://kingrings.com/downloads/PrintDialogIssue.mp4](https://kingrings.com/downloads/PrintDialogIssue.mp4)

iyut:
## Description
Fixing the Firefox issue when the browser is set to open PDF file using another program instead of using native PDF opener. When PDF file is opened in another program, the iframe content will be empty.

This changes is trying to identify how the firefox open the PDF by checking if the iframe content empty or not.

Empty iframe content  = firefox use another program to open PDF.
Unempty iframe content = firefox use native PDF opener from browser.

The empty iframe content will not catch any error. Thus we dont call the print() and let the browser open the PDF using another program.

### Related issue(s)
Closes #2528 

1. Create an order and create a label using WC services.
2. Open Mozilla Firefox browser.
3. Setting the PDF file to be opened by another program.
26b0bea0a66b410eaa5a73ec9b181500
4. Try to print the generated label PDF in Firefox.
d028e47af1f938855d94cd652e3c3621
5. The PDF will be opened in another program and no blank page being printed.

### Checklist

- [x] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.24 - 2022-xx-xx =
+* Fix   - Empty document is opened when Firefox is set to open PDF file using another program.
+
 = 1.25.23 - 2022-02-10 =
 * Tweak - Make "Name" field optional if "Company" field is not empty.
 * Fix   - Added "Delete California tax rates" tool.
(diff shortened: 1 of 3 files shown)


dustinparker (approved):
This fixed the problem! Looks good

PR #2541, opened 2022-04-01
Fix Shipping Issue 224 - Display error in WC Cart Block & Checkout Block

### LINKED ISSUE #222: Not originally calculating rates on cart page

### LINKED ISSUE #223: Physical product with no shipping dimension in the cart
One of the most common shipping issues we encounter is when customer's don't input shipping dimensions for any or all of their products.

If the current account is an admin, could we display a notice like this: http://cld.wthms.co/19Rx4/crRossy5 ? Maybe with some simpler wording, and some links to edit the products in question and/or a link to a guide on how to add product dimensions (perhaps an FaQ item in our connect docs).

### LINKED ISSUE #224: Alert the customer if their USPS account number isn't valid
Would it be possible to test if the customer's account number is valid when they add in their own? And if it isn't valid, have an alert show above the USPS settings?

It could be very helpful to do this in the settings on save before they start experiencing issues at checkout.

iyut:
## Description

Add static functions : `WC_Connect_Functions::is_cart()` and `WC_Connect_Functions::is_checkout()`. These functions will **return true**, if the page is a **cart page**, **checkout page**, or any pages that has **cart block** and **checkout block**.

This PR also fixes an error notice that is not displayed on cart block or checkout block.

### Related issue(s)
Fixes [#224](https://github.com/Automattic/woocommerce-shipping-issues/issues/224) for WC Shipping Issue.
Fixes [#223](https://github.com/Automattic/woocommerce-shipping-issues/issues/223) for WC Shipping Issue.
Fixes [#222](https://github.com/Automattic/woocommerce-shipping-issues/issues/222) for WC Shipping Issue.

1. Install [WC Blocks plugin](https://wordpress.org/plugins/woo-gutenberg-products-block/).
2. Enable WC Shipping & Tax debug option.
3. Create 2 pages : "Cart Block" and "Checkout Block". Add the cart block and checkout block in the content editor using gutenberg block.
image
4. Add product to the cart and go to native woocommerce checkout page.
5. Add an invalid zip code.
image
6. Go to checkout block page. The invalid zipcode error is shown in Checkout block.
image

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-functions.php
+++ b/classes/class-wc-connect-functions.php
@@ -97,23 +97,6 @@ if ( ! class_exists( 'WC_Connect_Functions' ) ) {
 				return has_block( 'woocommerce/cart' );
 			}
 
-			$page = get_post();
-			if ( ! $page ) {
-				return false;
-			}
-
-			$blocks = parse_blocks( $page->post_content );
-			if ( ! $blocks ) {
-				return false;
-			}
-
-			foreach ( $blocks as $block ) {
-				$block_name = $block['blockName'];
-				if ( 'woocommerce/cart' === $block_name ) {
-					return true;
-				}
-			}
-
 			return false;
 		}
 
@@ -126,23 +109,6 @@ if ( ! class_exists( 'WC_Connect_Functions' ) ) {
 			// To support WP < 5.0.0, we need to check if `has_block` exists first as has_block only being introduced on WP 5.0.0.
 			if ( function_exists( 'has_block' ) ) {
 				return has_block( 'woocommerce/checkout' );
-			}
-
-			$page = get_post();
-			if ( ! $page ) {
-				return false;
-			}
-
-			$blocks = parse_blocks( $page->post_content );
-			if ( ! $blocks ) {
-				return false;
-			}
-
-			foreach ( $blocks as $block ) {
-				$block_name = $block['blockName'];
-				if ( 'woocommerce/checkout' === $block_name ) {
-					return true;
-				}
 			}
 
 			return false;
--- a/classes/class-wc-connect-functions.php
+++ b/classes/class-wc-connect-functions.php
@@ -92,6 +92,11 @@ if ( ! class_exists( 'WC_Connect_Functions' ) ) {
 		 * @return bool
 		 */
 		public static function has_cart_block() {
+			// To support WP < 5.0.0, we need to check if `has_block` exists first as has_block only being introduced on WP 5.0.0.
+			if ( function_exists( 'has_block' ) ) {
+				return has_block( 'woocommerce/cart' );
+			}
+
 			$page = get_post();
 			if ( ! $page ) {
 				return false;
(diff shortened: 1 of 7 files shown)


classes/class-wc-connect-functions.php
```diff
@@ -61,11 +61,73 @@ public static function is_store_api_call() {
...
+
+			foreach ( $blocks as $block ) {
+				$block_name = $block['blockName'];
+				if ( 'woocommerce/cart' === $block_name ) {
+					return true;
+				}
+			}
+
+			return false;
+		}
```
dustinparker:
```suggestion
		public static function has_cart_block() {
			// To support WP < 5.0.0, we need to check if `has_block` exists first as has_block only being introduced on WP 5.0.0.
			if ( function_exists( 'has_block' ) ) {
				return has_block( 'woocommerce/cart' );
			}

			return false;
		}
```

dustinparker: Is it possible we could shorten this function to the suggested? It appears the `parse_blocks` function wasn't added until WP 5.0.0 either.

classes/class-wc-connect-shipping-method.php
```diff
@@ -638,9 +638,8 @@ public static function format_rate_title( $rate_title ) {
 		 * @param string $type    Notice type.
 		 */
 		public function debug( $message, $type = 'notice' ) {
-			if ( is_cart() || is_checkout() || isset( $_POST['update_cart'] ) || WC_Connect_Functions::has_cart_or_checkout_block() ) {
+			if ( is_cart() || is_checkout() || isset( $_POST['update_cart'] ) || WC_Connect_Functions::has_cart_or_checkout_block() || WC_Connect_Functions::is_store_api_call() ) {
```
dustinparker:
```suggestion
			if ( WC_Connect_Functions::is_cart() || WC_Connect_Functions::is_checkout() || isset( $_POST['update_cart'] ) || WC_Connect_Functions::is_store_api_call() ) {
```

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -379,7 +379,12 @@ public function _error( $message ) {
 			}
 
 			// if on checkout page load (not ajax), don't set an error as it prevents checkout page from displaying
-			if ( ( is_cart() || ( is_checkout() && is_ajax() ) ) && ! wc_has_notice( $message, 'error' ) ) {
+			if ( (
+					( is_cart() || ( is_checkout() && is_ajax() ) ) ||
```
dustinparker, line 383: Is it possible that `is_ajax()` should be `! is_ajax()`? The original comment seems to say that if the checkout page is loaded, not via ajax, it should set the error, but the code says if it `is_checkout` and it `is_ajax`, set the error. Or am I not understanding it correctly?

classes/class-wc-connect-functions.php
iyut: Good catch @dustinparker! thanks for catching this. I will shorten this function.

dustinparker (approved):
Everything looks good!

classes/class-wc-connect-taxjar-integration.php
iyut: yeah, from what i see, the code wants the error to be displayed if the ajax is called.

PR #2545, opened 2022-04-20
Fix/issue 2543 - Cache the 400 and 404 error for TaxJar API and validate the TaxJar request

### LINKED ISSUE #2543: Reduce number of tax rate requests
Our tax API endpoint is currently receiving a lot of `404` requests. It seems like these could be greatly reduced by the use of appropriate caching.

[Here is the method call](https://github.com/Automattic/woocommerce-services/blob/47ab8331da19b100f9a8b01e37546b9b9ad2f697/classes/class-wc-connect-taxjar-integration.php#L1021) where the extension decides whether to use a cached response or make a new request. It looks like only successful requests are cached meaning if a request results in a `404` or another kind of erroneous HTTP code, it will not be cached and will be retried.

## Acceptance criteria

1. Research when the tax requests are made and confirm the above assumptions.
   * In particular, reproduce a scenario where a `404` response is received. Does changing a form field cause the exact same request to be repeated? Does changing a form field irrelevant to tax rate calculations cause a repeated request? Etc.
2. Implement appropriate measures to prevent excessive requests from being made.
   * Maybe don't make requests that are sure to fail at all by using pre-request validation?
   * Maybe leave caching `200` requests as it is and cache erroneous responses for a shorter time, e.g. 5 minutes?

iyut:
## Description

Create a function for pre-request validation of TaxJar API calls. And cache the response for 5 minutes if the TaxJar response is 400 or 404.

### Related issue(s)
Closes #2543 

1. Enable the Automated taxes from `/wp-admin/admin.php?page=wc-settings&tab=tax`
image

2. Add any product into the cart.
3. Go to checkout page and add the address like this : 
image

4. Refresh the page again so it will try to call the TaxJar api again.
5. Go to WooCommerce >> Status page and click "WooCommerce Shipping & Tax" tab. And the TaxJar log will have something like this :
```
xx-xx-2022 @ xx:xx:xx - Retrieved the error from the cache. (WCS Tax)
xx-xx-2022 @ xx:xx:xx - Error retrieving the tax rates. Received (400): {"status":400,"error":"Bad Request","detail":"to_zip 90215 is not used within to_state CA"} (WCS Tax)
```

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.27 - 2022-xx-xx =
+* Tweak - Validate the TaxJar request before calling the api and cache 404 and 400 TaxJar response error for 5 minutes.
+
 = 1.25.26 - 2022-04-19 =
 * Fix   - Display error on cart block and checkout block from WC Blocks plugin.
 * Fix   - TaxJar does not calculate Quebec Sales Tax when shipping from Canadian address.
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1162,6 +1162,53 @@ class WC_Connect_TaxJar_Integration {
 	}
 
 	/**
+	 * Validate TaxJar API request json value and add the error to log.
+	 *
+	 * @param $json
+	 *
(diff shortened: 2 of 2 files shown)


dustinparker (approved):
Looks good!

PR #2546, opened 2022-04-22
Fix/issue 2544 - Add the correct tax code for taxable, non-taxable, and shipping-only-taxable product. 

### LINKED ISSUE #2544: Investigate issues related to taxed vs. untaxed shipping based on cart items.
It seems there are at least a few open issues related to errors in how shipping is taxed (or not) based on cart contents and various states' rules about tax on shipping services.

There was a recent report that if the user's store is based in PA, the cart contains a non-taxable product, and the customer is shipping to PA, there is tax incorrectly charged for shipping. [Pennsylvania law](https://www.taxjar.com/blog/retail/shipping-pennsylvania-taxable) states that shipping should only be taxed if _any_ of the cart contents are also taxable.

Possible related issues:
#2428 
#1226
And it looks like this desired behavior was added in #2424 but I'm not sure if something else has changed in the codebase to cause a regression, or if that fix didn't quite fix all of the edge cases here.

cc @harriswong

iyut:
## Description
Make non-taxable line items will always be listed in TaxJar request.   

### Related issue(s)
Closes #2544 

1. Set the store location in Pennsylvania. Example : 
image
2. Create a non-taxable product and taxable product.
3. Add those products to the cart.
4. Go to checkout page and set the customer location based on Pennsylvania. Example : 
screenshot-localhost_8050-2022 04 22-21_17_12
5. The tax will be **calculated**.
6. Remove the taxable product.
7. The tax in the cart will **not be calculated**.
8. Add the product with taxable for shipping-only into the cart.
9. The tax in the cart will be calculated for shipping only. 

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 1.25.27 - 2022-xx-xx =
+* Fix   - Cart with non-taxable product still calculate the tax.
+
 = 1.25.26 - 2022-04-19 =
 * Fix   - Display error on cart block and checkout block from WC Blocks plugin.
 * Fix   - TaxJar does not calculate Quebec Sales Tax when shipping from Canadian address.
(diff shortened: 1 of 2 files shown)


dustinparker (approved):
Looks good!

PR #2555, opened 2022-05-23
Enable shipping tax by default for Florida intrastate shipping - Tweak/issue 2531

### LINKED ISSUE #2531: BUG: TAX: Florida sales tax on shipping if NO local delivery is available.
We have [a forum user](https://wordpress.org/support/topic/tax-shipping-cost-in-florida/#post-15422906) who's nexus is in Florida. For shipping, tax should be added if there is **no option given** to pick up locally. Automated taxes always defaults to no shipping tax. 

>Delivery charges are not subject to tax when the charge is separately stated on an invoice or bill of sale **and** the purchaser has the option of picking the item up or arranging for third-party  #transportation services on its own.
https://wordpress.org/support/topic/tax-shipping-cost-in-florida/#post-15422906

Is there a way of allowing this specification into the automated taxes?

dustinparker:
## Description

Normally TaxJar enables taxes on shipping by default for Florida to Florida shipping, but because WooCommerce uses a single account, a nexus cannot be added for Florida (or any state) which means the shipping tax is not enabled. So, we will enable it here by default and give merchants the option to disable it if needed via filter.
### Related issue(s)

Fixes #2531 

1. Set your store address to a Florida address:

store settings

2. Set your tax settings to the following:

tax settings

3. Add a shipping zone for Florida and add a Flat Rate method as shown:

flat rate settings

4. Add a "Beanie with logo" to cart and go to checkout.
5. Set a Florida destination address.
6. You should see the following tax rate and total tax returned:

tax rate

checkout view

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
-= 1.25.29 - 2022-xx-xx =
-* Add - Tool to clear cached Tax server responses from the transients.
+= 1.26.0 - 2022-xx-xx =
+* Add   - Tool to clear cached Tax server responses from the transients.
+* Tweak - Enable shipping tax by default if is Florida interstate shipping.
 
 = 1.25.28 - 2022-05-12 =
 * Fix   - Notice: Undefined index: 'from_country' when validating TaxJar request.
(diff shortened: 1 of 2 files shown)


iyut (approved):
I tested this by following the steps in the description. Here is the result : 
image

LGTM!

PR #2561, opened 2022-06-14
Fix/issue 2249 - Add filter around TaxJar class to give a user an option to override TaxJar result

### LINKED ISSUE #2249: No way to override rates from TaxJar if they are incorrect
Following up to the thread here:
p1604238751225700-slack-C7U3Y3VMY

In this particular instance the customer would like to override tax rates for Tenerife to 0%. This is due to although Tenerife is located within the EU, they do not get charged VAT. TaxJar is retuning VAT rates and applying them to customers with addresses in Tenerife. 

This means there are two problems:

* TaxJar is returning incorrect rates.
* There is no way to override rates returned from TaxJar.

Ticket is here: 
3428621-zen

I tagged this as a High priority Bug due to I believe it's on us to look into the incorrect rates we're providing our customers, and to provide a work around if possible.

iyut:
## Description
The current TaxJar class doesnt have a way for the user to override TaxJar result. This PR add a couple of filter that the user can use to override the TaxJar result.

### Related issue(s)

Closes #2249 

1. Add this snippet in your `theme/functions.php` file : 
```
function change_taxjar_tax( $tax_value, $taxes, $body ) {
	if ( 'DE-BE' === $body['to_state'] && '10115' === $body['to_zip'] ) {
		$tax_value = 0.19;
	}
	return $tax_value;
}

add_filter( 'woocommerce_services_override_tax_rate', 'change_taxjar_tax', 10, 3 );
```
2. Enable `Automatic Tax Rate` from **WooCommerce >> Settings >> Tax**.
3. Add any product to the cart and checkout with this shipping address :
`Berlin, Germany, 10178`.
4. Check the tax rate in the cart. It will always be 19%. 

### Checklist

- [ ] unit tests


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1008,13 +1008,20 @@ public function calculate_tax( $options = array() ) {
 			$body['line_items'] = $line_items;
 		}
 
+		$skip_taxjar_request_taxes = apply_filters( 'woocommerce_services_skip_taxjar_request', false, $taxes, $body );
```
bartech: No need to send `$taxes` in this filter, at this point `$taxes` will always be array with placeholder values.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1008,13 +1008,20 @@ public function calculate_tax( $options = array() ) {
...
+
+		if ( false !== $skip_taxjar_request_taxes ) {
+			// 4th parameter set to true indicate the this filter is skipped the TaxJar request.
+			return apply_filters( 'woocommerce_services_taxjar_calculate_tax_result', $skip_taxjar_request_taxes, null, $body, true );
```
bartech: Suggest to rename this filter to `woocommerce_services_custom_tax_result` pass `$taxes` as first parammeter to convey what shape returned value should be. Think using same filter name in different places will be more confusing. It will require developer to exemine values passed by filter to know in which place same filter was triggered.
```suggestion
return apply_filters( 'woocommerce_services_custom_tax_result', $taxes, $body );
```

bartech (changes requested):

iyut:
@bartech , I have made some changes for this PR and the description on how to test it. Could you try to pull and have a test again? And also let me know what you think.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1008,7 +1045,9 @@ public function calculate_tax( $options = array() ) {
 			$body['line_items'] = $line_items;
 		}
 
-		$response = $this->smartcalcs_cache_request( wp_json_encode( $body ) );
+		// Use this filter to custom the response.
+		$custom_response = apply_filters( 'woocommerce_services_custom_response', array(), $taxes, $body );
```
bartech: Suggest to remove this filter also this PR. Let's move it to another PR. Will be good to use similar approach as with `maybe_override_taxjar_tax`

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -887,6 +887,43 @@ public function append_base_address_to_customer_taxable_address( $address ) {
...
 
+	/**
+	 * This method is used to override the TaxJar result.
+	 *
+	 * @param object $taxjar_resp_tax TaxJar response object.
+	 * @param array  $body            Body of TaxJar request.
+	 *
+	 * @return object
+	 */
+	public function override_taxjar_tax( $taxjar_resp_tax, $body ) {
```
bartech: Let's rename this method to `maybe_override_taxjar_tax`

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -887,6 +887,43 @@ public function append_base_address_to_customer_taxable_address( $address ) {
...
+		$new_tax_rate = apply_filters( 'woocommerce_services_override_tax_rate', $taxjar_resp_tax->rate, $taxjar_resp_tax, $body );
+
+		if ( $new_tax_rate !== $taxjar_resp_tax->rate ) {
```
bartech: Let's revert condition check and do early return if `$new_tax_rate === $taxjar_resp_tax->rate` before comparing let's cast each value to float by using `floatval`

bartech (changes requested):

bartech (approved):

Issue #2536, opened 2022-03-12
Remove `colors` package


frosso:
## Description
Background: p6jPRI-5cd-p2

Some time ago, the author of some popular npm packages [Marak](https://href.li/?https://www.npmjs.com/~marak) decided to delete some packages they authored. To prevent future incidents, let's see if we can find an alternative.

It looks like this package is only used in the `tasks/release.js` task.
Maybe it can be as simple as finding an alternative package or just duplicating the `colors.cyan`/`colors.green`/`colors.red` wrappers behavior.

bartech:
Resolved in #2566

PR #2563, opened 2022-06-16
Replace colors package with chalk in release.js script

bartech:
## Description

Replace colors npm package with chalk in `tasks/release.js` script for building plugin zip.

### Related issue(s)
#2536 

run `npm i && npm run release`
Release script should work as before change and show colored messages. 

### Checklist

- [x] unit tests

No changelog entry as this is development dependency change.


bartech:
Not sure if I should commit changes in `npm-shrinkwrap.json` as this is what was changed after I added chalk and removed colors as dependency,

iyut (changes requested):
LGTM, only minus with changelog. Can you add the changelog please?

iyut:
> Not sure if I should commit changes in `npm-shrinkwrap.json` as this is what was changed after I added chalk and removed colors as dependency,

Since we're replacing one of their library, I believe we need to commit the npm-shrinkwrap as well. But it will be better if you could confirm with them from A8C slack group first.

bartech:
@iyut 
> Can you add the changelog please?

Though changelog is only for customer facing features or bugs. This change is only developer side don't think we want to inform plugin users about internals.

iyut:
> This change is only developer side don't think we want to inform plugin users about internals.

Do you mind to check in changelog file if there is any log for internal changes? I thought there are a couple of internal changes being listed in there.

bartech:
I'm closing this PR as this fix is also included in PR #2566 that address E2E and JS Test fixes.
I branched from fix/issue-2536 branch and think this will be easiest to continue.
Please continue review on new PR.

PR #2573, opened 2022-07-22
Fix/issue 2570 - Always let the user pick the package box

### LINKED ISSUE #2570: Always allow to manually confirm/select package while printing label
**Describe the bug**

When the destination address is auto verified as valid by the system it will offer you to select the package as shown below:

https://d.pr/i/Ikv173+

This is not the case where the destination address needs to be reviewed / we need to use suggested address. The system skips the package section and directly offers the shipping rates:

https://d.pr/i/6Fys6w+

**Expected behavior**

If the destination address cannot be auto-validated / needs an update, or you need to use a suggested address, once you're done with the address, the system should still let you select the shipping package once the destination address is correctly set up.

**Steps to replicate**
1. Attempt to create an order with a valid/verified address
2. Notice that the shipping from and destination addresses are automatically validated, and you're offered to select a package
3. Now, create an order with a typo in the address
4. Notice that once you fix the address under `Destination Address` tab, you're automatically offered the shipping rates - instead of the tab to select the packaging

**Screen Recording:**

https://d.pr/i/hB7WQF+

**Workaround**

* Note that you can still manually check the dropdown option and select a different package
* It is just that automatically skipping the package selection interface adds an additional click/step to the process

iyut:
## Description
Removing the default box selection function to always let the user pick the package box.

### Related issue(s)

Fixes #2570

1. Attempt to create an order with a valid/verified address
2. Notice that the shipping from and destination addresses are automatically validated, and you're offered to select a package
3. Now, create an order with a typo in the address
4. Notice that once you fix the address under Destination Address tab, you wil always need to pick the package box manually.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
 + 1.26.3 - 2022-xx-xx =
+* Tweak - Always let the user to pick the package box.
 * Add   - Add filter to override TaxJar result.
 
 = 1.26.2 - 2022-07-04 =
--- a/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
@@ -347,12 +347,7 @@ export const openPrintingFlow = ( orderId, siteId ) => ( dispatch, getState ) =>
 
 	waitForAllPromises( promisesQueue ).then( () =>
 		tryGetLabelRates( orderId, siteId, dispatch, getState )
-	).then( () => {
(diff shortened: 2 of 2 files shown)


dustinparker (approved):
Works as expected!

harleec61:
Works as expected, but not as we requested in the original report. The original request was to simply not skip the package selection tab when there's an address verification step. Everyone was happy with the persistent setting of the "Package Details" field, unless there is the need to change packages. Over 95% of our shipments use the same package, so now we're forced to reselect the same package 100's of times per day. Is it possible to not skip the Packaging tab but don't force selection of a package? We typically only need to change the weight, not the package.

PR #2595, opened 2022-11-08
Replace calypso/form-radio with internal FormRadio

harriswong:
## Description
Similar to https://github.com/Automattic/woocommerce-services/pull/2344, There is no comparable WordPress nor WooCommerce FormRadio. I moved the Calypso submodule's FormRadio over.

I also appended `wcs-client` to import so we know this has been migrated.

I considered using the WordPress component `RadioControl` (https://developer.wordpress.org/block-editor/reference-guides/components/radio-control/#development-guidelines). But the `RadioControl` actually renders all the radios for the selection. Versus the `FormRadio` here represents 1 radio. In various places, we have our own `RadioButton` that wraps `FormRadio`. It's not a simple swap-in replacement. I think we can revisit and cleanup the whole `client/components` folder once Calypso is removed. 

### Related issue(s)
Relates to https://github.com/Automattic/woocommerce-services/issues/2333

1. wp-admin > Order > Pick an order > Click "Create new label" or print label
2. Make sure the address verifying step's radio button is ok
image
3. Then, make sure the radio buttons in the packaging step is okay
image

### Checklist

- [ ] unit tests


frosso (approved):
Looks straightforward 👍 any styles to be migrated?

harriswong:
Nope, there is only 1 file
image

frosso (approved):
Thanks for confirming! 👍

Issue #2002, opened 2020-04-13
Spike: Determine which WCS flows need E2E tests


c-shultz:

c-shultz:
@Ferdev, already created some good issues for this, so these should be included in the master thread

**Tests:**
https://github.com/Automattic/woocommerce-services/issues/1915
https://github.com/Automattic/woocommerce-services/issues/1918
https://github.com/Automattic/woocommerce-services/issues/1917

**Infrastructure Improvements:**
https://github.com/Automattic/woocommerce-services/issues/1914
https://github.com/Automattic/woocommerce-services/issues/1898

c-shultz:
For getting started with Looker, there's some good in the comments here: https://github.com/Automattic/woocommerce-shipping-issues/issues/22

harriswong:
I agree to focusing on identifying and testing the critical flows, but not every use-cases. End to end test takes long time to run and it also has a high maintenance cost. I vote for the "test pyramid" approach where we focus more on the bottom of the pyramid than the top.

c-shultz:
As part of this, let's also look at what Proton has done to run core tests in Docker to make debugging tests much easier. We should determine how easy this would be for us to adopt and how much value we think we can get from it.

c-shultz:
+1 to focusing on the critical flows. Let's also update the linked issues as part of this to be really clear about what steps/flow the E2E should follow, so none of our PRs get lost in testing every possible use case.

#1915
#1918
#1917
#1916

Ferdev:
hey @c-shultz, was this comment intended for another issue? https://github.com/Automattic/woocommerce-services/issues/2002#issuecomment-613142168

c-shultz:
Lol. yep, I'm not sure what happened there. I had a few issues open at the same time.

harriswong:
### User flow:
A user trying to accomplish a goal and how this user click through WCS. Some of these user flow are the following. If any of these failed, the user can't buy label and that means WCS is not working properly:
- Purchasing label
- Reprinting label
- Refunding label
	- Purchasing label again after refund
- Track label
- WCS settings
- WCS banner from WC

### Order status
How the user interact with WCS also depends on the order status. New order and existing order should provide different user flow. Some of the following overlaps the user flow listed above. But I wanted to list them out and consolidate after.
#### New order
- Click create shipping label, click buy shipping label
- Shipping label modal
	- Filling the form and click "buy shipping label"
	- Test 1 item 
[...]

PR #2598, opened 2022-11-09
Add/wcc server issue 1974 - Add EndShipper information to be sent to WCC Server

iyut:
## Description
**STILL IN PROGRESS**
This PR is related with [WCC Server PR #2002](https://github.com/Automattic/woocommerce-connect-server/pull/2002) 
The purpose of this changes is to send WC Store Address, WP.com Email, and origin address to WCC Server as EndShipper information

### Related issue(s)
Closes [WCC Server Issue 1974](https://github.com/Automattic/woocommerce-connect-server/issues/1974)

1. Pull your WCC Server `master` branch in your local repo.
2. Checkout your WCC Server Environment with `git checkout add/issue-1974`.
3. Run `nvm i` and `npm i` on WCC Server repo to install the latest **EasyPost API** library.
4.  Create a new order.
5. Try to create a shipping label from order edit page in admin area.
6. Check `wcc_endshipper` table from `http://localhost:18080` and make sure there is a no new endshipper record there.
7. Now, try to change `wcs_version` from `2.0.0` to `1.26.2` on this [this line](https://github.com/Automattic/woocommerce-connect-server/blob/bf07d440ef7ad2ebcfca9ac612aaa9d9ebe5cab6/lib/shipping/purchase-labels.js#L81)
8. And create a new order and shipping label.
9. `wcc_endshipper` table will have a new record.

### Checklist

- [ ] unit tests


classes/class-wc-connect-service-settings-store.php
```diff
@@ -110,6 +110,31 @@ public function can_user_manage_payment_methods() {
...
 		}
 
+		/**
+		 * Gets the store's location settings.
+		 *
+		 * @return array
+		 */
+		public function get_endshipper_address() {
+			$master_user    = WC_Connect_Jetpack::get_master_user();
+			$connected_data = WC_Connect_Jetpack::get_connected_user_data( $master_user->ID );
```
dustinparker:
```suggestion
			$email       = '';
			$master_user = WC_Connect_Jetpack::get_master_user();

			if ( is_a( $master_user, 'WP_User' ) ) {

				$connected_data = WC_Connect_Jetpack::get_connected_user_data( $master_user->ID );

				if ( is_array( $connected_data ) && ! empty( $connected_data['email'] ) ) {
					$email = $connected_data['email'];
				}
			}
```

classes/class-wc-connect-service-settings-store.php
```diff
@@ -110,6 +110,31 @@ public function can_user_manage_payment_methods() {
...
+
+			$store_settings = array(
+				'company'   => html_entity_decode( get_bloginfo( 'name' ), ENT_QUOTES ), // HTML entities may be saved in the option.
+				'address_1' => WC()->countries->get_base_address(),
+				'address_2' => WC()->countries->get_base_address_2(),
+				'city'      => WC()->countries->get_base_city(),
+				'state'     => WC()->countries->get_base_state(),
+				'country'   => WC()->countries->get_base_country(),
+				'postcode'  => WC()->countries->get_base_postcode(),
+				'email'     => $connected_data['email'],
```
dustinparker:
```suggestion
				'email'     => $email,
```

classes/class-wc-connect-service-settings-store.php
dustinparker: this might be too cautious but I thought it might make sense to be sure those Jetpack methods return the expected types before using.

dustinparker (approved):
Everything works as expected! Nice job! I just added a code suggestion that might help avoid some bugs down the road.

dustinparker (approved):
Just reviewed the latest changes. Looks good!

changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.1.0 - 2022-xx-xx =
+* Add - USPS EndShipper feature.
```
waclawjacek: Let's go with something that will be a bit more descriptive for users, like:

```suggestion
* Add - Compatibility with USPS's end shipper requirements.
```

classes/class-wc-connect-shipping-label.php
```diff
@@ -274,6 +274,12 @@ protected function format_address_for_api( $address ) {
 			return $address;
 		}
 
+		protected function get_endshipper_address() {
+			$endshipper = $this->format_address_for_api( $this->settings_store->get_endshipper_address() );
+
+			return $endshipper;
+		}
+
```
waclawjacek: It looks like this method can be deleted and `format_address_for_api` call inlined into `get_origin_address()` below.

```suggestion
```

client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
```diff
@@ -1036,6 +1036,15 @@ const pollForLabelsPurchase = ( orderId, siteId, dispatch, getState, labels ) =>
 	downloadAndPrint( orderId, siteId, dispatch, getState, labels );
 };
 
+const completingEndShipperAddress = ( form ) => {
```
waclawjacek:
```suggestion
const appendEndShipperAddress = ( form ) => {
```

classes/class-wc-connect-service-settings-store.php
```diff
@@ -110,6 +110,40 @@ public function can_user_manage_payment_methods() {
 				( is_a( $master_user, 'WP_User' ) && $current_user->ID === $master_user->ID );
 		}
 
+		/**
+		 * Gets the store's location settings.
+		 *
+		 * @return array
+		 */
+		public function get_endshipper_address() {
```
waclawjacek: Let's stay consistent in nomenclature, i.e. let's call this either "end shipper" or "endshipper". I think "end shipper" is a valid term so let's replace this with:

```suggestion
		public function get_end_shipper_address() {
```

classes/class-wc-connect-service-settings-store.php
```diff
@@ -110,6 +110,40 @@ public function can_user_manage_payment_methods() {
...
+			$master_user = WC_Connect_Jetpack::get_master_user();
+
+			if ( is_a( $master_user, 'WP_User' ) ) {
+
+				$connected_data = WC_Connect_Jetpack::get_connected_user_data( $master_user->ID );
+
+				if ( is_array( $connected_data ) && ! empty( $connected_data['email'] ) ) {
+					$email = $connected_data['email'];
+				}
+			}
```
waclawjacek: Let's extract this part to a separate method, e.g. `get_end_shipper_email()`.

classes/class-wc-connect-service-settings-store.php
```diff
@@ -110,6 +110,40 @@ public function can_user_manage_payment_methods() {
...
+		/**
+		 * Gets the store's location settings.
+		 *
+		 * @return array
+		 */
+		public function get_endshipper_address() {
+			$email       = '';
+			$master_user = WC_Connect_Jetpack::get_master_user();
+
+			if ( is_a( $master_user, 'WP_User' ) ) {
```
waclawjacek:
```suggestion
			if ( $master_user instanceof WP_User ) ) {
```

classes/class-wc-connect-service-settings-store.php
```diff
@@ -110,6 +110,40 @@ public function can_user_manage_payment_methods() {
...
+
+				$connected_data = WC_Connect_Jetpack::get_connected_user_data( $master_user->ID );
+
+				if ( is_array( $connected_data ) && ! empty( $connected_data['email'] ) ) {
```
waclawjacek, line 120: Should we exit if we couldn't retrieve an email?

```suggestion
				if ( empty( $connected_data['email'] ) ) {
					throw new SOME_EXCEPTION;
```

classes/class-wc-connect-service-settings-store.php
```diff
@@ -110,6 +110,40 @@ public function can_user_manage_payment_methods() {
...
+				'city'      => WC()->countries->get_base_city(),
+				'state'     => WC()->countries->get_base_state(),
+				'country'   => WC()->countries->get_base_country(),
+				'postcode'  => WC()->countries->get_base_postcode(),
+				'email'     => $email,
+				'phone'     => '',
+				'name'      => '',
+			);
+
+			return apply_filters( 'endshipper_store_settings', $store_settings, array() );
```
waclawjacek: I think this was accidentally left over.

```suggestion
			return apply_filters( 'endshipper_store_settings', $store_settings );
```

client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
```diff
@@ -258,14 +258,15 @@ export const isCustomsFormRequired = createSelector(
 				'An origin address phone number is required for this shipment.'
 			);
 		} else if (
+			'US' === country &&
```
waclawjacek, line 261: Why do we have this check here now?

client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
```diff
@@ -258,14 +258,15 @@ export const isCustomsFormRequired = createSelector(
...
 			10 !==
 			phone
 				.split( /\D+/g )
 				.join( '' )
 				.replace( /^1/, '' ).length
 		) {
 			errors.phone = translate(
-				'Customs forms require a 10-digit phone number. ' +
+				'Origin address require a 10-digit phone number. ' +
```
waclawjacek: Do we hijack the customs flow to validate the origin address?

```suggestion
				'Origin phone number has to be 10-digit. ' +
```

waclawjacek (commented):
Just posting some comments. Thanks for the hard work!

classes/class-wc-connect-service-settings-store.php
```diff
@@ -110,6 +110,40 @@ public function can_user_manage_payment_methods() {
...
+				'company'   => html_entity_decode( get_bloginfo( 'name' ), ENT_QUOTES ), // HTML entities may be saved in the option.
+				'address_1' => WC()->countries->get_base_address(),
+				'address_2' => WC()->countries->get_base_address_2(),
+				'city'      => WC()->countries->get_base_city(),
+				'state'     => WC()->countries->get_base_state(),
+				'country'   => WC()->countries->get_base_country(),
+				'postcode'  => WC()->countries->get_base_postcode(),
+				'email'     => $email,
+				'phone'     => '',
+				'name'      => '',
```
waclawjacek, line 146: Could you please describe where these will come from? Thanks!

classes/class-wc-connect-service-settings-store.php
iyut: Sure, let me do that. I think it will make the code cleaner.

classes/class-wc-connect-service-settings-store.php
iyut: Yes, let's use "end shipper". Just want to confirm, in WCC Server, I have created a new table for end shippper with the name : `wcc_endshipper`. Should we change it to `wcc_end_shipper` too?

client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
iyut: I thought 10 digit phone number only related to US phone number?

classes/class-wc-connect-service-settings-store.php
iyut: Yes, I have added a comment on top of this code to explain it.

client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
iyut: Before this changes, origin phone number only required if customs form is also required. Now since origin phone number is always required I thought we could change the `errors.phone` text here with origin address related message.

classes/class-wc-connect-service-settings-store.php
waclawjacek: I think that'd be nice. That way we don't have to wonder if we separated these words or not.

classes/class-wc-connect-service-settings-store.php
```diff
@@ -110,6 +110,45 @@ public function can_user_manage_payment_methods() {
...
+		}
+
+		/**
+		 * Gets the store's location settings.
+		 *
+		 * @return array
+		 */
+		public function get_end_shipper_address() {
+			$store_settings = array(
+				'company'   => html_entity_decode( get_bloginfo( 'name' ), ENT_QUOTES ), // HTML entities may be saved in the option.
```
waclawjacek, line 135: Should we use the site name as the company name? Maybe we should also use the company name from the origin form? cc @laurajohnsonwoo @harriswong

classes/class-wc-connect-service-settings-store.php
```diff
@@ -110,6 +110,45 @@ public function can_user_manage_payment_methods() {
...
+				'country'   => WC()->countries->get_base_country(),
+				'postcode'  => WC()->countries->get_base_postcode(),
+				'email'     => $this->get_end_shipper_email(),
+
+				// Phone and name data will be added later from origin address.
+				'phone'     => '',
+				'name'      => '',
+			);
+
+			return apply_filters( 'end_shipper_store_settings', $store_settings );
```
waclawjacek: Just to avoid running into collisions with any potential other extensions doing this.

```suggestion
			return apply_filters( 'wcship_end_shipper_store_settings', $store_settings );
```

client/extensions/woocommerce/woocommerce-services/lib/initialize-labels-state/index.js
```diff
@@ -106,6 +106,9 @@ export default data => {
 				normalizationInProgress: false,
 				allowChangeCountry: Boolean( canChangeCountries ),
 			},
+			endShipper: {
+				values: formData.end_shipper,
```
waclawjacek, line 110: I noticed the other fields use other keys than just `values`. Do you know how this works?

classes/class-wc-connect-service-settings-store.php
laurajohnsonwoo: As the Company Name is optional, if it's in the Origin Form then I would like to use that, but if not, the merchant's name should suffice. Do we always have that though? 

I hesitate to use the site name as it might not be accurate, but it's a good backup option if we need to use it!

harriswong: Maybe we need a new section inside settings to allow endshipper-specific info as a follow-up ticket? By default, we will fall back to "what Woo thinks" is the best. But with the settings page, it will provide an option to overwrite our guess.

client/extensions/woocommerce/woocommerce-services/lib/initialize-labels-state/index.js
iyut: I see : `isNormalized`, `normalized`, `ignoreValidation`, `selectNormalized`, `normalizationInProgress`, `allowChangeCountry` keys on the other object. 

And from what I see, these keys : `ignoreValidation`, `selectNormalized`, `normalizationInProgress`, `allowChangeCountry` are meant for flags in the address UI. And since end shipper doesn't have UI at the moment, IMO we won't need to use that.

`normalized` is address info that has been normalized and saved into database (e.g. normalized origin address is saved on `wp_options` under `wc_connect_origin_address` option name ).

`isNormalized` is used for a flag whether the current address is used normalized address or not.

I did some simple tests when sending an endshipper address like this : 
```
street1: 60 29TH ST,
street2: # 343,
city: SAN FRANCISCO,
state: CA,
zip:94110,
country: US
```
endShipper API seems responding with normalized address like below : 
```
street1: 60 29TH ST # 343,
street2: null,
city: SAN FRANCISCO,
state: CA,
zip:94110-4929,
country: US
```

Wondering if we need to save this normalized address in `wp_options` like origin address do? Or should we just leave it as we don't save the end shipper address in the WCC Server DB?

classes/class-wc-connect-service-settings-store.php
iyut: Thanks for the advice, I have separated `endshipper` word into `end_shipper` or `endShipper` on some variables.
I have done this on both wcc services and wcc server.

classes/class-wc-connect-service-settings-store.php
iyut: It seems there is no exception catch around this code yet. So if we throw any exception, it will show a fatal error instead.

What I can suggest probably, we add a try/catch around [this code](https://github.com/Automattic/woocommerce-services/blob/4b160bcfb1917b92450a86893cbc292efd374af7/classes/class-wc-connect-shipping-label.php#L554-L577). Here is how we can implement it : 

```
public function meta_box( $post, $args ) {
			try {
				$connect_order_presenter = new WC_Connect_Order_Presenter();
				$order                   = WC_Connect_Compatibility::instance()->init_theorder_object( $post );
				...
				...
				...

				do_action( 'enqueue_wc_connect_script', 'wc-connect-create-shipping-label', $payload );
			} catch ( Exception $e ) {
				?>
				<strong><?php echo sprintf( esc_html__( 'Error: %1$s', 'wocommerce-services' ), $e->getMessage() ); ?></strong>
				<?php
			}
		}
```
WDYT?

waclawjacek:
It looks like we won't be merging this PR after all, since the Connect Server now supports figuring out the end shipper without any changes to the extension. We will be figuring out a dedicated UI in the future.

cc @laurajohnsonwoo @harriswong Please reopen if this is incorrect!

PR #2601, opened 2022-11-09
Copy calypso Notice component into WCS&T

harriswong:
## Description
Copied from wordpress/wp-content/plugins/woocommerce-services/wp-calypso/client/components/notice.

## Considerations
I tried using WP's `<Notice>` component (https://developer.wordpress.org/block-editor/reference-guides/components/notice/#development-guidelines). But the style looks too different.

This is our current one: 
image

This is the WP's style one:
image

I decided to stay with our existing one by reusing the Calypso's component.

### Related issue(s)

1. Go to label printing dialog
2. The easiest to test this is to force the `<Notice>` to show in the code. 
3. Go to [this line](https://github.com/Automattic/woocommerce-services/blob/0afd96fbf88967598cc98ee4179deaf66b1f1d1e/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/address-step/fields.js#L140) and change the condition to `true ||`.
4. Confirm notice showing okay.

### Checklist

- [x] unit tests


waclawjacek (approved):
LGTM! :shipit:

But! I am now getting this when running `npm start`:

```
WARNING in ./client/components/notice/index.jsx
Module Warning (from ./node_modules/thread-loader/dist/cjs.js):

/Users/waclaw/Sites/fulfill/wp-content/plugins/woocommerce-services/client/components/notice/index.jsx
  122:6   error  Visible, non-interactive elements with click handlers must have at least one keyboard listener  jsx-a11y/click-events-have-key-events
  122:6   error  Static HTML elements with event handlers require a role                                         jsx-a11y/no-static-element-interactions
  122:12  error  `tabIndex` should only be declared on interactive elements                                      jsx-a11y/no-noninteractive-tabindex

✖ 3 problems (3 errors, 0 warnings)

 @ ./client/components/error-notice/index.js 19:0-39 24:29-35
 @ ./client/apps/print-test-label/view.js
 @ ./client/apps/print-test-label/index.js
 @ ./client/main.js
 @ multi ./client/main.js
ℹ ｢wdm｣: Compiled with warnings.
```

Please create an issue to fix these.

harriswong:
I took a look at Calypso's trunk https://github.com/Automattic/wp-calypso/blob/5df33323fb7e3a35334284486cc303c437e32add/client/components/notice/index.jsx#L136 but the code looks a bit different.

I patched this by adding role and onKeyPress. The warnings should be gone now. Tested again on the address validator, looks good.
image

Rebased latest trunk. @waclawjacek , please take a look again!

client/components/notice/index.jsx
```diff
@@ -0,0 +1,132 @@
...
+				<span className="notice__content">
+					<span className="notice__text">{ text ? text : children }</span>
+				</span>
+				{ text ? children : null }
+				{ showDismiss && (
+					<span tabIndex="0" className="notice__dismiss" onClick={ onDismissClick } onKeyDown={onDismissClick} role="button" aria-label={ translate( 'Dismiss' ) }>
+						<Gridicon icon="cross" size={ 24 } />
+						<ScreenReaderText>{ translate( 'Dismiss' ) }</ScreenReaderText>
+					</span>
+				) }
```
waclawjacek, line 126: Could we maybe just turn it into a button instead of providing tabIndex, role, and onKeyDown?

harriswong: We could make it closer to what Calypso uses: https://github.com/Automattic/wp-calypso/blob/5df33323fb7e3a35334284486cc303c437e32add/client/components/notice/index.jsx#L132-L139

waclawjacek: Yes please, let's!

harriswong:
Use button instead. Tested a few notices below:
image
image

client/components/notice/style.scss
```diff
@@ -57,15 +57,15 @@
 		}
 	}
 
-	.notice__dismiss {
+	.notice__wcs-dismiss {
```
harriswong: Using a custom WCS&T dismiss class to overwrite the position and colour of the dismiss button. Reason being is because WCS&T loads a few other custom styles which also include `.notice__dismiss`. For example,
image


In the above, we can see that the `.global-notices` and the media query overwrites the location of the `X`. We could modify padding inside the global-notices classes but that could undo changes like [this](https://github.com/Automattic/woocommerce-services/pull/2130).

Since other plugin or WCS&T itself can have the same Calypso's notice component style, WCS&T can have its own padding overwrite.

client/components/notice/index.jsx
harriswong: Please take a look again.

waclawjacek (changes requested):
There is a regression in terms of accessibility. I am testing this in the "Create shipping label" modal by disabling wifi and then clicking here:

Markup on 2022-11-29 at 15:57:48

Then, once the error pops up, I click the static text in the error message and press `Tab` on my keyboard.

On `trunk`, this results in:

Markup on 2022-11-29 at 15:55:36

(Not ideal but there is some indication of what is focused.)


On this branch, this results in:

Markup on 2022-11-29 at 15:56:33


(So no change between before and after pressing `Tab`.)

harriswong:
Should we then go back to use the old code where it has the `tabIndex="0"`?

harriswong:
@waclawjacek To elaborate, if we use the old Calypso code, then the functionality doesn't change. The linter will complain and we will add tabIndex, role, and onKeyDown to span, not too bad.

If we use the newer Calypso's code but not the whole package, we have collisions with styles, missing tab indexes. 

My preference is to rollback [this part](https://github.com/Automattic/woocommerce-services/pull/2601#discussion_r1033921809) and keep the index, role and keydown. Thoughts?

waclawjacek:
Yeah, I agree - let's go with replicating what the version of Calypso used by us up until now does.

harriswong:
Reverted 7c32565a52ea3b1bae26b454c29b3e6e72babfee and did a quick test on the notice.
image

@waclawjacek please take a look again

waclawjacek (approved):
LGTM! :shipit:

PR #2603, opened 2022-11-16
Copy SimplifiedSegmentedControl from Calypso into WCS&T

harriswong:
## Description
Copy `<SimplifiedSegmentedControl>` from calypso into ours.

### Related issue(s)

### Considerations
I looked into replacing the `<SimplifiedSegmentedControl>` with WP's `TabPanel` https://developer.wordpress.org/block-editor/reference-guides/components/tab-panel/. Something like this,
```js
				<TabPanel
					className="packages__mode-select"
					activeClass="active-tab"
					initialTabName={mode}
					onSelect={ switchMode }
					tabs={[
						{ name: 'edit', title: 'Edit package' },
						{ name: 'add-custom', title: 'Custom package' },
						{ name: 'add-predefined', title: 'Service package' },
					]}
				>
					{ (  ) => {
						if (showSegmentedControl) {
							return  <EditPackage { ...props } />
						} else if ( showEdit )
							return  <EditPackage { ...props } />
						else if (showPredefined)
							return  <PredefinedPackages { ...props } />
					} }
				</TabPanel>
```
image

The `SimplifiedSegmentedControl` component and `TabPanel` works differently. We will need to rewrite some parts of the `PackageDialog`. If we are rewriting some parts, it may be better to wait until we delete the calypso submodule, and upgrade nodejs version. We may also need to figure out whether or not WCS&T should be using WC components or WP components. In the event we choose to use WC components, then the rewrite will be different than using `TabPanel` above.

All that considered, I think copying the `<SimplifiedSegmentedControl>` over is the best value step toward deleting the wp-calypso submodule goal.

1. Go to http://localhost/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings
2. Under "Packaging", click "Edit", an "Edit Package" dialogue should show up
image
3. Next, go to an order and print a label. In "Packaging", click "Add package"
4. The "Custom package" and "Service package" should continue to show
image

### Checklist

- [x] unit tests


waclawjacek (approved):
LGTM! :shipit:

PR #2392, opened 2021-03-18
feat: spinner icon on service data fetch

frosso:
## Description

I thought it would be nice to display a spinner icon on the status page when the data is refreshing in the background. Otherwise there's no feedback of the action that the user has taken.

### Steps to reproduce & screenshots/GIFs

- Go to the status page
- Click refresh
- During the refresh, a spinner icon is displayed

### Checklist

- [ ] unit tests

- [x] `changelog.txt` entry added

PR #2605, opened 2022-11-17
Replace calypso's spinner

harriswong:
## Description
Removes `<Spinner>` and use existing `Gridicon` with CSS animation.

### Related issue(s)
Relates to https://github.com/Automattic/woocommerce-services/issues/2333.

### Considerations
We don't have the <Spinner> components from WC https://woocommerce.github.io/woocommerce-admin/#/components/ in our `'woocommerce/components/`. We could add `@woocommerce/components` but I would rather wait until we decide which component (WC/WP/our own) to use.

We I tried using WP `<Spinner>` https://developer.wordpress.org/block-editor/reference-guides/components/spinner/ but couldn't get it to show. 

Using existing `GridIcon` + CSS reduces component dependency and we already use [a similar technique here](https://github.com/Automattic/woocommerce-services/pull/2392). 

I decided to use GridIcon + CSS.

1. Go purchase a label
2. Go through each step. As it loads, you should see a spinner similar to the following:
[spinner.webm](https://user-images.githubusercontent.com/572862/202322622-a8bbb12f-41b2-4e43-aca4-a52a403916c8.webm)

### Checklist

- [x] unit tests

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/style.scss
+++ b/client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/style.scss
@@ -77,11 +77,6 @@
 	.gridicon.is-warning {
 		color: var( --color-warning );
 	}
-	.gridicon.label-purchase-modal__step-spinner {
-		fill: var( --color-neutral-dark );
-		animation: 3s linear infinite;
-		animation-name: rotate-spinner;
-	}
 
 	.is-error:not( .notice ) {
 		color: var( --color-error );
@@ -228,4 +223,3 @@
 .label-purchase-modal__shipping-summary-section {
 	padding-bottom: 15px;
 }
-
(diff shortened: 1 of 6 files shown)


client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/style.scss
```diff
@@ -77,6 +77,11 @@
 	.gridicon.is-warning {
 		color: var( --color-warning );
 	}
+	.gridicon.label-purchase-modal__step-spinner {
+		fill: var( --color-neutral-dark );
```
waclawjacek: WDYT about preserving the spinner's pink color here?

image

```suggestion
		fill: var( --color-accent );
```

As much as I like the decalypsoification in this PR and your code, I am wondering if maybe we should copy the original spinner from Calypso here. 🤔 It is a very small component and I think it looks better (IMHO!)

What do you think?

waclawjacek (commented):
The PR is good to go but I'm suggesting using a different color (the original pink) and maybe porting the Calypso component instead? (It's very tiny.)

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/style.scss
harriswong: +1 on the Calypso spinner. I will clone it over.

waclawjacek (approved):
LGTM!

PR #2606, opened 2022-11-17
Replace TokenField with FormTokenField

harriswong:
## Description
Replace Calypso's `TokenField` with WP's component `FormTokenField`.

### Related issue(s)
Relates to https://github.com/Automattic/woocommerce-services/issues/2333

1. Go to WooCommerce -> Settings -> Shipping Zones -> USPS Live Rates
2. Check that the "Shipping Classes" drop down still works.

Before with Calypso:
image

After, with WP's component: 
image

### Checklist

- [x] unit tests


waclawjacek (commented):
This is how it looks on my machine. Any idea why this is the case?

Screen Capture on 2022-11-28 at 20-03-05

harriswong:
🤔  Did you wait for the whole `npm i` to finish running?

waclawjacek:
Yep!

waclawjacek:
Could you please give this another look?

harriswong:
Hey @waclawjacek , I just pulled and fixed the conflict. I still don't see what you were seeing:
image

harriswong:
I was wondering if it's the longer class name you have, but I still don't see the same issue:
image

waclawjacek (changes requested):
@harriswong Could you please update your WP, WC, WCS&T versions to the most recent ones and check what you see then? I am still getting the same thing.

harriswong:
I can reproduce this. I will have to dig a little bit into it.

harriswong:
Fixed and now it looks like:
image

Updated the styles by using the FormTokenField from storybook https://wordpress.github.io/gutenberg/?path=/docs/components-formtokenfield--default. I think there are some opportunities here for us to clean up the whole calypso styles embedded here https://github.com/Automattic/woocommerce-services/blob/trunk/assets/stylesheets/style.scss. But that could be done after Calypso is removed. 

image

waclawjacek (changes requested):
Aha, great job! Looks and works well. But I feel I have to request changes because I think we should sort out the border situation; otherwise, the field will look out of place alongside other fields.

## Current look

Markup on 2023-05-03 at 14:11:57

## Previous look

image

harriswong:
- Matched border colour
- Removed border
Please take a look again. @waclawjacek

harriswong:
No longer removing Calysp from WCS&T (https://github.com/Automattic/woocommerce-services/issues/2333). Putting this on hold.

waclawjacek (approved):
Reviewing anyway so this doesn't get stale! :) LGTM! :shipit: Thank you for the updates.

harriswong:
Closing. No longer needed.

PR #2618, opened 2022-12-09
Fix/issue 2578 - Make sure the packaging step is not skipped

### LINKED ISSUE #2578: Allow for setting a default package.
We've received a feature request:

When #2573 was merged in, it forces a package to be chosen for every single label purchase. Some stores may need to print large numbers of labels every day all with the same package. It would be helpful if there were a way to choose a default package in the plugin settings so that it is automatically selected when creating a label. It would be preferred that the default (or last used) package be pre-selected in the select box during label purchase and that the `Packaging` section still be opened so that the user can confirm the package selection before continuing on to `Shipping Rates`.

iyut:
## Description
When [#2573](https://github.com/Automattic/woocommerce-services/pull/2573) was merged in, it forces a package to be chosen for every single label purchase. Some stores may need to print large numbers of labels every day all with the same package.

This PR will change that behavior. It will set the package automatically but also will not skip the package step. The reason it is being designed this way in this PR is to prevent the user skip the packages step. So they now the packages that they set before they purchase the label.

### Related issue(s)

Fixes #2578 

1. Attempt to create an order with a valid/verified address
2. Notice that the shipping from and destination addresses are automatically validated, and you're offered to select a package
3. Now, create an order with a typo in the address
4. Notice that once you fix the address under Destination Address tab, you will go to Packaging tab instead of directly go to Shipping Rates tab.

### Checklist

- [ ] unit tests

### DIFF
--- a/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
@@ -434,9 +434,8 @@ export const removeIgnoreValidation = ( orderId, siteId, group ) => {
 
 const checkPackagesStep = ( orderId, siteId, dispatch, getState ) => {
 	const { expanded } = getShippingLabel( getState(), orderId, siteId ).form[ 'packages' ];
-	if ( true !== expanded ) {
+	if ( !expanded ) {
 		dispatch( toggleStep( orderId, siteId, 'packages' ) );
-		return;
 	}
 };
 
--- a/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
+++ b/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
@@ -347,12 +347,7 @@ export const openPrintingFlow = ( orderId, siteId ) => ( dispatch, getState ) =>
 
 	waitForAllPromises( promisesQueue ).then( () =>
 		tryGetLabelRates( orderId, siteId, dispatch, getState )
-	).then( () => {
-		const { packageId, boxId } = getDefaultBoxSelection( orderId, siteId, getState ) || {};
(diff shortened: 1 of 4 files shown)


client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
```diff
@@ -427,6 +432,14 @@ export const removeIgnoreValidation = ( orderId, siteId, group ) => {
 	};
 };
 
+const checkPackagesStep = ( orderId, siteId, dispatch, getState ) => {
+	const { expanded } = getShippingLabel( getState(), orderId, siteId ).form[ 'packages' ];
+	if ( true !== expanded ) {
```
dustinparker:
```suggestion
	if ( !expanded ) {
```

dustinparker: is there a reason we need to do `true !==` instead of `!`?

client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
```diff
@@ -427,6 +432,14 @@ export const removeIgnoreValidation = ( orderId, siteId, group ) => {
...
+		dispatch( toggleStep( orderId, siteId, 'packages' ) );
+		return;
```
dustinparker:
```suggestion
```

dustinparker (approved):
Everything works as expected. Nice work!

client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
iyut: no particular reason. Just want to make sure that tab will expand only if the `expanded` has value `true`.

bartech (approved):

PR #2626, opened 2023-01-19
Only validate US zip codes when they are 5 or 10 digits long - Fix/issue 2335

### LINKED ISSUE #2335: fix: focus is stolen from checkout input fields on error
## Description

See discussion here: https://github.com/Automattic/woocommerce-services/pull/2310#issuecomment-780554304

The Tax integration can display error messages at checkout. e.g.: on invalid zip.
When an error is rendered, the focus is stolen from the active input.

This is troublesome if the user at checkout is not using the browser's autocomplete or is not fast enough to edit the problematic field (esp. on mobile).

To reproduce:
- Enable automated taxes
- Enter a valid address
- Change the ZIP code to `99999`
- Try changing the ZIP code, one digit at a time, slowly
- Focus is stolen from the input every time.

This behavior does not occur without the plugin enabled. However, the focus-stealing behavior might be part of WooCommerce Core.
I would categorize this as an accessibility issue.

dustinparker:
## Description

WC core has a timeout after each `keydown` event in the fields which then triggers a checkout update. That update process then triggers our [TaxJar request validation method](https://github.com/Automattic/woocommerce-services/blob/94a2100ebbcf31efc20064c1d0622516d1b2663d/classes/class-wc-connect-taxjar-integration.php#L1242). The TaxJar address validation method has [a condition](https://github.com/Automattic/woocommerce-services/blob/94a2100ebbcf31efc20064c1d0622516d1b2663d/classes/class-wc-connect-taxjar-integration.php#L1265-L1269) (for US addresses only) that returns an error if the zipcode is not a valid US zipcode. Currently if a customer types one digit of the zipcode and delays one second, the input focus will be stolen and the screen will scroll up to show the zipcode error notice. This can be really frustrating.

This PR adjusts the TaxJar US zipcode validation to run only after exactly 5 or 10 digits have been entered to avoid the error notice from being displayed (and input focus stolen) every time a customer delays one second between typing the first 5 or 10 digits of the zipcode.

### Related issue(s)

Fixes #2335 

1. Checkout the `fix/issue-2335` branch of WC Services and the Connect Server `master` branch.
2. Make sure the store is using a valid US address and that automated taxes are enabled.
3. Add a product to cart and go to checkout.
4. Enter a valid US address.
5. Clear the zipcode field and type `1234` numbers waiting 1 second between each digit.
6. The input focus should not be stolen and no zipcode error notice should show.
7. Now type an `a` (the zipcode input should now equal `1234a`).
8. The input focus should be stolen and the screen should scroll up to reveal a zipcode error notice.
9. Now type `-123` slowly waiting a second between digits. 
10. The zipcode input should now contain `1234a-123`.
11. The input focus should not be stolen and no zipcode error notice should show.
12. Now type a `4` (the zipcode input should now equal `1234a-1234`).
13. The input focus should be stolen and the screen should scroll up to reveal a zipcode error notice.

### Checklist

- [x] unit tests


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1262,7 +1262,7 @@ public function validate_taxjar_request( $json ) {
 			return false;
 		}
 
-		if ( 'US' === $json['to_country'] && ! WC_Validation::is_postcode( $json['to_zip'], $json['to_country'] ) ) {
+		if ( 'US' === $json['to_country'] && ! empty( $json['to_zip'] ) && in_array( strlen( $json['to_zip'] ), array( 5, 10 ) ) && ! WC_Validation::is_postcode( $json['to_zip'], $json['to_country'] ) ) {
```
iyut, line 1270: Let's add a comment on what we're trying to do here. Perhaps something like : 
`// Only display the zip error if zip code length is 5 or 10.`

iyut (approved):
I have tested it, it works great. LGTM!

classes/class-wc-connect-taxjar-integration.php
dustinparker, line 1270:
```suggestion
		// Apply this validation only if the destination country is the US and the zip code is 5 or 10 digits long.
		if ( 'US' === $json['to_country'] && ! empty( $json['to_zip'] ) && in_array( strlen( $json['to_zip'] ), array( 5, 10 ) ) && ! WC_Validation::is_postcode( $json['to_zip'], $json['to_country'] ) ) {
```

changelog.txt
```diff
@@ -5,6 +5,7 @@
 
 = 2.2.0 - 2023-01-19 =
 * Add   - Add option to let user pick whether to save the last package & service or not.
+* Fix   - Adjust checkout US zipcode validation to run only when exactly 5 or 10 digits are typed.
```
harriswong, line 8: @dustinparker I saw this commit merged but we already released `2.2.0`. Do we need to release 2.2.2. including this PR? If so, we will need to change this changelog (and readme).

waclawjacek: I have already released this ([diff](https://github.com/Automattic/woocommerce-services/compare/aaedea0ac6c67616adfc3c651b91dc0cfd192316...2.2.2)) by moving the changelog entry but it would be great if you could please confirm if this 👆 is correct.

dustinparker: Sorry guys, I didn't get this notification for some reason until today. Yes, this is just a fix so 2.2.2 is fine.

PR #2644, opened 2023-03-09
Fix E2E tests

### LINKED ISSUE #2632: Upgrade e2e-tests's workflow to a newer ubuntu version
### Description

GitHub is deprecating Ubuntu 18.04 (https://github.com/actions/runner-images/issues/6002). 

> Deprecation will begin on 2022/08/08 and the image will be fully unsupported by 2023/04/01

Our workflow (https://github.com/Automattic/woocommerce-services/blob/trunk/.github/workflows/e2e_tests.yml#L8) uses Ubuntu 18.04. We should bump this up. We should also check other workflows.

### Acceptance criteria
A higher version of Ubuntu such that pipeline continues to run

waclawjacek:
## Description

Update Ubuntu and WordPress to latest. Kill service running on port 8084 so the E2E service can run. 

+ minor changes such as using .nvmrc

### Checklist

- [ ] unit tests

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -10,7 +10,8 @@ jobs:
       E2E_WP_VERSION: 6.1.0
       TRAVIS_MARIADB_VERSION: 10.8.3
     steps:
-      - run: git config --global url.https://github.com/.insteadOf git://github.com/
+      - name: Set Git to use HTTPS instead of SSH
+        run: git config --global url.https://github.com/.insteadOf git://github.com/
       - uses: actions/checkout@v3
         with:
           submodules: recursive
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -5,18 +5,18 @@ on: [push, workflow_dispatch]
 jobs:
   e2e_build_and_test:
     name: Build Container and Run E2E Tests
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     env:
-      WP_VERSION: 6.1.0
+      E2E_WP_VERSION: 6.1.0
       TRAVIS_MARIADB_VERSION: 10.8.3
     steps:
-      - uses: actions/checkout@v2
+      - run: git config --global url.https://github.com/.insteadOf git://github.com/
+      - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e_tests.yml
```diff
@@ -5,21 +5,34 @@ on: [push, workflow_dispatch]
...
-          node-version: '10.16.0'
-      - run: git config --global url.https://github.com/.insteadOf git://github.com/
-      - run: npm ci
-      - run: npm run dist
-      - run: npm run test:e2e-docker-up
-      - run: npm run test:e2e-docker-ping
-      - run: npm run test:e2e
-      - run: npm run test:e2e-docker-down
+            node-version-file: '.nvmrc'
+      - name: Kill process running on port 8084 to unblock it
```
harriswong, line 21: Do you know what is running on 8084 in test? Can we use another port instead of killing the process?

harriswong: I think this is the `mono-xsp4` service? Taking suggestion from [here](https://github.com/actions/runner-images/issues/2821#issuecomment-789512742). What do you think about the following?

```
- name: disable and stop mono-xsp4.service
  run: |
       sudo systemctl stop mono-xsp4.service || true
       sudo systemctl disable mono-xsp4.service || true
```

This way, we know what we are disabling and in the event Ubuntu changes the service port (ie. to 8080 in the future or something), we could still disable the service regardless of the port.

harriswong (approved):
LGTM

.github/workflows/e2e_tests.yml
waclawjacek: Yep, that's it! Interesting! Let's try that. I guess the "kill the thing on this port" approach is more straightforward and future-proof but I'd rather have it break so we can know what's going on. Adding this now.

waclawjacek: I gave it a try but it didn't work out of the box. Let's go with the "kill 8084" approach for now. WDYT?

harriswong: Sounds good! Thanks for trying. Maybe that's why there were 2 thumbs down in the other thread 😆

waclawjacek: 🤷

PR #2657, opened 2023-06-20
Adjust docker-compose file for different PHP version support

samnajian:
## Description
This PR:
- Adjusts the docker-compose file to easily switch the Dockerfile in use.
- Adds Dockerfile-php8.1 + SOAP extension 

### Testing
- See [this line](https://github.com/Automattic/woocommerce-services/pull/2657/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R11) for new `Dockerfile-php81` file and adjust the line to change to this Dockerfile or add an [override file](https://docs.docker.com/compose/extends/#understanding-multiple-compose-files) 
- Run `docker compose up -d  --no-deps --build wordpress` to build with the new Dockerfile
- Verify you are on PHP8.1

### Changelog
No changelog entry needed as it's on infra change


harriswong (approved):
The changes here is for dev. I didn't test this. The code LGTM

PR #2663, opened 2023-07-04
Fix failing PHP tests

samnajian:
## Description
- Our test suit had issues running MariaDB with `mariadb:latest` and this PR pins the version number to the latest working version.
- Makes sure the pnpm version used matches the pinned version in WC plugin.

### DIFF
--- a/.github/workflows/php_tests.yml
+++ b/.github/workflows/php_tests.yml
@@ -62,13 +62,17 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: '8.2'
+      - name: Get WC pinned pnpm version
+        id: get_wc_pnpm_version
+        run: echo "WC_PNPM=$(node -p "require('/tmp/woocommerce/plugins/woocommerce/package.json').engines.pnpm")"  >> $GITHUB_OUTPUT
       - name: Setup PNPM
         uses: pnpm/action-setup@v2
         with:
-          version: '^7.13.3'
+          version: ${{ steps.get_wc_pnpm_version.outputs.WC_PNPM }}
       - uses: actions/setup-node@v3
         with:
           node-version: 'v16'
+
       - name: Build WooCommerce essentials for running tests
         run: |
           cd /tmp/woocommerce/plugins/woocommerce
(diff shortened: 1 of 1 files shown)


.github/workflows/php_tests.yml
```diff
@@ -62,13 +62,17 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: '8.2'
+      - name: Get WC pinned pnpm version
+        id: get_wc_pnpm_version
+        run: echo "WC_PNPM=$(node -p "require('/tmp/woocommerce/plugins/woocommerce/package.json').engines.pnpm")"  >> $GITHUB_OUTPUT
```
waclawjacek: This works but I think the quotes should be escaped. What's happening right now is this (string concatenation):

```
waclaw@Waclaws-MacBook-Pro woocommerce % echo "hey"foo"hey"
heyfoohey
```

How about this instead?

```suggestion
        run: echo "WC_PNPM=$(node -p \"require('/tmp/woocommerce/plugins/woocommerce/package.json').engines.pnpm\")" >> "$GITHUB_OUTPUT"
```

waclawjacek (changes requested):
What a great solution! Suggesting what I think is an improvement (adding some escaping).

samnajian:
@waclawjacek The applied suggestion didn't work, not sure what I should do with it, would it be fine to revert the last commit?

.github/workflows/php_tests.yml
```diff
@@ -62,13 +62,17 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: '8.2'
+      - name: Get WC pinned pnpm version
+        id: get_wc_pnpm_version
+        run: echo "WC_PNPM=$(node -p \"require('/tmp/woocommerce/plugins/woocommerce/package.json').engines.pnpm\")" >> "$GITHUB_OUTPUT"
```
waclawjacek: Sorry, my bad! I've just learned that using `$()` creates a new quoting context. So this should work (it does when I test it locally after defining `GITHUB_OUTPUT`):

```suggestion
        run: echo "WC_PNPM=$(node -p "require('/tmp/woocommerce/plugins/woocommerce/package.json').engines.pnpm")" >> "$GITHUB_OUTPUT"
```

waclawjacek (changes requested):

waclawjacek (approved):

PR #2664, opened 2023-07-04
Add HAZMAT checkbox and bind to redux state

samnajian:
## Description
This PR adds HAZMAT section to package selection tab if the selected package is from USPS service.
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/261

### What to test:

- Verify you see the hazmat section ONLY when selecting a USPS package.
- Verify once a package is purchased, you see the HAZMAT keyword on the label.
- Switching from a USPS package to a none-USPS package shouldn't send the `hazmat` in the payload to the connect-server.


harriswong:
👀

harriswong (changes requested):
Custom packages don't have the "hazmat" checkbox. We should probably add that because hazmat isn't limited to just USPS's boxes?
image

waclawjacek:
> Custom packages don't have the "hazmat" checkbox.

The notion of a box is carrier-agnostic in the sense of "which carrier this will be shipped with". It's just a box that will be used; you can ship USPS shipments in DHL boxes, etc. :)

harriswong:
There is a usability issue I wanna raise.

As a user, I want to ship a battery but I can't get any labels. It keeps saying "No rates available, please double check dimensions and weight or try using different packaging.". I don't know what package to use? I just want to ship my battery. I tried switching "Package details" to a few others but still no rate returned. As a user, I don't know what's going on.

client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
```diff
@@ -1392,3 +1400,20 @@ export const openDetailsDialog = ( orderId, siteId, labelId ) => {
 export const closeDetailsDialog = ( orderId, siteId ) => {
 	return { type: WOOCOMMERCE_SERVICES_SHIPPING_LABEL_CLOSE_DETAILS_DIALOG, orderId, siteId };
 };
+
+export const setHazmatType = ( { hazmatType, siteId, orderId } ) => {
```
harriswong: Curious why the param here is an object but the rest of the actions aren't?

harriswong (approved):
LGTM 
I tested
- Separate packages. One with hazmat, the other doesn't. Works.
- I tested international shipping + hazmat, but I can't get any labels to print. 🤷 
- I tested international shipping without hazmat, works. DHL also shows. 👍 
- I also have this console warning when printing a label: 
[2664-video1.webm](https://github.com/Automattic/woocommerce-services/assets/572862/aa69539b-5e1b-4bfb-8283-f8699fb54d25)

If these are expected, then :shipit:

samnajian:
@harriswong 
> I tested international shipping + hazmat, but I can't get any labels to print. 🤷

Wondering if you're testing against the relevant branch from the connect-server?

> I also have this console warning when printing a label:

I'm getting the same on the trunk, aren't you?

harriswong:
> Wondering if you're testing against the relevant branch from the connect-server?

Yes.

> I'm getting the same on the trunk, aren't you?

No.

client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js
samnajian: Adjusted in https://github.com/Automattic/woocommerce-services/pull/2664/commits/d95c7e20b0fe3c5768673e375c7ca68464d88948

harriswong:
>> I'm getting the same on the trunk, aren't you?
> No.

After switching back and forth, I think `trunk` shows that error for me now too. We can create a follow-up if it doesn't exist already. Looks like that warning isn't new and it doesn't block this to merge.

PR #2670, opened 2023-07-28
Add/qit workflow

samnajian:
## Description
- Adds GH Action workflow for QIT 
- Add composer dependencies to check coding styles locally


harriswong (approved):
Tested these 2 commands, they LGTM
- `composer run check-security`
- `composer run qit:security`

The GitHub action returned an error https://github.com/Automattic/woocommerce-services/actions/runs/5719181046/job/15496545653?pr=2670 in the "Run security test" phase. Is this expected?
image

:shipit: if this is expected. Otherwise, please fix and then ship.

samnajian:
@harriswong that's not expected, that's why I suggested we can add the fix in this PR if you agree with it :)

harriswong:
@samnajian We can create a follow-up to fix it.

harriswong:
@waclawjacek some changes here may conflict with your Jetpack DNA work.

samnajian:
> @samnajian We can create a follow-up to fix it.

@harriswong once we merge this into trunk without fixing these issues we'll end up getting trunk red as well. I've added a fix for it already and fixing some failing php tests now.

harriswong (approved):
The last 5 commits look good to me.

PR #2689, opened 2023-10-19
Load sift js snippet

harriswong:
## Description
This PR adds the Sift.js snippet. It follows the documentation here https://sift.com/developers/docs/php/javascript-api/overview. 

The logic is very similar to WooCommerce Payments https://github.com/Automattic/woocommerce-payments/blob/c7b5b934684cc8c3d256e2ec84bc4ddf1071edac/includes/class-wc-payments-fraud-service.php#L170-L205

### Related issue(s)
273-gh-Automattic/woocommerce-shipping-issues
2246-gh-Automattic/woocommerce-connect-server

1. You will need this PR to accompany the test. 2247-gh-Automattic/woocommerce-connect-server
2. Go to label printing page. Open up your browser's network tab
3. Confirm that you see a request made to `s.js`, and Request URL is `https://cdn.sift.com/s.js`.
4. If you have access to the Sift dashboard, you should see something like this: pNPgK-6DZ-p2#comment-25368 

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -202,6 +202,15 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 		}
 
 		/**
+		 * Retrieve Sift configurations.
+		 *
+		 * @return object|WP_Error
+		 */
+		public function get_sift_configurations() {
+			return $this->request( 'GET', '/payment/sift' );
+		}
+
+		/**
 		 * Gets shipping rates (for labels) from the WooCommerce Shipping & Tax Server
 		 *
 		 * @param array $request - array(
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -206,7 +206,7 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 		 *
 		 * @return object|WP_Error
 		 */
-		public function get_sift_configurations() {
+		public function get_sift_configuration() {
 			return $this->request( 'GET', '/payment/sift' );
 		}
 
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -818,7 +818,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			add_action( 'woocommerce_email_after_order_table', array( $this, 'add_tracking_info_to_emails' ), 10, 3 );
 			add_filter( 'woocommerce_admin_reports', array( $this, 'reports_tabs' ) );
 			add_action( 'woocommerce_checkout_order_processed', array( $this, 'track_completed_order' ), 10, 3 );
-			add_action( 'admin_print_footer_scripts', [ $this, 'add_sift_js_tracker' ] );
+			add_action( 'admin_print_footer_scripts', array( $this, 'add_sift_js_tracker' ) );
 
 			$tracks = $this->get_tracks();
 			$tracks->init();
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -1614,10 +1614,10 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 				return;
 			}
 
-			$fraud_config = [
+			$fraud_config = array(
 				'beacon_key' => $sift_configurations->beacon_key,
-				'user_id' => $connected_data['ID']
-			];
+				'user_id'    => $connected_data['ID']
+			);
 
 			?>
 			<script type="text/javascript">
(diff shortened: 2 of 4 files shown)


classes/class-wc-connect-api-client.php
```diff
@@ -201,6 +201,15 @@ public function get_payment_methods() {
 			return $this->request( 'POST', '/payment/methods' );
 		}
 
+		/**
+		 * Retrieve Sift configurations.
+		 *
+		 * @return object|WP_Error
+		 */
+		public function get_sift_configurations() {
```
waclawjacek: Nitpick:

```suggestion
		public function get_sift_configuration() {
```

woocommerce-services.php
```diff
@@ -818,6 +818,7 @@ public function attach_hooks() {
 			add_action( 'woocommerce_email_after_order_table', array( $this, 'add_tracking_info_to_emails' ), 10, 3 );
 			add_filter( 'woocommerce_admin_reports', array( $this, 'reports_tabs' ) );
 			add_action( 'woocommerce_checkout_order_processed', array( $this, 'track_completed_order' ), 10, 3 );
+			add_action( 'admin_print_footer_scripts', [ $this, 'add_sift_js_tracker' ] );
```
waclawjacek: [WP coding standards array syntax](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#declaring-arrays):

```suggestion
			add_action( 'admin_print_footer_scripts', array( $this, 'add_sift_js_tracker' ) );
```

woocommerce-services.php
```diff
@@ -1594,6 +1595,45 @@ public function add_plugin_action_links( $links ) {
...
 
+		/**
+		 * Adds the Sift JS page tracker if needed. See the comments for the detailed logic.
+		 *
+		 * @return  void
+		 */
+		public function add_sift_js_tracker() {
+			$sift_configurations = $this->api_client->get_sift_configurations();
+			$master_user     = WC_Connect_Jetpack::get_master_user();
+			$connected_data  = WC_Connect_Jetpack::get_connected_user_data( $master_user->ID );
```
waclawjacek, line 1610:
```suggestion
			if ( ! $master_user ) {
				return;
			}
			
			$connected_data  = WC_Connect_Jetpack::get_connected_user_data( $master_user->ID );
```

woocommerce-services.php
```diff
@@ -1594,6 +1595,45 @@ public function add_plugin_action_links( $links ) {
...
 		}
 
+		/**
+		 * Adds the Sift JS page tracker if needed. See the comments for the detailed logic.
+		 *
+		 * @return  void
+		 */
+		public function add_sift_js_tracker() {
+			$sift_configurations = $this->api_client->get_sift_configurations();
+			$master_user     = WC_Connect_Jetpack::get_master_user();
```
waclawjacek, line 1605: Just noting here that the master user is the person whose Jetpack account is associated with the site. This might not be the person purchasing labels at a given time.

woocommerce-services.php
```diff
@@ -1594,6 +1595,45 @@ public function add_plugin_action_links( $links ) {
...
+
+			if ( is_wp_error( $sift_configurations ) || empty( $sift_configurations->beacon_key ) || empty( $connected_data['ID'] ) ) {
+				// Don't add sift tracking if we can't have the parameters to initialize Sift
+				return;
+			}
+
+			$fraud_config = [
+				'beacon_key' => $sift_configurations->beacon_key,
+				'user_id' => $connected_data['ID']
+			];
```
waclawjacek:
```suggestion
			$fraud_config = array(
				'beacon_key' => $sift_configurations->beacon_key,
				'user_id'    => $connected_data['ID']
			);
```

waclawjacek (approved):
LGTM! Just an extra condition for when `master_user` is empty.

classes/class-wc-connect-api-client.php
harriswong: Thanks! Fixed.

woocommerce-services.php
harriswong: Good catch, thanks!

woocommerce-services.php
harriswong: `$master_user` can be `false`, nice catch. Thanks!

woocommerce-services.php
harriswong: Thanks!

woocommerce-services.php
harriswong: Thanks for checking! The Jetpack account is the account on WPCOM, which is the account WCS&T bills the labels under. The whole transaction should be under this same Jetpack/WPCOM user_id. 

The `WC_Connect_Jetpack::get_master_user()` is the account associated with these cards (screenshot) right?
image

Should I be using another ID?

waclawjacek: Yep, that's the account! I was just saying that if a merchant has employees who are also allowed to purchase labels under their own (local to the store) user accounts, the user ID sent to Sift will still point at the merchant, not the employee's account.

There isn't exactly another ID we could rely on though - these other user accounts might not be connected with WPCOM accounts. Besides, as you pointed out, this is about the cards that are being used so it makes sense to keep this the card owner's user ID anyway I guess!

waclawjacek (approved):

woocommerce-services.php
harriswong: Perfect, thanks for double checking!

harriswong:
Tested the scenario where the server is not ready to provide sift's configuration. I did this by triggering an error thrown by `get_sift_configuration()`. I confirmed that the plugin will function as if sift is not loaded.

PR #2690, opened 2023-10-23
Add notice about disconnection when uninstalling Jetpack

waclawjacek:
## Description

Now that Jetpack will no longer be required by WCS&T, it could be uninstalled from a site. However when that happens, Jetpack will automatically remove its tokens from a site, therefore disconnecting WCS&T from WPCOM.

Although we will display our good old "Connect your site" banners after a page refresh/navigating to a page where these banners are displayed (e.g. Plugins page, WC Settings page, Edit Order page) so merchants using label printing might not miss this, this might be an issue for users of the tax rates or live shipping rates.

To make sure merchants keep their sites connected, we're adding a notice after Jetpack is uninstalled.

The notice will only show up after Jetpack is uninstalled, and only if WCS&T has been connected. It will also not allow connecting if already connected thanks to leveraging the NUX flow's `register_woocommerce_services_jetpack` action that is only registered when the `before_jetpack_connection` banner should be displayed.

### Related issue(s)

Fixes https://github.com/woocommerce/shipping/issues/84

https://github.com/Automattic/woocommerce-services/assets/1759681/80e15010-c980-48c3-8807-ebaf2ea7c783

### Testing instructions

1. Set up your local environment as described in this PR: https://github.com/Automattic/woocommerce-services/pull/2439 (note the Woo-branded WPCOM auth page is only available when sandboxed - that PR has information on this)
2. Make sure your site is connected to WPCOM.
3. Reproduce the steps in the video above.


waclawjacek:
I've updated the copy per paJDVv-2JR-p2#comment-3331

client/jetpack-deleted-notice.js
```diff
@@ -0,0 +1,90 @@
+/* global wcServicesJetpackDeletedNotice */
+/**
+ * External dependencies
+ */
+import jQuery from 'jquery';
```
kloon, line 5: We are only using jQuery 2 x times in this file with the rest being vanilla JS, perhaps it is worth removing jQuery completely and sticking to just vanilla JS in the two places.

client/jetpack-deleted-notice.js
```diff
@@ -0,0 +1,90 @@
...
+	const [ isVisible, setIsVisible ] = useState( false );
+
+	useEffect( () => {
+		const handleDeleteSuccess = ( event, response ) => {
+			if ( 'jetpack' === response.slug ) {
+				setIsVisible( true );
+			}
+		};
+
+		jQuery( document ).on( 'wp-plugin-delete-success', handleDeleteSuccess );
```
kloon, line 22: This can be done using vanilla JS as `document.addEventListener( 'wp-plugin-delete-success', handleDeleteSuccess );`

client/jetpack-deleted-notice.js
```diff
@@ -0,0 +1,90 @@
...
+
+		return () => {
+			jQuery( document ).off( 'wp-plugin-delete-success', handleDeleteSuccess );
```
kloon, line 25: This can be done using vanilla JS as `document.removeEventListener( 'wp-plugin-delete-success', handleDeleteSuccess );`

client/jetpack-deleted-notice.js
```diff
@@ -0,0 +1,90 @@
+/* global wcServicesJetpackDeletedNotice */
+/**
+ * External dependencies
+ */
+import jQuery from 'jquery';
+import { translate as __ } from 'i18n-calypso';
+import { Button, Modal } from '@wordpress/components';
+import ReactDOM from "react-dom";
+import React from "react";
```
kloon, line 9: These imports should not be needed in a WP environment, any reason why you are importing it here?

client/jetpack-deleted-notice.js
```diff
@@ -0,0 +1,90 @@
+/* global wcServicesJetpackDeletedNotice */
+/**
+ * External dependencies
+ */
+import jQuery from 'jquery';
+import { translate as __ } from 'i18n-calypso';
```
kloon, line 6: What is the difference between using `translate` from `i18n-calypso` vs using `__` from `@wordpress/i18n`? My inclination says we should stick to using the `@wordpress` packages.

kloon (commented):
Leaving some comments for now, I am having a hard time getting a dev site up and running for this extension so seeing as this is time sensitive as well, feel free to pick another reviewer for now. I will continue to work through setting up and following along with the other reviewer.

client/jetpack-deleted-notice.js
waclawjacek: I wanted to use `document.addEventListener` instead but it didn't pick up the event which is emitted using jQuery's `.trigger()` function. I decided to stick with this.

client/jetpack-deleted-notice.js
waclawjacek: Oh yeah, that's what I wanted to do! :D But it didn't work.

client/jetpack-deleted-notice.js
waclawjacek: Oh, I didn't know! I'll try that.

client/jetpack-deleted-notice.js
waclawjacek: I tried using `@wordpress/i18n` despite this being the only usage of that package in WCS&T but I thought since we're working on the new extension, it might make porting the code easier. But unfortunately, this caused problems with building the extension so I just used what the rest of the extension is using, which is `translate` from Calypso.

client/jetpack-deleted-notice.js
waclawjacek: I just checked and without these, I'm getting:

```
  57:4  error  'React' must be in scope when using JSX  react/react-in-jsx-scope
  61:5  error  'React' must be in scope when using JSX  react/react-in-jsx-scope
  65:5  error  'React' must be in scope when using JSX  react/react-in-jsx-scope
  69:5  error  'React' must be in scope when using JSX  react/react-in-jsx-scope
  73:5  error  'React' must be in scope when using JSX  react/react-in-jsx-scope
  84:2  error  'ReactDOM' is not defined                no-undef
  85:3  error  'React' must be in scope when using JSX  react/react-in-jsx-scope
```

It looks like the build setup we're using requires us to use these.

waclawjacek:
Thank you for the review, @kloon! I've responded to your comments.

FangedParakeet (approved):
I agree with @kloon that the jQuery dependence is a little annoying as well as surprising, but I don't doubt @waclawjacek's efforts to launch this notice without it. FWIW, most of our WCS&T client scripts also have jQuery as a dependency (as well as [the file](https://code.a8c.com/source/wpcom-git/browse/trunk/wp-admin/js/updates.js$11) that [triggers the `wp-plugin-delete-success` event](https://code.a8c.com/source/wpcom-git/browse/trunk/wp-admin/js/updates.js$1081)), so maybe this isn't such a tremendous tragedy after all.

In any case, I can't see anything worthy of too much condemnation here and this worked just as expected for me on a JN connected to the sandbox. Nice one! Happy to give this one the green light. 🚢

client/jetpack-deleted-notice.js
kloon: OK thanks for trying, looks like a build setup issue then. Let's not spend too much time on this and proceed as is.

client/jetpack-deleted-notice.js
kloon: Gotcha, did some quick research now and it seems jQuery has its own events layer that cannot be accessed via standard JS events methods. :shakes_fist: @ jQuery

samnajian:
@kloon we don't need this in Woo Shipping as JetPack doesn't remove the token anymore, do we?

kloon:
> @kloon we don't need this in Woo Shipping as JetPack doesn't remove the token anymore, do we?

That is correct.

PR #2691, opened 2023-10-24
Bump supported WC versions to L-2 and its WP & PHP requirements

waclawjacek:
## Description

In line with [this change to WooPayments](https://developer.woocommerce.com/2021/05/12/woocommerce-payments-is-adopting-a-new-version-support-policy/), we are adopting a WC L-2 version support policy. This means that we will support the current release of WC, as well as the last two major releases. For example, if the current version is WC 8.2.1, we will support the latest patch release (the last digit in the version number) for the following WC major verisons:

- WC 8.2.x
- WC 8.1.x
- WC 8.0.x

As well as these WooCommerce versions' WP and PHP dependencies. The WP and PHP requirements of WC 8.0 are PHP 7.3 and WP 6.2.

**Before merging, we should sort out announcing this change**: p1698148432153039-slack-C3ECCGUVC Once done, this can be converted from Draft to a published PR.

### Checklist

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.4.0 - 2023-xx-xx =
+* Update - Bump supported WC versions to L-2, and WP and PHP versions to WC L-2's requirements.
+
 = 2.3.7 - 2023-10-23 =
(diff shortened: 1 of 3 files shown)


jeffpaul:
Looks like this is outdated and can be closed

PR #2702, opened 2023-11-28
Fix failing e2e tests

samnajian:
## Description
This PR fixes the issue with activating/deactivating e2e test and makes the WP version used for CI e2e and the local e2e tests the identical to make it more transparent which version is used.

The issue was that [we've flagged the plugin compatible with WP6.2+](https://wordpress.org/plugins/woocommerce-services/), whereas the e2e was randomly sometimes installing v6.1 and failing to upgrade it, which resulted in WP completely disabling the activation of the plugin.

Can-activate-WooCommerce-Shipping-Tax-extension-if-it-is-deactivated


samnajian:
Now the `wc/v1/connect/normalize-address` endpoint throws, which is causing the test to fail.

package.json
```diff
@@ -19,10 +19,11 @@
     "build": "echo yes | npm run release && cp release/woocommerce-services.zip ./",
     "up": "docker-compose up --build --force-recreate -d && ./bin/docker-setup.sh",
     "down": "docker-compose down",
-    "test:e2e-docker-up": "npx wc-e2e docker:up",
+    "test:e2e-docker-up": "env WP_VERSION=6.4.1 TRAVIS_MARIADB_VERSION=10.8.3 npx wc-e2e docker:up",
     "test:e2e-docker-down": "npx wc-e2e docker:down",
     "test:e2e-docker-ping": "bash ./tests/bin/wait-for-e2e-build.sh",
-    "test:e2e": "npx wc-e2e test:e2e",
+    "test:e2e": "WC_E2E_SCREENSHOTS=1 npx wc-e2e test:e2e",
```
kloon: Will this upload the screenshots somewhere? If so that would be great to help determine what causes the failure.

kloon (approved):
Thanks for catching this, good to ship!

package.json
samnajian: ah, that's a left over, I'd remove it and we can have an issue to upload them somewhere as they are very handy for e2e tests.

PR #2727, opened 2024-04-04
Don't register functionality if both Woo Shipping and Woo Tax are active

### LINKED ISSUE #2726: Stop plugin execution when woocommerce-shipping or woocommerce-tax is active
In order to prepare for the separation of shipping and tax we need to add a check to ensure that WCS&T does not work when either `woocommerce-shipping` or `woocommerce-tax` are active.

We also need to display a message indicating that you cannot have WCS&T active alongside Woo Shipping or Woo Tax.

waclawjacek:
### Related issue(s)

Fixes #2726.

Screenshot 2024-04-04 at 11 36 08

### Testing instructions

1. Install & activate both Woo Shipping and WCS&T.
2. See the notice in the screenshot.
3. Go to Edit Order and see that WCS&T didn't register its functionality but Woo Shipping did.

The same should work for Woo Tax instead of Woo Shipping, as well as both Woo Shipping and Woo Tax, with an appropriate change in the message.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
 = 2.5.5 - 2024-xx-xx =
-* Add - Don't register functionality if Woo Shipping is active.
+* Add - Don't register functionality if Woo Shipping or Woo Tax is active.
 
 = 2.5.4 - 2024-03-25 =
 * Tweak - WordPress 6.5 compatibility.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.5.5 - 2024-xx-xx =
+* Add - Don't register functionality if Woo Shipping is active.
+
 = 2.5.4 - 2024-03-25 =
 * Tweak - WordPress 6.5 compatibility.
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
 = 2.5.5 - 2024-xx-xx =
-* Add - Don't register functionality if Woo Shipping or Woo Tax is active.
+* Add - Prevent upcoming Woo Shipping and Woo Tax plugins from running in parallel with this plugin unless both are active, then they will take over for this plugin.
 
 = 2.5.4 - 2024-03-25 =
 * Tweak - WordPress 6.5 compatibility.
--- a/readme.txt
+++ b/readme.txt
@@ -79,6 +79,9 @@ The source code is freely available [in GitHub](https://github.com/Automattic/wo
 
 == Changelog ==
 
+= 2.5.5 - 2024-xx-xx =
+* Add - Don't register functionality if Woo Shipping is active.
+
 = 2.5.4 - 2024-03-25 =
 * Tweak - WordPress 6.5 compatibility.
 
--- a/readme.txt
+++ b/readme.txt
@@ -80,7 +80,7 @@ The source code is freely available [in GitHub](https://github.com/Automattic/wo
 == Changelog ==
 
 = 2.5.5 - 2024-xx-xx =
-* Add - Don't register functionality if Woo Shipping is active.
+* Add - Don't register functionality if Woo Shipping or Woo Tax is active.
 
 = 2.5.4 - 2024-03-25 =
 * Tweak - WordPress 6.5 compatibility.
--- a/readme.txt
+++ b/readme.txt
@@ -80,7 +80,7 @@ The source code is freely available [in GitHub](https://github.com/Automattic/wo
 == Changelog ==
 
 = 2.5.5 - 2024-xx-xx =
-* Add - Don't register functionality if Woo Shipping or Woo Tax is active.
+* Add - Prevent upcoming Woo Shipping and Woo Tax plugins from running in parallel with this plugin unless both are active, then they will take over for this plugin.
 
 = 2.5.4 - 2024-03-25 =
 * Tweak - WordPress 6.5 compatibility.
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -338,13 +338,22 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 					}
 				}
 			);
-			/*
-			 * Used to let Woo Shipping know WCS&T will handle the plugins' coexistence
-			 * by displaying an appropriate notice and not registering its functionality.
-			 */
-			add_filter( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', '__return_true' );
 			add_action( 'plugins_loaded', array( $this, 'jetpack_on_plugins_loaded' ), 1 );
+			add_action( 'plugins_loaded', array( $this, 'maybe_handle_coexistence_with_woo_shipping_and_woo_tax' ), 5 );
 			add_action( 'plugins_loaded', array( $this, 'on_plugins_loaded' ) );
+		}
+
+		/**
+		 * Used to let Woo Shipping and Woo Tax know WCS&T will handle the plugins' coexistence
+		 * by displaying an appropriate notice and not registering its functionality.
+		 */
+		public function maybe_handle_coexistence_with_woo_shipping_and_woo_tax() {
+			$is_woo_shipping_active = in_array( 'woocommerce-shipping/woocommerce-shipping.php', get_option( 'active_plugins' ) );
+			$is_woo_tax_active      = in_array( 'woocommerce-tax/woocommerce-tax.php', get_option( 'active_plugins' ) );
+
+			if ( $is_woo_shipping_active && $is_woo_tax_active ) {
+				add_filter( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', '__return_true' );
+			}
 		}
 
 		public function get_logger() {
@@ -582,22 +591,11 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		public function on_plugins_loaded() {
 			$this->load_textdomain();
 
-			$is_woo_shipping_active = in_array( 'woocommerce-shipping/woocommerce-shipping.php', get_option( 'active_plugins' ) );
-			$is_woo_tax_active      = in_array( 'woocommerce-tax/woocommerce-tax.php', get_option( 'active_plugins' ) );
-
-			if ( $is_woo_shipping_active || $is_woo_tax_active ) {
+			if ( apply_filters( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', false ) ) {
 				add_action(
 					'admin_notices',
-					function () use ( $is_woo_shipping_active, $is_woo_tax_active ) {
-						if ( $is_woo_shipping_active && $is_woo_tax_active ) {
-							$active_plugins = __( 'Woo Shipping and Woo Tax plugins are already active.', 'woocommerce-services' );
-						} elseif ( $is_woo_shipping_active ) {
-							$active_plugins = __( 'Woo Shipping plugin is already active.', 'woocommerce-services' );
-						} else {
-							$active_plugins = __( 'Woo Tax plugin is already active.', 'woocommerce-services' );
-						}
-
-						echo '<div class="error"><p><strong>' . sprintf( esc_html__( '%s Please deactivate WooCommerce Shipping & Tax.', 'woocommerce-services' ), esc_html( $active_plugins ) ) . '</strong></p></div>';
+					function () {
+						echo '<div class="error"><p><strong>' . esc_html__( 'Woo Shipping and Woo Tax plugins are already active. Please deactivate WooCommerce Shipping & Tax.', 'woocommerce-services' ) . '</strong></p></div>';
 					}
 				);
 				return;
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -582,11 +582,22 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		public function on_plugins_loaded() {
 			$this->load_textdomain();
 
-			if ( in_array( 'woocommerce-shipping/woocommerce-shipping.php', get_option( 'active_plugins' ) ) ) {
+			$is_woo_shipping_active = in_array( 'woocommerce-shipping/woocommerce-shipping.php', get_option( 'active_plugins' ) );
+			$is_woo_tax_active      = in_array( 'woocommerce-tax/woocommerce-tax.php', get_option( 'active_plugins' ) );
+
+			if ( $is_woo_shipping_active || $is_woo_tax_active ) {
 				add_action(
 					'admin_notices',
-					function () {
-						echo '<div class="error"><p><strong>' . esc_html__( 'Woo Shipping plugin is already active. Please deactivate WooCommerce Shipping & Tax.', 'woocommerce-shipping' ) . '</strong></p></div>';
+					function () use ( $is_woo_shipping_active, $is_woo_tax_active ) {
+						if ( $is_woo_shipping_active && $is_woo_tax_active ) {
+							$active_plugins = esc_html__( 'Woo Shipping and Woo Tax plugins are already active.', 'woocommerce-services' );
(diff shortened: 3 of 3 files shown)


waclawjacek:
I'm not sure why the E2E tests are failing. I can run them locally. 🤔

woocommerce-services.php
```diff
@@ -577,6 +582,27 @@ public function jetpack_on_plugins_loaded() {
...
+					function () use ( $is_woo_shipping_active, $is_woo_tax_active ) {
+						if ( $is_woo_shipping_active && $is_woo_tax_active ) {
+							$active_plugins = __( 'Woo Shipping and Woo Tax plugins are already active.', 'woocommerce-services' );
+						} elseif ( $is_woo_shipping_active ) {
+							$active_plugins = __( 'Woo Shipping plugin is already active.', 'woocommerce-services' );
+						} else {
+							$active_plugins = __( 'Woo Tax plugin is already active.', 'woocommerce-services' );
+						}
+
+						echo '<div class="error"><p><strong>' . sprintf( esc_html__( '%s Please deactivate WooCommerce Shipping & Tax.', 'woocommerce-services' ), esc_html( $active_plugins ) ) . '</strong></p></div>';
```
kallehauge: Blocking: AFAIR all `sprintf` should have an a translators comment, so translators knows what they're working with 😄 

```suggestion
						/* translators: %s: the reason WCS&T has to be deactivated */
						echo '<div class="error"><p><strong>' . sprintf( esc_html__( '%s Please deactivate WooCommerce Shipping & Tax.', 'woocommerce-services' ), esc_html( $active_plugins ) ) . '</strong></p></div>';
```

kallehauge (changes requested):
The use of `wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax` is very smart! 👏 

According to @laurajohnsonwoo in our weekly meeting, then we want to prioritise that we're not breaking tax calculations, so I'm not sure if we should disable WCS&T if they do not have Woo Tax installed?

waclawjacek:
> I'm not sure if we should disable WCS&T if they do not have Woo Tax installed?

Hmm, good point! WDYT, @laurajohnsonwoo, @kloon?

I suppose we could check what features a merchant has enabled and display different prompts based on that.

waclawjacek:
I've updated the PR with the approach suggested by @kloon here: p1712745585197439/1712743452.598759-slack-C04KWSNPSE5

> * WCS&T and Woo Shipping enabled, then disabled Woo Shipping.
> * WCS&T and Woo Tax enabled, then disable Woo Tax.
> * WCS&T and Woo Tax and Woo Shipping enabled, then disable WCS&T.

waclawjacek:
E2E tests continue to fail but not when trying locally:

```
npm run test:e2e-docker-up && npm run test:e2e
```

or

```
npm run test:e2e-docker-up && npm run test:e2e-dev
```

waclawjacek:
@kallehauge Do you think you could review despite the E2E tests failing? The same happens in other PRs, e.g.: https://github.com/Automattic/woocommerce-services/actions/runs/8606963325/job/23586481611

Added an issue to fix E2E tests: https://github.com/Automattic/woocommerce-services/issues/2729

changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.5.5 - 2024-xx-xx =
+* Add - Don't register functionality if Woo Shipping or Woo Tax is active.
```
kallehauge: Suggestion (not blocking): from a merchant/developer PoV, then we could perhaps tease the the future deprecation a bit more and clarify what even "Woo Shipping" and "Woo Tax" is when they read this changelog; otherwise they might just be confused about the plugins being shipping and tax functionality in WCS&T?

```suggestion
* Add - Prevent upcoming Woo Shipping and Woo Tax plugins from running in parallel with this plugin unless both are active, then they will take over for this plugin.
```

readme.txt
```diff
@@ -79,6 +79,9 @@ The source code is freely available [in GitHub](https://github.com/Automattic/wo
 
 == Changelog ==
 
+= 2.5.5 - 2024-xx-xx =
+* Add - Don't register functionality if Woo Shipping or Woo Tax is active.
```
kallehauge: Suggestion (not blocking): from a merchant/developer PoV, then we could perhaps tease the the future deprecation a bit more and clarify what even "Woo Shipping" and "Woo Tax" is when they read this changelog; otherwise they might just be confused about the plugins being shipping and tax functionality in WCS&T?

```suggestion
* Add - Prevent upcoming Woo Shipping and Woo Tax plugins from running in parallel with this plugin unless both are active, then they will take over for this plugin.
```

**Duplicate of other comment**

woocommerce-services.php
```diff
@@ -339,9 +339,23 @@ function() {
...
+		 * Used to let Woo Shipping and Woo Tax know WCS&T will handle the plugins' coexistence
+		 * by displaying an appropriate notice and not registering its functionality.
+		 */
+		public function maybe_handle_coexistence_with_woo_shipping_and_woo_tax() {
+			$is_woo_shipping_active = in_array( 'woocommerce-shipping/woocommerce-shipping.php', get_option( 'active_plugins' ) );
+			$is_woo_tax_active      = in_array( 'woocommerce-tax/woocommerce-tax.php', get_option( 'active_plugins' ) );
+
+			if ( $is_woo_shipping_active && $is_woo_tax_active ) {
+				add_filter( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', '__return_true' );
+			}
```
kallehauge: [Subjective, non-blocking] Could we potentially simplify the solution to "just" return if Woo Shipping and Woo Tax is activated and then call the method directly when we call `apply_filters`?

```php
apply_filters( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', $this->maybe_handle_coexistence_with_woo_shipping_and_woo_tax() );
```

I see a few potential benefits:

1. It's very minor, but it is a slight performance increase since we can get rid of 1 action call (`plugins_loaded`) and 1 filter (`wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax`).
2. By not adding any `add_filter( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', ... );` ourselves, we reduce the chance of race conditions to happen if others hook into the filter.
    * _This argument is **super** theoretical since we do not expect others to use our hooks._
3. Opinionated, then it makes the implementation simpler and therefore easier to read and maintain for other developers.

That way we still use the same logic as as the fallback value for our filter and it'll work exactly like now, but we don't have to use a `add_filter` ourselves that is being executed in an additional `plugins_loaded` action? 😄  

```suggestion
			return $is_woo_shipping_active && $is_woo_tax_active;
```

kallehauge (approved):
Do we need to run `WC_Connect_Loader::jetpack_on_plugins_loaded` when both Woo Shipping and Woo Tax are activated?

Phrased differently: We have these 3 actions (one of them being introduced in this PR):

```php
add_action( 'plugins_loaded', array( $this, 'jetpack_on_plugins_loaded' ), 1 );
add_action( 'plugins_loaded', array( $this, 'maybe_handle_coexistence_with_woo_shipping_and_woo_tax' ), 5 );
add_action( 'plugins_loaded', array( $this, 'on_plugins_loaded' ) );
```

Why is `jetpack_on_plugins_loaded` running in an earlier priority? Without having dug into the code, then the other plugins should take care for their own WPCOM connection, so we don't have to run the code?

**Approval**: I'm approving since the PR works and if the Jetpack comment is a non-issue, then it can be merged as-is 👍

waclawjacek:
> Why is `jetpack_on_plugins_loaded` running in an earlier priority?

Unfortunately, there is this restriction: https://github.com/Automattic/woocommerce-payments/pull/637/files#diff-ad243b7f0e0ae2cc54963fec106a5c54ece9c5bb8b272acaeece1836d71418a2R36

After reading your (very valid) comment, I have considered loading the "coexistence" stuff with a negative priority (hey, WP docs say "integer") but that'd mean loading our stuff _before_ WP core has done some of the stuff that it registers with priority `0` itself.

But anyway, we don't need to run this check during `plugins_loaded`. I'll post more info in my response to this comment: https://github.com/Automattic/woocommerce-services/pull/2727#discussion_r1562739028

woocommerce-services.php
waclawjacek: > It's very minor, but it is a slight performance increase since we can get rid of 1 action call (plugins_loaded) and 1 filter (wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax).

Hey, you're right! Since we're reading the option and not actually waiting for plugins to load, we can do the check as soon as `get_option()` is available.

> By not adding any add_filter( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', ... ); ourselves, we reduce the chance of race conditions to happen if others hook into the filter. 

I'm not sure I understand. What is the difference if we don't send any value to the hook?

> Opinionated, then it makes the implementation simpler and therefore easier to read and maintain for other developers.

Yes! Very cool.

waclawjacek: > That way we still use the same logic as as the fallback value for our filter and it'll work exactly like now, but we don't have to use a add_filter ourselves that is being executed in an additional plugins_loaded action? 😄

Oh, wait; I committed the change you suggested but we do have to register the `add_filter` call. Otherwise Woo Shipping and Woo Tax won't get the message.

waclawjacek:
@kallehauge I've updated the code to not register the Jetpack config unless needed. Could you please re-review this PR?

woocommerce-services.php
kallehauge: Nice! That all sounds awesome! 😄 

>I'm not sure I understand. What is the difference if we don't send any value to the hook?

What I meant was that instead of having both a place where "call" the hook and registering our own as well:

```php
// __construct():
if ( $this->are_woo_shipping_and_woo_tax_active() ) {
	/**
	 * Used to let Woo Shipping and Woo Tax know WCS&T will handle the plugins' coexistence
	 * by displaying an appropriate notice and not registering its functionality.
	 */
	add_filter( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', '__return_true' );
	return;
}

// on_plugins_loaded():
apply_filters( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', false )
```

It might be possible to combine them into this:

```php
apply_filters( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', $this->are_woo_shipping_and_woo_tax_active() );
```

The primary (very theoretical) upside is that we reduce the chance for race conditions with other places that wants to use the same hook.
It's really not important since I do not expect 3. party developers to add hooks as well, which is why it's super theoretical 😄 

The second benefit is just a small performance improvement since and - depending on your pov - a simplified codebase

---

> Oh, wait; I committed the change you suggested but we do have to register the add_filter call. Otherwise Woo Shipping and Woo Tax won't get the message.

Ah, sorry! It was meant as a partial example since I couldn't modify the code in multiple places 🙈

woocommerce-services.php
```diff
@@ -577,6 +588,11 @@ public function jetpack_on_plugins_loaded() {
 		public function on_plugins_loaded() {
 			$this->load_textdomain();
 
+			if ( apply_filters( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', false ) ) {
```
kallehauge, line 604: [Non-blocking] I'm mentioning this with the intent of knowledge sharing the standard more than anything since the rest of the plugin doesn't really adhere to it anyway, but [WP standards wants us to document do_action and apply_filters](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters) (IMO most importantly for the "since" tag to help determine backwards compatibility when using the hook)

```suggestion
			/**
			 * Allow third party logic to determine if this plugin should initiate its logic.
			 *
			 * The primary purpose here is to allow a smooth transition between the new Woo Shipping / Woo Tax plugins
			 * and WooCommerce Shipping & Tax (this plugin), by letting them take over all responsibilities if all three
			 * plugins are activated at the same time.
			 *
			 * @since {{next-release}}
			 * 
			 * @param bool $status The value will determine if we should initiate the plugins logic or not.
			 */
			if ( apply_filters( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', false ) ) {
```

woocommerce-services.php
```diff
@@ -1748,6 +1764,17 @@ public function render_schema_notices() {
 				<?php
 			}
 		}
+
+		public function are_woo_shipping_and_woo_tax_active() {
```
kallehauge, line 1786: [Non blocking] WP standards is normally pretty strict with adding DocBlocks, but it's not a strictly followed rule on the plugin anyway.

woocommerce-services.php
```diff
@@ -1748,6 +1764,17 @@ public function render_schema_notices() {
...
+			$is_woo_shipping_active = in_array( 'woocommerce-shipping/woocommerce-shipping.php', get_option( 'active_plugins' ) );
+			$is_woo_tax_active      = in_array( 'woocommerce-tax/woocommerce-tax.php', get_option( 'active_plugins' ) );
+
+			return $is_woo_shipping_active && $is_woo_tax_active;
+		}
+
+		public function display_woo_shipping_and_woo_tax_are_active_notice() {
```
kallehauge, line 1800: [Non blocking] WP standards is normally pretty strict with adding DocBlocks, but it's not a strictly followed rule on the plugin anyway.

woocommerce-services.php
```diff
@@ -338,8 +338,19 @@ function() {
...
+			if ( $this->are_woo_shipping_and_woo_tax_active() ) {
+				/**
+				 * Used to let Woo Shipping and Woo Tax know WCS&T will handle the plugins' coexistence
+				 * by displaying an appropriate notice and not registering its functionality.
+				 */
+				add_filter( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', '__return_true' );
+				return;
+			}
+
+			add_action( 'plugins_loaded', array( $this, 'jetpack_on_plugins_loaded' ), 1 );
```
kallehauge, line 355: [Note] If we want to leave the plugin as "untouched" with issues as possible, then I don't think we should spend time simplifying the `wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax` logic further like I mentioned in my last comment (https://github.com/Automattic/woocommerce-services/pull/2727#discussion_r1576962027) since I think we'd need to move `jetpack_on_plugins_loaded()` inside `on_plugins_loaded()` and that means we'd change the priority it'd be loaded in as well (IMO it's not worth the potential issues we'd be introducing)

woocommerce-services.php
```diff
@@ -338,8 +338,19 @@ function() {
...
 			);
-			add_action( 'plugins_loaded', array( $this, 'jetpack_on_plugins_loaded' ), 1 );
+
 			add_action( 'plugins_loaded', array( $this, 'on_plugins_loaded' ) );
+
+			if ( $this->are_woo_shipping_and_woo_tax_active() ) {
+				/**
+				 * Used to let Woo Shipping and Woo Tax know WCS&T will handle the plugins' coexistence
+				 * by displaying an appropriate notice and not registering its functionality.
+				 */
```
kallehauge: [Non-blocking] If we're leaning up against the WP standards for `do_action` and `apply_filters` when adding hooks like this, then the summary should be 1 line and if it provides value with further explanation, then the description should have a new line between it and the summary 😄 

https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters

kallehauge (approved):
Looks nice! Thank you for implementing the improvements (and spotting the anonymous function removal in `display_woo_shipping_and_woo_tax_are_active_notice` :chefs-kiss:).

I've left some light formatting comments, but the rest of the plugin doesn't adhere to it anyway, so LGTM! 👍

woocommerce-services.php
waclawjacek: Very cool, thanks!

woocommerce-services.php
waclawjacek: Aha, gotcha! That is a nice improvement! I'll leave this part intact though for the sake of: https://github.com/Automattic/woocommerce-services/pull/2727#discussion_r1576972404

PR #2733, opened 2024-05-03
Cap max allowed shipments to prevent 504 on order page

### LINKED ISSUE #2731: large orders causing 504 in orders dash
User with a large order (13k qty of a single item) is getting a 504/ OOM when they try to open the order screen to review the order. Doesn't happen for any other order (but no other orders on site same qty).

slack: p1713454127420779-slack-C7U3Y3VMY
> I think the issue is that the code is trying to load all the order items for the meta box, and then timed out before it could finish: https://github.com/Automattic/woocommerce-services/blob/trunk/classes/class-wc-connect-shipping-label.php#L556-L576

issue doesn't happen when product is switched to virtual 

p2 where being addressed: p9F6qB-eRR-p2 
site: riotrfid.com
SA: https://wordpress.com/wp-admin/network/admin.php?page=store-admin&action=search&username=riotrfid

samnajian:
## Description
This PR  caps the max allowes shipments to make should too many number of shipments don't cause performance issue for client nor server side.

### Related issue(s)
Fixes #2731

- With the trunk branch, create and order with a product, and add the quantity of 140,000 for this order item.
- Visit this order details page and see that the page breaks or doesn't load properly.
- Now switch to this branch, and refresh the order detail page and verify the page loads without any problem.

https://github.com/Automattic/woocommerce-services/assets/789421/86e1aa16-07b7-41a8-977b-8183dcec030f

### Checklist

- [x] unit tests

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -175,7 +175,6 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 					),
 				);
 			}
-
 			$formatted_packages = array();
 
 			foreach ( $packages as $package_obj ) {
@@ -226,10 +225,48 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 				}
 
 				$refunded_qty = $order->get_qty_refunded_for_item( $item->get_id() );
+				$remaining_quantity = $item['qty'] - absint( $refunded_qty );
+				/**
+				 * The threshold at which we will start batching items together.
+				 * As an example, if a single order item has a quantity 60, which is more than the default threshold
+				 * value of 20 we will start batching together.
+				 */
+				$threshold = apply_filters( 'wc_connect_shipment_item_quantity_threshold', 20 );
+				/**
+				 * Max number of shipments allowed to be created for this item should quantity of this order items
+				 * exceeds `wc_connect_shipment_item_quantity_threshold`
+				 */
+				$max_shipments = apply_filters( 'wc_connect_max_shipments_if_quantity_exceeds_threshold', 5 );
+				/**
+				 * Default quantity per item if the `wc_connect_shipments_item_threshold_quantity` is not exceeded.
+				 */
+				$quantity_per_shipment = 1;
 
-				for ( $i = 0; $i < ( $item['qty'] - absint( $refunded_qty ) ); $i ++ ) {
-					$items[] = $item_data;
+				$weight_per_item = $item_data['weight'];
+				$should_cap_shipments = $remaining_quantity > $threshold;
+
+				if ( $should_cap_shipments ) {
+					$quantity_per_shipment = floor( $remaining_quantity / $max_shipments );
+					for ( $i = 0; $i < $max_shipments; $i ++ ) {
+						$item_data['weight'] = $quantity_per_shipment * $weight_per_item;
+						$remaining_quantity -= $quantity_per_shipment;
+
+						if( $remaining_quantity > $quantity_per_shipment ) {
+							$item_data['quantity'] = $quantity_per_shipment;
+						} else {
+							$item_data['quantity'] = $quantity_per_shipment + $remaining_quantity;
+						}
+
+						$items[] = $item_data;
+					}
+				} else {
+					for ( $i = 0; $i < $remaining_quantity; $i ++ ) {
+						$items[] = $item_data;
+					}
 				}
+
+
+
 			}
 
 			return $items;
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -237,10 +237,6 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 				 * exceeds `wc_connect_shipment_item_quantity_threshold`
 				 */
 				$max_shipments = apply_filters( 'wc_connect_max_shipments_if_quantity_exceeds_threshold', 5 );
-				/**
-				 * Default quantity per item if the `wc_connect_shipments_item_threshold_quantity` is not exceeded.
-				 */
-				$quantity_per_shipment = 1;
 
 				$weight_per_item = $item_data['weight'];
 				$should_cap_shipments = $remaining_quantity > $threshold;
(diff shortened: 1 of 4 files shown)


samnajian:
@kloon Please let me know if I should adjust the default filter values.

classes/class-wc-connect-shipping-label.php
```diff
@@ -226,10 +225,48 @@ public function get_all_items( WC_Order $order ) {
 				}
 
 				$refunded_qty = $order->get_qty_refunded_for_item( $item->get_id() );
+				$remaining_quantity = $item['qty'] - absint( $refunded_qty );
+				/**
+				 * The threshold at which we will start batching items together.
+				 * As an example, if a single order item has a quantity 60, which is more than the default threshold
+				 * value of 20 we will start batching together.
+				 */
+				$threshold = apply_filters( 'wc_connect_shipment_item_quantity_threshold', 20 );
```
kloon, line 234: Do we have an indication at what number of items the issues starts presenting itself? 20 seems like a very low number. If we have data to we could work based on that to determine a threshold closer to the actual limit.

samnajian: actually the number can be pretty high, like in thousands, but this is not based on data, as this is very much dependent on the server's allocated resources and also the resources available for usage by the browser, but on my local machine this number can be very high.
On usability side, let's say we make the cut for excess of 60 per item, then we'll be showing 60 rows of for this single item, please see how _quantity_ is converted to a row on the shipping label modal.

classes/class-wc-connect-shipping-label.php
```diff
@@ -226,10 +225,48 @@ public function get_all_items( WC_Order $order ) {
...
+				/**
+				 * Max number of shipments allowed to be created for this item should quantity of this order items
+				 * exceeds `wc_connect_shipment_item_quantity_threshold`
+				 */
+				$max_shipments = apply_filters( 'wc_connect_max_shipments_if_quantity_exceeds_threshold', 5 );
+				/**
+				 * Default quantity per item if the `wc_connect_shipments_item_threshold_quantity` is not exceeded.
+				 */
+				$quantity_per_shipment = 1;
```
dustinparker: We can remove this, right?

dustinparker:
The logic looks correct. I haven't been able to test in the UI or PHP tests as I'm running into some script/dependency errors and need to install a fresh local environment overnight (Calypso and deps take forever to reinstall).

classes/class-wc-connect-shipping-label.php
```diff
@@ -226,10 +225,48 @@ public function get_all_items( WC_Order $order ) {
...
+				$weight_per_item = $item_data['weight'];
+				$should_cap_shipments = $remaining_quantity > $threshold;
+
+				if ( $should_cap_shipments ) {
+					$quantity_per_shipment = floor( $remaining_quantity / $max_shipments );
+					for ( $i = 0; $i < $max_shipments; $i ++ ) {
+						$item_data['weight'] = $quantity_per_shipment * $weight_per_item;
+						$remaining_quantity -= $quantity_per_shipment;
+
+						if( $remaining_quantity > $quantity_per_shipment ) {
```
dustinparker:
```suggestion
						if( $remaining_quantity >= $quantity_per_shipment ) {
```

dustinparker: This fixes a bug when the remaining quantity can be evenly divided by the threshold (e.g. 160):

screenshot-localhost_8050-2024 05 07-07_33_54

classes/class-wc-connect-shipping-label.php
```diff
@@ -226,10 +225,48 @@ public function get_all_items( WC_Order $order ) {
...
+				 */
+				$quantity_per_shipment = 1;
+
+				$weight_per_item = $item_data['weight'];
+				$should_cap_shipments = $remaining_quantity > $threshold;
+
+				if ( $should_cap_shipments ) {
+					$quantity_per_shipment = floor( $remaining_quantity / $max_shipments );
+					for ( $i = 0; $i < $max_shipments; $i ++ ) {
+						$item_data['weight'] = $quantity_per_shipment * $weight_per_item;
```
dustinparker, line 251: Should we maybe round the weight to 2 decimal places?

screenshot-localhost_8050-2024 05 07-07_49_56

classes/class-wc-connect-shipping-label.php
dustinparker: Once the `$remaining_quantity` gets below the `$quantity_per_shipment`, the weight isn't being calculated correctly. We need to move the `$item_data['weight']` assignment below the `$item_data['quantity']` assignment and use the newly assigned quantity like so:

```
$item_data['weight'] = round($item_data['quantity'] * $weight_per_item, 2);
```

The final for loop would then look like this:

```
					for ( $i = 0; $i < $max_shipments; $i++ ) {
						$remaining_quantity -= $quantity_per_shipment;

						if ( $remaining_quantity >= $quantity_per_shipment ) {
							$item_data['quantity'] = $quantity_per_shipment;
						} else {
							$item_data['quantity'] = $quantity_per_shipment + $remaining_quantity;
						}

						$item_data['weight'] = round( $item_data['quantity'] * $weight_per_item, 2 );

						$items[] = $item_data;
					}
```

dustinparker (changes requested):
Overall the functionality is a huge improvement for large quantity orders. Great work! I left a few comments regarding some bugs I found.

samnajian:
@dustinparker I've addressed your comments and suggestions in a083615d794d008d98a4316d566453b1d9a1559a, please take another look.

kloon:
> The logic looks correct. I haven't been able to test in the UI or PHP tests as I'm running into some script/dependency errors and need to install a fresh local environment overnight (Calypso and deps take forever to reinstall).

@dustinparker add the below to your ~/.gitconfig file or checkout the repo using https URL instead if the git url.
```
[url "https://github.com"]
	insteadOf = git://github.com
```

kloon (approved):
The latest changes work as intended and all tests pass. We just need a changelog entry before merging.

dustinparker (approved):
Everything works as expected. LGTM! We just need to copy over the changelog entry to the readme.txt.

anastas10s-afk:
Issue reported in 8351887-zen.

kloon:
@anastas10s-afk The following snippet will allow the merchant to create up to 10 shipments
```php
add_filter( 'wc_connect_max_shipments_if_quantity_exceeds_threshold', function( $max_shipments ) {
    return 10;
} 10, 1 );
```

PR #2743, opened 2024-06-06
Define Feature Announcement component and open it on click [122]

samnajian:
## Description
This PR implements the Feature Announcement modal and adds in on the order detail page.

### Notes:
- Please note the eligility check is left to [another issue](https://github.com/woocommerce/shipping/issues/126)  and just to make the current PR testable some parts of it were added to this PR.
- The functionality on the frontend ( update, snooze ) will be added in [another issue](https://github.com/woocommerce/shipping/issues/123).

### Related issue(s)

https://github.com/woocommerce/shipping/issues/122

- Checkout this branch.
- Build the project.
- Set [this boolean to true](https://github.com/Automattic/woocommerce-services/pull/2743/files#diff-d4f6c48565de5a19d0ea35c131cab7b53f92578a1ac94bd9bc3f8edb6472f274R620-R622).
- Visit an order page and click on **Create shipping label** button.
- Verify the Feature Announcement modal opens up and overlaps the purchase modal.
- Verify the modal looks adheres to the Figma design mentioned on the issue.

### Checklist

- [ ] unit tests


client/components/migration/feature-announcement.jsx
```diff
@@ -0,0 +1,133 @@
...
+		onRequestClose={closeModal}
+	>
+		<Flex gap={8}>
+			<FlexItem>
+				<Flex>
+					<span>
+						{translate('Update')}
+					</span>
+					<h2>
+						{translate('A new dedicated WooCommerce Shipping extension is now available')}
```
harriswong, line 51: 💅 – Cosmetic change 
The text section doesn't have full white background extended to the bottom
image

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/index.js
```diff
@@ -20,77 +20,70 @@ import RatesStep from './rates-step';
...
 		loaded,
 		form: loaded && shippingLabel.form,
 		showPurchaseDialog: shippingLabel.showPurchaseDialog,
-		isCustomsFormRequired: isCustomsFormRequired( state, orderId, siteId ),
+		isCustomsFormRequired: isCustomsFormRequired(state, orderId, siteId),
 	};
 };
 
 const mapDispatchToProps = dispatch => {
-	return bindActionCreators( { exitPrintingFlow }, dispatch );
```
harriswong: ❓️ – Explanations needed

This PR has a few of these white space/new line changes. I don't think they are necessary since WP prefers these spaces between brackets. I am not sure why `eslint` isn't complaining about it 🤷

client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
```diff
@@ -720,3 +720,12 @@ export const getStateForCurrentPackage = ( state, orderId ) => {
 		''
 	)
 }
+
+export const isEligableToMigrate = ( state, orderId, siteId = getSelectedSiteId( state ) ) => {
```
harriswong: ❔ – Curiosity – “Good to go, but I’d like to learn more.”
Is `orderId` being used? Should we remove it if it's not needed?

client/components/migration/feature-announcement.jsx
```diff
@@ -0,0 +1,133 @@
...
+import { Dashboard, Preformatted, LessonPlan, Shipping } from './icons';
+import {
+	isEligableToMigrate,
+} from 'woocommerce/woocommerce-services/state/shipping-label/selectors';
+
+const FeatureAnnouncement = ({ translate, isEligable }) => {
+	const [isOpen, setIsOpen] = useState(isEligable);
+	const [isUpdating, setIsUpdating] = useState(false);
+
+	const closeModal = useCallback(() => {
```
harriswong: ❔ – Curiosity – “Good to go, but I’d like to learn more.”
`useCallback` is used to optimize performance. To me, this feels like pre-mature optimization to skip re-rendering. Also, the 2nd parameter for dependencies is missing. How does React's `useCallback` if the dependency is empty? Does it always return the cache function or does it never return the cache function? How does it check when to use the cached version?

client/components/migration/feature-announcement.jsx
```diff
@@ -0,0 +1,133 @@
...
+
+	const update = useCallback(() => {
+		// Todo: implement update
+		setIsUpdating(true);
+		setTimeout(() => setIsUpdating(false), 2000);
+	});
+
+	return <>{isOpen && (<Modal
+		className="migration__announcement-modal"
+		shouldCloseOnEsc={false}
```
harriswong, line 40: ❔ – Curiosity – “Good to go, but I’d like to learn more.”

Should pressing `esc` not close this modal?

client/components/migration/feature-announcement.jsx
```diff
@@ -0,0 +1,133 @@
...
+			<FlexItem>
+				<Flex>
+					<span>
+						{translate('Update')}
+					</span>
+					<h2>
+						{translate('A new dedicated WooCommerce Shipping extension is now available')}
+					</h2>
+					<p>{translate(
+						'WooCommerce Shipping and WooCommerce Tax are now two dedicated extensions. We\'ll automatically deactivate WooCommerce Shipping & Tax and carry over your settings when you update.')}< /p>
```
harriswong, line 54: 🚧 – Request changes – “It’s quite a blocking problem.”

`JSX element 'p' has no corresponding closing tag.ts(17008)` From my editor

harriswong (commented):
Tested and the modal popped up. I have a few questions, LGTM otherwise.

client/components/migration/feature-announcement.jsx
samnajian: This only happens on my end if the console is open i.e. the height of the page is very small, does this always happen on your end?

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/index.js
samnajian: I was expecting eslint to check my code, but seemingly it doesn't not, I was not sure what's the preferred formatting on this repo, I'd update if I you suggest a reference we follow for this repo.

client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js
samnajian: Addresses in fab5a81b3ce2c6178046cd54c90985b7e4b365ff

client/components/migration/feature-announcement.jsx
samnajian: We want the users to migrate, I'd rather remove one more way of skipping the migration, but I'm happy to change this if you think we'd better allow them to use escape.
Beside using escape when 2 modals are open may get confusing.

client/components/migration/feature-announcement.jsx
samnajian: the `p` tag is being closed and my IDE doesn't nag about it and it builds and works fine.

client/components/migration/feature-announcement.jsx
samnajian: When dependencies array is left empty or not provided, the function is always cached between renders., which is fine at least for now as these functions are simply a stub and are to be implemented properly when working on https://github.com/woocommerce/shipping/issues/123

client/components/migration/feature-announcement.jsx
harriswong: Yea. 
image

client/components/migration/feature-announcement.jsx
harriswong: > When dependencies array is left empty or not provided, the function is always cached between renders.

I would like to learn more about this. Can you find me a reference please? It seems to contradict this https://react.dev/reference/react/useCallback#every-time-my-component-renders-usecallback-returns-a-different-function.

client/components/migration/feature-announcement.jsx
harriswong: I am not sure. But if we want to prevent users from exiting this modal and force them to go through it. Then should we remove the "x" button at the top right?

client/components/migration/feature-announcement.jsx
harriswong: Ok, then let's stick with your editor.

client/components/migration/feature-announcement.jsx
samnajian: Addressed in 8629ceb98894b4e21c63f6eb2abe057d3da09f58

client/components/migration/feature-announcement.jsx
samnajian: > If you forget the dependency array, useCallback will return a new function every time:

I was wrong!
Would it work for you if we left this to the other PR that's gonna take care of the functionality of these buttons? Hopefully the implementer of that feature doesn't make the same mistake.

client/components/migration/feature-announcement.jsx
samnajian: Then they can't use the plugin's normal functionality, it will be a hard force! removing the **escape** button functionality is only removing one way of skipping the upgrade while still allowing them to close it if they want. We can ask for opinions from @MatthiasReinholz on this, or we can keep it as is for now and collect their feedback when they test it, I'm not convinced this a big problem.

samnajian:
@harriswong Please feel free to give this PR another look if you're fine with both of the followings:
 
https://github.com/Automattic/woocommerce-services/pull/2743#discussion_r1633839387
https://github.com/Automattic/woocommerce-services/pull/2743#discussion_r1633836208

client/components/migration/feature-announcement.jsx
harriswong: I agree that the `shouldCloseOnEsc` and `shouldCloseOnClickOutside` questions do not block this PR. +1 to discuss it and implement afterwards if needed.

harriswong (changes requested):
🚧 – Request changes

I tested mobile view (390x844) and it looks visually squeezed together
image
image

Figma doesn't have a mobile design so I don't know what the expectation is. However, I do think it is unreadable. A quick fix could be to move the picture to the top and have the text at the bottom. Thoughts?

client/components/migration/feature-announcement.jsx
harriswong: No problem! Thanks for checking. Then, I suggest we remove `useCallback` for the `closeModal`, `snooze` and `update` functions to something like
```js
	const closeModal = () => setIsOpen(false);

	const snooze = () => {
		// Todo: implement maybe later
	};

	const update = () => {
		// Todo: implement update
		setIsUpdating(true);
		setTimeout(() => setIsUpdating(false), 2000);
	};
```

Whoever is working on these 3 functions can then decide to add the `useCallback` should there be a need for any performance optimization. Thoughts?

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/index.js
harriswong: `.eslintignore` ignores this folder so `eslint` won't report any errors. I don't think we need to remove the whitespace/newline because they look like they were adhering to the WordPress/Woo Standard. Removing the space seems to be a step away from the standard, which is why I am calling it out as a code reviewer.

> I was not sure what's the preferred formatting on this repo

I usually default to the WordPress/Woo standards. Ideally, `eslint --fix` will fix this for us. I don't want to come off as being nitpicky and I also don't have a strong argument given that this file is ignored in `.eslintignore`. We can check with the team or you can use your best judgement.

client/components/migration/feature-announcement.jsx
samnajian: >  should there be a need for any performance optimization

We don't always need to measure to find out about performance optimizations if:
1) We know there is a probable issue.
2) The so-to-say optimization comes with little to zero cost in terms of development time or readability, like in here.

Anyways I'm removing it for now, the other PR may add it though.

samnajian:
> mobile

This what I see on simulated mobile anyways ( without this Feature announcement modal ) 
I can't see how the extension has been usable on mobile so far, and not sure about our policy for mobile support.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/index.js
samnajian: > Ideally, eslint --fix will fix this for us.

It doesn't fix the issues, not it does it for the newly added file (client/components/migration/feature-announcement.jsx) 
I remember we had a discussion about this on Feb 22nd 1Ju-42fxF1FiOymv0mrRAnJZfFsWZ3jPTGu06BGKX8Co-gdoc and we agreed to leave these things to our linters!

samnajian: The commit 2e2f50b2daf9899358328ea113ff7764566c46f5  reverts the changes in the old file, but the linter didn't do anything for the new file.

client/components/migration/feature-announcement.jsx
harriswong: I had a 1:1 chat with @samnajian. Adding some clarity to the issue here (discovered by my VSCode editor.

The issue here is `< /p>` gives me the following errors:
image
image

The fix is to update `< /p>` to `</p>` but I can fix that later when I work on this file.

harriswong (approved):
I had a 1:1 chat with @samnajian. Regarding the mobile view, we will tackle that ~will~ with a follow up.

LGTM

harriswong:
I will merge it for you per our discussion on the call.

PR #2745, opened 2024-06-07
Install and activate a list of plugins during migrations

harriswong:
This PR installs and activates  "WooCommerce Shipping" and WooCommerce Tax" upon clicking the "Update" button. It also deactivates WCS&T afterwards.

The order is:
1. Install WooCommerce Shipping.
2. Install WooCommerce Tax.
3. Activates WooCommerce Shipping.
4. Activates WooCommerce Tax.
5. Deactivate WCS&T

## TODO
- Upon clicking "Update", it should update a flag to indicate that the user is migrated. This will be done via another PR. 
- Error handling if install/activation fails. Where do the errors go?

### Technical considerations
I contemplated using `@wordpress/apiFetch` or not. I tried using it at first thinking that:
1. It fits better with the wp-data way.
6. It handles nonces.
Unfortunately, WCS&T does not include the `@wordpress/apiFetch` package. I tried installing this (first commit) and can verify that it also doesn't automatically handle nonce. 

At the end, I removed the package and opted to use the native `fetch` instead. In this repo, we have a library that talks to the connect server. This client uses a nonce. This PR utilizes the same nonce. 

### Related issue(s)
N/A

### Testing activation
1. Go to `woocommerce-services/classes/class-wc-connect-service-settings-store.php` and set the following to `true`
```
			return true;
		}
```
2. Make sure you have "WooCommerce Shipping" and "WooCommerce Tax" in your plugins folder. We can't test the "install" part because they aren't part of dotorg yet.
3. Go to an individual order where you can see the WCS&T shipping banner.
4. Click "Create shipping label" 
image
5. Click "Update now"
image
6. You should see 3 requests go by the network tab. `install`, `activate` and a `woocommerce-service`
image
7. Confirm that you are redirected to the plugins page and confirm that WCS&T is deactivated, WooCommerce Shipping and WooCommerce tax are activated. 
image

### Testing installation
To test installation, we can test it against the `hello-dolly` plugin. 
1. Go to http://localhost/wp-admin/plugins.php and delete your "Hello Dolly" plugin.
2. Open `client/components/migration/feature-announcement.jsx`, change line 34 from `const plugins = 'woocommerce-shipping,woocommerce-tax';` to `const plugins = 'hello-dolly';`. This will now install and activate hello dolly.
3. Repeat step 3-7 above and confirm "Hello Dolly" is installed and activated. 

### Checklist

- [ ] unit tests

### DIFF
--- a/client/components/migration/feature-announcement.jsx
+++ b/client/components/migration/feature-announcement.jsx
@@ -31,14 +31,16 @@ const FeatureAnnouncement = ({ translate, isEligable }) => {
 	};
 
 	const update = () => {
-		const plugins = 'hello-dolly'; //this needs to be a CSV string.
+		const plugins = 'woocommerce-shipping,woocommerce-tax'; //this needs to be a CSV string.
+		const headers = {
+			"Content-Type": "application/json",
+			'X-WP-Nonce': getNonce()
+	   };
+
 		const installPluginAPICall = () =>
 			fetch( '/wp-json/wc-admin/plugins/install', {
 				method: 'POST',
-				headers: {
-					 "Content-Type": "application/json",
-					 'X-WP-Nonce': getNonce()
-				},
+				headers,
 				body: JSON.stringify({
 					plugins
 				})
@@ -47,21 +49,28 @@ const FeatureAnnouncement = ({ translate, isEligable }) => {
 		const activatePluginAPICall = () =>
 			fetch( '/wp-json/wc-admin/plugins/activate', {
 				method: 'POST',
-				headers: {
-					 "Content-Type": "application/json",
-					 'X-WP-Nonce': getNonce()
-				},
+				headers,
 				body: JSON.stringify({
 					plugins
 				})
 			} );
 
+		const deactivateWCSTPluginAPICall = () => {
+			fetch( '/wp-json/wp/v2/plugins/woocommerce-services/woocommerce-services', {
+				method: 'POST',
+				headers,
+				body: JSON.stringify({
+					status: 'inactive'
+				})
+			} );
+		}
 
 		const installAndActivatePlugins = async() => {
 			try {
 				setIsUpdating(true);
 				await installPluginAPICall();
 				await activatePluginAPICall();
+				await deactivateWCSTPluginAPICall();
 				setIsUpdating(false);
 			} catch (e) {
 				//TODO: error handling.
--- a/client/components/migration/feature-announcement.jsx
+++ b/client/components/migration/feature-announcement.jsx
@@ -7,7 +7,7 @@ import { useState } from '@wordpress/element';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { localize } from 'i18n-calypso';
-import { getNonce } from 'api/request'; // client/api/request.js
+import { getNonce, getBaseURL } from 'api/request'; // client/api/request.js
 
 /**
  * Internal dependencies
@@ -36,9 +36,10 @@ const FeatureAnnouncement = ({ translate, isEligable }) => {
 			"Content-Type": "application/json",
 			'X-WP-Nonce': getNonce()
 	   };
+	   console.log('getBaseURL', getBaseURL());
 
 		const installPluginAPICall = () =>
-			fetch( '/wp-json/wc-admin/plugins/install', {
+			fetch( getBaseURL() + 'wc-admin/plugins/install', {
 				method: 'POST',
 				headers,
 				body: JSON.stringify({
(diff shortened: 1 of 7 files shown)


harriswong:
### June 7 , 2024 update
This PR has a button to demo the "install and activate plugin" function. By clicking the "Activate plugin" button, it will install and activate the "hello-dolly" plugin.

image

harriswong:
### June 11th update - Bind the buttons to the actual modal

Updated the draft to bind the install/activate to the "update" button.
image

client/components/migration/feature-announcement.jsx
```diff
@@ -30,9 +31,55 @@ const FeatureAnnouncement = ({ translate, isEligable }) => {
...
+
+
+		const installAndActivatePlugins = async() => {
+			try {
+				setIsUpdating(true);
+				await installPluginAPICall();
+				await activatePluginAPICall();
+				await deactivateWCSTPluginAPICall();
+				setIsUpdating(false);
+				window.location = '/wp-admin/plugins.php';
```
samnajian: While it works, I was wondering if we ran [admin_url](https://developer.wordpress.org/reference/functions/admin_url/) to get this url and the localize is on PHP side and use it here would be better? Please feel free to update or leave it as is.

samnajian (commented):
For some reason the endpoints are not accessible, do we need to do anything to enable them? And if so, using these endpoints are probably not the best way of implementing this feature.

https://github.com/Automattic/woocommerce-services/assets/789421/53b0160a-366f-4516-bbee-02f04daf15ec

Besides while testing, I noticed  even despite the failure I'm redirected to the plugins page, that's why I had to comment out the redirection out to be able to see what's happening, we'd probably need to do something for it, na?

```
  setIsUpdating(true);
  await installPluginAPICall();
  await activatePluginAPICall();
  await deactivateWCSTPluginAPICall();
  setIsUpdating(false);
  // window.location = '/wp-admin/plugins.php';

```

I see you've asked about error management, but we probably would better at least not direct should any of the ajax calls fail, what do you think?

Other than this, how about chaining the API calls and not do them in parallel so if any of them throws it's easier to handle it? ( This is only an idea a question, you won't need to change it as I'm not sure either ).

harriswong:
> For some reason the endpoints are not accessible

🤔 `404` means that the URL isn't found. Maybe it's not locating the proper relative path. I will take a look.

> I noticed even despite the failure I'm redirected to the plugins page, that's why I had to comment out the redirection out to be able to see what's happening, we'd probably need to do something for it, na?

Good catch, I think it would be best to circuit break on these. If any of the call throws 4xx or 5xx, then jump to the exception/error handling that will be taken care of in https://github.com/woocommerce/shipping/issues/134.

> how about chaining the API calls and not do them in parallel so if any of them throws it's easier to handle it?

Thanks. They are run with `await` and they shouldn't be running in parallel. The 3 calls 
```
  await installPluginAPICall();
  await activatePluginAPICall();
  await deactivateWCSTPluginAPICall();
```
should be running in sequence. I think we are okay on this but please let me know otherwise.

harriswong:
Hey @samnajian, what's your WooCommerce version please?

samnajian:
> Hey @samnajian, what's your WooCommerce version please?

It's Version 8.9.3. 
My local dev was loading very slow and acting weirdly, just prior to testing this feature I nuked my WP installation and re-install everything from scratch, thought this might help, basically I tested this feature with a clean install.

samnajian:
> Thanks. They are run with `await` and they shouldn't be running in parallel. The 3 calls

Ah you're right, since all 3 were failing very fast, it gave me the impression that the they are dispatched in parallel.

samnajian:
> Good catch, I think it would be best to circuit break on these. If any of the call throws 4xx or 5xx, then jump to the exception/error handling that will be taken care of in [woocommerce/shipping#134](https://github.com/woocommerce/shipping/issues/134).

👍

harriswong:
> > Hey @samnajian, what's your WooCommerce version please?
> 
> It's Version 8.9.3. My local dev was loading very slow and acting weirdly, just prior to testing this feature I nuked my WP installation and re-install everything from scratch, thought this might help, basically I tested this feature with a clean install.

Hey @samnajian. I can't reproduce this even with a new docker container, using 8.9.3. To help me see what the issue is, do you see any errors in your error logs please?

client/components/migration/feature-announcement.jsx
harriswong: Thanks, that's a good idea. Updated in https://github.com/Automattic/woocommerce-services/pull/2745/commits/d370fdf03e4e087cdb1b1aacf6fca3c294b5279e and https://github.com/Automattic/woocommerce-services/pull/2745/commits/96fc7c042466b4ceb4a685366928a473d7c0764c.

harriswong:
Hey @samnajian, I updated the codes with the following, please take a look again.
- The API URLs are now using relative paths. Hopefully that addresses the 404 errors you were experiencing.
- "install, activate, deactivate" are now put into a "circuit breaker". If any 1 of these breaks, it will throw an exception. https://github.com/Automattic/woocommerce-services/pull/2745/commits/bad3bdc776f9613cfadb6984a822e3140ecb9243

samnajian (approved):
LGTM, thanks for the updates.

harriswong:
:O It worked? Sweeeet!

PR #2747, opened 2024-06-12
Set a flag to indicate migration has started or has completed

harriswong:
## Description
Upon clicking "Update now", this PR sets the migration state to `MIGRATION_STATE_STARTED` (`2`) before the migration begins. Then, the plugin install, activates WooCommerce Shipping and WooCommerce Tax. Followed by deactivating WCS&T. When it's done, it sets the migration state to `MIGRATION_STATE_COMPLETED` (`3`).

### Related issue(s)
NA

1. Pull this branch, make sure you already have your WCS&T, Woo Shipping, Woo Tax setup in the plugins folder. Note, we can't test "install" because these aren't found on .org yet. If you want to test the install steps, you can test it with the "hello-dolly" plugin as described here https://github.com/Automattic/woocommerce-services/pull/2745.
2. Open `classes/class-wc-connect-service-settings-store.php` and change `is_eligible_for_migration()` to return `true`. Something like this `return $migration_state !== WC_Connect_API_Constants::MIGRATION_STATE_COMPLETED && true;`. The `true` flag mimic the connect server's flag.
3. Go to any order page and click "Create shipping label" 
image
4. You should see the migration modal, click F12 to see the network tab in console. And click "Update now"
image
5. Confirm you are being redirected to plugins page and both WooCommerce Shipping and WooCommerce Tax are activated; while WCS&T is deactivated. Confirm that you are seeing the 4 requests in the network tab. 
image
6. [bonus] Check your database table `wp_options` where `name` =`wc_connect_options`. In the serialized `option_value`, you should see `"wcshipping_migration_state";i:3;`.

### Re-running the test
This PR sets a flag `wcshipping_migration_state` so once you "completed" migration, the modal won't show up again. To get around this, add `WC_Connect_Options::delete_option('wcshipping_migration_state');` to the `is_eligible_for_migration()` function so it will always delete the option first. For example, my local dev testing function looks like this:
```php
			WC_Connect_Options::delete_option('wcshipping_migration_state');
			//TODO: Check against the connect-server flag
			return $migration_state !== WC_Connect_API_Constants::MIGRATION_STATE_COMPLETED && true;
		}
```

### Checklist

- [ ] unit tests


woocommerce-services.php
```diff
@@ -249,6 +256,18 @@ class WC_Connect_Loader {
...
 		public static function plugin_deactivation() {
 			wp_clear_scheduled_hook( 'wc_connect_fetch_service_schemas' );
+
+			/**
+			 * When we deactivate the plugin after wcshipping_migration_state has started,
+			 * that means the migration is done. We can mark it as completed before we deactivate the plugin.
+			 */
+			$migration_state = WC_Connect_Options::get_option( 'wcshipping_migration_state' );
+			if ( ! class_exists(WC_Connect_API_Constants::class) ) {
+				require_once __DIR__ . '/classes/class-wc-connect-api-constants.php';
```
harriswong: When deactivating, somehow the files aren't loaded. I have to manually load them here. This is also the reason why I moved the constants into its own top layer file.

client/components/migration/feature-announcement.jsx
```diff
@@ -63,10 +63,24 @@ const FeatureAnnouncement = ({ translate, isEligable }) => {
...
+				body: JSON.stringify({
+					migration_state: 2 // Check WC_Connect_API_Constants::MIGRATION_STATE_STARTED
+				})
+			} );
 
 
 		const installAndActivatePlugins = async() => {
-			const tasks = [installPluginAPICall, activatePluginAPICall, deactivateWCSTPluginAPICall];
+			const tasks = [
+				markMigrationStartedAPICall,
```
harriswong, line 78: Mark the migration has started, followed by installing, then activating.

classes/class-wc-rest-connect-migration-flag-controller.php
```diff
@@ -0,0 +1,44 @@
...
+	protected $rest_base = 'connect/migration-flag';
+
+	private function is_valid_state($state) {
+		$valid_states = [
+			WC_Connect_API_Constants::MIGRATION_STATE_NOT_STARTED,
+			WC_Connect_API_Constants::MIGRATION_STATE_STARTED,
+			WC_Connect_API_Constants::MIGRATION_STATE_COMPLETED,
+			WC_Connect_API_Constants::MIGRATION_STATE_FAILED
+		];
+		if (in_array($state, $valid_states, true)) {
```
harriswong: I had this as `$state >= 1 && $state <= 4` because it's simpler and more efficient than `in_array`, but that smelled a bit, especially if we started removing constants and skipping integer sequences. I thought about adding an enum class but that's too complex for this.

At the end, I used a simple array of constants just so that we could verify the input.

classes/class-wc-rest-connect-migration-flag-controller.php
```diff
@@ -0,0 +1,44 @@
...
+				'invalid_migration_state',
+				__( 'Invalid migration state. Can not update migration state.', 'woocommerce-services' ),
+				array( 'status' => 400 )
+			);
+			return $error;
+		}
+
+		$result = WC_Connect_Options::update_option( 'wcshipping_migration_state', $migration_state );
+
+		return new WP_REST_Response( array( 'result' => $result ), 200 );
```
harriswong: I thought about returning `new WP_Error()` if the `$result` is `false`. At the same time, I think we can let it fail gracefully with a 200 with `result=>false`. The UI can figure out what we want to do with the failed `update_option` call. Thus, I used the latter approach.

We can look at error handling around installation and activation holistically in woocommerce shipping issue 134.

client/components/migration/feature-announcement.jsx
```diff
@@ -63,10 +63,24 @@ const FeatureAnnouncement = ({ translate, isEligable }) => {
 					status: 'inactive'
 				})
 			} );
+		const markMigrationStartedAPICall = () =>
+			fetch( getBaseURL() + 'wc/v1/connect/migration-flag', {
+				method: 'POST',
+				headers,
+				body: JSON.stringify({
+					migration_state: 2 // Check WC_Connect_API_Constants::MIGRATION_STATE_STARTED
```
harriswong: I thought about using a constant for `2` but also thought that we would have to create a mapping to map the PHP side, which seems a bit too complex for what we are trying to do here. I left a comment to indicate where the constant is coming from instead.

classes/class-wc-connect-api-constants.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * A class for constants used by the API.
+ */
+class WC_Connect_API_Constants {
```
waclawjacek: WDYT about not using the word "API"? Is this something that is related to the API?

classes/class-wc-rest-connect-migration-flag-controller.php
waclawjacek: WDYT about moving the `is_valid_state` method to the class that's defining the consts? That way when you're updating the consts, you can also update this method in the same place, reducing the risk of forgetting to update the `is_valid_state` method.

classes/class-wc-rest-connect-migration-flag-controller.php
```diff
@@ -0,0 +1,44 @@
...
+		if (!$this->is_valid_state($migration_state)) {
+			$error = new WP_Error(
+				'invalid_migration_state',
+				__( 'Invalid migration state. Can not update migration state.', 'woocommerce-services' ),
+				array( 'status' => 400 )
+			);
+			return $error;
+		}
+
+		$result = WC_Connect_Options::update_option( 'wcshipping_migration_state', $migration_state );
```
waclawjacek, line 26: Is the option something that will be used by WCShip? Asking because:

- Placing it in its own option rather than in wc_connect_options would make it easier to access, but
- Then the name might have to be reconsidered (`wcservices_` or something instead of `wcshipping_` )

classes/class-wc-rest-connect-migration-flag-controller.php
waclawjacek: Mmmmmmmmmm I don't like returning `200` and `result = false` :< Might be personal preference but this is something that might take me by surprise as I'm working on consuming code. Can we just not return the `result`, or return `500` if `update_option` fails?

Caveat: `update_option` returns `false` if `$value === $old_value`

classes/class-wc-connect-service-settings-store.php
```diff
@@ -618,7 +618,9 @@ public function get_package_lookup() {
 		}
 
 		public function is_eligible_for_migration() {
-			return false;
+			$migration_state = WC_Connect_Options::get_option( 'wcshipping_migration_state' );
+			//TODO: Return $migration_state check && connect-server flag
```
waclawjacek, line 621: Is there an issue for this? Could you create it if there isn't one?

classes/class-wc-connect-api-constants.php
```diff
@@ -0,0 +1,11 @@
...
+	// These are used for WCS&T to WCShipping migration.
+	public const MIGRATION_STATE_NOT_STARTED = 1;
+	public const MIGRATION_STATE_STARTED = 2;
+	public const MIGRATION_STATE_COMPLETED = 3;
+	public const MIGRATION_STATE_FAILED = 4;
```
waclawjacek: WDYT about something like this?

```suggestion
class WC_Connect_WCST_To_WCShipping_Migration_State_Enum {

	public const NOT_STARTED = 1;
	public const STARTED = 2;
	public const COMPLETED = 3;
	public const FAILED = 4;
```

client/components/migration/feature-announcement.jsx
waclawjacek: Sweet!

client/components/migration/feature-announcement.jsx
waclawjacek: Agreed!

woocommerce-services.php
```diff
@@ -249,6 +256,18 @@ class WC_Connect_Loader {
...
+			}
+			if ( $migration_state === WC_Connect_API_Constants::MIGRATION_STATE_STARTED ) {
```
waclawjacek: How about we have a different migration state after plugin installation? That way we're able to distinguish "started but not installed" (if for whatever reason it is not changed to "failed" inbetween) and "started and installed". Then we can transition to completed only from "installed".

waclawjacek (commented):
Things are working as expected and the code is looking good! Posting as a comment for now because I have some questions. Also please make sure to reformat the code to match the WP Code Style (spacing and all).

classes/class-wc-connect-api-constants.php
harriswong: Yea, I was thinking this constants file either to be a "file for all WCS&T constants" or "file for connect client/API related constants", thus the naming.

Instead of creating a general constants file, I like the `WC_Connect_WCST_To_WCShipping_Migration_State_Enum` suggestion below. I will update it with an enum class.

classes/class-wc-rest-connect-migration-flag-controller.php
harriswong: I like the WC_Connect_WCST_To_WCShipping_Migration_State_Enum suggestion below. I will update it with an enum class.

classes/class-wc-connect-api-constants.php
harriswong: Updated in 85a7e614

classes/class-wc-connect-api-constants.php
harriswong: Updated in 85a7e614

classes/class-wc-rest-connect-migration-flag-controller.php
harriswong: This is a good callout, I thought about using its own options name with the WP `update_option` directly, but later opt-ed for the `WC_Connect_Options::update_option()`. 

I leaned towards using the `WC_Connect_Options::update_option()` because it follows the existing pattern to encapsulate all connect-server-related configs under 1 option name. 

> Is the option something that will be used by WCShip?

I am not sure but I don't think so. During the migration, this flag will be updated a few times and I don't think the migration script will sync this very well either. It won't be reliable even if we migrated this option over.

Maybe it is better to use its own variable?  What do you think?

classes/class-wc-rest-connect-migration-flag-controller.php
harriswong: > Caveat: update_option returns false if $value === $old_value

What about returning 200 if it's `true`, and 304 (not modified) for `false`?

woocommerce-services.php
harriswong: I drew out a finite state automata :D I think we can have 8 states and we can start from any of them and will always get to the end. This includes error handling as well. What do you think about adding this to woocommerce shipping issue 134? I can update that ticket to handle how we want to tackle each migration state in the automata?

image

classes/class-wc-connect-service-settings-store.php
harriswong: Added this to woocommerce shipping issue 136.

classes/class-wc-rest-connect-migration-flag-controller.php
harriswong: Updated in 85a7e6148804edfe858e56f13d0e1fe94ad05611

harriswong:
👋 Hi @waclawjacek, I responded to all the comments. There are 2 remaining items to be discussed:
1. Should we use a standalone option instead of using `WC_Connect_Options::update_option`?
2. For the various states, are we okay to group that work into issue 134?

Please take another look!

classes/class-wc-connect-wcst-to-wcshipping-migration-state-enum.php
```diff
@@ -0,0 +1,24 @@
...
+		$valid_states = array(
+			self::NOT_STARTED,
+			self::STARTED,
+			self::COMPLETED,
+			self::FAILED,
+		);
+		if ( in_array( $state, $valid_states, true ) ) {
+			return true;
+		}
+		return false;
```
waclawjacek: Nitpick, change not needed:
```suggestion
		return in_array( $state, $valid_states, true );
```

classes/class-wc-connect-service-settings-store.php
waclawjacek: Can't see it. Are you sure it's in woocommerce/woocommerce-shipping#136?

classes/class-wc-rest-connect-migration-flag-controller.php
waclawjacek: If this isn't used by any other plugin, I think it's fine to keep as it is now; maybe even better for consistency. Thanks!

woocommerce-services.php
```diff
@@ -249,6 +256,18 @@ class WC_Connect_Loader {
 
 		public static function plugin_deactivation() {
 			wp_clear_scheduled_hook( 'wc_connect_fetch_service_schemas' );
+
+			/**
```
waclawjacek: Super nitpicky and no need to change but since `/**` are PHPDoc DocBlocks and [not everything can be documented with a DocBlock](https://docs.phpdoc.org/3.0/guide/getting-started/what-is-a-docblock.html#what-can-be-documented-in-your-source-code), this could be a `/*` comment. :shrug: Doesn't really hurt though.

woocommerce-services.php
```diff
@@ -249,6 +256,18 @@ class WC_Connect_Loader {
...
+			 * When we deactivate the plugin after wcshipping_migration_state has started,
+			 * that means the migration is done. We can mark it as completed before we deactivate the plugin.
+			 */
+			$migration_state = WC_Connect_Options::get_option( 'wcshipping_migration_state' );
+			if ( ! class_exists( WC_Connect_WCST_To_WCShipping_Migration_State_Enum::class ) ) {
```
waclawjacek: Do we need this extra check? I imagine `require_once` already does something like this internally.

classes/class-wc-rest-connect-migration-flag-controller.php
waclawjacek: Nice, I appreciate the specificity. I think `304` is for "Not Changed" as in "there's nothing new, you don't need to redownload this thing" rather than "you requested a change but I didn't make it" though. What about one of the 5xx codes?

woocommerce-services.php
waclawjacek: Love it! Yeah, let's do it in another issue. My only concern is that we make sure we don't transition sites from "started" to "completed" when nothing really happened due to an error (and failure to update the state to "failed").

waclawjacek (approved):
LGTM! :shipit: Great work.

classes/class-wc-connect-service-settings-store.php
harriswong: https://github.com/woocommerce/shipping/issues/136

woocommerce-services.php
harriswong: > we make sure we don't transition sites from "started" to "completed" when nothing really happened due to an error (and failure to update the state to "failed").

Yea 100%. This is also why I think we should tackle this problem in https://github.com/woocommerce/shipping/issues/134 instead. 134 deal with the whole error handling. I will update the description there so that the issue needs to detect which of these 8 (or more) states the migration was in, and carry on from there.

classes/class-wc-connect-wcst-to-wcshipping-migration-state-enum.php
harriswong: 😆 Nice call out, there is no point in another condition for this. Fixed in a02e0fa1

woocommerce-services.php
harriswong: Makes sense. Changing `/**` to `/*` since this inline comment can not be a DocBlock. Fixed in 02b759b3

woocommerce-services.php
harriswong: No we don't. I took the existing pattern/practice in the WCS&T code where the import starts with a `if ( ! class_exists()` check.

It is useful when we have 2 different file paths with the same class name (collision). 

Here, we don't have this problem because `WC_Connect_WCST_To_WCShipping_Migration_State_Enum` is a class I just added to WCS&T, there is no chance of collision. `require_once` suffices.

harriswong: Updated in d14b131c

classes/class-wc-rest-connect-migration-flag-controller.php
harriswong: This one is a bit tricky. `update_option()` returns `false` if it didn't update or if there is nothing to update. `false` here implies it either failed or it didn't do anything. 
For example: If it hits this condition, then it would be a valid 304: https://github.com/WordPress/wordpress-develop/blob/2132c7bacb1626af19737ffea0bcce9e42eee997/src/wp-includes/option.php#L903

Without being too specific and too complicated, I think I agree `500` is a good choice, if the `connect/migration-flag` failed to update, we need to know something is wrong, `500` would be a good one.

harriswong: Fixed in 73f0d9f6

Tested the error: 
image

harriswong:
👋 Hey @waclawjacek , I  have addressed all your comments, please take another look!

waclawjacek (approved):
LET'S GOOOOOOOO! :shipit: (╯°□°)╯︵ ┻━┻

classes/class-wc-rest-connect-migration-flag-controller.php
waclawjacek: Ooooh, I see! I guess the solution would be to compare the values before updating and then if they're different, only then update. This would make the `false`s returned by `update_option` only indicate actual failure.

I suppose we shouldn't run into these situations when the UI governs these flows so it should be OK to keep as it is.

PR #2749, opened 2024-06-17
Chore: phpcbf and phpcs will now check against WP and WC standards

harriswong:
## Description
Running `composer phpcbf <file>` doesn't fix the file. Same for `phpcs`. This PR fixes the `lint-staged` script as well as the `phpcbf` and `phpcs` composer scripts. These will now check against the following standards `--standard=WooCommerce-Core,WordPress-Core,WordPress-Extra`.

### Related issue(s)
N/A

### Steps to test
1. Run `composer phpcbf woocommerce-services.php`
2. Confirm that you see it auto fixed 4 things and 21 remaining items
```

PHPCBF RESULT SUMMARY
------------------------------------------------------------------------------------------------------------------------------------
FILE                                                                             FIXED  REMAINING
------------------------------------------------------------------------------------------------------------------------------------
..wordpress/wp-content/plugins/woocommerce-services/woocommerce-services.php     4      21
------------------------------------------------------------------------------------------------------------------------------------
A TOTAL OF 4 ERRORS WERE FIXED IN 1 FILE
------------------------------------------------------------------------------------------------------------------------------------
```

### Checklist

- [ ] unit tests

- [ ] `changelog.txt` entry added
- [ ] `readme.txt` entry added

PR #2692, opened 2023-10-26
NUX: Fix banner display on Edit Orders pages when using HPOS

waclawjacek:
## Description

When HPOS is enabled (which has been the default for new sites since WC 8.2 release on Oct 10th, 2023), the admin screen IDs for "Edit Order" pages are different. This causes WCS&T's NUX banner to not show up. This applies to both the order listing and single order edit pages.

### Steps to reproduce & screenshots/GIFs

Screenshot 2023-10-26 at 11 33 05

1. Set up a fresh site. Install WCS&T and WC. Don't connect with Jetpack.
2. Go to the Edit Order page (listing or single order page).
3. If you're on HPOS, the banner won't show up without this PR.
4. You can toggle HPOS usage in WC > Settings > Advanced > Features. You have to have order sync enabled on that page to be able to toggle this setting.

### Checklist

- [ ] unit tests

- [x] `changelog.txt` entry added
- [x] `readme.txt` entry added

PR #2751, opened 2024-06-18
Add admin notice on orders page for WCShipping migration

harriswong:
## Description
Create an admin notice on the orders page (http://localhost/wp-admin/edit.php?post_type=shop_order). 

The notice will have a dimissible button and a "Confirm update" button. The button does not do anything yet. 

### Related issue(s)
N/A

### Steps to test
1. Open `wordpress/wp-content/plugins/woocommerce-services/classes/class-wc-connect-service-settings-store.php`
2. Change `is_eligible_for_migration()` to return `true`
```php
			return true;
		}
```
3. Go to orders page http://localhost/wp-admin/edit.php?post_type=shop_order
4. Confirm that the admin notice is there
image

### Checklist

- [ ] unit tests


woocommerce-services.php
```diff
@@ -895,6 +895,7 @@ public function attach_hooks() {
 			add_filter( 'woocommerce_admin_reports', array( $this, 'reports_tabs' ) );
 			add_action( 'woocommerce_checkout_order_processed', array( $this, 'track_completed_order' ), 10, 3 );
 			add_action( 'admin_print_footer_scripts', array( $this, 'add_sift_js_tracker' ) );
+			add_action(' current_screen', array( $this, 'edit_orders_page_actions' ) );
```
harriswong: We need to make sure `get_current_screen()` function is loaded.

woocommerce-services.php
```diff
@@ -1801,6 +1821,20 @@ public function are_woo_shipping_and_woo_tax_active() {
...
+		public function display_wcst_to_wcshipping_migration_notice() {
+			echo wp_kses_post(
+				'<div class="notice notice-error is-dismissible wcst-wcshipping-migration-notice"><p style="margin-bottom:0px">' .
+				 sprintf(
+					/* translators: %s: documentation URL */
+					__( 'WooCommerce Shipping & Tax is splitting into two dedicated extensions: WooCommerce Shipping and WooCommerce Tax. To minimize disruption, your settings and data will be carried over to the new extensions when you upgrade. <a href="%s" style="color:#3858E9;">Learn more about this change.</a>', 'woocommerce-services' ),
+					'https://wordpress.org/plugins/woocommerce-services/'
+				)  .
+				'</p>
+				<button style="color:#3858E9;margin: 12px 0;border: 1px solid #3858E9;padding: 11.5px 12px 11.5px 12px;border-radius: 2px;background-color:#fff;">Confirm update</button>
```
harriswong, line 1901: "Confirm update" doesn't do anything yet. I pinged design to double check on how we want to bind this action.

woocommerce-services.php
```diff
@@ -1801,6 +1821,20 @@ public function are_woo_shipping_and_woo_tax_active() {
 		public function display_woo_shipping_and_woo_tax_are_active_notice() {
 			echo '<div class="error"><p><strong>' . esc_html__( 'Woo Shipping and Woo Tax plugins are already active. Please deactivate WooCommerce Shipping & Tax.', 'woocommerce-services' ) . '</strong></p></div>';
 		}
+
+		public function display_wcst_to_wcshipping_migration_notice() {
+			echo wp_kses_post(
```
harriswong, line 1893: `woocommerce-services/assets/stylesheets/style.scss` isn't loaded on the orders page. We could add some hooks to load a separate stylesheet for this notice, but I chose to use inline styles instead because this notice is a standalone component and its style is not reusable by others yet. If needed, we can always refactor afterwards.

woocommerce-services.php
Ferdev: there's a space before the action name:

```suggestion
			add_action( 'current_screen', array( $this, 'edit_orders_page_actions' ) );
```

woocommerce-services.php
```diff
@@ -1440,6 +1441,25 @@ private function get_i18n_json() {
...
 		}
 
+		public function edit_orders_page_actions() {
+			if ( is_admin() ) {
+				$screen = get_current_screen();
+				if ( ! $screen || ! isset( $screen->id ) ) {
+					return false;
+				}
+
+				if ( 'edit-shop_order' !== $screen->id ) {
```
Ferdev: It doesn't work for me, because the $screen->id in the orders page, in my setup is called `woocommerce_page_wc-orders`, I'm not sure why is different? I'm using WooCommerce `v8.9.3`, not sure if that changes a lot in `v9.0.0`

Ferdev (changes requested):
I doesn't load the banner in my machine. Maybe because of the WooCommerce version? Left a comment with more info

woocommerce-services.php
harriswong: Thanks. 

😭 Soon, https://github.com/Automattic/woocommerce-services/pull/2749 will fix this for me!

harriswong: Fixed here 9a0e90c9

woocommerce-services.php
harriswong: 🤔 Mine is also WC v8.9.3. 

It looks like this is related to HPOS https://github.com/Automattic/woocommerce-services/pull/2692/files. I will borrow the logic there.

harriswong: Yup I can confirm HPOS returns "woocommerce_page_wc-orders" instead. Fixed in efc2229f. Tested against 2 installations, one with HPOS, and the other without.

harriswong:
Hey @Ferdev, I fixed the condition for screen->ID. Please take another look.

woocommerce-services.php
harriswong: Nope, that did not fix it completely. The modal now opens inside a single order as well.  I tried updating it to
```diff
if (OrderUtil::custom_orders_table_usage_is_enabled()) {
+                                       // Orders list page when using HPOS.
+                                       if (wc_get_page_screen_id( 'shop_order' ) === $screen->id) {
+                                               $should_display_wcst_to_wcshipping_migration_notice = true;
+                                       }
+                               } else {
+                                       // Orders list page when not using HPOS.
+                                       if ('shop_order' === $screen->post_type && in_array( $screen->base, array( 'edit' ), true )) {
+                                               $should_display_wcst_to_wcshipping_migration_notice = true;
+                                       }
+                               }
+
+                               if ($should_display_wcst_to_wcshipping_migration_notice) {
```
But that doesn't fix it 100% either. This also makes it very complicated.

harriswong: Ok, I added a `is_on_order_list_page()` function to check if we are on the order list page. The admin notice should show only when it's on the order listing page (HPOS or not).

Fixed in 0e270e03.

woocommerce-services.php
harriswong: 👋 Hey @Ferdev , based on the design feedback p1718744494706379-slack-C06FP5XT1S5 and p1718745005310189-slack-C06FP5XT1S5, I think we will have to change the "circle-x" to "x" (still confirming). 

If we need to replace it, I think I would pull all these css out into its own `.scss` file. This will make it much easier to overwrite the `button.is-dismissible` class. To create a css file that loads in the order listing page, I think the work involves:
- Add a new scss file like `assets/stylesheets/wcst-migration-wcshipping.scss`. 
- Figure out webpack: Webpack didn't build that `.scss` properly when I tried it, even though the [config](https://github.com/Automattic/woocommerce-services/blob/9e93c9fda94b1650a4a3d1fba8e1560a1d9fb8fa/webpack.config.js#L165) should build it. 
- I ended up adding an entry point in `.js`. Then do `import '../assets/stylesheets/wcst-migration-wcshipping.scss';`. This seems to work. We will then have to `wp_enqueue_script` and `wp_enqueue_style` these extra scripts for the css. I'm not sure if there is a better way.
- Consider if we should build our own dismiss button if it's hard to overwrite the default styles. 

If there is a quick fix, I am happy to add that here in this PR. Otherwise, maybe a follow-up?

Ferdev: hmm, webpack not building that css file is weird, apparently the `sass-loader` is looking for stuff in that path. Could be the problem the file is missing the css extension? I've seen sometimes sass preprocessors not liking filenames with just the sass extension, ie, `my-css-file.scss` instead of just `my-css-file.css.scss`.

About creating that follow up: yeah, I think it's the best approach. This way we can unlock this PR and rebase mine with it :+1:

Ferdev (approved):
That worked great, thanks! :bow:

harriswong:
Thanks for merging this for me!!

PR #2752, opened 2024-06-20
Get the migration banner info from the connect server

Ferdev:
## Description

Retrieves the migration admin banner contents from the connect server, which also allows the client to show (or not) this info to the users. The client needs to pass a specific payload `settings.wcship_migration_supported` set to `true` to the server to let it know the client is prepared to receive the banner info.

Note: the feature also needs to be enabled in the server side, which currently is only enabled for the staging server.

### Related issue(s)

N/A

1. Set to `true` the `false` statement in this line in `woocommerce-services/classes/class-wc-connect-service-settings-store.php`:

``` php
	//TODO: Return $migration_state check && connect-server flag
	return WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED !== $migration_state && false;
}
```

2. Setup WooCommerce Shipping & Tax to connect to the staging WooCommerce Connect Server: https://api-staging.woocommerce.com

3. Go to the [`WooCommerce > Orders`](http://localhost/wp-admin/edit.php?post_type=shop_order) page.

4. The banner with the right content should be displayed.

### Checklist

- [ ] unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,12 @@ If you'd just like to check out the latest code and/or wish to contribute code, 
     * For Development: run `npm run up`, let the process finish, connect your site to Jetpack using Jurassic Tube or ngrok.
     * For testing or pre-production use: Run `npm run dist` which will build the files into the `dist` folder, and will be loaded by the plugin without any additional configuration
 
-## Security
+*note*: if `npm install` gets stuck, you may have to update your GitHub urls in your git config. Use this command to fix it:
+
+```git config --global url.https://github.com/.insteadOf git://github.com/```
+
+
+## SECURITY
 
 Need to report a security vulnerability? Go to [https://automattic.com/security/](https://automattic.com/security/) or directly to our security bug bounty site [https://hackerone.com/automattic](https://hackerone.com/automattic).
 
--- a/classes/class-wc-connect-service-schemas-store.php
+++ b/classes/class-wc-connect-service-schemas-store.php
@@ -271,7 +271,7 @@ if ( ! class_exists( 'WC_Connect_Service_Schemas_Store' ) ) {
 		public function get_wcship_wctax_upgrade_banner() {
 			$service_schemas = $this->get_service_schemas();
 			// check if $service_schemas->features->wcshippingtax_upgrade_banner exists
-			if ( ! is_object( $service_schemas ) || ! property_exists( $service_schemas, 'features' ) || ! property_exists( $service_schemas->features, 'wcshippingtax_upgrade_banner' ) ) {
+			if ( ! is_object( $service_schemas ) || ! property_exists( $service_schemas, 'features' ) || ! property_exists( $service_schemas->features, 'wcshippingtax_upgrade_banner' ) || empty( $service_schemas->features->wcshippingtax_upgrade_banner ) ) {
 				return null;
 			}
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -32,7 +32,12 @@ If you'd just like to check out the latest code and/or wish to contribute code,
     * For Development: run `npm run up`, let the process finish, connect your site to Jetpack using Jurassic Tube or ngrok.
     * For testing or pre-production use: Run `npm run dist` which will build the files into the `dist` folder, and will be loaded by the plugin without any additional configuration
 
-## Security
+*note*: if `npm install` gets stuck, you may have to update your GitHub urls in your git config. Use this command to fix it:
```
harriswong, line 35: ❤️

classes/class-wc-connect-service-schemas-store.php
```diff
@@ -262,5 +262,20 @@ public function get_predefined_packages_schema() {
...
+		 * @return object|null The banners schema or null if no such id was found
+		 */
+		public function get_wcship_wctax_upgrade_banner() {
+			$service_schemas = $this->get_service_schemas();
+			// check if $service_schemas->features->wcshippingtax_upgrade_banner exists
+			if ( ! is_object( $service_schemas ) || ! property_exists( $service_schemas, 'features' ) || ! property_exists( $service_schemas->features, 'wcshippingtax_upgrade_banner' ) ) {
+				return null;
+			}
+
+			return $service_schemas->features->wcshippingtax_upgrade_banner;
```
harriswong, line 278: [Optional - an observation]

Because `property_exists()` returns true even if the value is `null`, there is a chance this line returns `null` causing the `$banner = $schema->get_wcship_wctax_upgrade_banner();` call below to throw "property doesn't exist" error.

If we want to be extra safe, we could use ` ! empty()` or `isset()` here to only return when it isn't `null`.

woocommerce-services.php
```diff
@@ -1475,7 +1475,7 @@ private function get_i18n_json() {
 		 * @return boolean
 		 */
 		public function is_on_order_list_page() {
-			if ( !is_admin() ) {
+			if ( ! is_admin() ) {
```
harriswong, line 1478: Thanks :D

woocommerce-services.php
```diff
@@ -1889,18 +1890,26 @@ public function display_woo_shipping_and_woo_tax_are_active_notice() {
...
 		}
 
+		/**
+		 * Returns the notice for migrating from WCST to WC Shipping.
+		 *
+		 * @return bool
+		 */
 		public function display_wcst_to_wcshipping_migration_notice() {
+			$schema = $this->get_service_schemas_store();
+			$banner = $schema->get_wcship_wctax_upgrade_banner();
```
harriswong, line 1900: I think this is a good idea. The server can control the message and we don't need another WCS&T patch.

woocommerce-services.php
```diff
@@ -1889,18 +1890,26 @@ public function display_woo_shipping_and_woo_tax_are_active_notice() {
...
 			echo wp_kses_post(
 				'<div class="notice notice-error is-dismissible wcst-wcshipping-migration-notice"><p style="margin-bottom:0px">' .
```
harriswong: Should we change this to use `$banner->type` instead of hardcoding it to `notice-error`?

harriswong (approved):
I tested this against the server's `wcship_migration_supported` variable. Can confirm the admin notice in the order lists page.

I have some unblocking comments (feel free to ignore or tackle via a follow-up, etc), LGTM!

classes/class-wc-connect-service-schemas-store.php
Ferdev: good call, I'll fix it!

Ferdev:
Thanks Harris! Just implemented your suggestions, I think this is good to go!

PR #2755, opened 2024-06-21
Get the flag to display the migration banner from the connect server

Ferdev:
## Description
The Connect Server's endpoint `POST /services` includes a `features.wcshippingtax_upgrade_banner` property containing all the required info to display a banner with info about the migration to WooCommerce Shipping and WooCommerce Tax.

This PR leverages all control of enabling or disabling this banner to the Connect Server's new property mentioned above.

### Related issue(s)

N/A

1. Checkout this branch in your local setup.
2. Point the Connect Server to the staging instance: `https://api-staging.woocommerce.com`.
3. Visit the orders page: `http://localhost/wp-admin/admin.php?page=wc-orders`.
4. The banner should be displayed:

shot-2024-06-21-16-50-20

5. Change the Server's endpoint to production.
6. The banner should be no hidden after reloading the orders page, as it's not enabled in production yet.

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-service-schemas-store.php
+++ b/classes/class-wc-connect-service-schemas-store.php
@@ -264,18 +264,29 @@ if ( ! class_exists( 'WC_Connect_Service_Schemas_Store' ) ) {
 		}
 
 		/**
+		 * Returns the WooCommerce Shipping and WooCommerce Tax migration enabled status
+		 *
+		 * @return bool
+		 */
+		public function is_wcship_wctax_migration_enabled() {
+			$service_schemas = $this->get_service_schemas();
+
+			return is_object( $service_schemas ) && property_exists( $service_schemas, 'features' ) && property_exists( $service_schemas->features, 'wcshippingtax_upgrade_banner' ) && ! empty( $service_schemas->features->wcshippingtax_upgrade_banner );
+		}
+
+		/**
 		 * Returns the WooCommerce Shipping and WooCommerce Tax upgrade banners
 		 *
 		 * @return object|null The banners schema or null if no such id was found
 		 */
 		public function get_wcship_wctax_upgrade_banner() {
 			$service_schemas = $this->get_service_schemas();
-			// check if $service_schemas->features->wcshippingtax_upgrade_banner exists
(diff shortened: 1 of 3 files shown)


harriswong:
👀  Driving by, going to help review this.

classes/class-wc-connect-service-schemas-store.php
```diff
@@ -263,19 +263,30 @@ public function get_predefined_packages_schema() {
...
-			// check if $service_schemas->features->wcshippingtax_upgrade_banner exists
-			if ( ! is_object( $service_schemas ) || ! property_exists( $service_schemas, 'features' ) || ! property_exists( $service_schemas->features, 'wcshippingtax_upgrade_banner' ) || empty( $service_schemas->features->wcshippingtax_upgrade_banner ) ) {
-				return null;
+
+			if ( $this->is_wcship_wctax_migration_enabled() ) {
+				return $service_schemas->features->wcshippingtax_upgrade_banner;
 			}
 
-			return $service_schemas->features->wcshippingtax_upgrade_banner;
+			return null;
```
harriswong, line 289: [Observation]

I have an observation because this part is updated, I don't think this would happen because our connect server will get updated before we push this plugin out. There is a slight chance `$this->get_service_schemas();` is cached and the feature isn't pulled from the server and this would return `null`.

When that happens, the $banner [here](https://github.com/Automattic/woocommerce-services/blob/wcs-migration/woocommerce-services.php#L1900) can be `null` and $banner->type, etc will throw undefined property error.

For that line, maybe we can do something defensive like
```
$banner = $schema->get_wcship_wctax_upgrade_banner();
if (empty($banner)) {
  return;
}
```

woocommerce-services.php
```diff
@@ -1523,7 +1523,7 @@ public function edit_orders_page_actions() {
 			// Add the WCS&T to WCShipping migratio notice, creating a button to update.
 			$settings_store = $this->get_service_settings_store();
 			$schemas_store  = $this->get_service_schemas_store();
```
harriswong: [Optional]
We can also remove `$schemas_store = $this->get_service_schemas_store();`, doesn't look like it is in used anymore.

harriswong (approved):
LGTM! I tested `wcship_migration_supported: true,` and `wcship_migration_supported: false`,. I can confirm that the connect server controls the banner. 👍 

I left 2 comments. The `null` one could be a problem but we can also have a follow-up or merge after it's patched!

woocommerce-services.php
Ferdev: Good call, totally forgot to remove that var!

PR #2756, opened 2024-06-21
Load the FeatureAnnouncement modal to the order listing page

harriswong:
## Description
This PR makes the migration admin notice's "Confirm update" button functional.
image

Upon clicking the button, the migration will install, and activate WCShipping and WCTax. 
image

This is the same as clicking the "Update now" directly in the modal, introduced in this PR https://github.com/Automattic/woocommerce-services/pull/2745.

### [main.js](https://github.com/Automattic/woocommerce-services/blob/trunk/client/main.js#L54) vs wcshipping-migration-admin-notice.js
To load `<FeatureAnnouncement>` on the order list page, we could use the existing `getRouteClass` pattern found in [main.js](https://github.com/Automattic/woocommerce-services/blob/trunk/client/main.js#L54). This way, we wouldn't need to create our own webpack's entry point. In order to use this pattern, we need to follow the `enqueue_wc_connect_script()`  code to load the javascript and style, etc. There are also a lot of variables added in here that we don't need. For the sake of the migration modal, I thought it would be simpler (and we'd have more control) to create our own [entry point](https://github.com/Automattic/woocommerce-services/pull/2756/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fR58) and render the `<FeatureAnnouncement>` component inside the PHP admin notice banner. This way, we also have more control over the modal without worrying about breaking other pages. The downside is loading the `isEligable` props which is from the redux store. It turned out it wasn't too bad because we only needed the [shipping label store](https://github.com/Automattic/woocommerce-services/pull/2756/files#diff-cafa06d18d10a101b583ffbe68a81c29f97cacbf981e3e79167354c18786efa2R21). As a result, I chose to stick with `client/wcshipping-migration-admin-notice.js` instead of following the `main.js` pattern.

### Clicking "Confirm update" to trigger React `<FeatureAnnouncement>` button action
Instead of creating our own React component for the admin notice. I continued to use the WP's one from the PHP's side. To trigger the migration inside the PHP's admin notice, I added an event listener to render the `<FeatureAnnouncement>` component with the store context. Then, the script continues and will click the "Update now" button to trigger the migration. 

### Related issue(s)
NA

### Steps to test
1. Go to `woocommerce-services/classes/class-wc-connect-service-settings-store.php` and replace the following `is_eligible_for_migration` function. This function decides if the modal should popup or not
```
			//TODO: Return $migration_state check && connect-server flag
			return WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED !== $migration_state && true;
		}
```
2. Make sure you have "WooCommerce Shipping" and "WooCommerce Tax" in your plugins folder. We can't test the "install" part because they aren't part of dotorg yet.
3. Go to the order listing page.
4. Press F12 and check the Network tab, click the "Confirm update" button in the admin notice banner ![image]
(https://github.com/Automattic/woocommerce-services/assets/572862/0a5880a9-a046-4075-b14d-7171528628cb)
5. Confirm the 3 requests did go through and confirm the modal popped up and the button in the modal reads "Updating...".
image
6. Confirm you are being redirected to the plugins page with WCS&T deactivated, WCShipping and WCTax activated.

### Checklist

- [ ] unit tests


harriswong:
### Jun 21 update
#### Done
- I added a new entrypoint to compile the feature announcement codes into its own js and css.
- Modal loads on order list page

#### TODO
- "Update now" button in the modal is broken. It's saying 404 not found on the order listing page. I think WCS&T isn't loaded on this page, so the controller isn't registered. This is gonna be a bit tricky to load the controller and all its dependency :/ 
- The modal shouldn't popup immediately on the order list page, it should only pop up after clicking the admin notice's "Confirm update".
- Clicking "Confirm update" should automatically click "Update now"

client/wcshipping-migration-admin-notice.js
```diff
@@ -0,0 +1,43 @@
...
+const args = container.dataset.args && JSON.parse( container.dataset.args ) || {};
+setNonce( args.nonce );
+setBaseURL( args.baseURL );
+const ShippingLabelStore = ShippingLabel(args);
+const store = createStore(ShippingLabelStore.getReducer(), ShippingLabelStore.getInitialState());
+
+const wcstWCShippingMigrationNoticeButton = document.getElementById('wcst-wcshipping-migration-notice__click');
+
+// Clicking "Confirm update" will start the migration. This is the same as popping up the modal and clicking the "Update" button there.
+["click", "keydown"].forEach(eventName =>
```
harriswong: ARIA, needs to bind both `click` and `keydown` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#sect1

client/wcshipping-migration-admin-notice.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+import '../assets/stylesheets/migration_to_wcshipping_admin_notice.scss';
+import FeatureAnnouncement from 'components/migration/feature-announcement';
+import ShippingLabel from 'wcs-client/apps/shipping-label';
+import { setNonce, setBaseURL } from 'wcs-client/api/request';
+
+const container = document.getElementById('wcst_wcshipping_migration_admin_notice_feature_announcement');
+const args = container.dataset.args && JSON.parse( container.dataset.args ) || {};
+setNonce( args.nonce );
+setBaseURL( args.baseURL );
```
harriswong, line 20: `getBaseURL` is used inside the `<FeatureAnnouncement>` component. It needs to be set here, similar to how we do `setBaseURL(...)` in `main.js`.

woocommerce-services.php
```diff
@@ -1898,22 +1900,52 @@ public function display_woo_shipping_and_woo_tax_are_active_notice() {
...
 			);
 		}
+
+		public function register_wcshipping_migration_modal() {
+			$plugin_version = self::get_wcs_version();
+			wp_register_style( 'wcst_wcshipping_migration_admin_notice', $this->wc_connect_base_url . 'woocommerce-services-wcshipping-migration-admin-notice-' . $plugin_version . '.css', array(), null ); // phpcs:ignore WordPress.WP.EnqueuedResourceParameters.MissingVersion
+			wp_register_script( 'wcst_wcshipping_migration_admin_notice', $this->wc_connect_base_url . 'woocommerce-services-wcshipping-migration-admin-notice-' . $plugin_version . '.js', array(), null );
+			wp_enqueue_script( 'wcst_wcshipping_migration_admin_notice' );
+			wp_enqueue_style( 'wcst_wcshipping_migration_admin_notice' );
+			wp_enqueue_script( 'wc_connect_admin' );
```
harriswong, line 1985: > wp_enqueue_script( 'wc_connect_admin' );

I enqueued this whole script because `<FeatureAnnouncement>` needs the global `global.wcsPluginData.adminPluginPath`, which is a variable in the `wc_connect_admin` script. Loading this whole script is easier.

Alternatively, I could have extracted the `global.wcsPluginData.adminPluginPath`, and loads it in the [$encoded_arguments](https://github.com/Automattic/woocommerce-services/pull/2756/files#diff-4cb5834c68b324f4fbecc437b25f62021e55ecc28db9adb93053eca0391f5cc5R1912) up there. If I do this, I would have to edit `<FeatureAnnouncement>` and change:
 `window.location = global.wcsPluginData.adminPluginPath;`
to something like:
`window.location = global.wcsPluginData.adminPluginPath || someState.adminPluginPath;`.
This doesn't seem worth it because the condition is more complicated (prone to errors). So I chose to load the whole `wc_connect_admin` script into the order list page instead.

client/components/migration/feature-announcement.jsx
```diff
@@ -7,7 +7,7 @@ import { useState } from '@wordpress/element';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { localize } from 'i18n-calypso';
-import { getNonce, getBaseURL } from 'api/request'; // client/api/request.js
+import { getNonce, getBaseURL } from 'wcs-client/api/request';
```
waclawjacek: Nice.

woocommerce-services.php
waclawjacek: Thanks for the explanation. This also seems OK performance-wise because I think `/client/main.js` won't load the specific functionality (like the label purchase modal) unless certain conditions are met, correct? 🤔

client/wcshipping-migration-admin-notice.js
```diff
@@ -0,0 +1,43 @@
...
+		// Pop open feature announcement modal.
+		ReactDOM.render(
+			<Provider store={store}>
+				<FeatureAnnouncement />
+			</Provider>,
+			container
+		);
+
+		// Click the update button in the modal to start the migration.
+		const update_button = document.getElementsByClassName('migration__announcement-update-button');
```
waclawjacek: Maybe it'd be more precise if we used an ID instead?

client/wcshipping-migration-admin-notice.js
```diff
@@ -0,0 +1,43 @@
...
+import FeatureAnnouncement from 'components/migration/feature-announcement';
+import ShippingLabel from 'wcs-client/apps/shipping-label';
+import { setNonce, setBaseURL } from 'wcs-client/api/request';
+
+const container = document.getElementById('wcst_wcshipping_migration_admin_notice_feature_announcement');
+const args = container.dataset.args && JSON.parse( container.dataset.args ) || {};
+setNonce( args.nonce );
+setBaseURL( args.baseURL );
+const ShippingLabelStore = ShippingLabel(args);
+const store = createStore(ShippingLabelStore.getReducer(), ShippingLabelStore.getInitialState());
```
waclawjacek, line 22: What are we actually using this for? Do we need to use the reducer?

waclawjacek (approved):
LGTM! :shipit: Posted some questions but things work and look good. Please fix the code style before merging.

harriswong:
> Please fix the code style before merging.

👋 Hey @waclawjacek , can you help me pinpoint where these issues are please? I am solely relying on the phpcs precommit hooks. Do you mean the code conflict?

client/wcshipping-migration-admin-notice.js
harriswong: Thanks! I thought about using `ID` but I couldn't find `ID` in https://developer.wordpress.org/block-editor/reference-guides/components/button/#props. 
But! Now that you brought it up again, I wonder if `ID` is "native" so it should just work? Let me give this a try.

client/wcshipping-migration-admin-notice.js
harriswong: > What are we actually using this for?

The `store` is needed because `<FeatureAnnouncement>` connects to the "label store" for the [isEligableToMigrate ](https://github.com/Automattic/woocommerce-services/blob/1bce661e84bf2efbc27e9bfbdd4cb2896f1f5510/client/components/migration/feature-announcement.jsx#L197-L199)selector.
In order for this selector to work without modification, the store needs to have the `extensions > woocommerce > woocommerceServices > sites >...` path 
image
I could have created a different store by loading a different set of initial states similar to [this](https://github.com/Automattic/woocommerce-services/blob/1bce661e84bf2efbc27e9bfbdd4cb2896f1f5510/client/apps/shipping-label/index.js#L118-L137). But I chose to load the existing store instead.

> Do we need to use the reducer?

No. The `<FeatureAnnouncement>` only consumes the data from the redux store. It doesn't dispatch any action to update the store's state. Reducer isn't needed specifically for this modal. However! [createStore()](https://redux.js.org/api/createstore) requires the first arguments to be a reducer. So I use the existing ShippingLabelStore's one. I supposed we could put an empty reducer if we don't want to use the existing one? Something like:
```js
function shippingLabelReducer(state = [], action) {
   return state;
}
```

Let me know what you think!

woocommerce-services.php
harriswong: Thanks @waclawjacek! Actually, you just made me realize that `client/main.js` is the registered `wc_connect_admin` script which turns into `'woocommerce-services-*.js`. 
If I am already loading `wc_connect_admin`, maybe we didn't need its own entry point. I have 3 options.

#### Option 1.
Leave it as is because this is the migration modal only. Loading `'woocommerce-services-*.js` for `adminPluginPath` is heavy but it is also cached, so it should be faster for the individual order page. The `client/wcshipping-migration-admin-notice.js` is separated as if it was chunk-split and won't be loaded inside the individual order page, which is good.

#### Option 2.
Remove `client/wcshipping-migration-admin-notice.js` and move that into `main.js`. Use `enqueue_wc_connect_script()` pattern to create the `wcc-` container inside the admin notice in order to load this `<FeatureAnnouncement>` modal.

#### Option 3.
Don't load `wp_enqueue_script( 'wc_connect_admin' );`. Pull `adminPluginPath` and find a way to load this in the order list page, and then inject it to the `<FeatureAnnouncement>` modal.

I am opting for **Option 1** because there is a slight performance hit on the order list page, but does not impact overall total performance since `woocommerce-service-*.js` is cached and will be used in the individual order page, it will be loaded regardless. On size, it will actually be better because these codes won't be loaded on the individual order page.

Let me know your thoughts!

client/wcshipping-migration-admin-notice.js
harriswong: ID works, updated in 402998bb

harriswong:
👋 Hi @waclawjacek, I fixed the conflict and addressed the concerns! Please let me know what you think.

harriswong:
👋 Hey @waclawjacek, please hold off on the review. I am anticipating a big merge conflict coming in https://github.com/Automattic/woocommerce-services/pull/2758. I will fix those conflicts after it's merged and will ping you again.

harriswong:
Hey @waclawjacek, I fixed the merge conflict and the functionality of this PR should work. Please take another look.

~Note, the dismissal button from https://github.com/Automattic/woocommerce-services/pull/2758 doesn't seem to work anymore after the merge. I ran out of energy tonight to dig into it. I can create a separate issue/PR to address that regression.~  
**edited: This is fixed. Please check the comment below**

harriswong:
> Note, the dismissal button from https://github.com/Automattic/woocommerce-services/pull/2758 doesn't seem to work anymore after the merge. I ran out of energy tonight to dig into it. I can create a separate issue/PR to address that regression.

Hey @waclawjacek. I fixed(4df565cb15248501999fac0890ce72d1742c3ab2) the dismissible button issue after the last merge conflict. Here is a summary of what I did:
- Removed the dependencies `jQuery` and `wp-utils` in `wcst_wcshipping_migration_admin_notice`. We will use plain javascript instead.
- Replaced the javascript cookie logic with a plain javascript cookie logic. The cookie sets a `1` and will expire in 3 days.
- The WP admin notice's "is-dismissible" button is added after the DOM is ready. I tried a few things but the dismiss button wasn't found (not loaded in DOM). So I cloned the markup of the button and added it directly to our admin notice instead. Then, I bind the 3-days-cookie event to it. This way, the button is loaded into the DOM.
- Updated the scss on the banner so that it looks the same as figma.

waclawjacek:
Thanks for the changes! I won't be able to review them as I'm headed AFK. @Automattic/escargot, could someone please give the PR a look?

samnajian:
> Thanks for the changes! I won't be able to review them as I'm headed AFK. @Automattic/escargot, could someone please give the PR a look?

I can review it.

client/wcshipping-migration-admin-notice.js
```diff
@@ -0,0 +1,61 @@
...
+
+const wcstWCShippingMigrationNoticeButton = document.getElementById('wcst-wcshipping-migration-notice__click');
+const wcstMigrationNoticeDimissButton = document.querySelector('.wcst-wcshipping-migration-notice button.notice-dismiss');
+
+// Helper function to set simple name=>value cookie in days.
+const setCookie = (name, value, days) => {
```
samnajian: [Optional] You could use `window.wpCookies`  to set cookies.

client/wcshipping-migration-admin-notice.js
```diff
@@ -0,0 +1,61 @@
...
+		);
+
+		// Click the update button in the modal to start the migration.
+		const update_button = document.getElementById('migration__announcement-update-button');
+		update_button.click();
+	});
+
+	// Dismiss it for 3 days, then remove it from view.
+	wcstMigrationNoticeDimissButton.addEventListener(eventName, () => {
+		setCookie('wcst-wcshipping-migration-dismissed', 1, 3);
```
samnajian: [ Optional ] what do you think about assigning `3` to a variable with a transparent name with the following benefits:
1) It won't be lost in the code, but explicitly defined in this file in a higher level.
2) A proper name explains what it is.

client/wcshipping-migration-admin-notice.js
```diff
@@ -0,0 +1,61 @@
...
+		const wcstMigrationAdminNotice = document.querySelector('.wcst-wcshipping-migration-notice');
+		wcstMigrationAdminNotice.remove();
+	});
+});
```
samnajian: Missing empty line at the end of file.

woocommerce-services.php
samnajian: I'd rather go with **Option 1** as it's a migration specific change.

samnajian (approved):
LGTM!

- Please see the added comments and feel free to apply them before merging.
- There is a merge conflict which needs care before merging.
- It functions how the issue requires you to do, but it's a bit weird that the modal opens and then you're redirected shortly to the plugins page, it was confusing to me, it can be confusing to others as well, I'd call @MatthiasReinholz to have his input, but this can probably be another issue and this PR doesn't have to wait as it addresses what it's supposed to do.

harriswong:
> It functions how the issue requires you to do, but it's a bit weird that the modal opens and then you're redirected shortly to the plugins page, it was confusing to me, it can be confusing to others as well, I'd call @MatthiasReinholz to have his input, but this can probably be another issue and this PR doesn't have to wait as it addresses what it's supposed to do.

100% agree. This is also discussed via the design request as well: p1718745005310189-slack-C06FP5XT1S5

client/wcshipping-migration-admin-notice.js
harriswong: Fixed in 42b2e131

harriswong:
~After resolving the merge conflicts, the API calls inside `migration-runner.js` is returning `undefined` for `nonce`. I am looking into it.~

**edited: Fixed in 2fbd631d**

client/wcshipping-migration-admin-notice.js
harriswong: Updated to a constant in bd141b6c

client/wcshipping-migration-admin-notice.js
harriswong: Thanks. I like this suggestion. Replaced it with the wp's one bd141b6c

harriswong:
I tested the following:
- In the order listing page, click "Confirm update", I can confirm that it's installing and activating the plugins. Then it deactivates WCS&T and redirects the user to the plugin page.
- I confirmed that clicking "dismiss" sets the cookie for 3 days
image
- In the order page itself, clicking "Update now" also installs, and activates the plugins. Then, it deactivates WCS&T and redirects to the plugins page. 👍 

@samnajian, please take another look and please feel free to merge for me if it looks good.

samnajian (approved):
LGTM! Thanks for updates.

PR #2760, opened 2024-06-28
Add tracking data for the migration process

Ferdev:
-------

## Description
This PR adds tracking data for all the different states the migration process can go through.

The tracking code has been added in the controller that takes care of updating the migration status in the database whenever it changes during the migration process.

### Related issue(s)

N/A

1. After checking out this branch, replace this method in `class-wc-connect-service-settings-store.php`:

``` php

	$migration_pending = WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED !== $migration_state;
	$migration_enabled = $this->service_schemas_store->is_wcship_wctax_migration_enabled();

	return $migration_pending && $migration_enabled;
}
```

This will remove the current migration status from the database, so we can be sure the migration popup will display again.

2. Create a new order and try to create a new shipping label.

3. When the create shipping label button is pressed, the migration popup should appear. Click it.

4. The migration flow will start, and when it ends the tracks API should be called. If logging is enabled in your WooCommerce Shipping & Tax setup, you should see an entry in the log like this one:

``` text
Tracked the following event: woocommerceconnect_migration_flag_updated, referer: http://localhost:8889/wp-admin/post.php?post=34&action=edit
```

This is logged right before the data is sent to the tracks server.

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-rest-connect-migration-flag-controller.php
+++ b/classes/class-wc-rest-connect-migration-flag-controller.php
@@ -53,7 +53,13 @@ class WC_REST_Connect_Migration_Flag_Controller extends WC_REST_Connect_Base_Con
 		$result = WC_Connect_Options::update_option( 'wcshipping_migration_state', $migration_state );
 
 		if ( $result ) {
-			$this->tracks->record_user_event( 'migration_flag_updated', array( 'migration_state' => $migration_state ) );
+			$this->tracks->record_user_event(
+				'migration_flag_state_update',
+				array(
+					'migration_state' => $migration_state,
+					'updated'         => $result,
+				)
+			);
 			return new WP_REST_Response( array( 'result' => 'Migration flag updated successfully.' ), 200 );
 		}
 
--- a/classes/class-wc-rest-connect-migration-flag-controller.php
+++ b/classes/class-wc-rest-connect-migration-flag-controller.php
@@ -11,6 +11,31 @@ if ( class_exists( 'WC_REST_Connect_Migration_Flag_Controller' ) ) {
 class WC_REST_Connect_Migration_Flag_Controller extends WC_REST_Connect_Base_Controller {
 	protected $rest_base = 'connect/migration-flag';
 
+	/**
+	 * Tracks instance.
+	 *
+	 * @var WC_Connect_Tracks
+	 */
+	protected $tracks;
+
+	public function __construct( $api_client, $settings_store, $logger, $tracks ) {
+		parent::__construct( $api_client, $settings_store, $logger );
+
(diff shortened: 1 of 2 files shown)


classes/class-wc-rest-connect-migration-flag-controller.php
```diff
@@ -11,6 +11,31 @@
...
+	 *
+	 * @var WC_Connect_Tracks
+	 */
+	protected $tracks;
+
+	public function __construct( $api_client, $settings_store, $logger, $tracks ) {
+		parent::__construct( $api_client, $settings_store, $logger );
+
+		error_log( 'WC_REST_Connect_Migration_Flag_Controller::__construct' );
+		error_log( print_r( $tracks, true ) );
```
samnajian: Are these error loggings expected to remain here?

classes/class-wc-rest-connect-migration-flag-controller.php
```diff
@@ -31,6 +56,7 @@ public function post( $request ) {
 		$result = WC_Connect_Options::update_option( 'wcshipping_migration_state', $migration_state );
 
 		if ( $result ) {
+			$this->tracks->record_user_event( 'migration_flag_updated', array( 'migration_state' => $migration_state ) );
```
samnajian: Do you think it's fine to only track it when the migration state has updated? what's against always tracking when this endpoint is called? Wouldn't that give us more insight on what's happening on the merchant sites when it gets to migration?

samnajian (commented):
It works as described, just left some questions.

classes/class-wc-rest-connect-migration-flag-controller.php
Ferdev: we can also track when errors are triggered, but I was thinking about this metrics more like product metrics, instead of debugging info. The errors defined in that class are not supposed to happen: we should use a different tool (ie. sentry? or whatever we use to catch errors) and fix them, but I don't think we should reflect errors in the product metrics. Do you think we should track them too? I'm ok with implementing that, anyway, if you think it would be valuable :+1:

samnajian: I agree that other tools could be a better fit for tracking none-updating requests, however we don't use them at this point so and tracking too much is always better than tracking too less, how about adding the update value ($result) to the tracked data, something like the following:

```
$this->tracks->record_user_event( 'migration_flag_state_update', array( 'migration_state' => $migration_state, 'updated => $result ) );
```
I'd leave it to you to decide if you want to change it though.

Ferdev: that's a great idea, I'll add it to the tracking data :+1:

PR #2761, opened 2024-07-05
Conditionally load shipping functionality based on if WC Shipping is active

### LINKED ISSUE #2766: Do not load shipping related classes when WC Shipping activated
We need to ensure that when it is detected that `woocommerce-shipping` plugin is active no shipping related classes are loaded, but tax related functionality should stay in tact and functioning as per normal.

kallehauge:
## Description

Just playing around in context of paJDVv-3NH-p2

Fixes: [2766](https://github.com/Automattic/woocommerce-services/issues/2766)

We need to:

* [x] Update name of status page to (conditionally?) say "WooCommerce Tax" or "WooCommerce Tax & Shipping".
  * TL;DR: We shouldn't: https://github.com/Automattic/woocommerce-services/pull/2761/commits/d6fbe3e5141f5c28f6e2b984326ab0e0172962d5 
* [x] The status page should also, conditionally, display shipping settings and content.
  *  https://github.com/Automattic/woocommerce-services/pull/2776
* [x] We need to rename the tab from "connect" in WC Shipping since that's conflicting
  *  https://github.com/woocommerce/woocommerce-shipping/pull/578
* [ ] We need to update our connect / ToS banners.
* [x] Verify that mobile will continue to work
    * The REST endpoints continue to be functional.
* [x] Since we expect WCS&T to be around for a while: maybe add a banner / notification saying WCS&T doesn't show WC Shipping labels if the merchant disabled the plugin afterwards (and that we do not support backwards migration)
    * The migration project already has enough "WC Shipping is activated" type communication. 

What's working without issue:
* WPCOM connection!
* I'm able to fetch taxes without UI or debug.log warnings/errors
* I'm able to purchase shipping labels when using WC Shipping in parallel

#### WCS&T
The first test is to ensure that WCS&T is functioning as it should on its own, things to test:
- Establishing a connection
- All label functionality
- Automated taxes
- Settings
- Debug info

#### WC Shipping combined with WCS&T
The second test would be to see how well WC Shipping works with WCS&T installed already. There are a couple of starting scenarios to test ie: establishing a connection with WCS&T and then activating WC Shipping and ensuring the connection is kept established, this is the most likely scenario that would be used in the wild, but we should also test for instances where WC Shipping is installed first and then you install WCS&T, does it allow for establishing a new connection for tax purposes in WCS&T.

Things to test include:
- Connections, WC Shipping and WCS&T have separate connections/tos agreements.
- Automated taxes, does it continue functioning
- Label functionality, does it utilise the label purchase flow from WC Shipping instead of WCS&T.
- WC Shipping settings, does the settings of WC Shipping show and work instead of that of WCS&T.
- Debugging info, are there no conflicts between the debug info from WCS&T and WC Shipping

Please test around this these instruction in any way you can think.

### Checklist

- [ ] unit tests

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -258,6 +258,9 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 
 		/**
 		 * Deletes WC Admin notices.
+		 *
+		 * @todo We'd need to make sure the DHL Live Rates class reachable during uninstall.
+		 *       We should be able to require everything in the 'pre_uninstall_plugin' action, so it shouldn't be an issue.
 		 */
 		public static function delete_notices() {
 			if ( self::can_add_wc_admin_notice() ) {
@@ -340,18 +343,6 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			);
 
 			add_action( 'plugins_loaded', array( $this, 'on_plugins_loaded' ) );
-
-			if ( $this->are_woo_shipping_and_woo_tax_active() ) {
-				/**
-				 * Used to let Woo Shipping and Woo Tax know WCS&T will handle the plugins' coexistence.
-				 *
-				 * WCS&T does it by not registering its functionality and displaying an appropriate notice
-				 * in WP admin.
-				 */
-				add_filter( 'wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax', '__return_true' );
-				return;
-			}
-
 			add_action( 'plugins_loaded', array( $this, 'jetpack_on_plugins_loaded' ), 1 );
 		}
 
@@ -582,7 +573,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 				'connection',
 				array(
 					'slug' => WC_Connect_Jetpack::JETPACK_PLUGIN_SLUG,
-					'name' => __( 'WooCommerce Shipping & Tax', 'woocommerce-services' ),
+					'name' => __( 'WooCommerce Tax', 'woocommerce-services' ),
 				)
 			);
 		}
@@ -598,6 +589,8 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			 * plugins are activated at the same time.
 			 *
 			 * @since {{next-release}}
+			 *
+			 * @todo Replace this notice with something that says "This will purely be a Tax plugin by <date>".
 			 *
 			 * @param bool $status The value will determine if we should initiate the plugins logic or not.
 			 */
@@ -703,6 +696,11 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		}
 
 		public function init_core_wizard_shipping_config() {
+			// @todo Verify what happens if WCS&T was installed without this, and WC Shipping is then deactivated.
+			if ( $this->is_wc_shipping_activated() ) {
+				return;
+			}
+
 			$store_currency = get_woocommerce_currency();
 
 			if ( 'USD' === $store_currency ) {
@@ -827,10 +825,6 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			require_once __DIR__ . '/classes/class-wc-connect-debug-tools.php';
 			new WC_Connect_Debug_Tools( $this->api_client );
 
-			require_once __DIR__ . '/classes/class-wc-connect-settings-pages.php';
-			$settings_pages = new WC_Connect_Settings_Pages( $this->api_client, $this->get_service_schemas_store() );
-			$this->set_settings_pages( $settings_pages );
-
 			$schema   = $this->get_service_schemas_store();
 			$settings = $this->get_service_settings_store();
 			$logger   = $this->get_logger();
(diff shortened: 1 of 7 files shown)


Ferdev (commented):
I started reviewing this and apparently everything's working well. One thing I've found confusing is the status tab. It appears to merge both WooCommerce Shipping and WooCommerce Shipping & Tax status pages. While the WooCommerce Shipping & Tax doesn't show any shipping related info, I think it would be clearer for the user if WooCommerce Shipping & Tax used its own tab (the status tab is called "WooCommerce Shipping")

Ferdev (approved):
The code looks good. There are some @todo's in the code, I'm not sure if we want to implement them now or leave them for future upgrades.

I haven't been able to test the live rates, specially with DHL as mentioned in one of the @todo's. @kallehauge maybe you can give me some directions?

I left a comment earlier about splitting the status tabs, as right now are combined (WCS&T and Woo Shipping) and look a bit confusing.

But everything works well. I tried purchasing labels witch WooCommerce Shipping & Tax, migrating them after activating WooCommerce Shipping, checking them in the new plugin and they looked fine. I then purchased new labels using WooCommerce Shipping and disabled WooCommerce Shipping to check what WooCommerce Shipping & Tax does, and for that order, it just shows the create shipping label button.

Maybe would be worth it to get another check from a person with more WordPress knowledge, like you @samnajian

woocommerce-services.php
```diff
@@ -258,6 +258,9 @@ public static function plugin_uninstall() {
 
 		/**
 		 * Deletes WC Admin notices.
+		 *
+		 * @todo We'd need to make sure the DHL Live Rates class reachable during uninstall.
+		 *       We should be able to require everything in the 'pre_uninstall_plugin' action, so it shouldn't be an issue.
```
Ferdev: I'm not sure how to test this. Can you give me some testing steps @kallehauge , please? Thanks!

Ferdev (dismissed):
I just checked that if you have live rates enabled, you can see their status in the Status page of WooCommerce Shipping & Tax. Is the only mention of shipping features I've seen in that screen for WooCommerce Shipping & Tax. I think we should hide that info if WooCommerce Shipping is enabled alongside WooCommerce Shipping & Tax

Ferdev (commented):
I just was able to test live rates and the behavior is correct: when WooCommerce Shipping is disabled, you can use live rates without issues. Once WooCommerce Shipping is enabled, the live rates shipping methods from WooCommerce Shipping & Tax dissapear from the shipping zones where they were configured, and you can't use them anymore. If you disable WooCommerce Shipping, they're back.

woocommerce-services.php
```diff
@@ -703,6 +699,11 @@ protected function add_method_to_shipping_zone( $zone_id, $method_id ) {
 		}
 
 		public function init_core_wizard_shipping_config() {
+			// @todo Verify what happens if WCS&T was installed without this, and WC Shipping is then deactivated.
```
samnajian: We already have a notice in place which nags in case WC Shipping is active and WCS&T is active as well.
We can adjust to tell them to update the WCS&T (to the version that can co-exist with WC Shipping) or remove it to be able to use WC Shipping.

samnajian:
> I then purchased new labels using WooCommerce Shipping and disabled WooCommerce Shipping to check what WooCommerce Shipping & Tax does, and for that order, it just shows the create shipping label button.

That's how it should work as we don't have data migration from WC Shipping to WCS&T.

woocommerce-services.php
```diff
@@ -1778,16 +1797,24 @@ public function render_schema_notices() {
 			}
 		}
 
+		/**
+		 * Check if WooCommerce Shipping has been activated.
+		 *
+		 * @return bool
+		 */
+		public function is_wc_shipping_activated() {
+			return in_array( 'woocommerce-shipping/woocommerce-shipping.php', get_option( 'active_plugins' ) );
```
samnajian, line 1818: Using `wp_get_active_and_valid_plugins` would be a better option as it checks for `network_plugins` and more.

samnajian (approved):
LGTM, left some comments and since this is going to be treated as the base branch, I'm approving, well done!

harriswong (commented):
Tested
WCS&T activated, Woo Shipping deactivated
- Printing USPS labels
- Refunding USPS label

WCS&T and Woo Shipping activated
- Tested opening old labels from old orders. I printed multiple labels for a single order, so now they are showing up as "split shipment". They look a bit strange, I created a separate issue https://github.com/woocommerce/woocommerce-shipping/issues/576. 
- Checkout cart kept spinning with both plugins on. It's fine with just Woo Shipping:
image
- Tested printing a USPS ground advantage label with signature required. Requested refunds. 👍 


When both plugins are enabled, I noticed these errors on the UI. I am using PHP 8.2.7
```
Deprecated: Function is_staging_site is deprecated since version 3.3.0! Use in_safe_mode instead. in /var/www/wordpress/wp-includes/functions.php on line 6085

Deprecated: Function is_staging_site is deprecated since version 3.3.0! Use in_safe_mode instead. in /var/www/wordpress/wp-includes/functions.php on line 6085
```

❓ Should this branch merge into `wcs-migration` instead of `trunk`?

woocommerce-services.php
```diff
@@ -827,31 +828,68 @@ public function load_admin_dependencies() {
...
+
+			$tracks = $this->get_tracks();
+			$tracks->init();
+
+			$this->taxjar->init();
+			$this->paypal_ec->init();
+
+			// Primary condition to initiate shipping.
+			if ( ! $this->is_wc_shipping_activated() ) {
+				$this->init_shipping();
```
harriswong, line 887: ❔ – Curiosity

It is hard to determine what code should go into this function. But the code in this `attach_hooks()` function makes sense, all shipping-related things are now inside `init_shipping()` 👍

woocommerce-services.php
```diff
@@ -1778,16 +1797,24 @@ public function render_schema_notices() {
...
+		public function is_wc_shipping_activated() {
+			return in_array( 'woocommerce-shipping/woocommerce-shipping.php', get_option( 'active_plugins' ) );
+		}
+
 		/**
 		 * Returns if both Woo Shipping and Woo Tax are active.
 		 *
 		 * @return bool
 		 */
 		public function are_woo_shipping_and_woo_tax_active() {
```
harriswong, line 1826: Do we still need this function?

woocommerce-services.php
kallehauge: @kloon was it you and I that briefly had a discussion about the correctness of checking / not checking network plugins when we first started doing this in WCS&T?

woocommerce-services.php
kallehauge: Yup! You can read a bit more about the case we need to support here: https://github.com/woocommerce/woocommerce-shipping/pull/586

TL;DR: The only WC Tax support we have right now in WCS&T is when both WC Tax and WC Shipping is activated, so we should keep this around to maintain some sort of backwards compatibility for WC Tax 😄

woocommerce-services.php
kallehauge: No need @Ferdev ! My concern was: if someone uninstalled WCS&T, then if we had not initiated shipping logic, then this shouldn't cause a fatal error because the class doesn't exist.

TL;DR: It's not an issue since we require the class right before we call it, and the method is static and doesn't depend on anything outside core Woo code, so we're good here! 👏

PR #2762, opened 2024-07-05
Clicking "Maybe later" will dismiss feature announcement for 3 days

harriswong:
## Description
Clicking "Maybe later" will dismiss the modal for 3 days. 
image

### Technical decisions
If the user clicks "x" or "Maybe later", I consider this as "not eligible for migration". Thus, I added a cookie check in `is_eligible_for_migration()` as well. 

### Related issue(s)
N/A

### Steps to test
1. Pull this branch, go to `woocommerce-services/classes/class-wc-connect-service-settings-store.php`, change `is_eligible_for_migration()` to
```
...
		}
```
The idea is to delete the `wcshipping_migration_state` option every time so we can re-test infinitely. 
2. Go to an order page and click "Create new label"
3. Observe the feature announcement modal
image
4. Click "Maybe later", you should see the cookie `wcst-wcshipping-migration-dismissed` being set.
image
5. Close the shipping modal, click "Create new label" again, and confirm no feature announcement migration modal.
6. Refresh the page (ctrl-r), click "Create new label", and confirm no feature announcement migration modal.
7. Mimic the 3-days cookie expiration by deleting the cookie `wcst-wcshipping-migration-dismissed`, refresh the page, and confirm feature announcemnt migration modal pop up again.

### Checklist

- [ ] unit tests


client/components/migration/feature-announcement.jsx
```diff
@@ -18,17 +18,26 @@ import {
 	wcshippingMigrationState,
 } from 'woocommerce/woocommerce-services/state/shipping-label/selectors';
 import { installAndActivatePlugins } from './migration-runner';
+import { TIME_TO_REMMEMBER_DISMISSAL_SECONDS } from './constants';
 
 const FeatureAnnouncement = ( { translate, isEligable, previousMigrationState } ) => {
 	const [isOpen, setIsOpen] = useState(isEligable);
 	const [isUpdating, setIsUpdating] = useState(false);
 
+	useEffect(() => {
```
harriswong: If we click "Maybe later" to close the modal, close the label printing modal, then re-open label printing modal, the feature announcement will show up again. This is because the `props` `isEligable` hasn't refresh from PHP yet.

This side effect checks against the cookie when it mounts so that it doesn't reopen the feature announcement modal.

client/components/migration/feature-announcement.jsx
```diff
@@ -18,17 +18,26 @@ import {
...
+		const isMigrationDismissed = window.wpCookies.get('wcst-wcshipping-migration-dismissed') && parseInt(window.wpCookies.get('wcst-wcshipping-migration-dismissed'));
+		if ( isMigrationDismissed ) {
+			setIsOpen(false);
+		}
+	}, []);
```
kallehauge: Minor code standard suggestion 😄

```suggestion
	useEffect( () => {
		const isMigrationDismissed = window.wpCookies.get( 'wcst-wcshipping-migration-dismissed' ) && parseInt( window.wpCookies.get( 'wcst-wcshipping-migration-dismissed' ) );
		if ( isMigrationDismissed ) {
			setIsOpen( false );
		}
	}, [] );
```

client/components/migration/feature-announcement.jsx
```diff
@@ -18,17 +18,26 @@ import {
...
+
 	const closeModal = () => {
 		setIsOpen(false);
 	};
 
 	const snooze = () => {
-		// Todo: implement maybe later
+		window.wpCookies.set('wcst-wcshipping-migration-dismissed', 1, TIME_TO_REMMEMBER_DISMISSAL_SECONDS)
+		setIsOpen(false);
```
kallehauge: Minor code standard suggestion 😄

```suggestion
		window.wpCookies.set( 'wcst-wcshipping-migration-dismissed', 1, TIME_TO_REMMEMBER_DISMISSAL_SECONDS );
		setIsOpen( false );
```

kallehauge (approved):
Code LGTM and works as described 👏 

I have one observation that is strictly a slightly different approach of cookies vs localStorage, not a blocking recommendation. Since none of this is cached (read: typically front-end page views etc) then we don't really have any concerns to work around, so it's not worth spending time updating this, but I could see an approach where:

1. The server gives React its migration state
2. React will then: show the modal if the server says they haven't migrated and we do not have a localStorage item that functions the same way as the cookie.

Effectively it would work the same, so it's 100% not blocking, but I like to share that I often consider Cookies for the server and localStorage and sessionStorage client specific like: "should we delay this one?" 😛

client/components/migration/feature-announcement.jsx
harriswong: Thanks for catching these! I rely on the githooks and for some reason it's not catching these things 😞 .

client/components/migration/feature-announcement.jsx
harriswong: Appreciate the suggestion, this allows me to commit directly via GitHub. Thank you!

harriswong:
I didn't consider using localStorage, thanks for the suggestion! Coincidentally, we do use the cookie on the server side https://github.com/Automattic/woocommerce-services/blob/8333367c31bc1edfaf026d6d4b28656055d5da8c/woocommerce-services.php#L1927. Sounds like this works out!

I will proceed to merge.

PR #2763, opened 2024-07-08
Save transients after a migration has been completed

Ferdev:
-------

## Description
When the WooCommerce Shipping & Tax migration to WooCommerce Shipping and WooCommerce Tax is completed, we need to display migration status messages for each entry of these plugins in the plugins list screen. One issue is the migration status is stored in the WooCommerce Shipping & Tax options object, and the class helper that access that options object is not available for WooCommerce Shipping and WooCommerce Tax as WooCommerce Shipping & Tax is disabled. This PR solves that issue by creating a couple of transients that tell WooCommerce Shipping and WooCommerce Tax that the migration has been completed successfully. The transients last one day until they're deleted.

### Related issue(s)

N/A

1. Checkout this branch and make sure you're eligible to run the migration:

	- Reset the `wcshipping_migration_state` option with `WC_Connect_Options::delete_option( 'wcshipping_migration_state' );`
	- Use the WooCommerce Connect Server staging endpoint: `https://api-staging.woocommerce.com`

2. Go to the orders list, and click the migration button at the top of the screen. Wait for the migration to finish.

shot-2024-07-06-13-51-19

3. Install a transients manager plugin and check if the transients `wcshipping_migration_completed` and `wctax_migration_completed` have been created.

### Checklist

- [ ] unit tests

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -266,7 +266,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			require_once __DIR__ . '/classes/class-wc-connect-wcst-to-wcshipping-migration-state-enum.php';
 
 			$migration_state = get_option( 'wcshipping_migration_state' );
-			if ( $migration_state !== WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED ) {
+			if ( ! $migration_state || intval( $migration_state ) !== WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED ) {
 				update_option( 'wcshipping_migration_state', WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED );
 			}
 		}
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -265,11 +265,9 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			 */
 			require_once __DIR__ . '/classes/class-wc-connect-wcst-to-wcshipping-migration-state-enum.php';
 
-			$migration_state = WC_Connect_Options::get_option( 'wcshipping_migration_state' );
+			$migration_state = get_option( 'wcshipping_migration_state' );
 			if ( $migration_state !== WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED ) {
-				WC_Connect_Options::update_option( 'wcshipping_migration_state', WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED );
-				set_transient( 'wcshipping_migration_completed', true, DAY_IN_SECONDS );
-				set_transient( 'wctax_migration_completed', true, DAY_IN_SECONDS );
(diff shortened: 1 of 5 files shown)


harriswong:
> One issue is the migration status is stored in the WooCommerce Shipping & Tax options object, and the class helper that access that options object is not available for WooCommerce Shipping and WooCommerce Tax as WooCommerce Shipping & Tax is disabled.

For `wcshipping_migration_state`, would it be better if we change all code:
`WC_Connect_Options::get_option()` --> `get_option()`
`WC_Connect_Options::update_option()` --> `update_option()` 

This way, we decouple the WCS&T class helper from the option itself, making the option available to anything that has access to `wp_options` table.

Ferdev:
Yeah, good idea Harris! I just pushed a new commit with that implementation

woocommerce-services.php
```diff
@@ -265,9 +265,9 @@ public static function plugin_deactivation() {
 			 */
 			require_once __DIR__ . '/classes/class-wc-connect-wcst-to-wcshipping-migration-state-enum.php';
 
-			$migration_state = WC_Connect_Options::get_option( 'wcshipping_migration_state' );
+			$migration_state = get_option( 'wcshipping_migration_state' );
 			if ( $migration_state !== WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED ) {
-				WC_Connect_Options::update_option( 'wcshipping_migration_state', WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED );
+				update_option( 'wcshipping_migration_state', WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED );
```
harriswong, line 270: The `get_option()` doesn't type cast and `"12 !== 12`. After the migration is done, the banner will continue to show. 
image

harriswong (changes requested):
I tested the "Update now" flow 
image, but the banner kept showing, i think we need to typecast. 

Also, `wordpress/wp-content/plugins/woocommerce-services/classes/class-wc-connect-options.php` line 55 `'wcshipping_migration_state',` isn't needed anymore, we can remove. 

Thanks!

Ferdev:
:wave: I've updated the code to cast o int all migration state ocurrences wherever they're read from the database. I'd forgotten that scalar values are returned as strings by `get_option`! What I wonder is why I didn't have issues with the popup not hidding... :thinking: Anyway, just updated the code, let me know what you think!

harriswong (approved):
There is a conflict with `wcs-migration`, but code LGTM. I tested the modal migration from the order listing page and the individual order page 👍 

`WC_Connect_Options` serialized the option so we didn't need to check for type. But the WP option one is straight from the DB table so we do need to check for type. Instead of doing type-safe comparison, `!==`, I was thinking maybe we can do `!=` instead. Then, we don't need all these `intval()` everywhere. The condition for `$migration_pending` is to return a boolean so I think using a typeless comparison would be fine too. 

This is good to merge once the merge conflict is resolved and if the migration button still works after the merge conflict!

Issue #2720, opened 2024-02-29
Enforce HS Tariff code field requirement for EU countries in WCS&T UI


laurajohnsonwoo:
We need to update the WCS&T label purchase customs section to meet updated EU customs requirements. 

For cross-border shipments bound for EU countries: 

1. Remove "Optional" language from HS Tariff code field 
2. Make field required 

**Reference**: 
> Per USPS, "Both HS code and country of origin (for commercial items) are conditional requirements of the destination countries. See IMM 123.711 for more details.
> 
> The HS Code is required by EU countries.  As an example, I have copied below the top of the Germany ICL which states this:
>  
> Country Conditions for Mailing — Germany
> 
> All items containing goods must include the corresponding 6-digit harmonized system code for each article listed on the customs declaration form. For more information, see 537 and the Observations section below."

- [Customs Forms and Online Shipping Labels](https://pe.usps.com/text/imm/immc1_009.htm)
- [International Mail Services](https://pe.usps.com/cpim/ftp/manuals/imm/immc1.pdf)
- [Harmonized Tariff Schedule page](https://hts.usitc.gov/)

Related PR https://github.com/woocommerce/woocommerce-shipping/pull/198

paulostp:
I believe this is related: https://github.com/Automattic/woocommerce-services/issues/2712

sophiegyo:
We may have a similar issue with Switzerland - this from 8272375-zen.
The merchant in that ticket shared this:
https://pe.usps.com/text/imm/immc5_010.htm

iyut:
I will close this issue as [this PR](https://github.com/Automattic/woocommerce-services/pull/2764) has been merged.

PR #2764, opened 2024-07-11
Fix/issue 2712 - Make The Tariff Number Field a Required Field when the Destination is Within EU

### LINKED ISSUE #2712: Customs info missing, but errors are unclear
I had a chatter yesterday who ran into the following error…

> Customs info error: missing required field for non-document shipment: customs_items.hs_tariff_number

image

We were both a little confused at first by the message, as it isn't a friendly error message, but it also prompts you to go to your Credit Card settings.

We did eventually realize it's for the HS Tariff number in the Customs section of label creation. The other confusion is that the field is labeled as optional, but in this case was required for the label.

image

We were then able to quickly find the HS Tariff number and get them on their way, but I'd promised them I'd follow up on this confusion and see how we can make this better for future scenarios like this for other merchants.

So first, when the HS Tariff number is required is it possible to show that for the field? And if this error is triggered, can we change that error message to better direct them on how to find their HS Tariff number(s) and update the Customs section?

iyut:
## Description
Add condition to make the tariff number field become a required field when the destination is within EU.

### Related issue(s)

Fixes #2712   #2720 

1. Add a new order with the destination to EU countries.
2. Try to create the labels from order admin page.
3. Upon the customs, the HS Tariff Number field will be required field : 
image

4. Add the HS Tariff Number and continue the process until buy the label.
5. The label will be purchased successfully.
6. Try to create a new order again but change the destination to outside EU countries ( such as : UK ).
7. And try to repeat the process from no. 2.
8. The HS Tariff Number field will be optional field. You can leave it blank.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
 = 2.6.2 - 2024-xx-xx =
-* Fix - Tariff number field on customs should be required field when the destination country is within EU.
+* Fix - Require HS Tariff number on customs form for EU destination countries.
 
 = 2.6.1 - 2024-07-02 =
 * Tweak - WooCommerce 9.0 and WordPress 6.6 compatibility.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.6.2 - 2024-xx-xx =
+* Fix - Tariff number field on customs should be required field when the destination country is within EU.
+
 = 2.6.1 - 2024-07-02 =
 * Tweak - WooCommerce 9.0 and WordPress 6.6 compatibility.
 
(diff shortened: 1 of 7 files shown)


changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.6.2 - 2024-xx-xx =
+* Fix - Tariff number field on customs should be required field when the destination country is within EU.
```
dustinparker:
```suggestion
* Fix - Require HS Tariff number on customs form for EU destination countries.
```

readme.txt
```diff
@@ -79,6 +79,9 @@ The source code is freely available [in GitHub](https://github.com/Automattic/wo
 
 == Changelog ==
 
+= 2.6.2 - 2024-xx-xx =
+* Fix - Tariff number field on customs should be required field when the destination country is within EU.
```
dustinparker:
```suggestion
* Fix - Require HS Tariff number on customs form for EU destination countries.
```

dustinparker (approved):
Everything works as expected. LGTM!

Abdalsalaam (approved):
LGTM, It works as expected 👍🏼

PR #2765, opened 2024-07-11
Add missing tracking data for the migration process

Ferdev:
## Description
When the plugin is deactivated and the migration is marked as completed, a call to Tracks was missing. We need to track that metric, so this PR takes care of that.

Also removes the "plugin successfully deactivated" message from the plugins list entry, as has been moved to the WooCommerce Shipping plugin.

### Related issue(s)
N/A

1. Create a test site, add this plugin and enable it. Ensure that is connected to WooCommerce Connect Server staging server.
2. Create a WooCommerce Shipping build using latest `wcs-migration` branch and add it to the test site. Don't enable it.
3. Create a build of WooCommerce Tax and add it to the site, but don't enable it.
4. Make sure none of the options in the following query are set:

```sql
select *
from wp_options
where option_name in (
  'wcshipping_migration_state',
  'wcshipping_migration_completed_message_shown',
  'wcst_plugin_deactivated_message_shown'
);
```

Remove them if they exist.

5. Create a few orders (hint, you can use https://github.com/woocommerce/wc-smooth-generator/releases/ for that) and go to the orders page. The message to start the migration should appear at the top of the page. Click the button to start the process.
6. When the migration is completed, a call to Tracks should've been made when WooCommerce Shipping & Tax was deactivated.

### Checklist

- [ ] unit tests

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -267,7 +267,17 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 
 			$migration_state = get_option( 'wcshipping_migration_state' );
 			if ( ! $migration_state || intval( $migration_state ) !== WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED ) {
-				update_option( 'wcshipping_migration_state', WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED );
+				$result = update_option( 'wcshipping_migration_state', WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED );
+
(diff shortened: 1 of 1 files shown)


harriswong (approved):
I can see `WC_Connect_Jetpack::tracks_record_event(....)` being called using the jetpack package, but I didn't confirm the event being recorded in MC tracks.

I searched in MC tracks for `woocommerceconnect_opted_in` to see if the `WC_Connect_Tracks` class is recording track events, seems like it does. So this PR's `$tracks->record_user_event(..)` should work. 

LGTM! 👍

PR #2767, opened 2024-07-12
Remove db migration and deactivation

harriswong:
## Description
This PR changes the state machine to the following
image

1. It skips the DB updates from the migration-runner.js
2. It skips the "deactivating WCS&T" step
3. Once we activated "Woo Shipping", the migration is done and we will redirect the user to the plugins page.
4. Because WCS&T is no longer being activated, it means that the `WC_REST_Connect_Migration_Flag_Controller` controller `wc/v1/connect/migration-flag` will continue to work. The `migration-runner` will now be able to make an API call to update the `wcshipping_migration_state` without relying on `plugin_deactivation` hook. 

This PR also removes `woocommerce-tax` from the installation API call. 

### Related issue(s)
Closes https://github.com/woocommerce/shipping/issues/150

### Steps to test
1. Pull this branch, 
2. Go to plugins page `http://localhost/wp-admin/plugins.php?plugin_status=all&paged=1&s`, enabled WCS&T, deactivate Woo Shipping (wcs-migration branch) and Woo Tax.
3. Open `wordpress/wp-content/plugins/woocommerce-services/classes/class-wc-connect-service-settings-store.php` and add `delete_option( 'wcshipping_migration_state' );` in the `is_eligible_for_migration()`, something like this:
```
+			delete_option( 'wcshipping_migration_state' );
			$migration_state = get_option( 'wcshipping_migration_state' );
```
The idea is to delete the `wcshipping_migration_state` flag so that we can keep testing the banner.
4. Go to the order listing page http://localhost/wp-admin/edit.php?post_type=shop_order, you should see this banner
image
5. Open the network tab, click "Confirm update". You should see `migration-flag` API called twice to update status from `2` to `12`. You should also be able to see `install` and `activate` API calls go through. In the plugins page, you should see only WCS&T and Woo Shipping activated, where Woo Tax remains deactivated.
image

### Checklist

- [ ] unit tests

### DIFF
--- a/client/components/migration/migration-runner.js
+++ b/client/components/migration/migration-runner.js
@@ -43,7 +43,7 @@ const MIGRATION_ENUM_TO_STATE_NAME_MAP = {
     12: 'stateDone'
 }
 
-const plugins = 'woocommerce-shipping,woocommerce-tax'; //this needs to be a CSV string.
+const plugins = 'woocommerce-shipping'; //this needs to be a CSV string.
 
 const getHeaders = () => ({
     "Content-Type": "application/json",
--- a/client/components/migration/migration-runner.js
+++ b/client/components/migration/migration-runner.js
@@ -180,7 +180,7 @@ const migrationStateTransitions = {
     stateDone: { // Done state.
         success: null,
         fail: null,
-        callback: null
+        callback: stateErrorHandlingAPICall( MIGRATION_STATE_ENUM.COMPLETED )
     }
 }
 
--- a/client/components/migration/migration-runner.js
+++ b/client/components/migration/migration-runner.js
@@ -148,7 +148,7 @@ const migrationStateTransitions = {
         callback: stateErrorHandlingAPICall( MIGRATION_STATE_ENUM.ERROR_INSTALLING ),
     },
     stateActivating: {
-        success: 'stateDeactivating', // TODO: This should be stateDBMigrating to migrate DB.
+        success: 'stateDone',
         fail: 'stateErrorActivating',
         callback: fetchAPICall( activatePluginAPICall ),
     },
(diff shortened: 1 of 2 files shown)


client/components/migration/migration-runner.js
```diff
@@ -180,7 +180,7 @@ const migrationStateTransitions = {
     stateDone: { // Done state.
         success: null,
         fail: null,
-        callback: null
+        callback: stateErrorHandlingAPICall( MIGRATION_STATE_ENUM.COMPLETED )
```
harriswong, line 183: WCS&T is no longer deactivated. `WC_REST_Connect_Migration_Flag_Controller controller` and its `wc/v1/connect/migration-flag` will work. We no longer need to rely on `plugin_deactivation` to update the "completed state". We will update the status here instead.

woocommerce-services.php
```diff
@@ -267,22 +267,18 @@ public static function plugin_deactivation() {
 			require_once __DIR__ . '/classes/class-wc-connect-tracks.php';
 			require_once __DIR__ . '/classes/class-wc-connect-wcst-to-wcshipping-migration-state-enum.php';
 
-			$migration_state = get_option( 'wcshipping_migration_state' );
-			if ( ! $migration_state || intval( $migration_state ) !== WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED ) {
```
harriswong, line 271: 👋 Hey @Ferdev, this was introduced in https://github.com/Automattic/woocommerce-services/pull/2765. 

This PR no longer needs to rely on the `plugin_deactivation()` to update the `wcshipping_migration_state` flag to "completed". When the migration modal finishes and redirects the users to `http://localhost/wp-admin/plugins.php`, the `wcshipping_migration_state` will be set to `WC_Connect_WCST_To_WCShipping_Migration_State_Enum::COMPLETED`.

I changed the condition for the tracking to "if it is completed, then do tracking". Let me know if this doesn't work.

Ferdev (approved):
This looks and works as expected. Great job, sir! :shipit:

PR #2770, opened 2024-07-16
Fix modal image not loading on orders listing [570]

samnajian:
## Description
This PR fixes the issue where then absolute path to the dist folder is not present properly when the asset is getting requested.
The PR fixes the issue by providing the right path to the dist folder to the js files.

### Related issue(s)
https://github.com/woocommerce/woocommerce-shipping/issues/570

### Steps to reproduce
[copied from the issue]
1. Go to your local's override file and comment out the `WOOCOMMERCE_CONNECT_DEV_SERVER_URL` constant: `// define('WOOCOMMERCE_CONNECT_DEV_SERVER_URL', 'http://localhost:8085'); //WCS&T`. This disable dev server and uses `dist/`. 
2. Build the files into `/dists`. I used `npm run build`. 
3. Return true from WC_Connect_Service_Settings_Store::is_eligible_for_migration
4. Refresh your page and go to order list page, click "Confirm update", confirm the image is loaded.
5. Visit an order page, click on `Create shipping label` and verify the migration modal properly loads the image.

### Dev notes
To make testing/debugging easier, we don't want the order list "Confirm update" to redirect us every time. I go to `wordpress/wp-content/plugins/woocommerce-services/client/components/migration/migration-runner.js` and modify:
1.  I comment out `// window.location = global.wcsPluginData.adminPluginPath;` so it no longer redirects.
2. I replace `fetch( getBaseURL() + 'wc-admin/plugins/install', {` to `fetch( getBaseURL() + 'wc-admin/plugins/install2', {`. This trips the installer and it will always fail at the installation stage, so it won't deactivate itself. 

### Checklist

- [ ] unit tests

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -21,10 +21,6 @@ import { setNonce, setBaseURL } from 'api/request';
 import wpcomApiMiddleware from 'state/data-layer/wpcom-api-middleware';
 import localApiMiddleware from 'lib/local-api-middleware';
 
-// Modify webpack pubilcPath at runtime based on location of WordPress Plugin.
-// eslint-disable-next-line no-undef
-__webpack_public_path__ = global.wcsPluginData.assetPath;
-
 // We need to lazy load the moment locale files.
 // First we try language code with region if it's different then fall back to language code only.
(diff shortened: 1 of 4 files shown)


Ferdev (approved):
Code looks good, and works as expected in both the orders list and the create shipping label button :+1:

PR #2775, opened 2024-07-23
Introduce new "wc_services_will_disable_shipping_logic" hook

kallehauge:
## Description

We've previously agreed to rely on a hook called `wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax` that will return `true` if both WC Tax and WC Shipping are activated and that all logic in WCS&T will stop working if that's the case, so our individual plugins can take over.

The #2761 PR will force the hook to always be "true" which causes two issues:
1. It's purely down to a race-condition that we're not blocking both Tax and Shipping from being initiated.
    * _The "bail early" logic is defined in `on_plugins_loaded` which comes _after_ we force the hook to always be true._
2. WC Tax will always assume that it can initiate its logic while WCS&T _also_ initiates its tax logic which causes issues.

We should instead add a new filter (`wc_services_will_disable_shipping_logic`) that represents WCS&T supporting conditionally loading shipping, so WC Shipping can listen for that as well as the pre-existing `wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax` condition.

_Note: we could remove `wc_services_will_handle_coexistence_with_woo_shipping_and_woo_tax` but that would prevent us from being able to have months old versions of WCS&T be able to work with both individual plugins present and it's the only WC Tax specific support WCS&T has right now, so it doesn't make sense to remove it._ 

### Related issue(s)

- Related to #2761 

### How to test

_**Important!** Make sure WC Tax isn't activated when you go through this test to make sure we're testing the new `wc_services_will_disable_shipping_logic` filter._

* Spin up a local environment using this branch
* Activate a version of WC Shipping that includes https://github.com/woocommerce/woocommerce-shipping/pull/586
* Activate automated taxes and verify that they still apply in checkout
* Go to your new order you just created while verifying automated taxes works
* Purchase a shipping label that uses the new WC Shipping experience.

### Checklist

- [ ] unit tests

### DIFF
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -346,7 +346,7 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 			add_action( 'plugins_loaded', array( $this, 'jetpack_on_plugins_loaded' ), 1 );
 
 			/**
-			 * Used to let WC Tax know WCS&T will handle the plugins' coexistence.
+			 * Used to let WC Tax know WCS&T will handle the plugin's coexistence.
 			 *
 			 * WCS&T does it by not registering its functionality and displaying an appropriate notice
 			 * in WP admin.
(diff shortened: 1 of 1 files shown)


woocommerce-services.php
```diff
@@ -1270,8 +1283,8 @@ public function add_tracking_info_to_emails( $order, $sent_to_admin, $plain_text
 
 			// Abort if no $order was passed, if the order is not marked as 'completed' or if another extension is handling the emailing.
 			if ( ! $order
-				 || ! $order->has_status( 'completed' )
-				 || ! WC_Connect_Extension_Compatibility::should_email_tracking_details( $order->get_id() ) ) {
+				|| ! $order->has_status( 'completed' )
+				|| ! WC_Connect_Extension_Compatibility::should_email_tracking_details( $order->get_id() ) ) {
```
kallehauge, line 1290: We don't actually do anything here - it's just the pre-commit cleanup logic that is making changes.

Ferdev (approved):
Looks good! Working as expected :+1:

PR #2776, opened 2024-07-23
Status page should respect shipping conditional loading

### LINKED ISSUE #2777: Hide shipping related data on the status page if WC Shipping is activated
Reported here by @Ferdev: https://github.com/Automattic/woocommerce-services/pull/2761#pullrequestreview-2176877690

> One thing I've found confusing is the status tab. It appears to merge both WooCommerce Shipping and WooCommerce Shipping & Tax status pages. While the WooCommerce Shipping & Tax doesn't show any shipping related info, I think it would be clearer for the user if WooCommerce Shipping & Tax used its own tab (the status tab is called "WooCommerce Shipping")

kallehauge:
## Description

The status page needs to respect the new "conditional shipping loading" condition where WC Shipping is able to take over the entire shipping experience.

This PR aims to hide all shipping status on the status page by hiding the following sections:

**Health > WooCommerce Shipping & Tax Data**: Although it includes "Tax Data" in its name, then the services data we fetch are all shipping related, so it's a bit misleading. TL;DR: We should just hide it.

**Debug > Shipping Log**: We should, hopefully, not log anything shipping related, but even if we do, then this just hides it from the UI which is good enough.

**Print**: No need for us to allow test label prints when we have disabled all shipping features.

### Related issue(s)

- Related to #2761 

### How to test

1. Spin up plugin locally using this branch
2. Create a WPCOM connection
3. Go to `/wp-admin/admin.php?page=wc-status&tab=connect`
4. Verify that all of the sections from the screenshot appears
5. Activate WC Shipping (any version will do)
6. Refresh `/wp-admin/admin.php?page=wc-status&tab=connect`
7. Verify that the following sections are now hidden:
    a. Health > WooCommerce Shipping & Tax Data
    b. Debug > Shipping Log
    c. Print

#### Before PR when WC Shipping is activated

Screenshot 2024-07-23 at 03 58 39

#### With PR when WC Shipping is activated

Screenshot 2024-07-23 at 04 36 58

### Checklist

- [x] unit tests

### DIFF
--- a/classes/class-wc-connect-help-view.php
+++ b/classes/class-wc-connect-help-view.php
@@ -119,6 +119,10 @@ if ( ! class_exists( 'WC_Connect_Help_View' ) ) {
 			return $health_items;
 		}
 
+		protected function is_shipping_loaded() {
+			return ! in_array( 'woocommerce-shipping/woocommerce-shipping.php', get_option( 'active_plugins' ) );
+		}
+
 		protected function get_services_items() {
 			$available_service_method_ids = $this->service_schemas_store->get_all_shipping_method_ids();
 			if ( empty( $available_service_method_ids ) ) {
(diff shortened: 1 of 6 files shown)


samnajian (approved):
LGTM! Works as expected.

PR #1672, opened 2019-09-22
Initial implementation of Packaging screen design

bor0:
Fixes #1669.

After discussion with Ola we don't need the checkbox `Save this package for future use`.

Issue #325, opened 2016-05-10
Packages: Adding a second package uses read-only styles in modal


kellychoffman:
I added my first box using these sizes and dimensions:

<img width="487" alt="screen shot 2016-05-10 at 12 33 28 pm" src="https://cloud.githubusercontent.com/assets/5835847/15158078/79306d26-16ab-11e6-9f45-52c4ca7e2fed.png">

I successfully saved and then went to add a second package. This is what came up:

<img width="484" alt="screen shot 2016-05-10 at 12 32 21 pm" src="https://cloud.githubusercontent.com/assets/5835847/15158103/920ffafa-16ab-11e6-8df0-3e36a65bbc0a.png">

cc: @nabsul

PR #2812, opened 2024-09-11
Keep live rates enabled even with WC Shipping active

### LINKED ISSUE #2817: Keep live rates enabled for applicable sites even with WC Shipping active
When the latest release of WooCommerce Shipping & Tax is installed alongside WooCommerce Shipping, live rates are disabled.

The goal of this issue is to restore access to live rates for merchants who had them enabled prior to migrating to WooCommerce Shipping.

Context: p1726007823968179/1725352540.023879-slack-C04KWSNPSE5

Related to https://github.com/woocommerce/woocommerce-shipping/issues/776 because package management needs to be handled out to fix live rates.

waclawjacek:
## Description

Context for development decisions: p1726023174650669-slack-C04KWSNPSE5

This PR keeps live rates enabled even when both WCS&T and WCShip are active. It does this by bringing back the registration of live rates shipping method registration, as well as bringing back the WCS&T settings page (hidden but available through the shipping method settings and in the future, also [through the WCShipping settings page](https://github.com/woocommerce/woocommerce-shipping/pull/748)).

Additionally, it uses package data from WCShipping if a store has been migrated from WCS&T to WCShipping. Since the two extensions use different package data formats, this PR introduces remapping of package data formats on both reads and writes to `wc_connect_options`.

## Breaking changes

The changes described below do not affect how WCShipping stores data, only how WCS&T remaps WCShipping's package data for its own purposes.

However, since WCS&T is writing data to WCShipping's options field when an update is requested from the WCS&T settings page that this PR reinstates for package management, it does affect the values that end up in `wcshipping_options[packages]`.

To clean up the package data in WCShipping (not a required step; just for the sake of clarity), separate PRs will have to be implemented there.

### 1) Outer dimensions of custom packages

This is a feature that seems long deprecated but still supported in the codebase.

In WCS&T, predefined packages have outer dimensions (dimensions of the box used for calculating shipping rates) and inner dimensions (dimensions of the inside of the box used for box packing cart contents).

The WCS&T `wc_connect_options` option allows the key "outer_dimensions" to be defined on custom packages too. This [used to be a feature of the extension](https://github.com/Automattic/woocommerce-services/issues/325) but [it was removed in Sep 2019](https://github.com/Automattic/woocommerce-services/pull/1672).

Although it is no longer possible to set outer dimensions in WCS&T, the field was never unset if a custom package had it set (AFAIK). This means that some stores might have this key lying around for some old custom packages, although it is not presented anywhere anymore.

We could continue passing this key around however since the package edit modal in both WCS&T and WCShipping does not allow you to edit outer dimensions - it only supports inner dimensions (labelled simply as "Dimensions") - there is no way to edit them. This means that merchants effectively have no way to edit the dimensions used for calculating live rates for some very old packages, if they have any still around.

Because of the above, I decided to not support this field in this PR and to only carry over the `inner_dimensions` field which is the one that merchants can edit (and is the only one set for custom packages created after Sep 2019).

### 2) Custom package type value stored under `type`/`isLetter`/`is_letter`

WCS&T used to define packages as a boolean `is_letter`. If a package is a letter, it means it's an envelope. Otherwise, it's a box.

WCShipping introduces a new field called `type` which can have the value of `box` or `envelope`. To support old packages, the field `is_letter` is still kept around, as `isLetter`. In addition to that, a package might also have `is_letter` set.

Since it seems these are generally in sync for the packages defined by WCShipping, I have opted to only keep the `type` field after consulting with the team (p1726489562638679-slack-C04KWSNPSE5).

### Testing instructions

Testing live rates pre-migration:

1. Install WCS&T on a non-migrated-to-WCShipping site (or reset the WCShip migration wizard by running the following code):
   ```
   delete_option( 'wcshipping_migration_state' );
   delete_option( 'wcst_data_migration_required' );
   delete_option( 'wcshipping_installation_completed_shown' );
   delete_option( 'wcst_data_migration_processes_to_run' );
   delete_option( 'wcshipping_version' );
   ```
3. Add some packages in WC > Settings > WooCommerce Shipping.
4. Ensure your site is eligible for USPS live rates.
5. Add a shipping zone with a live rates shipping method.
6. Verify you can access and edit the shipping method's settings.
7. Install and activate WooCommerce Shipping.
8. Verify that live rates are working on the frontend.
   - Please check both the blocks and the classic version of the cart and checkout pages.
9. Verify that label purchases are working.
   - The label purchase modal should display the appropriate shipping costs based on the selected live rates shipping method.
10. Verify the status pages are working for both WCS&T and WCShip.
11. Verify plugin settings are working for both WCS&T and WCShip.

Testing live rates post-migration (after going through all of the above tests):

1. Migrate the settings from WCS&T to WCShip.
2. Go to `admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings` and verify you can see the packages you previously had in WCS&T.
3. Add packages of different dimensions/weight/types on the new live rates settings page (WC > Settings > Shipping > WooCommerce Shipping > Live rates > Select packages). Verify it shows up correctly in the "purchase label" modal.
4. Add packages of different dimensions/weight/types in the "purchase label" modal. Verify it shows up correctly in the live rates settings.
5. Verify that live rates are working on the frontend.
   - Please check both the blocks and the classic version of the cart and checkout pages.
9. Verify that label purchases are working.
   - The label purchase modal should display the appropriate shipping costs based on the selected live rates shipping method.
10. Verify the status pages are working for both WCS&T and WCShip.
11. Verify plugin settings are working for both WCS&T and WCShip.

Testing WCShipping deactivation (after going through all of the above tests):

1. Verify that after deactivating WCShipping, WCS&T displays the packages added/modified/removed during your testing.

### Checklist

- [x] unit tests


woocommerce-services.php
```diff
@@ -747,44 +747,6 @@ protected function add_method_to_shipping_zone( $zone_id, $method_id ) {
 			$this->nux->dismiss_pointer( 'wc_services_add_service_to_zone' );
 		}
 
-		public function init_core_wizard_shipping_config() {
-			if ( $this->is_wc_shipping_activated() ) {
-				return;
-			}
```
waclawjacek, line 753: This methods looks for options set by the WC setup wizard to preconfigure shipping methods. The options it looks for were [removed in WC 3.6](https://github.com/woocommerce/woocommerce/pull/21645) and are no longer set by newer versions of WC. This plugin requires WC ≥ 8.8.

Instead of conditionally loading this, I removed it altogether.

woocommerce-services.php
```diff
@@ -914,32 +876,33 @@ public function attach_hooks() {
...
 
-			add_filter( 'woocommerce_admin_reports', array( $this, 'reports_tabs' ) );
-
-			// Changing the postcode, currency, weight or dimension units affect the returned schema from the server.
-			// Make sure to update the service schemas when these options change.
-			// TODO: Add other options that change the schema here, or figure out a way to do it automatically.
-			add_action( 'update_option_woocommerce_store_postcode', array( $this, 'queue_service_schema_refresh' ) );
-			add_action( 'update_option_woocommerce_currency', array( $this, 'queue_service_schema_refresh' ) );
-			add_action( 'update_option_woocommerce_weight_unit', array( $this, 'queue_service_schema_refresh' ) );
-			add_action( 'update_option_woocommerce_dimension_unit', array( $this, 'queue_service_schema_refresh' ) );
```
waclawjacek, line 941: Moved to `init_shipping_labels()`

woocommerce-services.php
```diff
@@ -914,32 +876,33 @@ public function attach_hooks() {
 			$this->taxjar->init();
 			$this->paypal_ec->init();
 
-			// Primary condition to initiate shipping.
+			/*
+			 * Only register shipping label-related logic if WC Shipping is not active.
+			 */
 			if ( ! $this->is_wc_shipping_activated() ) {
-				add_action( 'rest_api_init', array( $this, 'wc_api_dev_init' ), 9999 );
```
waclawjacek, line 919: No longer needed. See https://github.com/Automattic/woocommerce-services/pull/2812/files#r1753818509

woocommerce-services.php
```diff
@@ -950,22 +913,33 @@ public function init_shipping() {
 				add_action( 'wc_connect_shipping_zone_method_added', array( $this, 'save_defaults_to_shipping_method' ), 10, 3 );
 				add_action( 'woocommerce_shipping_zone_method_deleted', array( $this, 'shipping_zone_method_deleted' ), 10, 3 );
 				add_action( 'woocommerce_shipping_zone_method_status_toggled', array( $this, 'shipping_zone_method_status_toggled' ), 10, 4 );
-
-				// Initialize user choices from the core setup wizard.
-				// Note: Avoid doing so on non-primary requests so we don't duplicate efforts.
-				if ( ! defined( 'DOING_AJAX' ) && is_admin() && ! isset( $_GET['noheader'] ) ) {
-					$this->init_core_wizard_shipping_config();
-				}
```
waclawjacek, line 958: No longer needed. See https://github.com/Automattic/woocommerce-services/pull/2812/files#r1753818509

classes/wc-api-dev/class-wc-rest-dev-data-continents-controller.php
waclawjacek, line 1: This controller was added to support WC versions that didn't have the "v3" controllers. These controllers [were added in WC 3.5.0](https://github.com/woocommerce/woocommerce-rest-api/blob/88ff745b300488d9d2145fa1ab73a2715558530a/src/Controllers/Version3/class-wc-rest-data-continents-controller.php#L8).

We require WC ≥ 8.8.

classes/wc-api-dev/class-wc-rest-dev-data-controller.php
waclawjacek, line 1: This controller was added to support WC versions that didn't have the "v3" controllers. These controllers [were added in WC 3.5.0](https://github.com/woocommerce/woocommerce-rest-api/blob/88ff745b300488d9d2145fa1ab73a2715558530a/src/Controllers/Version3/class-wc-rest-data-controller.php#L8).

We require WC ≥ 8.8.

woocommerce-services.php
waclawjacek, line 919: Removed instead of loading. See: https://github.com/Automattic/woocommerce-services/pull/2812/files#r1753874334

woocommerce-services.php
```diff
@@ -1121,21 +1095,6 @@ public function rest_api_init() {
 			add_filter( 'rest_request_before_callbacks', array( $this, 'log_rest_api_errors' ), 10, 3 );
 		}
 
-		/**
-		 * If the required v3 REST API endpoints haven't been loaded at this point, load the local copies of said endpoints.
-		 * Delete this when the "v3" REST API is included in all the WC versions we support.
-		 */
-		public function wc_api_dev_init() {
```
waclawjacek, line 1128: Removed instead of loading. See: https://github.com/Automattic/woocommerce-services/pull/2812/files#r1753874334

classes/wc-api-dev/class-wc-rest-dev-data-continents-controller.php
waclawjacek, line 1: This controller was added to support WC versions that didn't have the "v3" controllers. These controllers [were added in WC 3.5.0](https://github.com/woocommerce/woocommerce-rest-api/blob/88ff745b300488d9d2145fa1ab73a2715558530a/src/Controllers/Version3/class-wc-rest-data-continents-controller.php#L8).

We require WC ≥ 8.8.

classes/wc-api-dev/class-wc-rest-dev-data-controller.php
waclawjacek, line 1: This controller was added to support WC versions that didn't have the "v3" controllers. These controllers [were added in WC 3.5.0](https://github.com/woocommerce/woocommerce-rest-api/blob/88ff745b300488d9d2145fa1ab73a2715558530a/src/Controllers/Version3/class-wc-rest-data-continents-controller.php#L8).

We require WC ≥ 8.8.

waclawjacek:
There is another thing that needs to be figured out in this PR - support for available packages which are taken into account when calculating live rates. I'm currently working on that.

waclawjacek:
One more thing we need to figure out is that when live rates are calculated, existing packages are taken into account.

When WCShipping is installed, it migrates packages from WCS&T and stores them under a new option name.

With WCShipping present, WCS&T's settings page is no longer registered, and so merchants can no longer edit the available packages on that page. Even if they could, WCS&T would be updating the old option which is still referenced by the live rates logic (provided by WCS&T) but not by the label purchase logic (provided by WCShipping).

In WCShip, the only way to edit packages is in the label purchase modal.

@Automattic/escargot How do you suggest we approach this? Some ideas that came to mind are:

1. Point merchants to the label purchase modal for editing labels and update WCS&T to use WCShip's package list when WCShip is installed.
   - **Drawback:** Requires going to a random order edit screen to access.
2. Keep WCS&T's package configuration screen but make it reference the new settings.
   - **Drawback:** Requires restoring the package manager to the settings page. Should be easy when done on a separate settings page, conditionally linked to from WCShip's settings.
   - **Drawback:** The data structure of package settings might be different now.

kloon:
How about we remove the link, and then just add a check to use the new saved packages instead of the old one for the calculations if it is present? Seem like the most minimal amount of changes?

waclawjacek:
I'll try this approach. 👍

waclawjacek:
Unfortunately, another issue is that in the label purchase modal in WCShipping, there is no notion of enabled/disabled package - all are available.

In WCS&T, you could enable selected carrier-provided and define custom packages. Your choices would then be reflected in which packages are presented in WCS&T's label purchase modal, as well as sent to the Connect Server to decide what packages are used for box packing during live rates calculation.

A way to define these packages seems necessary. How do you think we should implement it?

kallehauge:
> add a check to use the new saved packages instead of the old one for the calculations if it is present?

I created this "normalise" method for our custom packages that maps new field names and removes all new package keys since WCS will reject the requests if it finds unknown key/parameters:

https://github.com/woocommerce/woocommerce-shipping/pull/744/files#diff-5afc41abf1925e1fbd86294fa5dddcbaadb6d48cf1ce7e8e6442ccc75f16a00bR441-R471

waclawjacek:
This PR is ready for review again.

Please verify especially the packages data structure to make sure the data format is correct (item no. 8 in the description).

package.json
```diff
@@ -3,6 +3,7 @@
 	"version": "2.8.1",
 	"scripts": {
 		"start": "cross-env NODE_ENV=development CALYPSO_CLIENT=true webpack-dev-server --hot --inline --watch --content-base dist --port 8085 --host 0.0.0.0",
+		"watch": "cross-env NODE_ENV=development CALYPSO_CLIENT=true webpack --watch",
```
waclawjacek, line 6: I needed this for my local dev. 😅 Can be removed.

woocommerce-services.php
```diff
@@ -1866,6 +1842,7 @@ public function enqueue_wc_connect_script( $root_view, $extra_args = array() ) {
 				'wcs_server_connection' => $is_alive,
 				'accountSettings'       => $account_settings->get(),
 				'packagesSettings'      => $packages_settings->get(),
+				'is_wcshipping_active'  => self::is_wc_shipping_activated(),
```
waclawjacek, line 1857: We need to expose this so that the WCS&T settings page knows what to display.

client/apps/shipping-settings/view-wrapper.js
```diff
@@ -66,34 +66,42 @@ class LabelSettingsWrapper extends Component {
...
-					<LiveRatesCarriersList carrierIds={ liveRatesTypes } />
-					<CarrierAccounts accounts={ carrierAccounts } />
-					<SubscriptionsUsage subscriptions={ subscriptions } />
+					<GlobalNotices id="notices" notices={ notices.list }/>
+					<MigratorSettings/>
+					{ isWCShippingActive && ( <div>
+						<h2>
+							<a href="admin.php?page=wc-settings&tab=shipping&section=woocommerce-shipping-settings">{ translate( 'WooCommerce Shipping' ) }</a> &gt; { translate( 'Live rates' ) }
+						</h2>
+					</div> ) }
```
waclawjacek, line 81: Heading with breadcrumbs (headcrumbs?)

client/apps/shipping-settings/view-wrapper.js
```diff
@@ -66,34 +66,42 @@ class LabelSettingsWrapper extends Component {
...
+					{ !isWCShippingActive && <LabelSettings onChange={ this.onChange }/> }
+					<Packages onChange={ this.onChange }/>
+					<LiveRatesCarriersList carrierIds={ liveRatesTypes }/>
+					{ !isWCShippingActive && <CarrierAccounts accounts={ carrierAccounts }/> }
+					{ !isWCShippingActive && <SubscriptionsUsage subscriptions={ subscriptions }/> }
```
waclawjacek, line 86: Hide sections that are not relevant if WCShipping is active (i.e. only display sections related to live rates).

package.json
kallehauge: No way! Keeper for sure! 😄

client/apps/shipping-settings/view-wrapper.js
harriswong: Got cha. For user to be able to access or edit packages, it needs to have this complementary PR https://github.com/woocommerce/woocommerce-shipping/pull/748 right? This `view-wrapper.js` changes here is to allow `http://localhost/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings` to load otherwise the breadcrumbs points to the shipping one: 
image

This PR make the settings page available if accessed via the URL directly; whereas https://github.com/woocommerce/woocommerce-shipping/pull/748 allows the user to access this via a button. This makes sense to me. 👍

woocommerce-services.php
```diff
@@ -883,18 +845,16 @@ public function load_admin_dependencies() {
 			add_action( 'admin_notices', array( WC_Connect_Error_Notice::instance(), 'render_notice' ) );
 			add_action( 'admin_notices', array( $this, 'render_schema_notices' ) );
 
-			if ( ! $this->is_wc_shipping_activated() ) {
```
harriswong, line 886: We are loading `http://localhost/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings` regardless so this condition is removed right? So that user can edit the live rate packages directly.

harriswong (approved):
Tested per test instructions. LGTM 🚀 ! Good call out on the blocks checkout!

I tested the following with both plugins enabled:
- Add shipping zone with a USPS live rate method
image
- Verified live rates are working on the front end (only when the item has a weight)
- Verified live rates work on classic and block checkout
- Tested label purchase modal, it utilizes WCShipping's one.
- Verified both status page work
- ⚠️ Test step 7) The "WooCommerce Shipping" link loads WCShipping only. WCS&T isn't there. I think that's intended.
image. 
- 8, i added a custom package and I can confirm it showing under "saved template" 👍 
image

Awesome work in such a short period of time! Thanks for this. LGTM.

client/apps/shipping-settings/view-wrapper.js
waclawjacek: Actually, they can also access the packages list from WC > Settings > Shipping > Shipping zones > USPS (WooCommerce Shipping) > Edit, and there's a link to package manager there pointing to ☝️ this page. But as a bonus, I'm adding a link to the WCShip settings page in the PR you mentioned so its easier to find.

woocommerce-services.php
waclawjacek: Yep. Removed so the WCS&T settings page is accessible again, however I did also remove the link to the old settings page within the class controlling that settings page: https://github.com/Automattic/woocommerce-services/pull/2812/files#diff-45ab2d6757efe6316bd243c6eb0cf9194e30ddd4c01b98de3ed0e4a9b0aec1a9

kloon:
I am having trouble getting live rates to actually show, I pinned it down to packages defined in WC Shipping not pulling through dimensions and weight in the WCS&T package editor.

Screenshot 2024-09-12 at 09 29 42
Screenshot 2024-09-12 at 09 32 04

woocommerce-services.php
```diff
@@ -914,22 +874,32 @@ public function attach_hooks() {
 			$this->taxjar->init();
 			$this->paypal_ec->init();
 
-			// Primary condition to initiate shipping.
-			if ( ! $this->is_wc_shipping_activated() ) {
+			/*
+			 * Only register shipping label-related logic if WC Shipping is not active.
+			 */
```
kallehauge: Not something we should spend time on, but technically this is a single inline comment 😄 

```suggestion
			// Only register shipping label-related logic if WC Shipping is not active.
```

kallehauge (commented):
I'm not leaving an "approved" comment because I'm unsure about package normalisation between WCS&T <> WC Shipping (cause it's being worked on), but it works great - also with https://github.com/woocommerce/woocommerce-shipping/pull/748 from the packages I had available 😄

waclawjacek:
I managed to make WCS&T use WCShipping’s package data, reformatted so WCS&T can support it without changing its logic. This approach is based on overwriting the name of the option accessed in WCS&T's stores using WP hooks, with some extra logic to reformat sprinkled on top.

The only place where this works right now is getting the live shipping rates. The shipping settings modal is not fixed yet.

I tried to place all the logic in WCS&T so update to WCShip is not necessary (first of all this would mean adding code to WCShip that isn't related to it, and secondly it'd mean things can get desynchronized when one plugin is updated and the other one isn't).

Tests don’t pass right now which is something to fix.

## Alternative approaches I tried

- Injecting WCShipping's objects into WCS&T's classes wouldn't work because they aren't under one hierarchy.
- I also wanted to make conditional calls to instances of WCShipping's classes if WCShipping is active. I don't remember if I didn't pursue this route because some classes call `add_action` in their constructors which would duplicate some hooks, or if it just turned out to be too much instantiation going on.
- Hooks are a solution that would work, like adding a hook called `wcservices_get_packages` to `calculate_rates` and then another place would do the overwriting, but the code feels clearer to me with direct calls to methods.

waclawjacek:
Changing PR state to "draft" because this is a work in progress.

waclawjacek:
I decided to continue going the route of using hooks because:

- This will allow the calling code to not have to contain compatibility-related function calls.
- The number of calls to the package-related functions isn't that large (plus all these places will all require this compatibility anyway) so scope of change isn't that large.
- It gives extra flexibility if it's ever needed.

The work to use hooks to filter package data is happening in #2819.

waclawjacek:
Nope! Back to using options in the end. IMO this results in cleaner code and catches all access to packages at once.

waclawjacek:
This is now ready for review again.

classes/class-wc-connect-compatibility-wcshipping-packages.php
```diff
@@ -0,0 +1,336 @@
...
+	/**
+	 * Registers hooks intercepting reads/writes to "wc_connect_options".
+	 *
+	 * This is done to replace the keys "packages" and "predefined_packages" with values from WCShipping's options
+	 * after doing some mapping.
+	 *
+	 * @return void
+	 */
+	public static function register_option_overwriting_hooks() {
+		// Intercept reads of "wc_connect_options[packages]" and "wc_connect_options[predefined_packages]".
```
harriswong, line 100: These comments are helpful, thank you!

classes/class-wc-connect-compatibility-wcshipping-packages.php
```diff
@@ -0,0 +1,336 @@
...
+		add_filter( 'option_wc_connect_options', array( self::class, 'intercept_packages_read' ) );
+		add_filter( 'option_wc_connect_options', array( self::class, 'intercept_predefined_packages_read' ) );
+
+		// Intercept updates to "wc_connect_options[packages]" and "wc_connect_options[predefined_packages]".
+		add_action( 'update_option_wc_connect_options', array( self::class, 'intercept_packages_update' ), 10, 2 );
+		add_action( 'update_option_wc_connect_options', array( self::class, 'intercept_predefined_packages_update' ), 10, 2 );
```
harriswong: The 2 `update_` functions here will always overwrite WooShipping's packages and predefined packages if the user edits live rate settings under WCS&T's code, regardless if there are any new changes from WCShipping. This may "delete" any new "saved templates" (the blue star) created from WCShipping's side. 

If WCS&T can update `wc_connect_options` and can overwrite `wcshipping_options` but WooShipping can only update `wcshipping_options` but can't modify `wc_connect_options`, then losing saved templates (blue star) is the only problem. We can fix this by adding another level of `array_merge` to combine them. However, if WooShipping can also update `wc_connect_options`, then this will get really messy and we should avoid doing that.

Also, when I clicked "edit" in one of my custom box via http://localhost/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings, these 2 hooks don't get fired. Is that the same for you too?

harriswong (commented):
Thanks for the PR! I have 2 questions but otherwise tested ok for me!

Tested live rates with new packages 👍 
image

Tested checking out the order and the shipping cost there 👍 
image

Confirmed the packages are brought over after migration 👍 
image

Questions:
Will updating "service packages" for live rates going to be a problem? 
image

classes/class-wc-connect-compatibility-wcshipping-packages.php
waclawjacek: > The 2 update_ functions here will always overwrite WooShipping's packages and predefined packages if the user edits live rate settings under WCS&T's code, regardless if there are any new changes from WCShipping. This may "delete" any new "saved templates" (the blue star) created from WCShipping's side.

They won't! If you look at [this method](https://github.com/Automattic/woocommerce-services/blob/cf4e4556b3d0c1e8799db4a94f96f3c24f1447c5/classes/class-wc-connect-compatibility-wcshipping-packages.php#L158), it first gets WCShip's current options value and then it only replaces the `packages` key.

Whatever WCS&T is writing there is based on current WCShip packages, too, because the read is intercepted as well.

> If WCS&T can update wc_connect_options and can overwrite wcshipping_options but WooShipping can only update wcshipping_options but can't modify wc_connect_options, then losing saved templates (blue star) is the only problem.

WCS&T modifies its entire `wc_connect_options` but the value of the `packages` key is replaced with whatever is in `wcshipping_options[packages]`. In pseudocode, it is doing:

```js
wc_connect_options = {
    ...wc_connect_options,
    packages: wcshipping_options.packages,
};
```

And after a merchant is done modifying the packages in the above object (note they came from the WCShip options), when WCS&T is writing to `wc_connect_options`, it also writes to just the `packages` field of WCShip:

```js
const wcshipping_options = get_from_db( 'wcshipping_options' )

save( 'wc_connect_options', submitted_form_data );

save( 'wcshipping_options', {
    ...wcshipping_options,
    packages: submitted_form_data.packages,
} );
```

> Also, when I clicked "edit" in one of my custom box via http://localhost/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings, these 2 hooks don't get fired. Is that the same for you too?

Please make sure you also have the WC Shipping plugin enabled. Without it, this compatibility layer will not be registered at all.

waclawjacek:
@harriswong Thank you for the review!

> Will updating "service packages" for live rates going to be a problem?

Could you please elaborate on what you mean by updating service packages?

I've responded to the other comments in places where you have posted them.

classes/class-wc-connect-compatibility-wcshipping-packages.php
harriswong: > Whatever WCS&T is writing there is based on current WCShip packages, too, because the read is intercepted as well.
... it also writes to just the packages field of WCShip:

Interesting... that's brilliant! Both read & write are intercepted and but the read-inside-the-write is also intercepted first, so you are reading the WCShipping, turns it to WCS&T settings, then write it, essentially syncing the new database config back to the old ones before writing. 👍 

Ok I need a moment to visualize this...
image
https://xkcd.com/555/

harriswong:
> Could you please elaborate on what you mean by updating service packages?

Yes i can. I was thinking if someone goes to http://localhost/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings and update the "Service packages" there, not "custom package". 
image

Then, this will save and update the WCS&T's configuration because this UI is loaded from WCS&T. That means the changes here won't get updated back in WCShipping's config.

But the explanation you have above https://github.com/Automattic/woocommerce-services/pull/2812#discussion_r1768593233 covered this. Every save action in WCS&T will saves a copy to WCShipping as well. 👍

waclawjacek:
That's right! An update to "Service packages" (a.k.a. carrier packages) changes the `predefined_packages` field the reads of/writes to which we are also intercepting.

harriswong:
> Please make sure you also have the WC Shipping plugin enabled. Without it, this compatibility layer will not be registered at all.

Double checked again and this time I ensured `$is_migration_to_wcshipping_completed` is `true`. Confirmed update is called. 👍

tests/php/classes/test-class-wc-connect-compatibility-wcshipping-packages.php
```diff
@@ -0,0 +1,284 @@
...
+
+		$this->assertEquals(
+			array_merge(
+				self::EXAMPLE_WC_CONNECT_OPTIONS,
+				array(
+					'packages'            => $expected_custom_wcshipping_packages_mapped_to_wcservices_format,
+					'predefined_packages' => self::EXAMPLE_WCSHIPPING_OPTIONS['predefined_packages'],
+				)
+			),
+			get_option( 'wc_connect_options' )
```
harriswong, line 179: Nice test, the setup starts with something in both `wc_connect_options` and `wcshipping_options`. This merges the two together when `wc_connect_options` is requested. 👍

tests/php/classes/test-class-wc-connect-compatibility-wcshipping-packages.php
```diff
@@ -0,0 +1,284 @@
...
+
+		$this->assertEquals(
+			array_merge(
+				self::EXAMPLE_WCSHIPPING_OPTIONS,
+				array(
+					'packages'            => $expected_custom_wcservices_packages_mapped_to_wcshipping_format,
+					'predefined_packages' => self::EXAMPLE_WC_CONNECT_OPTIONS['predefined_packages'],
+				)
+			),
+			get_option( 'wcshipping_options' )
```
harriswong, line 233: 👍

harriswong (approved):
Tested again and looks good to me. 👍 

I have 1 suggestion! What do you think about adding another test for this situation:
For this one, the test starts with
wc_connect_options = [101, 201]
wcshipping_options = [301, 401, 501]

Then we click "save" and add a new package, `update_option(wc_connect_options, [601]);` // adding a completely new config

Then, we verify the following:
wc_connect_options = [101, 201, 301, 401, 501, 601]
wcshipping_options =  [101, 201, 301, 401, 501, 601]

This will help make sure a new item introduced will get saved in both configs.

Just a suggestion though, please feel free to merge! :shipit:

waclawjacek:
Thanks for the review and testing! I've added the test case you suggested. Just a note that if the starting state is:

- `wc_connect_options[packages] == [101, 201]`
- `wcshipping_options[packages] == [301, 401, 501]`

The expected result will be `301, 401, 501, 601` and will not include `101, 201`.

This is because we read from `wcshipping_options` with the changes in this PR active. We don't merge the options on-the-fly.

In a real-world example, if packages are first migrated from WCS&T to WCShip, WCShip's package list should be equal to WCS&T's so then the list would be `[101, 201, 301, 401, 501]`. :smile:

Does that make sense?

By the way there is one quirk of this PR that I thought might not require fixing in this iteration - we can do it separately if needed. That quirk is that if you update the packages in WCShip's purchase modal, the WCS&T option will _not_ get updated. It is not a concern as long as this PR applies because we read WCShip's options anyway but if you disable WCShip, the package list in it will be from the last time you clicked "Save" in the WCS&T package manager we've reintroduced, not when you last clicked save in the WCShip purchase modal.

Please LMK if this sounds concerning. I don't think it is. :)

One way to fix this would be to either prevent changes to `wc_connect_options[packages]` so that one stays "frozen" until WCShip is uninstalled, or to make sure it is always updated when `wcshipping_options` is updated. We could go the former way but that would introduce an extra DB write.

waclawjacek:
@harriswong I managed to work around it. :) I've rewritten the filter of `update_option( 'wc_connect_options`' ) so it updates `wcshipping_options` with the new packages and predefined packages, and it does still update the `wc_connect_options` setting, but not with the new packages/predefined packages. This way when this PR goes out, the package setting will remain "frozen". If WCShipping is uninstalled, everything goes back to the pre-WCShipping installation state.

Could you please give it another look?

tests/php/classes/test-class-wc-connect-compatibility-wcshipping-packages.php
```diff
@@ -277,6 +285,40 @@ public function test_adding_a_package_by_updating_wc_connect_options_also_adds_i
...
+
+		$wc_connect_options = get_option( 'wc_connect_options' );
+		$this->assertEquals( array( 301, 401, 501, 601 ), array_column( $wc_connect_options['packages'], 'box_weight' ) );
+		$this->assertEquals( array( 'WCShipping predefined package', 'foo' ), $wc_connect_options['predefined_packages'] );
+
+		// Deregister the "read" hook so we can access the actual values in the DB without interception.
+		remove_filter( 'option_wc_connect_options', array( WC_Connect_Compatibility_WCShipping_Packages::class, 'intercept_packages_read' ) );
+		remove_filter( 'option_wc_connect_options', array( WC_Connect_Compatibility_WCShipping_Packages::class, 'intercept_predefined_packages_read' ) );
+
+		// Assert the original `wc_connect_options` remain unchanged throughout the entire test.
```
harriswong, line 310: 👍

classes/class-wc-connect-compatibility-wcshipping-packages.php
```diff
@@ -174,9 +181,9 @@ public static function intercept_packages_update( $old_wc_connect_options, $wc_c
 	 * @param mixed $old_wc_connect_options "wc_connect_options" value from the WP options table prior to the update.
 	 * @param mixed $wc_connect_options "wc_connect_options" value from the WP options table after updating.
 	 *
-	 * @return mixed
+	 * @return void
 	 */
-	public static function intercept_predefined_packages_update( $old_wc_connect_options, $wc_connect_options ) {
+	public static function intercept_predefined_packages_update( $wc_connect_options, $old_wc_connect_options ) {
```
harriswong: We should probably update the `@param` comments

harriswong (approved):
Thanks for the updates! I tested a migration and unserialized the before-and-after for `wc_connect_options` and `wcshipping_options`, I confirmed that updating WCShipping doesn't update WCS&T's options.

 I added a comment regarding the PHP docs but the last 3 commits look good to me. :shipit:

classes/class-wc-connect-compatibility-wcshipping-packages.php
waclawjacek: Done! Thanks.

harriswong: LGTM!

harriswong:
last commit LGTM :shipit: 

PS. Who would have thought that the `update_` and `pre_update_` hooks have the reverse parameter between the value and the old_value. Nice catch too!

PR #2834, opened 2024-12-18
Fix/issue 2562 - Add Retail Delivery Fee for Colorado

### LINKED ISSUE #2562: Colorado Retail Delivery Fee
Starting July 1st 2022, merchants who are subject to Colorado state sales tax are required to charge a $0.27 fee for taxable goods that are mailed, shipped, or otherwise delivered by motor vehicle to a purchaser in Colorado. See the new law [here](https://tax.colorado.gov/retail-delivery-fee)

The fee must be listed separately with the label "Retail Delivery Fee". It may not be possible to return an itemized and separated list of how taxes are broken down from TaxJar's API.

Is this something that should be explored so WooCommerce Shipping & Tax is compliant with tax laws for CO merchants?

iyut:
## Description
Adding a function for custom surcharge ( retail delivery fee ) for Colorado state. This changes also introduce a new filter hook that the user can use to disable the custom surcharge addition.

### Related issue(s)
Fixes #2562 

1. Add product to the cart.
2. Use any Colorado address for the shipping address.
3. The Retail Delivery Fee will be displayed as a free on the cart.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
-= 2.8.7 - 2024-xx-xx =
+= 2.9.0 - 2024-xx-xx =
 * Add   - Option to apply US Colorado Retail Delivery Fee tax by using `wc_services_apply_us_co_retail_delivery_fee` filter.
 
 = 2.8.6 - 2025-01-06 =
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
-= 2.9.0 - 2024-xx-xx =
+= 2.9.0 - 2025-xx-xx =
 * Add   - Option to apply US Colorado Retail Delivery Fee tax by using `wc_services_apply_us_co_retail_delivery_fee` filter.
 
 = 2.8.6 - 2025-01-06 =
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.8.6 - 2024-xx-xx =
+* Add   - Custom surcharge ( Retail Delivery Fee ) for shipment to Colorado, US.
+
 = 2.8.5 - 2024-12-10 =
 * Fix   - Fixed an issue that prevented editing an order when automated tax is enabled.
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
 = 2.8.6 - 2024-xx-xx =
-* Add   - Custom surcharge ( Retail Delivery Fee ) for shipment to Colorado, US.
+* Add   - Option to apply US Colorado Retail Delivery Fee tax by using `wc_services_apply_us_co_retail_delivery_fee` filter.
 
 = 2.8.5 - 2024-12-10 =
 * Fix   - Fixed an issue that prevented editing an order when automated tax is enabled.
--- a/classes/class-wc-connect-custom-surcharge.php
+++ b/classes/class-wc-connect-custom-surcharge.php
@@ -23,11 +23,22 @@ if ( ! class_exists( 'WC_Connect_Custom_Surcharge' ) ) {
 		 * @return boolean.
 		 */
 		public static function get_us_co_eligibility( $cart = null ) {
-			if ( $cart instanceof WC_Cart ) {
-				return true;
+			if ( false === ( $cart instanceof WC_Cart ) ) {
+				return false;
 			}
 
-			return false;
+			$has_taxable_product = false;
+
+			foreach ( $cart->get_cart() as $cart_item ) {
+				if ( $cart_item['data']->is_taxable() ) {
+					$has_taxable_product = true;
+					break;
+				}
+			}
+
+			$content_taxes = $cart->get_cart_contents_taxes();
+
+			return $has_taxable_product && ! empty( $content_taxes );
 		}
 	}
 }
--- a/classes/class-wc-connect-custom-surcharge.php
+++ b/classes/class-wc-connect-custom-surcharge.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * A class for working around the quirks and different versions of WordPress/WooCommerce
+ * This is for versions higher than 2.6 (3.0 and higher)
+ */
+
+// No direct access please.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( ! class_exists( 'WC_Connect_Custom_Surcharge' ) ) {
+	/**
+	 * WC_Connect_Compatibility class.
+	 */
+	class WC_Connect_Custom_Surcharge {
+
+		/**
+		 * Get US Colorado custom surcharge eligibility.
+		 *
+		 * @param WC_Cart|null $cart Cart object or null.
+		 *
+		 * @return boolean.
+		 */
+		public static function get_us_co_eligibility( $cart = null ) {
+			if ( $cart instanceof WC_Cart ) {
+				return true;
+			}
+
+			return false;
+		}
+	}
+}
--- a/classes/class-wc-connect-custom-surcharge.php
+++ b/classes/class-wc-connect-custom-surcharge.php
@@ -23,8 +23,20 @@ if ( ! class_exists( 'WC_Connect_Custom_Surcharge' ) ) {
 		 * @return boolean.
 		 */
 		public static function get_us_co_eligibility( $cart = null ) {
+			$fee_info = array();
+
 			if ( false === ( $cart instanceof WC_Cart ) ) {
-				return false;
+				return $fee_info;
+			}
+
+			// Do not apply the fee if all products are virtual.
+			if ( WC_Connect_Utils::is_all_products_are_virtual( $cart ) ) {
+				return $fee_info;
+			}
+	
+			// Do not apply the fee if all shipping methods use Local Pickup.
+			if ( 0 === count( array_diff( wc_get_chosen_shipping_method_ids(), apply_filters( 'woocommerce_local_pickup_methods', array( 'legacy_local_pickup', 'local_pickup' ) ) ) ) ) {
+				return $fee_info;
 			}
 
 			$has_taxable_product = false;
@@ -38,7 +50,14 @@ if ( ! class_exists( 'WC_Connect_Custom_Surcharge' ) ) {
 
 			$content_taxes = $cart->get_cart_contents_taxes();
 
-			return $has_taxable_product && ! empty( $content_taxes );
+			if ( $has_taxable_product && ! empty( $content_taxes ) ) {
+				$fee_info = array(
+					'fee'      => 0.27,
+					'fee_text' => __( 'Retail Delivery Fee', 'woocommerce_services' ),
+				);
+			}
+			
+			return $fee_info;
 		}
 	}
 }
--- a/classes/class-wc-connect-custom-surcharge.php
+++ b/classes/class-wc-connect-custom-surcharge.php
@@ -20,7 +20,7 @@ if ( ! class_exists( 'WC_Connect_Custom_Surcharge' ) ) {
 		 *
 		 * @param WC_Cart|null $cart Cart object or null.
 		 *
-		 * @return boolean.
+		 * @return array|false.
 		 */
 		public static function add_us_co_rdf( $cart = null ) {
 			if ( false === ( $cart instanceof WC_Cart ) ) {
@@ -42,7 +42,7 @@ if ( ! class_exists( 'WC_Connect_Custom_Surcharge' ) ) {
 			}
 			
 			return array(
-				'fee'      => 0.27,
+				'fee'      => 0.29,
 				'fee_text' => __( 'Retail Delivery Fee', 'woocommerce_services' ),
 			);
 		}
--- a/classes/class-wc-connect-custom-surcharge.php
+++ b/classes/class-wc-connect-custom-surcharge.php
@@ -13,24 +13,42 @@ if ( ! class_exists( 'WC_Connect_Custom_Surcharge' ) ) {
 	 * WC_Connect_Custom_Surcharge class.
 	 */
 	class WC_Connect_Custom_Surcharge {
+
 		/**
-		 * Add a 29 cent surcharge to your cart / checkout based on delivery US Colorado
-		 * Taxes, shipping costs and order subtotal are all included in the surcharge amount
-		 *
-		 * Fee won't be added if all products in the cart are virtual.
-		 * 
-		 * This fee is being DISABLED by default because there is a lot of consideration for this fee to be applied.
-		 * 
-		 * If the user wish to use this fee, they can enable it by using `wc_services_apply_us_co_retail_delivery_fee` filter.
-		 * Change the 'enabled' value from `false` to `true`. See the example below:
-		 * 
-		 * `add_filter( 'wc_services_apply_us_co_retail_delivery_fee', function( $fee_info, $cart ) { $fee_info['enabled'] = true; return $fee_info; }, 10, 2 );`
-		 *
-		 * Uses the WooCommerce fees API
-		 *
-		 * @param WC_Cart|null $cart WooCommerce Cart object.
+		 * Initialize the class and set up action hooks.
 		 */
-		public static function add_us_co_rdf( $cart = null ) {
+		public static function init() {
+			add_action( 'woocommerce_cart_calculate_fees', array( static::class, 'add_us_co_retail_delivery_fee' ), 10 );
+		}
+		/**
+		 * Add US Colorado Retail Delivery Fee Tax.
+		 * Uses the WooCommerce fees API `$cart->add_fee()`.
+		 *
+		 * Colorado Retail Delivery Fee Tax:
+		 * https://www.avalara.com/blog/en/north-america/2022/10/what-you-need-to-know-about-the-colorado-retail-delivery-fee-now.html
+		 *
+		 * RDF fee is DISABLED by default - not all business are required to charge the fee.
+		 * To apply the fee use `wc_services_apply_us_co_retail_delivery_fee` filter.
+		 * Change boolian flag to `true`
+		 * Example: `add_filter( 'wc_services_apply_us_co_retail_delivery_fee', '__return_true' );`
+		 *
+		 * @param WC_Cart $cart WooCommerce Cart object.
+		 */
+		public static function add_us_co_retail_delivery_fee( $cart ) {
+
+			/**
+			 * Filter should Retail Delivery Fee be applied.
+			 * Default: false.
+			 *
+			 * @since 2.8.6
+			 *
+			 * @param bool    Should the Retail Delivery Fee be applied.
+			 * @param WC_Cart WooCommerce cart object.
+			 */
+			if ( ! apply_filters( 'wc_services_apply_us_co_retail_delivery_fee', false, $cart ) ) {
+				return;
+			}
+
 			if ( is_admin() && ! defined( 'DOING_AJAX' ) ) {
 				return;
 			}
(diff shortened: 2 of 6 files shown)


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1499,4 +1502,109 @@ public function load_taxjar_admin_new_order_assets() {
...
+		}
+		
+		if( $this->maybe_all_products_are_virtual( $cart ) ) {
+			return;
+		}
+
+		/** Checks if Local Pickup is selected to not apply the fee */
+		if ( count( array_intersect( wc_get_chosen_shipping_method_ids(), apply_filters( 'woocommerce_local_pickup_methods', array( 'legacy_local_pickup', 'local_pickup' ) ) ) ) > 0 ) {
+			return;
+		}
```
bartech: We should check if all chosen methods `wc_get_chosen_shipping_method_ids()` are local_pickup, if anyone of them is not fee should be applied.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1499,4 +1502,109 @@ public function load_taxjar_admin_new_order_assets() {
...
+	 *
+	 * Change $state to set the state
+	 * Example: The state code for Colorado would be 'CO'
+	 * https://en.wikipedia.org/wiki/ISO_3166-2:US
+	 *
+	 * Uses the WooCommerce fees API
+	 *
+	 * @param WC_Cart $cart WooCommerce Cart object.
+	 */
+	public function add_custom_state_surcharge_fee( $cart ) {
```
bartech: We have to check if Colorado state Sales or Use Tax is charged on the items (see if we already have those taxes applied) as usually when items are exempt from state Sales or Use Tax they are exempt from Retail Delivery Fee.

bartech (changes requested):
I suggest to create separate file that have custom fees, maybe even folder to store logic per each case in separate file (function) that will return fee data or false.
In main hooked method check if `$custom_fees` contain item with `$customer->get_shipping_country() . ':" . $customer->get_shipping_state()`.
```
$custom_fees = array(
	'US:CO' => 'custom_function_name'
);
```
If matched include custom fee logic file and run `custom_function_name` to check if fee should be applied, if should be applied return fee data.
```
array(
	'value' => 0.27 // can be calculated
	'label' => __( 'Retail Delivery Fee', 'woocommerce_services' ),
);
```
If fee should not be applied return `false`.

classes/class-wc-connect-taxjar-integration.php
iyut: Hi @bartech , I'm not sure if I understand this. Can you explain a bit more on this?

This code is checking if there is any non "local pickup" method is being chosen or not. If it's a local pickup then the fee should not be applied. That's why it has `return` there.

classes/class-wc-connect-taxjar-integration.php
iyut: This is a nice catch. Thanks! I did try to dig into this and based on [this article](https://www.taxjar.com/blog/2022-08-what-is-the-colorado-retail-delivery-fee), and it said this : "If the entire retail sale is exempt from sales tax, the sale is also exempt from the retail delivery fee. If one item in the retail sale is exempt, but another item is subject to sales tax, the retailer should collect the retail delivery fee.". From WooCommerce point of view, it seems we don't need to apply this fee if either the sales tax is zero, or all products is non-taxable.

classes/class-wc-connect-taxjar-integration.php
bartech: As `wc_get_chosen_shipping_method_ids` is an array it can have multiple shipping methods picked (in case there are multiple shipping packages). Now code checks if any method is local pickup and we have to check if all of them are local pickup not just one of them. Even if part of the order is sent via carrier or delivered you apply the flat fee 27c

classes/class-wc-connect-taxjar-integration.php
bartech: Yes, if there is no tax on whole cart level no need to apply the fee.

classes/class-wc-connect-taxjar-integration.php
iyut: Ahh ok, I got it now and have made the changes.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1499,4 +1502,115 @@ public function load_taxjar_admin_new_order_assets() {
...
+		if ( is_admin() && ! defined( 'DOING_AJAX' ) ) {
+			return;
+		}
+		
+		if ( $this->maybe_all_products_are_virtual( $cart ) ) {
+			return;
+		}
```
bartech: This should be part of custom method. There might be a fee or custom tax that should be applied to virtual products.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1499,4 +1502,115 @@ public function load_taxjar_admin_new_order_assets() {
...
+
+		/** Do not apply the fee if all shipping methods use Local Pickup */
+		if ( 0 === count( array_diff( wc_get_chosen_shipping_method_ids(), apply_filters( 'woocommerce_local_pickup_methods', array( 'legacy_local_pickup', 'local_pickup' ) ) ) ) ) {
+			return;
+		}
```
bartech: This should be in custom method. It's specific to Colorado fee.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1499,4 +1502,115 @@ public function load_taxjar_admin_new_order_assets() {
...
+
+		if ( ! is_array( $this->get_custom_surcharges() ) ) {
+			return;
+		}
+
+		$customer = WC()->customer;
+
+		foreach ( $this->get_custom_surcharges() as $key => $custom_info ) {
```
bartech: Instead of looping over `$this->get_custom_surcharges()` better to get `$custom_surcharges` array and check if key exists. There  can be only one item with same index anyway. 
```php
$custom_surcharges = $this->get_custom_surcharges();
$key = $customer->get_shipping_country() . ':' . $customer->get_shipping_state();
if ( ! isset( $custom_surcharges[ $key ] )  {
	return;
}
```
If we want to apply multiple rules (fees) for same location we have to allow item to be array of methods.
```
'US:CO' => array( 'method_one', 'method_two' ),
```
IMHO it's better to use `:` as delimiter.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1499,4 +1502,115 @@ public function load_taxjar_admin_new_order_assets() {
...
+			if ( $customer->get_shipping_country() . '-' . $customer->get_shipping_state() !== $key ) {
+				continue;
+			}
+
+			// Making sure all info is not empty.
+			if ( ! isset( $custom_info['function'] ) || ! isset( $custom_info['fee'] ) || ! isset( $custom_info['fee_text'] ) || ! is_float( $custom_info['fee'] ) || ! is_callable( $custom_info['function'] ) ) {
+				continue;
+			}
+
+			$eligible_to_add_fee = call_user_func( $custom_info['function'], $cart );
```
bartech: Better if custom method returned the fee. This way fee can be defined dynamically by custom method. Return fee as array `'value' => 123, 'label' => 'Fee name'` or false if fee should not be applied. And here check if method is callable.

iyut: Yeah, I agree on this. Since we're applying multiple rules, I think we should set the fee in the custom method.

classes/class-wc-connect-taxjar-integration.php
iyut: Multiple rules seems like a great idea for scalability.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -170,6 +170,9 @@ public function init() {
 
 		add_filter( 'woocommerce_calc_tax', array( $this, 'override_woocommerce_tax_rates' ), 10, 3 );
 		add_filter( 'woocommerce_matched_rates', array( $this, 'allow_street_address_for_matched_rates' ), 10, 2 );
+		
+		// Add custom state surcharge.
+		add_action( 'woocommerce_cart_calculate_fees', array( $this, 'add_custom_state_surcharge_fee' ), 10 );
```
bartech: Let's rename to `maybe_add_custom_surcharge_fees` it doesn't have to be state specific.
No need for comment as method name say it all.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1499,4 +1502,86 @@ public function load_taxjar_admin_new_order_assets() {
...
+	public function add_custom_state_surcharge_fee( $cart ) {
+		if ( is_admin() && ! defined( 'DOING_AJAX' ) ) {
+			return;
+		}
+
+		if ( ! is_array( $this->get_custom_surcharges() ) ) {
+			return;
+		}
+
+		$custom_surcharges = $this->get_custom_surcharges();
```
bartech: Move the `$custom_surcharges = $this->get_custom_surcharges();` above the sanity check, no need to call method twice. Also suggest to rename to just `$surcharges`

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1499,4 +1502,86 @@ public function load_taxjar_admin_new_order_assets() {
...
+			 * @since 2.8.6
+			 *
+			 * @param array   Custom surcharge info.
+			 * @param string  Custom surcharge area string. Example: "US-CO".
+			 * @param array   Custom surcharge function or method.
+			 * @param WC_Cart WooCommerce cart object.
+			 */
+			$fee_info = apply_filters( 'wc_services_apply_custom_surcharge_fee', $fee_info, $key, $function_rule, $cart );
+			
+			if ( isset( $fee_info['fee_text'] ) && isset( $fee_info['fee'] ) && is_float( $fee_info['fee'] ) ) {
```
bartech: I would check if `$fee_info['fee']` `is_numeric` and cast it to float. Mostly for case when it's an int.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1499,4 +1502,86 @@ public function load_taxjar_admin_new_order_assets() {
...
+			 * Set to `false` to disable the custom surcharge.
+			 *
+			 * @since 2.8.6
+			 *
+			 * @param array   Custom surcharge info.
+			 * @param string  Custom surcharge area string. Example: "US-CO".
+			 * @param array   Custom surcharge function or method.
+			 * @param WC_Cart WooCommerce cart object.
+			 */
+			$fee_info = apply_filters( 'wc_services_apply_custom_surcharge_fee', $fee_info, $key, $function_rule, $cart );
```
bartech: We can call it just `$fee`.
Why we pass `$function_rule` in filter, how is this useful for the developer?

classes/class-wc-connect-custom-surcharge.php
```diff
@@ -0,0 +1,63 @@
...
+			}
+
+			$has_taxable_product = false;
+
+			foreach ( $cart->get_cart() as $cart_item ) {
+				if ( $cart_item['data']->is_taxable() ) {
+					$has_taxable_product = true;
+					break;
+				}
+			}
```
bartech: Think we don't have to check if products are taxable. `$cart->get_cart_contents_taxes()` check is enough, if products are not taxable `get_cart_contents_taxes` will return empty array. Maybe even check if  `0 < $cart->get_cart_contents_tax()` will be enough.

classes/class-wc-connect-custom-surcharge.php
```diff
@@ -0,0 +1,63 @@
...
+			$fee_info = array();
+
+			if ( false === ( $cart instanceof WC_Cart ) ) {
+				return $fee_info;
+			}
+
+			// Do not apply the fee if all products are virtual.
+			if ( WC_Connect_Utils::is_all_products_are_virtual( $cart ) ) {
+				return $fee_info;
+			}
```
bartech: Better replace with `$cart->needs_shipping()`. It's a loop over items in cart that check if each product `$product->needs_shipping()` which  is `! product->is_virtual()`

classes/class-wc-connect-custom-surcharge.php
```diff
@@ -0,0 +1,63 @@
...
+			foreach ( $cart->get_cart() as $cart_item ) {
+				if ( $cart_item['data']->is_taxable() ) {
+					$has_taxable_product = true;
+					break;
+				}
+			}
+
+			$content_taxes = $cart->get_cart_contents_taxes();
+
+			if ( $has_taxable_product && ! empty( $content_taxes ) ) {
```
bartech: Better to reverse the check and return early.

classes/class-wc-connect-custom-surcharge.php
```diff
@@ -0,0 +1,63 @@
...
+				$fee_info = array(
+					'fee'      => 0.27,
+					'fee_text' => __( 'Retail Delivery Fee', 'woocommerce_services' ),
+				);
+			}
+			
+			return $fee_info;
+		}
```
bartech:
```suggestion
		/**
		 * Add US Colorado "Retail Delivery Fee" Tax.
		 *
		 * @param WC_Cart|null $cart Cart object or null.
		 *
		 * @return array|false.
		 */
		public static function add_us_co_rdf( $cart = null ) {

			if ( false === ( $cart instanceof WC_Cart ) ) {
				return false;
			}
			
			// Retail Delivery Fee is only applicable if any product in order is taxable.
			if ( 0 === $cart->get_cart_contents_tax() ) {
				return false;
			}
			
			// Retail Delivery Fee is only applicable when order is shipped.
			if ( ! $cart->needs_shipping() ) {
				return false;
			}
	
			// Do not apply Retail Delivery Fee if all selected shipping methods are Local Pickup.
			if ( 0 === count( array_diff( wc_get_chosen_shipping_method_ids(), apply_filters( 'woocommerce_local_pickup_methods', array( 'legacy_local_pickup', 'local_pickup' ) ) ) ) ) {
				return false;
			}
			
			return array(
				'value' => 0.29,
				'label' => __( 'Retail Delivery Fee', 'woocommerce_services' ),
			);
		}
```

I renamed the method please adjust callable name in custom surcharges.
Adjusted fee to 29c based on [this](https://www.avalara.com/blog/en/north-america/2022/10/what-you-need-to-know-about-the-colorado-retail-delivery-fee-now.html#:~:text=The%20Colorado%20RDF%20has%20risen,%2C%202024%E2%80%93June%2030%2C%202025)

bartech (dismissed):

classes/class-wc-connect-taxjar-integration.php
iyut: From what I see, it's needed to differentiate between its rules. In case there are multiple rules under one state.

classes/class-wc-connect-taxjar-integration.php
iyut: Yeah, I just tested it again, it seems int value like 1, 2 will return `false` when using `is_float()`. I think using `is_numeric` will be fine and cast it to float later on.

classes/class-wc-connect-custom-surcharge.php
bartech: @iyut what about 29c fee? [here](https://www.avalara.com/blog/en/north-america/2022/10/what-you-need-to-know-about-the-colorado-retail-delivery-fee-now.html#:~:text=The%20Colorado%20RDF%20has%20risen,%2C%202024%E2%80%93June%2030%2C%202025)

classes/class-wc-connect-custom-surcharge.php
```diff
@@ -0,0 +1,50 @@
...
+	 * WC_Connect_Compatibility class.
+	 */
+	class WC_Connect_Custom_Surcharge {
+
+		/**
+		 * Add US Colorado "Retail Delivery Fee" Tax.
+		 *
+		 * @param WC_Cart|null $cart Cart object or null.
+		 *
+		 * @return boolean.
```
bartech: return type docblock should be array|false

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1499,4 +1501,87 @@ public function load_taxjar_admin_new_order_assets() {
...
+
+		$customer = WC()->customer;
+		$key      = $customer->get_shipping_country() . ':' . $customer->get_shipping_state();
+
+		if ( ! isset( $surcharges[ $key ] ) || ! is_array( $surcharges[ $key ] ) )  {
+			return;
+		}
+
+		foreach ( $surcharges[ $key ] as $function_rule ) {
+			$fee_info = call_user_func( $function_rule, $cart );
```
bartech: Lets add sanity check to make sure provided method is callable.

```php
if ( ! is_callable( $callback ) ) {
	// Log error.
	continue;
}

call_user_func( $callback );
```

classes/class-wc-connect-utils.php
```diff
@@ -104,5 +104,22 @@ public static function get_line_item_from_order( $product_id, $order ) {
...
 		}
+
+		/**
+		 * Check if all products are virtual or not.
+		 *
+		 * @param WC_Cart $cart WooCommerce Cart object.
+		 *
+		 * @return boolean.
+		 */
+		public static function is_all_products_are_virtual( $cart ) {
```
bartech: we don't need this anymore.

classes/class-wc-connect-custom-surcharge.php
iyut: Thanks for the reminder @bartech. I missed that!

bartech:
@iyut after reviewing all again, let's move whole logic into single function hooked directly to  `woocommerce_cart_calculate_fees` and make it opt-in by adding logic with filter at the top of the method.
```php
if ( ! apply_filters( 'wc_services_custom_tax_colorado_rdf', false, $cart ) ) {
	return;
}
```

then merchant can opt in by adding this snippet
```php
add_filter( 'wc_services_custom_tax_colorado_rdf', '__return_true' );
```

iyut:
@bartech , after having a second thought, I think it's still best if we put the whole custom surcharge logic in a different class.  The reason is, to make the future custom surcharge ( if there is any ) centralized in one class. What do you think? I have made the changes for this in [this commit](https://github.com/Automattic/woocommerce-services/pull/2834/commits/230a6d1c73a2f1143ab6ff69b7cf6eab8c8d6c44). LMKWYT?

bartech:
Yes, this is what I also thought to put everything in separate file.

classes/class-wc-connect-custom-surcharge.php
```diff
@@ -39,50 +57,48 @@ public static function add_us_co_rdf( $cart = null ) {
...
-				isset( $fee_info['enabled'] ) && true === $fee_info['enabled'] &&
-				isset( $fee_info['text'] ) &&
-				isset( $fee_info['value'] ) && is_numeric( $fee_info['value'] )
-			) {
-				$cart->add_fee( $fee_info['text'], floatval( $fee_info['value'] ), true, 'standard' );
-			}
+			// As of July 1, 2024 till June 30, 2025 RDF is 29 cents per order
+			// RDF is subject to sales tax.
+			// https://www.avalara.com/blog/en/north-america/2022/10/what-you-need-to-know-about-the-colorado-retail-delivery-fee-now.html.
+			$cart->add_fee( __( 'Retail Delivery Fee', 'woocommerce_services' ), 0.29, true, 'standard' );
```
iyut: @bartech , the reason I was putting the filter here is to let the user change the fee value easily through filter if the new fee is coming. Do you see any issue in the future if we put the filter here?

bartech: We can update the fee. User can also based on this method logic create their own and keep our disabled.

iyut: Yeah I guess they can update the method. Although, I'm still unsure if all user have the knowledge as we do. And if they want to change the text, they might as well need to use the whole method.

bartech: You can wrap Fee Name and Fee Value in filter as it was before as array. I was thinking about it too and was on the fence if this will be used by the merchant and if you think it will let's do that. Change I did was mostly to move bool flag filter to the top of the method instead of doing whole work of checking the logic to ditch everything at the end when enabled/disabled was a part of data for `add_fee`

readme.txt
```diff
@@ -81,6 +81,9 @@ The source code is freely available [in GitHub](https://github.com/Automattic/wo
 
 == Changelog ==
 
+= 2.8.7 - 2024-xx-xx =
```
dustinparker:
```suggestion
= 2.9.0 - 2025-xx-xx =
```

changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.8.7 - 2024-xx-xx =
```
dustinparker:
```suggestion
= 2.9.0 - 2025-xx-xx =
```

dustinparker (commented):
Everything works great on the frontend! I noticed that the fee doesn't get added to orders that are created via the backend. Technically the merchant can add this fee manually when creating orders in the backend, but they might forget to do so. Would you mind checking how much effort would be involved to automatically add this to a backend order when shipping is added?

classes/class-wc-connect-custom-surcharge.php
iyut: ok, let me add that.

iyut:
> Would you mind checking how much effort would be involved to automatically add this to a backend order when shipping is added?

@dustinparker , with development, testing, and reviewing, I would say 2 story points just to be safe. But I'm still thinking it's better for the user to create the fee manually.

dustinparker:
> > Would you mind checking how much effort would be involved to automatically add this to a backend order when shipping is added?
> 
> @dustinparker , with development, testing, and reviewing, I would say 2 story points just to be safe. But I'm still thinking it's better for the user to create the fee manually.

@kloon would you like us to automatically apply the Colorado Retail Delivery Fee to orders created via the admin when they add shipping to an order? Or should we leave it as a manual process?

kloon:
@dustinparker, Anything about admin side created orders is usually manual, so let's stick with that.

changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.9.0 - 2024-xx-xx =
```
dustinparker:
```suggestion
= 2.9.0 - 2025-xx-xx =
```

readme.txt
```diff
@@ -81,6 +81,9 @@ The source code is freely available [in GitHub](https://github.com/Automattic/wo
 
 == Changelog ==
 
+= 2.9.0 - 2024-xx-xx =
```
dustinparker:
```suggestion
= 2.9.0 - 2025-xx-xx =
```

dustinparker (approved):
Everything works as expected. LGTM!

PR #2837, opened 2025-01-31
Deprecate shipping features for new installations

Ferdev:
# 🚨 *WARNING: DO NOT MERGE* 🚨

We are not ready yet to release this PR. Please do not merge it until we have the green light from the team.

-------

## Description
Rename plugin to "WooCommerce Tax" and show shipping features only for existing users. This change:
- Adds 'legacy_site' flag to track existing vs new installations
- Limits shipping functionality to legacy sites only
- Updates plugin name and description for new installations

Fixes WOOSHIP-1339

### Note
This PR works in collaboration with this WooCommerce Connect Server PR: https://github.com/automattic/woocommerce-connect-server/pull/2735
Please, checkout that branch and configure your "WooCommerce Shipping & Tax" install to connect to your local WooCommerce Connect Server before testing this PR.

### Related issue(s)

https://github.com/woocommerce/shipping/issues/168

1. Create a new site. If you're using Jurassic Ninja, you'll have to expose your local connect server to the internet, or wait until the PR is merged and deployed to staging.
2. Checkout this branch and create a build of this plugin, and install it on your site.
3. The plugin's name should be "WooCommerce Tax", and only the tax functionality should be available.
4. If you don't have a site that used to use WooCommerce Shipping & Tax, run `WC_Connect_Options::update_option( 'legacy_site', true );` in your site. This will mimic a site that used to use WooCommerce Shipping & Tax.
5. Now the plugin name should have changed to "WooCommerce Tax (previously WooCommerce Shipping & Tax)".

### Checklist

- [ ] unit tests


woocommerce-services.php
```diff
@@ -256,6 +256,9 @@ class WC_Connect_Loader {
 
 		public const MIGRATION_DISMISSAL_COOKIE_KEY = 'wcst-wcshipping-migration-dismissed';
 
+		public const PLUGIN_NAME_FOR_LEGACY_SITES        = 'WooCommerce Tax (previously WooCommerce Shipping & Tax)';
+		public const PLUGIN_DESCRIPTION_FOR_LEGACY_SITES = 'Description: Hosted services for WooCommerce: automated tax calculation, shipping label printing, and smoother payment setup.';
```
kallehauge:
```suggestion
		public const PLUGIN_DESCRIPTION_FOR_LEGACY_SITES = 'Hosted services for WooCommerce: automated tax calculation, shipping label printing, and smoother payment setup.';
```

woocommerce-services.php
```diff
@@ -1,9 +1,9 @@
 <?php
 /**
- * Plugin Name: WooCommerce Shipping & Tax
+ * Plugin Name: WooCommerce Tax
  * Requires Plugins: woocommerce
  * Plugin URI: https://woocommerce.com/
- * Description: Hosted services for WooCommerce: automated tax calculation, shipping label printing, and smoother payment setup.
+ * Description: Automated tax calculation for WooCommerce.
```
kallehauge, line 6: I'm sure I'm just missing context since I know you've discussed the intent of the PR with e.g. @kloon, but the vibe I got from our team call today was that we continue to name it `WooCommerce Shipping & Tax` for now - since we're not ready to update the name on WPORG - but we will dynamically call it `WooCommerce Tax` if possible.

Again, I'm sure I'm just out of the loop, but I wanted to call it out to be sure we're on the same page 😄

kallehauge (commented):
Ah! Just noticed this PR was mentioned in https://github.com/Automattic/woocommerce-connect-server/pull/2735 😛

woocommerce-services.php
Ferdev: oops!! good catch! 😊

woocommerce-services.php
Ferdev: yep, the behavior you described is what Mathias shared with us

changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.8.8 - 2025-xx-xx =
```
kallehauge, line 3: I wonder if this would be the right time to cut a 3.0.0?

kallehauge: I wonder if this would be the right time to cut a 3.0.0?

woocommerce-services.php
kallehauge: If that's the case, should we keep the plugin file as meta data as it was before until we change the name on WordPress.org as well?
There could be arguments for this change being controversial to begin with, but keeping the value matching will probably be liked more from the WordPress.org team?

We can then dynamically change the name to WooCommerce Tax (previously WooCommerce Shipping & Tax) if needed.

Also, this might have been discussed with Matthias, but we still promote the plugin as WCS&T in WP Admin's plugin search and when searching for plugins on WordPress.org. I don't think we should call it just `WooCommerce Tax` before we officially change the name but instead, always, reference it as `WooCommerce Tax (previously WooCommerce Shipping & Tax)` to be on the safe side (my concern here is negative reviews of making false promises of what the plugin is capable of)

woocommerce-services.php
```diff
@@ -256,6 +256,9 @@ class WC_Connect_Loader {
 
 		public const MIGRATION_DISMISSAL_COOKIE_KEY = 'wcst-wcshipping-migration-dismissed';
 
+		public const PLUGIN_NAME_FOR_LEGACY_SITES        = 'WooCommerce Tax (previously WooCommerce Shipping & Tax)';
+		public const PLUGIN_DESCRIPTION_FOR_LEGACY_SITES = 'Hosted services for WooCommerce: automated tax calculation, shipping label printing, and smoother payment setup.';
```
kallehauge: It's years ago that I looked it up, but I believe the plugin's php header name and description will be translated if they exist somewhere in the pot file. Should these perhaps be wrapped in `__()` to match that behaviour? 

Note: AFAIR we can't wrap constant values in a function, but It looks like we only use them in one place, so moving them to `maybe_rename_plugin()` should allow us to add the translation functions

woocommerce-services.php
```diff
@@ -879,7 +888,7 @@ public function attach_hooks() {
 			$this->paypal_ec->init();
 
 			// Only register shipping label-related logic if WC Shipping is not active.
-			if ( ! self::is_wc_shipping_activated() ) {
+			if ( WC_Connect_Options::get_option( 'legacy_site' ) ) {
```
kallehauge: Do we need to keep `! self::is_wc_shipping_activated()` in place since we still need parallel support with WC Shipping for legacy sites?

_Actually, I think this could create conflicts for legacy sites if WC Shipping is also activated since they try to register the same endpoints?_

kallehauge (commented):
Just a general comment: should we add some sort of notification for new sites explaining that they're not getting shipping in this plugin? I think it will reduce the chances of merchants getting a sour taste

I'm asking since some might have the plugin automatically installed by host(s), have it installed from the old OBW, found the plugin on WordPress.org where it's still called "Shipping" - there might be more, but that's my main gotcha's 😄

woocommerce-services.php
Ferdev: ok, sounds good André, good point, sir! I'll restore the original make and will change it dynamically, just in case :+1:

changelog.txt
Ferdev: yeah, good call... I think yes, I'll upgrade the version :+1:

woocommerce-services.php
Ferdev: good catch!! Yeah, I'll update it.

woocommerce-services.php
Ferdev: Hmmm, not 100% I'll test it out and report back.

woocommerce-services.php
```diff
@@ -256,8 +256,10 @@ class WC_Connect_Loader {
 
 		public const MIGRATION_DISMISSAL_COOKIE_KEY = 'wcst-wcshipping-migration-dismissed';
 
-		public const PLUGIN_NAME_FOR_LEGACY_SITES        = 'WooCommerce Tax (previously WooCommerce Shipping & Tax)';
-		public const PLUGIN_DESCRIPTION_FOR_LEGACY_SITES = 'Hosted services for WooCommerce: automated tax calculation, shipping label printing, and smoother payment setup.';
+		public const PLUGIN_NAME_FOR_LEGACY_SITES        = __( 'WooCommerce Tax (previously WooCommerce Shipping & Tax)', 'woocommerce-services' );
+		public const PLUGIN_DESCRIPTION_FOR_LEGACY_SITES = __( 'Description: Hosted services for WooCommerce: automated tax calculation, shipping label printing, and smoother payment setup.', 'woocommerce-services' );
+		public const PLUGIN_NAME_FOR_NEW_SITES           = __( 'WooCommerce Tax', 'woocommerce-services' );
+		public const PLUGIN_DESCRIPTION_FOR_NEW_SITES    = __( 'Description: Automated tax calculation for WooCommerce.', 'woocommerce-services' );
```
kallehauge: I'm pretty sure this will give an error because functions aren't allowed in class constants (I made a quick test here: https://onlinephp.io/c/12300).

The issue being that constants (from PHP 8.1+ at least) allows expressions but it needs to be able to be evaluated at compile time (not runtime), so e.g. concatenating two constants would work fine, but functions are runtime so it will throw something in the likes of `Fatal error: Constant expression contains invalid operations`

We should be OK if we move the values directly to `maybe_rename_plugin` and thereby bypass constants 😄

woocommerce-services.php
```diff
@@ -630,6 +635,10 @@ public function jetpack_on_plugins_loaded() {
 		}
 
 		public function on_plugins_loaded() {
+			// Initialize legacy_site option if it doesn't exist yet
+			if ( WC_Connect_Options::get_option( 'legacy_site' ) === false ) {
+				WC_Connect_Options::update_option( 'legacy_site', true );
+			}
```
kallehauge: [Non-blocking; this condition works as expected so this is just a suggestion]

From a readability point of view, then I would assume that when we call `WC_Connect_Options::update_option( 'legacy_site', false );` in `maybe_mark_as_legacy_site()`, then this condition will always override the "legacy site state" to `true`.

The reason it doesn't is because `WC_Connect_Options::get_option()` uses [WP's `get_option`](https://developer.wordpress.org/reference/functions/get_option/) under the hood and it honestly has some weird behaviours built in like, if you call `update_option( 'name', false )` then it returns booleans as empty strings for falsy values and `'1'` for truthy values when fetched with `get_option( 'name' )`:

>false returns string(0) ""
>true returns string(1) "1"
>0 returns string(1) "0"
>1 returns string(1) "1"
>'0' returns string(1) "0"
>'1' returns string(1) "1"
>null returns string(0) ""

So, this condition works since `get_option` returns an empty string here for legacy sites (the false response of `WC_Connect_Options::update_option( 'legacy_site', false )`) but you'd basically have to know the inner workings of WP's `get_option` to know that.

---

This is purely a suggestion, but what if we rename `legacy_site` to something like `only_tax` and then save a truthy value in `maybe_mark_as_legacy_site()` that way:

1. We can complete remove this condition and no longer make a DB request on all `on_plugins_loaded` calls. We would only care if it was ever stored as truthy in `maybe_mark_as_legacy_site`.
    * _We could either call `'1' === WC_Connect_Options::get_option( 'only_tax' );` where we need it or wrap it in a new function and type cast the return value to a boolean_.
2. `only_tax` (or some other name) makes it sound like we expect a truthy value and will make it clear that this option is only about enabling tax so if we ever need to deprecate another feature in the future, then we can reuse `first_install` again and not confuse it with "legacy site" because the name is more descriptive

woocommerce-services.php
kallehauge: I've confirmed that it causes fatal errors in WCS&T 😄 

```
PHP Fatal error:  Constant expression contains invalid operations in /var/www/html/wp-content/plugins/woocommerce-services/woocommerce-services.php on line 259
``` 

image

kallehauge (changes requested):

woocommerce-services.php
Ferdev: :facepalm: that's what happens when you do edits right before going to bed. Sorry André, I'll fix it right away :bow:

woocommerce-services.php
Ferdev: that's a nice suggestion, thank you sir :bow: Implementing...

kallehauge (approved):
LGTM! 👏 

I would like to add a caveat that Laura / Matthias / Gerhard should 👍 this before we merge and release it (including the v3.0.0 release I guess? 🤔)

Ferdev:
Yeah, I was planning to share this PR with them before releasing it :+1:

Don't mind the re-review request, I just clicked the request review button accidentally :man_facepalming: 

Thanks!

kallehauge (approved):
No worries! Just makes for the fastest PR approval I've ever given, @Ferdev 😄

Ferdev:
As per conversation [here](https://a8c.slack.com/archives/C04KWSNPSE5/p1740479384083049), we won't be mergin this PR until we're ready to do so.

samnajian:
@Ferdev Is this banner heading expected?
image

samnajian:
@Ferdev 
Another issue with the current approach is that as soon as I disconnect my site from WPCOM via Jetpack plugin, the plugin names reverts back `WooCommerce Shipping & Tax`

samnajian (changes requested):
It works fine for the most part but the following scenario:

- Install currently released plugin.
- Make a label purchase.
- Delete and Install the build from this PR.
- On activation see the following (it reads `WooCommerce Tax`):

- Click on connect, and see `WooCommerce Tax (previously WooCommerce Shipping & Tax)` for the plugin name on plugins listing page.

### Expected behavior:
It reads `WooCommerce Tax (previously WooCommerce Shipping & Tax)` for both the plugin name and the connect banner even before the connection.

Ferdev:
Hey @samnajian , about that issue you're mentioning... I think we shouldn't display the `(formerly WooCommerce Shipping & Tax)` part in the banner... IMHO, is too long, that's why I left it like that there... In the plugins list I think is fine, but on the banner I think the name would be way too long. @george-vice I'd love to hear your thoughts as well on this

george-vice:
> Hey @samnajian , about that issue you're mentioning... I think we shouldn't display the `(formerly WooCommerce Shipping & Tax)` part in the banner... IMHO, is too long, that's why I left it like that there... In the plugins list I think is fine, but on the banner I think the name would be way too long. @george-vice I'd love to hear your thoughts as well on this

It's fine to keep it just WooCommerce Tax in the banner.

Ferdev:
Ok, then I think we're good to go, thanks @george-vice!

Ferdev:
I've been investigating for a while, and looks like removing the translation calls in the new plugin renaming function does not fix the issue with the QIT security test. There are merged PRs in this repo which had the very same issue, so is not something this PR introduced, so I think we're safe to merge it.

classes/class-wc-connect-settings-pages.php
```diff
@@ -82,9 +82,9 @@ public function output_shipping_settings_screen() {
 
 			if ( WC_Connect_Jetpack::is_offline_mode() ) {
 				if ( WC_Connect_Jetpack::is_connected() ) {
-					$message = __( 'Note: Your site is connected but WooCommerce Shipping & Tax is configured to work in offline mode. Please disable offline mode.', 'woocommerce-services' );
+					$message = __( 'Note: Your site is connected but WooCommerce Tax is configured to work in offline mode. Please disable offline mode.', 'woocommerce-services' );
 				} else {
-					$message = __( 'Note: WooCommerce Shipping & Tax is configured to work in offline mode. This site will not be able to obtain payment methods from WooCommerce Shipping & Tax production servers.', 'woocommerce-services' );
+					$message = __( 'Note: WooCommerce Tax is configured to work in offline mode. This site will not be able to obtain payment methods from WooCommerce Shipping & Tax production servers.', 'woocommerce-services' );
```
tpaksu: There's one you missed at the end.

classes/class-wc-connect-privacy.php
```diff
@@ -33,7 +33,7 @@ public function add_privacy_message() {
 			return;
 		}
 
-		$title   = __( 'WooCommerce Shipping & Tax', 'woocommerce-services' );
+		$title   = __( 'WooCommerce Shipping', 'woocommerce-services' );
```
tpaksu: This should be Tax I think.

tpaksu (commented):
Code looks good, found two instances of "Shipping" that was missed. Other than checking the modified code, didn't do a manual test.

Ferdev:
Thanks for catching that, @tpaksu ! Just pushed a new commit to fix them.

tpaksu (approved):

woocommerce-services.php
```diff
@@ -1811,17 +1811,25 @@ public function add_plugin_action_links( $links ) {
...
 		private static function _has_any_labels_db_check() {
 			global $wpdb;
+
 			// If $wpdb is not available yet, assume no labels (safer default).
-			if ( ! is_object( $wpdb ) || ! isset( $wpdb->postmeta ) ) {
+			if ( ! is_object( $wpdb ) ) {
 				return false;
 			}
+
+			$meta_table_to_check = $wpdb->prefix . 'wc_orders_meta';
```
samnajian: Why not use `OrderUtil::get_table_for_order_meta();` from `Automattic\WooCommerce\Utilities\OrderUtil` as in [here](https://github.com/woocommerce/woocommerce-shipping/blob/trunk/src/Migration/LegacyLabelMigrator.php#L113)?

samnajian (commented):
I left a single comment.

woocommerce-services.php
samnajian: The meta table is either `$wpdb->prefix . 'wc_orders_meta'` or  `$wpdb->postmeta`

Ferdev: oh, I wasn't aware of that function. Yeah, that works nicely, thanks for letting me know!

samnajian (approved):
LGTM!
Thanks for the updates.

PR #2840, opened 2025-02-19
Add Github action to update required headers

Abdalsalaam:
### Changes proposed in this Pull Request:
Add GitHub action to update requires headers by doing theses steps :
- Automatically get the latest WooCommerce release zip,
- unzip it
- extract the Requires at least and Requires PHP headers
- compares those values against the extension's headers
- and automatically creates a PR updating those headers as well as the WC requires at least header to the WC L-2 version and assigning our team to review.

https://github.com/Automattic/woocommerce-shipping-issues/issues/286

### DIFF
--- a/.github/workflows/update-requires-headers.yml
+++ b/.github/workflows/update-requires-headers.yml
@@ -1,0 +1,251 @@
+name: Update Requires Headers
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


dustinparker (approved):
LGTM!

PR #2845, opened 2025-04-01
Pass through settings with proxy requests as well

kloon:
## Description

When we make requests to api.woocommerce.com we include key settings to help us determine certain things on the server, when making proxy requests for taxes, we cannot include these as part of the body, so instead we should include this in another way.

- Check out this branch and build the plugin
- Enable automated taxes and ensure your store is set to US origin for testing
- Put an error log and die command [here](https://github.com/Automattic/woocommerce-services/blob/147ad8ffdd9253bdefb0f77597b8ec0a324dd64d/classes/class-wc-connect-api-client.php#L449)
```
print_r( $args );
die();
```
- Now go to the store frontend and and try and checkout with a US address
- It should show the request args and as part of that it should include a new `x-woo-settings` header containing things like the wc_version, wp_version, store_guid, etc in json format.

Screenshot 2025-04-01 at 16 34 25

### Checklist

- [x] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Shipping & Tax Changelog ***
 
+= 2.8.10 - 2025-xx-xx =
+* Tweak - Improve tax tracking.
+
 = 2.8.9 - 2025-04-07 =
 * Tweak - WordPress 6.8 & WooCommerce 9.8 Compatibility.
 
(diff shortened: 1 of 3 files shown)


Ferdev (approved):
Code looks good and works as expected!
shot-2025-04-10-14-36-39

PR #2850, opened 2025-05-21
[WOOSHIP-1400] update eligibility endpoint to return label purchase ability based on latest updates

samnajian:
## Description
This PR makes the eligibility endpoint aware of the label purchase functionality deprecation ( and removal for net new users ) in v3.0.0 by doing the following:

- Define a dedicated class for eligibility endpoint.
- Add check immediately to check if label purchase functionality is supported for the shop, without needing to pass order_id.

### Related issue(s)

Fixes WOOSHOP-1400

- Create a build from this repo.
- Upload it to a fresh test site.
- Verify sending a `GET` request to `{domain}/wp-json/wc/v1/connect/label/creation_eligibility` or `{domain}/wp-json/wc/v1/connect/label/{order_id}/creation_eligibility` returns the following:
```
{
    "is_eligible": false,
    "reason": "only_tax_functionality_is_supported"
}
```
- Create another test file that has purchase shipping label using order versions of this plugin ( less than v3.0.0 ).
- Upload the build from this branch to the new test site.
- Verify receiving the following for a `GET` request to `{domain}/wp-json/wc/v1/connect/label/{order_id}/creation_eligibility`, please note that you'd need to send a request that has an `order_id` in the URL.

```
{
    "is_eligible": true
}
```
- Verify purchasing labels on this second test site works normally.

### Checklist

- [x] unit tests


classes/class-wc-rest-connect-shipping-label-eligibility-controller.php
```diff
@@ -0,0 +1,245 @@
...
+			$this->namespace,
+			'/' . $this->rest_base . '/(?P<order_id>\d+)/creation_eligibility',
+			array(
+				array(
+					'methods'             => 'GET',
+					'callback'            => array( $this, 'get_creation_eligibility' ),
+					'args'                => array(
+						'order_id' => array(
+							'required' => true,
+							'type'     => 'is_numeric',
```
Ferdev: I think this should be `integer` instead of `is_numeric`

samnajian: Addressed in d0cbf253b73aa61b76d6febbd73c4f1caa9154a3

Ferdev (approved):
Hey @samnajian , code looks good and works well. Just one question: when you perform the get request over an eligible site, without passing the `<order_id>`, it feels weird to get a `"reason": "order_not_found"` attribute. Not a show-stopper, but just sharing my POV. Anyway, this is good to go :ship:

samnajian:
> Hey @samnajian , code looks good and works well. Just one question: when you perform the get request over an eligible site, without passing the <order_id>, it feels weird to get a "reason": "order_not_found" attribute. Not a show-stopper, but just sharing my POV. Anyway, this is good to go 🚢


To provide support for zero-change on mobile apps, the same endpoint is now supporting the check for label purchase functionality deprecation which actually doesn't need an order, and the `order_not_found` only happens if the shop supports label purchasing, which is probably when the app is checking for an order's eligibility where they should be passing order id. 

That said, I agree it's super optimal. But anyways the mobile app's are checking it for an order already, which should be fine.

PR #2856, opened 2025-06-25
Disable shipping features unless WooCommerce Shipping is active

PKMN-World:
## Description
This pull request disables the legacy shipping features in WooCommerce Tax unless the WooCommerce Shipping plugin is active.

After installing the WooCommerce Tax plugin, I noticed the shipping meta boxes cause a number of API requests to run on every load of the Edit Order page, amounting to a 2-4 second impact of load time, even when these boxes are set to hidden in the UI.

#### API Calls of the Edit Order page after a typical install of this plugin:

Screenshot 2025-06-24 at 11 21 42 PM

#### API Calls of the Edit Order page after the changes in this pull request:

Screenshot 2025-06-24 at 11 20 00 PM


Ferdev:
hey @PKMN-World , thank you so much for the pull request!. I've been checking the code and the functionality, and it looks fine 👌 But I have a suggestion to share. I think we should actually just remove the new setting you added here, and instead, the check this PR is doing to avoid loading the metaboxes, would check if the site has shipping disabled. What do you think?

Looking forward to hearing your thoughts on this.

PKMN-World:
Hi @Ferdev, thanks for the reply! What's the setting to check if the site has shipping disabled? Is that a global WooCommerce setting?

Just to add context: in my particular use case, my shop has shipping enabled: customers place orders and I ship them. However, I use a 3rd party shipping label service that fetches order data through the WooCommerce API. In short: I do ship orders, but don't need the shipping features of the WooCommerce Tax plugin.

Ferdev:
You can check the code [here](https://github.com/PKMN-World/woocommerce-services/blob/disable-legacy-shipping-features/woocommerce-services.php#L1021)

When `self::is_wc_shipping_activated()` returns `true` you can be sure WooCommerce Tax will have it's legacy shipping functionality disabled. You should be able to use that method anywhere in your code, as is public.

Please, let me know if you have any other questions!

PKMN-World:
Ahh okay. I missed that you were talking about the WooCommerce Shipping plugin. I understand now! Updated the code and PR description above, @Ferdev

Ferdev (approved):
Hey @PKMN-World , apologies for the delay, my days have been pretty packed lately.

This looks good, and it's a fine addition :heart: Thanks for your contribution!

PKMN-World:
@Ferdev Thanks! Any advice for the failing security test here?

PKMN-World:
@samnajian Any advice for the failing security test here?

Ferdev:
Hey, sorry @PKMN-World, missed your comment. I forgot to mention, I checked the security fail with our team in charge of that part, and the problem is that test requires some credentials only available to our partners. So it's not going to be possible to run it.

The only way I can think of is creating a new branch in our repo, and use that branch as the base branch. That way, we can create a new PR in our end to run all security checks.

PKMN-World:
@Ferdev @samnajian It doesn't look like I have permission to create a branch in this repo

PR #2861, opened 2025-07-08
Update Requires headers for WooCommerce compatibility

WooShippingBot:
This PR updates plugin headers to ensure compatibility with the latest WooCommerce release.

Please test the plugin and merge.


dustinparker:
@Abdalsalaam it looks like we need to adjust this to account for the WordPress version required by the minimum supported WooCommerce version.

PR #2870, opened 2025-07-28
chore: add changelog entries

ayushpahwa:
## Description

Adding changelog entries for release process

### Checklist

- [ ] unit tests


Ferdev (approved):
Actually, I'd rename this changelog entry to something in the lines of: Improves performance when WooCommerce Shipping is not active.

So what this patch actually does is saving several shippping-related requests that were made even when the WooCommerce Shipping was not installed. Sorry, I should've explained this beforehand 😄

PR #2873, opened 2025-08-07
[WOOSHIP-1541] chore: fix failing e2e test in wc services repo

ayushpahwa:
## Description

This PR fixes a regression introduced in #2856 where shipping label functionality was incorrectly blocked for merchants who should have access to it. The previous implementation checked if WooCommerce Shipping plugin was installed, but this breaks functionality for grandfathered merchants with existing shipping label. The fix replaces the `is_wc_shipping_activated()` check with `has_only_tax_functionality()`, which properly determines whether a merchant should have shipping features based on their configuration and existing labels.

### Related issue(s)

Fixes [#WOOSHIP-1541](https://linear.app/a8c/issue/WOOSHIP-1541/fix-failing-e2e-test-in-wc-services-repo)

1. Verify the [e2e test result](https://github.com/Automattic/woocommerce-services/actions/runs/16569101401/job/46856335042) on the `trunk` of the repo are failing. 
2. Verify the same test's results in the PR that it passes.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 3.0.8 - 2025-xx-xx =
 * Fix   - Improves performance when WooCommerce Shipping is not active.
+* Fix   - Restore shipping label functionality for merchants with shipping features enabled.
 
 = 3.0.7 - 2025-07-21 =
 * Fix   - Missing release files.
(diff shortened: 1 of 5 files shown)


tpaksu (approved):
The logic seems fine. If that was a PHPUnit test, I'd prefer to add the filter before the failing test, and remove it after the test is done, but on E2E, I guess this is not an option. I wonder if it'll affect the tests where they should test `has_only_tax_functionality = true` cases. Anyway, I'm approving it since there's no easy way to set it in a different way. Thanks for working on this! 🚢 it!

PR #2877, opened 2025-08-13
Implement migration survey for WOOSHIP-1511

Ferdev:
## Description

  This PR implements WOOSHIP-1511: a migration survey to understand why WCS&T users haven't migrated to WooCommerce Shipping. The survey appears after users complete a shipping label
  purchase to capture feedback on migration blockers.

  ### Related issue(s)

  Implements WOOSHIP-1511: https://linear.app/a8c/issue/WOOSHIP-1511/implement-survey-in-wcsandt-for-migration-blockers

  ### Key Features Implemented

  - **Backend Survey Logic**: PHP class handling survey eligibility, display limits, and tracking
  - **Frontend Modal Component**: React modal with primary/followup questions matching Linear spec
  - **Post-Purchase Timing**: Survey displays after label purchase completion (not during)
  - **Display Control**: Maximum 2 displays per user with 3-day cooldown period
  - **Analytics Integration**: Tracks survey events via WooCommerce Services Tracks
  - **Responsive Design**: Mobile-friendly modal with proper styling

  ### Survey Flow

  1. User completes shipping label purchase
  2. Survey appears after purchase dialog closes (500ms delay for print dialogs)
  3. Primary question with 7 predefined options
  4. Conditional followup textarea based on selection
  5. Submit or skip options with proper tracking
  6. Confirmation message with migration guidance

  ### Display Logic

  - **Eligibility**: Only shown if WooCommerce Shipping is not active
  - **Frequency**: Maximum 2 times per user with 3-day cooldown
  - **Completion**: Never shows again after user submits feedback
  - **Testing**: `?force_survey=1` parameter for development/testing

  **Testing Steps:**
  1. Go to WP Admin → Orders → Edit any order
  2. Add `?force_survey=1` to URL to force survey display
  3. Create a shipping label (or use existing label creation flow)
  4. Complete the purchase process
  5. Survey modal should appear after purchase completion
  6. Test both primary questions and followup responses
  7. Verify tracking in browser dev tools (AJAX calls to track display/submission)

  **Track Events to Monitor:**

  When testing, verify these events are properly fired in browser dev tools (Network tab → XHR requests) or analytics dashboard:

  1. **`migration_survey_survey_displayed`**
     - Fired when: Survey modal appears to user
     - Properties: None
     - Expected: Once per survey display

  2. **`migration_survey_survey_submitted`**
     - Fired when: User completes and submits survey
     - Properties:
       - `primary_reason`: Selected option from dropdown
       - `followup_*`: Any followup question responses (prefixed with `followup_`)
     - Expected: Once per successful submission

  3. **`migration_survey_survey_dismissed`**
     - Fired when: User clicks "Skip" or closes survey
     - Properties: None
     - Expected: Once per dismissal action

  ### Checklist

  - [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-migration-survey.php
+++ b/classes/class-wc-connect-migration-survey.php
@@ -135,7 +135,7 @@ if ( ! class_exists( 'WC_Connect_Migration_Survey' ) ) {
 				}
 			}
 
-			$this->track_event( 'survey_submitted', $track_properties );
+			$this->track_event( 'submitted', $track_properties );
 
 			wp_send_json_success();
 		}
@@ -150,7 +150,7 @@ if ( ! class_exists( 'WC_Connect_Migration_Survey' ) ) {
 			}
 
 			// Track dismissal event
-			$this->track_event( 'survey_dismissed', array() );
+			$this->track_event( 'dismissed', array() );
 
 			wp_send_json_success();
 		}
@@ -171,7 +171,7 @@ if ( ! class_exists( 'WC_Connect_Migration_Survey' ) ) {
 			update_user_meta( $user_id, 'wcs_migration_survey_last_shown', time() );
 
 			// Track survey display event
-			$this->track_event( 'survey_displayed', array() );
+			$this->track_event( 'displayed', array() );
 
 			wp_send_json_success();
 		}
--- a/classes/class-wc-connect-migration-survey.php
+++ b/classes/class-wc-connect-migration-survey.php
@@ -6,6 +6,8 @@
  * who haven't migrated from WCS&T to WooCommerce Shipping
  */
 
+use Automattic\WCServices\Utils;
+
 if ( ! class_exists( 'WC_Connect_Migration_Survey' ) ) {
 
 	class WC_Connect_Migration_Survey {
@@ -102,7 +104,7 @@ if ( ! class_exists( 'WC_Connect_Migration_Survey' ) ) {
 		 */
 		public function handle_survey_submission() {
 			// Verify nonce
-			if ( ! wp_verify_nonce( $_POST['nonce'], 'wcs_migration_survey' ) ) {
+			if ( ! wp_verify_nonce( Utils::get_sanitized_request_data( 'nonce' ), 'wcs_migration_survey' ) ) {
 				wp_die( 'Security check failed' );
 			}
 
@@ -143,7 +145,7 @@ if ( ! class_exists( 'WC_Connect_Migration_Survey' ) ) {
 		 */
 		public function handle_survey_dismissal() {
 			// Verify nonce
-			if ( ! wp_verify_nonce( $_POST['nonce'], 'wcs_migration_survey' ) ) {
+			if ( ! wp_verify_nonce( Utils::get_sanitized_request_data( 'nonce' ), 'wcs_migration_survey' ) ) {
 				wp_die( 'Security check failed' );
 			}
 
@@ -158,7 +160,7 @@ if ( ! class_exists( 'WC_Connect_Migration_Survey' ) ) {
 		 */
 		public function handle_survey_display_tracking() {
 			// Verify nonce
-			if ( ! wp_verify_nonce( $_POST['nonce'], 'wcs_migration_survey' ) ) {
+			if ( ! wp_verify_nonce( Utils::get_sanitized_request_data( 'nonce' ), 'wcs_migration_survey' ) ) {
 				wp_die( 'Security check failed' );
 			}
 
--- a/classes/class-wc-connect-migration-survey.php
+++ b/classes/class-wc-connect-migration-survey.php
@@ -1,0 +1,195 @@
+<?php
+/**
+ * Migration Survey Class
+ * 
+ * Handles the display and submission of migration survey for users
+ * who haven't migrated from WCS&T to WooCommerce Shipping
+ */
+
+if ( ! class_exists( 'WC_Connect_Migration_Survey' ) ) {
+
+	class WC_Connect_Migration_Survey {
+
+
+		/**
+		 * Maximum number of times to show the survey
+		 */
+		const MAX_DISPLAY_COUNT = 2;
+
+		/**
+		 * Cooldown period between survey displays (in seconds)
+		 */
+		const COOLDOWN_PERIOD = 259200; // 3 days
+
+		/**
+		 * Initialize the migration survey
+		 */
+		public function __construct() {
+			// Hook into script enqueue action
+			add_action( 'enqueue_wc_connect_script', array( $this, 'add_survey_data_to_script' ), 10, 2 );
+			
+			// AJAX handlers
+			add_action( 'wp_ajax_wcs_migration_survey_submit', array( $this, 'handle_survey_submission' ) );
+			add_action( 'wp_ajax_wcs_migration_survey_dismiss', array( $this, 'handle_survey_dismissal' ) );
+			add_action( 'wp_ajax_wcs_migration_survey_track_display', array( $this, 'handle_survey_display_tracking' ) );
+		}
+
+		/**
+		 * Check if survey should be shown to current user
+		 */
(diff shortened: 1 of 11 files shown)


woocommerce-services.php
```diff
@@ -873,9 +873,13 @@ public function load_dependencies() {
 			$tracks                 = new WC_Connect_Tracks( $logger, __FILE__ );
 			$shipping_label         = new WC_Connect_Shipping_Label( $api_client, $settings_store, $schemas_store, $payment_methods_store );
 			$nux                    = new WC_Connect_Nux( $tracks, $shipping_label );
-			$store_notices_notifier = new StoreNoticesNotifier( $taxes_logger->is_debug_enabled() );
+			$store_notices_notifier = class_exists( 'Automattic\WCServices\StoreNotices\StoreNoticesNotifier' )
```
ismaeldcom: Why do we need to check if the class exist? 🤔

classes/class-wc-connect-shipping-label.php
```diff
@@ -529,10 +529,6 @@ private function calculate_should_show_meta_box( $post ) {
 		 * @return bool True if shipping label is enabled from the settings.
 		 */
 		public function is_shipping_label_enabled() {
-			// For tax-only installations, shipping labels are disabled
```
ismaeldcom, line 532: Any reason to remove this? 🤔

classes/class-wc-connect-migration-survey.php
```diff
@@ -0,0 +1,195 @@
...
+			$this->track_event( 'survey_displayed', array() );
+
+			wp_send_json_success();
+		}
+
+		/**
+		 * Track survey events
+		 */
+		private function track_event( $event_name, $properties = array() ) {
+			if ( class_exists( 'WC_Connect_Tracks' ) && class_exists( 'WC_Connect_Logger' ) && function_exists( 'wc_get_logger' ) ) {
```
ismaeldcom: This is just an example, but I can see Claude being overly cautious, adding unnecessary checks, and making things far more complex than necessary.

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/index.js
```diff
@@ -26,45 +27,118 @@ import FeatureAnnouncement from 'components/migration/feature-announcement';
...
 
+	// Track when purchase dialog is shown
+	React.useEffect( () => {
+		if ( loaded && showPurchaseDialog ) {
+			setPurchaseDialogWasShown( true );
+		}
+	}, [ loaded, showPurchaseDialog ] );
+
+	// Show survey when purchase dialog is closed (after it was actually shown)
+	React.useEffect( () => {
```
ismaeldcom: Could we simplify this be just showing the survey using the `onClose` handler?

client/extensions/woocommerce/woocommerce-services/views/shipping-label/label-purchase-modal/index.js
```diff
@@ -26,45 +27,118 @@ import FeatureAnnouncement from 'components/migration/feature-announcement';
...
-					orderId={props.orderId}
-				/>
-				<PackagesStep siteId={props.siteId} orderId={props.orderId}/>
-				{props.isCustomsFormRequired && (<CustomsStep siteId={props.siteId} orderId={props.orderId}/>)}
-				<RatesStep siteId={props.siteId} orderId={props.orderId}/>
-			</div>
-			<Sidebar siteId={props.siteId} orderId={props.orderId}/>
-		</div>
-	</Modal>
+	const trackSurveyDisplay = async () => {
```
ismaeldcom: I think `trackSurveyDisplay` and `handleSurveySubmit` should be in the `MigrationSurveyModal` not here.

classes/class-wc-connect-migration-survey.php
```diff
@@ -0,0 +1,195 @@
...
+				try {
+					$wc_logger = wc_get_logger();
+					$logger = new WC_Connect_Logger( $wc_logger );
+					$tracks = new WC_Connect_Tracks( $logger, WCSERVICES_PLUGIN_FILE );
+					$tracks->record_user_event( 'migration_survey_' . $event_name, $properties );
```
ismaeldcom: Nitpick: We're including `survey_survey` on every event name.

ismaeldcom (commented):
Thanks for taking care of this. The solution works, and the survey looks good. However, I think the implementation could be improved.

I'm not sure if we intend to limit our time commitment to this extension as much as possible, but using Claude to take care of it results in _questionable quality_ 🙂

Apart from the review comments, I think it would be simpler to use `recordTracksEvent` on the JS side, rather than making AJAX calls with nonce validation to track events.  Also, the SCSS doesn't use any existing variables or styles.

classes/class-wc-connect-shipping-label.php
Ferdev: yeah, this was avoiding old users who were allowed to use the shipping functionality to keep using it. It's a bug introduced in latest version.

Ferdev:
Hey @ismaeldcom, I think i've addressed most of your suggestions. I was kind of in a hurry when I implemented this, and with my AFK closing by, didn't have much time to test it properly. I guess somebody else will take care of this while I'm away. But maybe, in the meantime, you can give it another look.

samnajian:
@ismaeldcom I've addressed QIT security issues, do you think we can merge this PR as is and create a follow up issue to clean up?

ismaeldcom:
>do you think we can merge this PR as is and create a follow up issue to clean up?

We can, although I don't think we should invest more time in something like this.

The code looks better now. However, I still think the display logic is a bit too complicated, and the survey feels off because of the use of custom styles instead of WP/WC components.


https://github.com/user-attachments/assets/ed75f568-d160-4168-af9e-07a1fb7f1dd3

ismaeldcom (commented):
>do you think we can merge this PR as is and create a follow up issue to clean up?

We can, although I don't think we should invest more time in something like this.

The code looks a bit better now. However, I still think the display logic is a bit too complicated, and the survey feels off because of the use of custom styles instead of WP/WC components.

@samnajian could you confirm that the events are being registered? It was working for me yesterday, but not today, not sure if it's due to my network issues 🤔 


https://github.com/user-attachments/assets/ed75f568-d160-4168-af9e-07a1fb7f1dd3

samnajian:
> @samnajian could you confirm that the events are being registered? It was working for me yesterday, but not today, not sure if it's due to my network issues 🤔


@ismaeldcom Only the `display` event triggers a separate tracking ajax call, the rest are tracked at the end of the method that handles the main action request, see https://github.com/Automattic/woocommerce-services/pull/2877/files#diff-57a181d83d0becfa757e1cc48709f7a14c6e9209bdbb6ccc2e9b8889faacbef7R138

ismaeldcom:
> Only the display event triggers a separate tracking ajax call, the rest are tracked at the end of the method that handles the main action request

I know, I meant if you're seeing them on [Tracks](https://mc.a8c.com/tracks/live/), because I don't.

samnajian:
> I know, I meant if you're seeing them on [Tracks](https://mc.a8c.com/tracks/live/), because I don't.

This is what I see 
image

However the `dismissal` is recorded as `submitted` as well!

ismaeldcom:
Thanks for checking it. The proxy doesn't work when I'm tethering.

Looks like in the last commit, [skipped became submitted with `primary_reason: skipped`](https://github.com/Automattic/woocommerce-services/blob/9132869a79293ee33439a4b9c005fb9e5e238627/client/components/migration-survey/modal.js#L135).

samnajian:
@ismaeldcom I just addressed some concerns, updated the button styling to match the plugin's design and removed redundancy in Tracks even names, please give it another check.
image
image

ismaeldcom (approved):
Now it works as expected, and looks closer to the rest of the UI.

PR #2879, opened 2025-08-18
Fix WOOSHIP-1535 - Don't load shipping settings if WC shipping or tax-only is enabled

Abdalsalaam:
## Description

* Fix   - Exclude shipping-related admin components when shipping functionality is disabled.

### Related issue(s)
[WOOSHIP-1535](https://linear.app/a8c/issue/WOOSHIP-1535/hide-print-feature-and-woo-shipping-settings-when-only-tax-flag-is-set)

1. Install and configure the current **WooCommerce Tax** plugin release along with **WooCommerce Shipping**.
2. Go to **WooCommerce > Status > WooCommerce Tax** then you will find **Print functionality** still
4. Update the plugin with this branch and re-do the previous step.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Tax Changelog ***
 
+= 3.1.1 - 2025-xx-xx =
+* Fix   - Exclude shipping-related admin components when shipping functionality is disabled.
+
 = 3.1.0 - 2025-09-16 =
 * Add   - Increase cache time for address validation errors.
 * Tweak - WooCommerce 10.2 Compatibility.
@@ -12,7 +15,6 @@
 
 = 3.0.9 - 2025-08-26 =
 * Add   - Migration survey to understand WooCommerce Shipping adoption blockers.
-* Fix   - Exclude shipping-related admin components when shipping functionality is disabled.
 
 = 3.0.8 - 2025-08-13 =
 * Fix   - Improves performance when WooCommerce Shipping is not active.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Tax Changelog ***
 
+= 3.0.9 - 2025-xx-xx =
+* Fix   - Exclude shipping-related admin components when shipping functionality is disabled.
+
 = 3.0.8 - 2025-08-13 =
 * Fix   - Improves performance when WooCommerce Shipping is not active.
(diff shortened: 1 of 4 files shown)


dustinparker:
@Abdalsalaam I'm still seeing the WooCommerce Shipping settings on a brand new WP site. These settings should not show if the merchant wasn't "grandfathered in" and doesn't have the WC Shipping plugin installed:

image

Abdalsalaam:
> @Abdalsalaam I'm still seeing the WooCommerce Shipping settings on a brand new WP site. These settings should not show if the merchant wasn't "grandfathered in" and doesn't have the WC Shipping plugin installed:
> image

I noticed something with new installs connected to a fresh WooCommerce account:
- Once the site is connected, the option `wc_connect_only_tax` is set to no. ( I want to check if this value is set by connect server ).
- From my understanding, this should probably be 1 for new installs/accounts as well.
- I tested multiple accounts and saw mixed results, so I’m still running more tests to confirm.

In addition, I found that the following banner is added by WooCommerce core, which effectively hides the WooCommerce Tax shipping meta box:
https://github.com/woocommerce/woocommerce/blob/b5640b450b21468a619e57f66298d2409426b635/plugins/woocommerce/client/admin/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js#L248
image

However, if you install the FedEx plugin first and then enable WC Tax, the shipping features are enabled, and a different banner (added by the WC Tax plugin) is displayed instead:
image


This behavior feels inconsistent. I think we may need to review how new installs are flagged and ensure the banners and the `wc_connect_only_tax` option align with the intended defaults.

====
note : this [PR](https://github.com/Automattic/woocommerce-services/pull/2883) related to the same issue

dustinparker:
Hey @Ferdev! :wave: 
Would you be able to provide some insight regarding how this should work on new installs? https://github.com/Automattic/woocommerce-services/pull/2879#issuecomment-3315900265

Ferdev:
hey @dustinparker ! yeah, sure, allow me to elaborate. When a new user installs this plugin in their site for the first time, the connect server checks whether or not this is the first time the user installed the plugin in their site. It returns that info to this plugin through the `features.first_install` flag in the services response. If it is the firs time it's been installed, the shipping part of this plugin shouldn't be loaded, but should allow using the WooCommerce Shipping plugin if it's installed.

dustinparker (approved):
This fixed the issue. LGTM!

PR #2881, opened 2025-08-21
[WOOSHIP-1511] chore: add active WC shipping plugin version to CS API calls

ayushpahwa:
## Description

Add the plugin version of WC shipping to the API call made to connect server. This will inform the connect server if WC shipping is also active along with the current plugin.

### Related issue(s)
Fixes [#WOOSHIP-1511](https://linear.app/a8c/issue/WOOSHIP-1511/implement-survey-in-wcsandt-for-migration-blockers)

1. Checkout locally and ensure both WC shipping and services are installed and activated. 
2. Ensure connect server is running locally and it's logs are accessible.
3. Add this console statement in any request handler. For eg: in `lib/services/index.js` in the function `validationHandler`, add `console.log("Request: ", JSON.stringify(request.payload, null, 2));`
4. Now navigate to `woocommerce>orders` page and notice the logs. It should show values for both `wcs_version` and `wcshipping_version` in the `settings` object.
5. Now go to `Installed plugins` and deactivate WC Shipping. Navigate back to `woocommerce>orders` page and notice that only value for both `wcs_version` is valid and `wcshipping_version` will be coming as `undefined`.

PS: ensure to test the similar PR in WC shipping for effective review.

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-api-client.php
+++ b/classes/class-wc-connect-api-client.php
@@ -468,6 +468,7 @@ if ( ! class_exists( 'WC_Connect_API_Client' ) ) {
 				'dimension_unit'       => strtolower( get_option( 'woocommerce_dimension_unit' ) ),
 				'weight_unit'          => strtolower( get_option( 'woocommerce_weight_unit' ) ),
 				'wcs_version'          => WC_Connect_Loader::get_wcs_version(),
+				'wcshipping_version'   => WC_Connect_Loader::get_wc_shipping_version(),
 				'jetpack_version'      => 'embed-' . WC_Connect_Jetpack::get_jetpack_connection_package_version(),
 				'is_atomic'            => WC_Connect_Jetpack::is_atomic_site(),
 				'wc_version'           => WC()->version,
--- a/woocommerce-services.php
+++ b/woocommerce-services.php
@@ -2108,6 +2108,23 @@ if ( ! class_exists( 'WC_Connect_Loader' ) ) {
 		}
 
 		/**
+		 * Get the version of WooCommerce Shipping plugin
+		 *
+		 * @return string
+		 */
+		public static function get_wc_shipping_version() {
+			if ( ! self::is_wc_shipping_activated() ) {
+				return 'undefined';
+			}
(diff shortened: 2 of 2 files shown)


classes/class-wc-connect-api-client.php
```diff
@@ -468,6 +468,7 @@ protected function get_settings_values() {
 				'dimension_unit'       => strtolower( get_option( 'woocommerce_dimension_unit' ) ),
 				'weight_unit'          => strtolower( get_option( 'woocommerce_weight_unit' ) ),
 				'wcs_version'          => WC_Connect_Loader::get_wcs_version(),
+				'wcshipping_version'   => WC_Connect_Loader::get_wc_shipping_version(),
```
samnajian, line 471: I guess it's worth checking how currently `wcshipping_version` is used on connect-server side, I guess in some cases it's used to determine if the request is coming from WC Shipping or not. 
Maybe simply passing `active_plugins => get_option( 'active_plugins' ) ` or `get_plugins(with a plugin's folder as argument)` would be a safer alternative.

ayushpahwa: I would actually argue to remove this ambiguous way. We should send the `plugin_source` key along with the API and if it is defined, that's the one we should be using. What do you think?
CleanShot 2025-08-22 at 20 40 13@2x

samnajian: We'd anyways need the version as well, `wcshipping_version` currently has 2 use cases:
- The actual version number of the plugin sending the request.
- The existence of `wcshipping_version`, implies request is coming from `wcshipping_version`
While the above is not perfect, we shouldn't refactor it at this stage.

ayushpahwa: The idea was not to remove the wcshipping_version but to add a better plugin source check. I worked on a draft over the weekend for this but it needs a lot of test fixes. I have pushed it [here](https://github.com/Automattic/woocommerce-connect-server/pull/2969) and will fix in the cooldown week. Till then I guess both of these PRs can be parked.

samnajian:
@ayushpahwa do you want to keep this PR or should we close it?

PR #2889, opened 2025-09-24
Add sanity check before using returned value from WC_Tax::find_rates() method.

bartech:
## Description

Add sanity check before using returned value from WC_Tax::find_rates() method.
Although method definition in docblock says this method returns `array` there is a filter in this method `woocommerce_find_rates` right before it returns and some 3rd party code can break the type contract.

### Related issue(s)

https://linear.app/a8c/issue/WOOSHIP-1625/fatal-error-in-woocommerce-tax-taxjar-integration-conflict-with-square

You can follow instruction described in the issue.
https://linear.app/a8c/issue/WOOSHIP-1625/fatal-error-in-woocommerce-tax-taxjar-integration-conflict-with-square
or simulate the issue by adding snippet like.
```php
add_filter( 'woocommerce_find_rates', '__return_false' );
```

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 3.1.1 - 2025-xx-xx =
 * Fix   - Incorrect tax rate saved in Woo Tax Table when Cart total is 0.
+* Fix   - Compatibility issue with plugins and themes that use woocommerce_find_rates filter.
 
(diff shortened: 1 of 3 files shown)


iyut (approved):
LGTM!

PR #2895, opened 2025-11-05
Fix/issue WOOSHIP-1717 - Change priority for print notices method

iyut:
## Description
Fix the error notices for being displayed too late (Currently the error will be displayed after the value is correct). It's fixed by changing the priority for `StoreNoticesController::maybe_display_notices()`.

### Related issue(s)

Closes [WOOSHIP-1717](https://linear.app/a8c/issue/WOOSHIP-1717/invalid-zip-code-error-appears-when-re-entering-valid-postcode-after#comment-15c8a563).

1. Go to checkout page with items in cart.
2. Enter a valid US ZIP code in the billing postcode field (e.g., 94539).
3. Wait for AJAX validation to complete - no error appears (correct behavior).
4. Delete the last digit from the postcode field to make it invalid (e.g., change 94539 to 9453).
5. The error will appear.
6. Add the deleted digit back to restore the valid postcode (e.g., 9453 → 94539)
7. The error will disappear

Observe the checkout page

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1628,9 +1628,9 @@ class WC_Connect_TaxJar_Integration {
 			$log_suffix    = 'in cached response.';
 		}
 
-		$to_state   = $response_body->tax->jurisdictions->state;
-		$to_country = $response_body->tax->jurisdictions->country;
-		$has_nexus  = $response_body->tax->has_nexus;
+		$to_state   = isset( $response_body->tax->jurisdictions->state ) ? $response_body->tax->jurisdictions->state : '';
+		$to_country = isset( $response_body->tax->jurisdictions->country ) ? $response_body->tax->jurisdictions->country : '';
+		$has_nexus  = isset( $response_body->tax->has_nexus ) ? $response_body->tax->has_nexus : false;
 
 		if ( 'CA' === $to_state && 'US' === $to_country && true === $has_nexus ) {
 			$this->_log(
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1646,7 +1646,7 @@ class WC_Connect_TaxJar_Integration {
 		}
 
 		if ( 'not set' === $to_state || 'not set' === $to_country || null === $has_nexus ) {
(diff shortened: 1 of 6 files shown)


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1628,9 +1628,9 @@ private function check_for_incorrect_california_tax_nexus( $response_body, $cach
 			$log_suffix    = 'in cached response.';
 		}
 
-		$to_state   = $response_body->tax->jurisdictions->state;
-		$to_country = $response_body->tax->jurisdictions->country;
-		$has_nexus  = $response_body->tax->has_nexus;
+		$to_state   = isset( $response_body->tax->jurisdictions->state ) ? $response_body->tax->jurisdictions->state : '';
+		$to_country = isset( $response_body->tax->jurisdictions->country ) ? $response_body->tax->jurisdictions->country : '';
+		$has_nexus  = isset( $response_body->tax->has_nexus ) ? $response_body->tax->has_nexus : false;
```
bartech: What do you think if we do this
```suggestion
		$to_state   = isset( $response_body->tax->jurisdictions->state ) ? $response_body->tax->jurisdictions->state : 'not set';
		$to_country = isset( $response_body->tax->jurisdictions->country ) ? $response_body->tax->jurisdictions->country : 'not set';
		$has_nexus  = isset( $response_body->tax->has_nexus ) ? $response_body->tax->has_nexus : null;
```

classes/class-wc-connect-taxjar-integration.php
bartech: And here we do 
```php
json_encode( $has_nexus ) // stringify true|false|null to 'true'|'false'|'null'.
```
This way we will know in logs when `$response_body->tax->has_nexus` is not set.

classes/class-wc-connect-taxjar-integration.php
iyut: yeah I think we can. And also we can elevate this further by adding `throw` if some of those are not set. Because this method is wrapped by `try..catch`. WDYT?

iyut: @bartech , can you check the latest commit on this PR? I have added your suggestion code and add throw exception code.

bartech (approved):
Looks good.

dustinparker (approved):
LGTM!

PR #2883, opened 2025-08-28
Fix regression causing slow loading times, loading shipping infrastructure when shipping is not enabled

PKMN-World:
## Description
This pull request fixes a regression introduced in https://github.com/Automattic/woocommerce-services/pull/2877 that allowed the shipping infrastructure to load and cause slow page load times in the admin panel when the shipping features of this plugin are not used. 

The original fix was introduced in https://github.com/Automattic/woocommerce-services/pull/2856, and then modified slightly in https://github.com/Automattic/woocommerce-services/pull/2873.

### Checklist

- [ ] unit tests

- [ ] `changelog.txt` entry added
- [ ] `readme.txt` entry added

PR #2897, opened 2025-11-08
Don't load shipping settings if WC shipping or tax-only is enabled

PKMN-World:
Follow on from https://github.com/Automattic/woocommerce-services/pull/2883

https://github.com/Automattic/woocommerce-services/pull/2879 did not fix the issue.

Screenshot 2025-11-08 at 10 35 37 AM

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,11 +1,13 @@
 *** WooCommerce Tax Changelog ***
 
+= 3.2.3 - 2025-xx-xx =
+* Fix   - Resolved issue where shipping features were incorrectly loading when in tax-only mode for some installations.
+
 = 3.2.2 - 2025-11-10 =
 * Add   - Allow round tax at subtotal level, instead of rounding per line.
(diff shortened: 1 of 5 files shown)


PKMN-World:
@Abdalsalaam

PKMN-World:
Also, the GitHub checks always fail on my pull requests. It doesn't seem like this repo is configured correctly to allow drop-in pull requests like this.

Abdalsalaam:
Thank you, @PKMN-World !

It seems there’s another scenarios for the only_tax mode that I wasn’t aware of. To address this, I’ve made some changes to let `has_only_tax_functionality` determine if it’s only_tax, instead of relying solely on the only_tax flag to fix my previous work.

dustinparker:
@PKMN-World thank you for your submission! 

The QIT tests are failing because the branch is remote. I've pushed the changes to a local branch and created a PR so that the QIT tests will pass here: https://github.com/Automattic/woocommerce-services/pull/2898. We will continue our review there.

PR #2898, opened 2025-11-10
Don't load shipping settings if WC shipping or tax-only is enabled (QIT fix attempt)

dustinparker:
## Description

This is a duplicate of [this existing PR](https://github.com/Automattic/woocommerce-services/pull/2897) which is failing on QIT authentication. I am seeing if the authentication will still fail from a local branch or if the remote branch is the cause of the auth failure.

### Checklist

- [ ] unit tests


Abdalsalaam (approved):
LGTM
Thank you!

PR #2900, opened 2025-11-30
Don't load shipping settings if WC shipping or tax-only is enabled

PKMN-World:
Follow on from https://github.com/Automattic/woocommerce-services/pull/2898.

My original change in `class-wc-connect-shipping-label.php` [was deleted in that pull request](https://github.com/Automattic/woocommerce-services/pull/2897/commits/f2b2fe4f655f4f7fcaba56f11867f98c953389c2#diff-d81ca9b0ed185e03f6de2b90adcc4a8220bf6ac280715c5da04a719d958b7669L533), which removed the fix.

Screenshot 2025-11-29 at 8 25 53 PM

### DIFF
--- a/classes/class-wc-connect-shipping-label.php
+++ b/classes/class-wc-connect-shipping-label.php
@@ -529,6 +529,12 @@ if ( ! class_exists( 'WC_Connect_Shipping_Label' ) ) {
 		 * @return bool True if shipping label is enabled from the settings.
 		 */
 		public function is_shipping_label_enabled() {
+
+			// For tax-only installations, shipping labels are disabled
+			if ( ! WC_Connect_Loader::is_wc_shipping_activated() && ! WC_Connect_Loader::has_only_tax_functionality() ) {
+				return false;
+			}
+
 			$account_settings = $this->account_settings->get();
 
 			if ( isset( $account_settings['formData']['enabled'] ) && is_bool( $account_settings['formData']['enabled'] ) ) {


PKMN-World:
@Abdalsalaam @Ferdev @dustinparker

Abdalsalaam:
Hello @PKMN-World 
it seems to me your store does not have `tax_only = true`, Can you try to add this filter to the functions.php file then check if the widgets/labels functionality will be initiated or not
```php
add_filter( 'wc_connect_has_only_tax_functionality', '__return_true' );
```

PKMN-World:
> Hello @PKMN-World it seems to me your store does not have `tax_only = true`, Can you try to add this filter to the functions.php file then check if the widgets/labels functionality will be initiated or not
> 
> ```
> add_filter( 'wc_connect_has_only_tax_functionality', '__return_true' );
> ```

Interesting! This does fix the issue for me. @Abdalsalaam What's causing that? I don't have the Shipping plugin installed.

Abdalsalaam:
> > Hello @PKMN-World it seems to me your store does not have `tax_only = true`, Can you try to add this filter to the functions.php file then check if the widgets/labels functionality will be initiated or not
> > ```
> > add_filter( 'wc_connect_has_only_tax_functionality', '__return_true' );
> > ```
> 
> Interesting! This does fix the issue for me. @Abdalsalaam What's causing that? I don't have the Shipping plugin installed.

Glad to hear its fixed! 
It seems the Shipping functionality is enabled for your site (Probably you connected your website to WooCommerce before adding only_tax flag ) and also its [not the first install](https://github.com/Automattic/woocommerce-services/pull/2879#issuecomment-3333641811)

PR #2903, opened 2025-12-04
Add tracks event for TaxJar

bartech:
## Description

We want to run experiment to see if we can skip all request to TaxJar to US when `from_state != to_state` to reduce amount of request that we assume should always give 0 tax rate. For this we are adding tracking event to see if `combined_tax_rate` is equal to 0 for all the cases.

Add a Tracks event immediately after the has_nexus[ check ](https://github.com/Automattic/woocommerce-services/blob/34d118847aa0af325e435bb1d895b5e3b1c0f2d8/classes/class-wc-connect-taxjar-integration.php#L1210)that captures the from/to states (and ideally the full addresses). Let it run for a week so we can see the actual distribution. That should give us enough signal to confirm whether narrowing API calls would silently break things for a meaningful number of merchants.

### Related issue(s)
Closes https://linear.app/a8c/issue/WOOSHIP-1821/add-tracks-even-tax-calculation-nexus-detected

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -77,6 +77,13 @@ class WC_Connect_TaxJar_Integration {
 	 */
 	private $backend_tax_classes;
 
+	/**
+	 * Tracks instance.
+	 *
+	 * @var WC_Connect_Tracks
+	 */
+	protected $tracks;
+
 	const PROXY_PATH               = 'taxjar/v2';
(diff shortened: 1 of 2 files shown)


bartech:
@samnajian should I add changelog entry for this change if this is temporary experiment?

samnajian (approved):
LGTM!

samnajian:
> @samnajian should I add changelog entry for this change if this is temporary experiment?

No need for a changelog entry for it.

PR #2906, opened 2025-12-18
Improve Woo Tax - Tax Rate Table Population Logic

bartech:
## Description
Revamped the Tax Rate Table population logic, improving handling of automated taxes, backend order calculations, and tax location normalization.

**Changes:**

- **Jurisdiction Handling:** Added logic to fetch and apply jurisdictions from TaxJar responses, simplifying tax labels and improving city/county assignment.

- **Removed code for Woo 2.6:** Refactored cart and backend tax calculation routines for WooCommerce 3.0+ to reduce redundant checks and improve reliability.
- **Normalized Addresses:** Introduced normalization of shipping and billing addresses using `wp_unslash` and consistent uppercasing.
- **Removed Legacy Code:** Dropped support for pre-WooCommerce 3.0 coupon and order item handling where unnecessary, including tooltip hacks and deprecated tax rate matching by street.
- **Schema Validation:** Tightened validation for address fields, enforcing type, length, and pattern checks more consistently.
- **Code Cleanups:** Replaced generic `json_encode` with `wp_json_encode`, unified style and comment formatting, and removed obsolete conditionals.

### Related issue(s)
Closes https://linear.app/a8c/issue/WOOSHIP-1819/improve-woo-tax-tax-rate-table-population-logic

### Test tax calculation for multiple scenarios:
- Add products to the cart for US, Canada (CA), and EU addresses.
- Verify automated taxes are applied correctly at checkout.
- Create backend orders and confirm taxes are calculated correctly.
- Confirm city/county information is properly assigned in tax labels.
- Ensure that zero-rate and non-taxable products are correctly excluded from tax calculations.

### Checklist

- [ ] unit tests

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -569,6 +569,8 @@ class WC_Connect_TaxJar_Integration {
 	 * Unchanged from the TaxJar plugin.
 	 * See: https://github.com/taxjar/taxjar-woocommerce-plugin/blob/96b5d57/includes/class-wc-taxjar-integration.php#L557
 	 *
+	 * @param int $order_id WC Order ID.
+	 *
 	 * @return void
 	 */
 	public function calculate_backend_totals( $order_id ) {
@@ -601,11 +603,16 @@ class WC_Connect_TaxJar_Integration {
 				$product_id    = $item->get_product_id();
 				$line_item_key = $product_id . '-' . $item_key;
 				if ( isset( $taxes['rate_ids'][ $line_item_key ] ) ) {
-					$rate_id  = $taxes['rate_ids'][ $line_item_key ];
-					$item_tax = new WC_Order_Item_Tax();
-					$item_tax->set_rate( $rate_id );
-					$item_tax->set_order_id( $order_id );
-					$item_tax->save();
+					$rate_ids = $taxes['rate_ids'][ $line_item_key ];
+					if ( ! is_array( $rate_ids ) ) {
+						continue;
+					}
+					foreach ( $rate_ids as $rate_id ) {
+						$item_tax = new WC_Order_Item_Tax();
+						$item_tax->set_rate( $rate_id );
+						$item_tax->set_order_id( $order_id );
+						$item_tax->save();
+					}
 				}
 			}
 		}
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -187,47 +187,38 @@ class WC_Connect_TaxJar_Integration {
 		$store_settings = $this->get_store_settings();
 		$store_country  = $store_settings['country'];
 
-		// TaxJar supports USA, Canada, Australia, and the European Union
+		// TaxJar supports USA, Canada, Australia, and the European Union.
 		if ( ! $this->is_supported_country( $store_country ) ) {
 			return;
 		}
 
-		// Add toggle for automated taxes to the core settings page
+		// Add toggle for automated taxes to the core settings page.
 		add_filter( 'woocommerce_tax_settings', array( $this, 'add_tax_settings' ) );
 
-		// Fix tooltip with link on older WC.
-		if ( version_compare( WOOCOMMERCE_VERSION, '4.4.0', '<' ) ) {
-			add_action( 'admin_enqueue_scripts', array( $this, 'fix_tooltip_keepalive' ), 11 );
-		}
-
-		// Settings values filter to handle the hardcoded settings
+		// Settings values filter to handle the hardcoded settings.
 		add_filter( 'woocommerce_admin_settings_sanitize_option', array( $this, 'sanitize_tax_option' ), 10, 2 );
 
-		// Bow out if we're not wanted
+		// Bow out if we're not wanted.
 		if ( ! $this->is_enabled() ) {
 			return;
 		}
 
-		// Scripts / Stylesheets
+		// Scripts / Stylesheets.
 		add_action( 'admin_enqueue_scripts', array( $this, 'load_taxjar_admin_new_order_assets' ) );
 
 		$this->configure_tax_settings();
 
-		// Calculate Taxes at Cart / Checkout
-		if ( class_exists( 'WC_Cart_Totals' ) ) { // Woo 3.2+
-			add_action( 'woocommerce_after_calculate_totals', array( $this, 'maybe_calculate_totals' ), 20 );
-		} else {
-			add_action( 'woocommerce_calculate_totals', array( $this, 'maybe_calculate_totals' ), 20 );
-		}
+		// Calculate Taxes at Cart / Checkout.
+		add_action( 'woocommerce_after_calculate_totals', array( $this, 'maybe_calculate_totals' ), 20 );
 
-		// Calculate Taxes for Backend Orders (Woo 2.6+)
+		// Calculate Taxes for Backend Orders.
 		add_action( 'woocommerce_before_save_order_items', array( $this, 'calculate_backend_totals' ), 20 );
 
-		// Set customer taxable location for local pickup
+		// Set customer taxable location for local pickup.
 		add_filter( 'woocommerce_customer_taxable_address', array( $this, 'append_base_address_to_customer_taxable_address' ), 10, 1 );
 
 		add_filter( 'woocommerce_calc_tax', array( $this, 'override_woocommerce_tax_rates' ), 10, 3 );
-		add_filter( 'woocommerce_matched_rates', array( $this, 'allow_street_address_for_matched_rates' ), 10, 2 );
+		add_filter( 'woocommerce_get_tax_location', array( $this, 'get_tax_location' ), 10, 3 );
 
 		add_filter( 'woocommerce_rate_label', array( $this, 'cleanup_tax_label' ) );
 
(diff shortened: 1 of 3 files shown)


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -729,6 +664,84 @@ public function cleanup_tax_label( $rate_name ) {
...
+	 *
+	 * @param  array       $location  Location address.
+	 * @param  string      $tax_class Tax class.
+	 * @param  WC_Customer $customer  Customer.
+	 *
+	 * @return array
+	 */
+	public function get_tax_location( $location, $tax_class, $customer ) {
+		$jurisdictions = $this->get_jurisdictions( $location, $customer );
+		$county        = $jurisdictions->county ?? '';
```
iyut, line 678: This might cause PHP warnings/errors when `get_jurisdictions()` returns false (which happens when customer is null OR the transient key doesn't match), the code tries to access properties on a boolean.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -729,6 +664,84 @@ public function cleanup_tax_label( $rate_name ) {
...
+	/**
+	 * Get normalized address from request body.
+	 *
+	 * @param array $request Request.
+	 *
+	 * @return array
+	 */
+	private function get_normalized_location( $request ) {
+		return array(
+			'to_country' => strtoupper( (string) $request['to_country'] ) ?? '',
```
iyut:
```suggestion
			'to_country' => isset($request['to_country']) ? strtoupper((string) $request['to_country']) : '',
```
The ?? '' will never execute because `strtoupper()` always returns a string (even empty string), never null.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1195,27 +1220,21 @@ private function is_nexus_address_valid( $address ): bool {
...
+			if ( 'string' === $rules['type'] && ! is_string( $value ) ) {
+				$errors[] = "[$field] field must be a string";
+				continue;
 			}
 
-			if ( isset( $rules['max_length'] ) && is_string( $value ) ) {
-				if ( strlen( $value ) > $rules['max_length'] ) {
-					$errors[] = "[$field] field exceeds maximum length of {$rules['max_length']}";
-				}
+			if ( is_string( $value ) && strlen( $value ) > $rules['max_length'] ) {
```
iyut:
```suggestion
			if ( isset( $rules['max_length'] ) && is_string( $value ) && strlen( $value ) > $rules['max_length'] ) {
```
To ensure `max_length` array key is set before we use it.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -809,24 +829,29 @@ protected function get_line_items( $wc_cart_object ) {
...
+					|| 'zero-rate' === sanitize_title( $product->get_tax_class() )
+				)
+			) {
 				$tax_code = '99999';
 			}
 
-			// Get WC Subscription sign-up fees for calculations
+			// Get WC Subscription sign-up fees for calculations.
 			if ( class_exists( 'WC_Subscriptions_Cart' ) ) {
 				if ( 'none' == WC_Subscriptions_Cart::get_calculation_type() ) {
```
iyut:
```suggestion
				if ( 'none' === WC_Subscriptions_Cart::get_calculation_type() ) {
```

readme.txt
```diff
@@ -70,6 +70,9 @@ This plugin relies on the following external services:
 
 == Changelog ==
 
+= 3.4.0 - 2025-xx-xx =
+* Revamp - Tax Rate Table population logic.
```
iyut, line 74: Are we allowed to use `Revamp`? I'm still not sure about this.

iyut (changes requested):
The changes does a good job on better code organization and adding more strict type comparison. But I've left some comments that could improve this PR.

readme.txt
bartech: IDK but this is what actually fit the changes done here. * Tweak would mean small change. * Change would also fit.

classes/class-wc-connect-taxjar-integration.php
bartech: `??` is equivalent to `isset( $value->prop ) ?  $value->prop : '';`

iyut (changes requested):
I see the code changes that it's intended to not save the zip code. What do you think if we put on the description of this PR and also put that on the changelog at least. I think the user should know about this. And potentially, we should ask this to A8C team too.

bartech:
> I see the code changes that it's intended to not save the zip code. What do you think if we put on the description of this PR and also put that on the changelog at least. I think the user should know about this. And potentially, we should ask this to A8C team too.

Yes, I'll reach out to change Woo Tax documentation and explain there that ZIP is not used in Tax Jurisdictions. Before that I want to have our internal validation that this change is good.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1670,62 +1690,68 @@ public function validate_taxjar_request( $json ) {
...
+	 * @param string $from_state   From state.
 	 *
 	 * @return mixed|WP_Error
 	 */
-	public function smartcalcs_cache_request( $json, $from_state ) {
+	public function smartcalcs_cache_request( $request_body, $from_state ) {
+		$json                = wp_json_encode( $request_body );
 		$cache_key           = 'tj_tax_' . hash( 'md5', $json );
+		$location            = $this->get_normalized_location( $request_body );
+		$location_catch_key  = 'tj_jurisdictions_' . hash( 'md5', $location['to_country'] . $location['to_state'] . $location['to_zip'] . $location['to_city'] . $location['to_street'] );
```
Abdalsalaam: It seems the jurisdiction cache key generation differs between where it's set and where it's retrieved :

The GET (`get_jurisdictions` line 722):
```php
$key = 'tj_jurisdictions_' . md5( strtoupper( implode( '', $location ) . $customer->get_shipping_address() ) );
```

This line uses `hash('md5', ...)` while GET uses md5(), these produce the same result, but the inputs are completely different

I think using 
```php
$location['to_country'] . $location['to_state'] . $location['to_zip'] . $location['to_city'] . $location['to_street'] )
``````
Is safer that `$customer->get_shipping_address()` since this can be filtered

File: classes/class-wc-connect-taxjar-integration.php [hunk 7]
```diff
@@ -1670,62 +1690,68 @@ public function validate_taxjar_request( $json ) {
...
 			$transient_set            = false;
 
-			if ( 200 == $response_code ) {
+			if ( 200 === $response_code ) {
 				set_transient( $cache_key, $response, $this->cache_time );
-			} elseif ( in_array( $response_code, $save_error_codes ) ) {
-				if ( 400 == $response_code
+				$jurisdictions = isset( $response_body_decoded->tax->jurisdictions ) ? $response_body_decoded->tax->jurisdictions : null;
+				if ( ! empty( $jurisdictions ) ) {
+					set_transient( $location_catch_key, $jurisdictions, $this->cache_time );
```
Abdalsalaam:
```suggestion
					set_transient( $location_cache_key, $jurisdictions, $this->cache_time );
```

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -729,6 +664,84 @@ public function cleanup_tax_label( $rate_name ) {
...
+	 * Replace WooCommerce tax locations with TaxJar jurisdictions.
+	 *
+	 * @param  array       $location  Location address.
+	 * @param  string      $tax_class Tax class.
+	 * @param  WC_Customer $customer  Customer.
+	 *
+	 * @return array
+	 */
+	public function get_tax_location( $location, $tax_class, $customer ) {
+		$jurisdictions = $this->get_jurisdictions( $location, $customer );
```
Abdalsalaam, line 677: `$jurisdictions` could be `false`, in PHP 7.4 accessing a property on false throws a warning.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1778,16 +1804,18 @@ public function smartcalcs_request( $json ) {
...
-		} elseif ( 404 == $response['response']['code'] || 400 == $response['response']['code'] ) {
-			$this->_error( 'Error retrieving the tax rates. Received (' . $response['response']['code'] . '): ' . $response['body'] );
+		} elseif ( 404 === $response_code || 400 === $response_code ) {
+			$this->_error( 'Error retrieving the tax rates. Received (' . $response_code . '): ' . $response['body'] );
 
 			return $response;
 		} else {
-			$this->_error( 'Error retrieving the tax rates. Received (' . $response['response']['code'] . '): ' . $response['body'] );
+			$this->_error( 'Error retrieving the tax rates. Received (' . $response_code . '): ' . $response['body'] );
 		}
```
Abdalsalaam: Missing return statement

Abdalsalaam (changes requested):

classes/class-wc-connect-taxjar-integration.php
bartech:
```php
$county = $jurisdictions->county ?? '';
```
Works same as 
```php
$county = isset( $jurisdictions->county) ? $jurisdictions->county : '';
```

Abdalsalaam: My bad, didn't see it 👍🏼

classes/class-wc-connect-taxjar-integration.php
dustinparker: Would be good to extract the cache key building to its own method to avoid mismatches going forward.

PR #2913, opened 2026-01-27
Make Sift configuration fetch non-blocking

ismaeldcom:
Closes WOOSHIP-601

### Description

- Update `WC_Connect_API_Client::get_sift_configuration()` to add non-blocking mode that returns immediately without making API requests
- Add background fetch via WP-Cron to retrieve Sift configuration asynchronously
- Cache successful responses for 1 month and errors for 5 minutes to prevent API hammering
- Replace inline JavaScript with WordPress script enqueuing API, using defer strategy for non-blocking rendering and browser caching
- Add unit tests for the new caching and non-blocking behavior

This change eliminates the synchronous HTTP request that was blocking admin page loads when Sift configuration wasn't cached. The first page load now returns immediately and schedules a background fetch, while subsequent loads serve from cache with deferred scripts.

Following the WooCommerce Shipping implementation, but also adding a background cache warming, to ensure we keep the performance impact to a minimum.

### Testing instructions
- Ensure you don't have WooCommerce Shipping enabled, because we prevent Sift from loading in that case.
- If you're using the local server, ensure you have a Sift config.
- Open the browser inspector on the network tab.
- Clear Sift transients with `wp transient delete wc_connect_sift_configuration` and reload the orders page.
- The Page should load instantly without blocking, and there shouldn't be any sift request.
- If you don't have cron enabled on the site, trigger it with `wp cron event run --due-now` and reload the page.
- Sift scripts should be fetched and `window._sift` initialized in the browser console.

### Checklist

- [x] unit tests

### DIFF
--- a/assets/javascript/sift.js
+++ b/assets/javascript/sift.js
@@ -1,0 +1,8 @@
+/* global wcServicesSiftConfig */
+
+if ( typeof wcServicesSiftConfig !== 'undefined' && wcServicesSiftConfig.beacon_key && wcServicesSiftConfig.user_id ) {
+	const _sift = window._sift || [];
+	_sift.push( [ '_setAccount', wcServicesSiftConfig.beacon_key ] );
+	_sift.push( [ '_setUserId', wcServicesSiftConfig.user_id ] );
+	_sift.push( [ '_trackPageview' ] );
+}
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Tax Changelog ***
 
+= 3.3.2 - 2026-xx-xx =
+* Fix   - Prevent admin slowdown by making Sift configuration non-blocking.
+
 = 3.3.1 - 2026-01-12 =
 * Fix   - Normalize state and country codes to uppercase in TaxJar integration.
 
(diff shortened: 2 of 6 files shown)


ayushpahwa (commented):
The change looks good and thanks for adding tests. 


I have a small question though, why do we need to cache the failure state? I think if we just cache the success and only delete the in progress key when a fetch succeeds, the code could be simplified. 

Success Flow

  1. Page load -> no cache -> WP_Error('sift_not_cached')
  2. Sets in-progress transient (5 min) ->schedules cron
  3. Cron runs -> API succeeds -> caches config (1 month) -> deletes in-progress flag
  4. Next page load -> returns cached config -> Sift loads

Failure Flow

  1. Page load -> no cache -> WP_Error('sift_not_cached')
  2. Sets in-progress transient (5 min) -> schedules cron
  3. Cron runs -> API fails -> nothing cached -> in-progress kept
  4. Next page load -> no cache -> WP_Error('sift_not_cached')
  5. schedule_background_sift_fetch() -> in-progress exists -> returns early
  6. (5 minutes pass, in-progress expires)
  7. Next page load -> back to step 1, retries

What do you think?

ismaeldcom:
Nice suggestion! I added the error case after testing it, but didn't think about simplifying the code. Addressed in 11a9efe

ayushpahwa (approved):
Changes look good, thanks for the updates!

PR #2918, opened 2026-02-02
Fixes/WOOTAX-29 - Prevent opted_out tracks event from being sent when user never opted in

iyut:
## Description

This PR fixes an issue where the `opted_out` tracks event was being sent when the plugin was deactivated, even if the user never opted in (never clicked "Connect" or accepted the Terms of Service).

**The Problem:**
- The `opted_out` method in `WC_Connect_Tracks` was unconditionally registered as a deactivation hook
- When the plugin was deactivated, it always sent the `opted_out` tracks event regardless of whether the user had actually opted in
- This resulted in semantically incorrect tracking data

**The Solution:**
- Added a check in the `opted_out` method to verify if the user has previously accepted the ToS (`tos_accepted` option is set to true)
- The `opted_out` event is now only sent if the user had previously opted in, making the tracking data accurate

### Related issue(s)

Closes [WOOTAX-29](https://linear.app/a8c/issue/WOOTAX-29/opted-out-event-tracked-before-tos-accepted)

1. Set up a fresh WP install with Jetpack connected
2. Delete all WCS options (to simulate fresh state)
3. Activate WCS plugin
4. Deactivate WCS without clicking "Connect" button
5. The `opted_out` tracks event is NOT sent (correct behavior)

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,10 +3,10 @@
 = 3.4.0 - 2026-xx-xx =
 * Add   - Optimize Woo Tax API calls by restricting same-state tax checks.
 * Add   - Country validation for TaxJar requests.
+* Add   - Add woocommerce_tax_line_item_location filter for per-item tax location override.
 * Fix   - Prevent admin slowdown by making Sift configuration non-blocking.
 * Fix   - Prevent opted_out tracks event from being sent when user never opted in.
 * Tweak - WooCommerce 10.5 Compatibility.
-* Add   - Add woocommerce_tax_line_item_location filter for per-item tax location override.
 
 = 3.3.1 - 2026-01-12 =
 * Fix   - Normalize state and country codes to uppercase in TaxJar integration.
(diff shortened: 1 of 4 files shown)


bartech (approved):
👍  This fixes the issue.

On Trunk with current test.

```php
WP_Test_WC_Connect_Tracks
 ✘ Opted out without opt in  3 ms
   ┐
   ├ WC_Connect_Logger::log('Tracked the following event: ...ed_out', '', false) was not expected to be called.
```

On this branch
```php
WP_Test_WC_Connect_Tracks
 ✔ Record user event  17 ms
 ✔ Opted in  2 ms
 ✔ Opted out  18 ms
 ✔ Opted out without opt in  3 ms
 ✔ Saved service settings  3 ms
 ✔ Shipping zone method added  2 ms
 ✔ Shipping zone method deleted  2 ms
 ✔ Shipping zone method enabled  2 ms
 ✔ Shipping zone method disabled  2 ms
```

PR #2922, opened 2026-02-06
Fix/issue wooship 1946 - Fix missing Sift tracker script 404 by inlining it via `wp_add_inline_script`

iyut:
## Description

After updating to WooCommerce Services / WooCommerce Tax 3.4.0, WP-Admin pages load a script tag pointing to a non-existent local asset: `/wp-content/plugins/woocommerce-services/assets/javascript/sift.js`. This results in a 404 error on every admin page load.

The root cause is that `assets/javascript/sift.js` was introduced in #2913 but the `assets/` directory is not included in the release build (`tasks/release.js` only copies `classes/`, `dist/`, `i18n/`, `images/`, `vendor/`, and `src/`).

This PR fixes the issue by inlining the Sift tracker initialization script via `wp_add_inline_script` on the `sift-science` CDN script handle, eliminating the need for a separate local JS file entirely. The standalone `assets/javascript/sift.js` file and the unused `WCSERVICES_JAVASCRIPT_URL` / `WCSERVICES_JAVASCRIPT_DIR` constants have been removed.

### Related issue(s)

Fixes [WOOSHIP-1946](https://linear.app/a8c/issue/WOOSHIP-1946/admin-console-404-error-for-missing-assetsjavascriptsiftjs-after)

1. Install WooCommerce Tax 3.4.0
2. Enable automated taxes in WooCommerce → Settings → Tax
3. Log into WP-Admin
4. Open browser dev tools → Console / Network

**Before:** `GET .../assets/javascript/sift.js?ver=3.4.0` returns `404 (Not Found)`

**After:** No 404 error. The Sift tracker script is inlined directly after the Sift CDN script.

### Checklist

- [ ] unit tests

### DIFF
--- a/assets/javascript/sift.js
+++ b/assets/javascript/sift.js
@@ -1,8 +1,0 @@
-/* global wcServicesSiftConfig */
-
-if ( typeof wcServicesSiftConfig !== 'undefined' && wcServicesSiftConfig.beacon_key && wcServicesSiftConfig.user_id ) {
-	const _sift = window._sift || [];
-	_sift.push( [ '_setAccount', wcServicesSiftConfig.beacon_key ] );
-	_sift.push( [ '_setUserId', wcServicesSiftConfig.user_id ] );
-	_sift.push( [ '_trackPageview' ] );
-}
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Tax Changelog ***
 
+= 3.4.1 - 2026-xx-xx =
+* Fix   - Fix missing Sift tracker script by inlining it instead of loading a non-distributed file.
+
 = 3.4.0 - 2026-02-03 =
 * Add   - Optimize Woo Tax API calls by restricting same-state tax checks.
 * Add   - Country validation for TaxJar requests.
--- a/readme.txt
+++ b/readme.txt
@@ -70,6 +70,9 @@ This plugin relies on the following external services:
 
 == Changelog ==
 
+= 3.4.1 - 2026-xx-xx =
+* Fix   - Fix missing Sift tracker script by inlining it instead of loading a non-distributed file.
+
 = 3.4.0 - 2026-02-03 =
 * Add   - Optimize Woo Tax API calls by restricting same-state tax checks.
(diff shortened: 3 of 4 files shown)


dustinparker:
Hey @iyut, I haven't been able to replicate the trunk issue for some reason. I don't see any JS files loaded in the console > sources and there are no 404s.

iyut:
@dustinparker , first of all, you need to use the release version (`npm run build`). Because the repo version has `assets/javascript/sift.js`. But it's removed on the release version. 

Or, probably because the plugin on your env, does not load `init_shipping_labels()`. You can edit the woocommerce-services.php on line 956, to be like this `if ( self::should_load_shipping_features() || true ) {`. You might also need to check the conditions on `add_sift_js_tracker()` as well. But I think the point is, the sift.js is not being included in released zip file. Because technically, all js file on WC Services should be generated on dist/ folder.

So the solution is actually : 
1. We either make the content inside sift.js to be generated on dist/ folder.
2. Or the one that I propose in this PR. Using `wp_add_inline_script()` since it's only 6 rows.

dustinparker (approved):
LGTM!

bartech (commented):
@iyut 
Inline scripts execute immediately during parsing.
If the dependent on script `sift-science` is deferred, isn't the inlined script executed too early?

iyut:
@bartech , yes, I've looked into it as well. From my understanding based on [this documentation](https://developers.sift.com/docs/v204/curl/javascript-api/overview), The `_sift` array seems acts as a command queue — we push commands `(_setAccount, _setUserId, _trackPageview)` onto a plain JS array, and when the deferred s.js CDN script finally loads, it picks up and processes whatever is already queued in `window._sift`.

If you looked into that documentation, it's declared `var _sift = window._sift = window._sift || [];` before the `cdn.sift.com/s.js` is loaded. Based on that, I assume it act as a command queue.

We can also remove the defer strategy since it's loaded on the footer.

what do you think?

erikgeurts:
It would be very convenient if a new release of this plugin containing this particular fix could be released rather sooner than later, since this 404 of sift.js is causing the few hairs I have left to turn even more grey.

PR #2923, opened 2026-02-07
Allow inclusive tax price while automated tax rates are enabled.

Laserology:
## Description

This PR allows the store owner to include tax into the cost of goods while still being able to take advantage of the woocommerce API for tax rates (automated taxes)

### Related issue(s)

https://github.com/woocommerce/woocommerce/issues/63179

Before: Open Woocommerce settings -> tax, see that "inclusive of tax" is not an available option.
After: Open Woocommerce settings -> tax, see that "inclusive of tax" can now be toggled properly.

image

image

### Checklist

- [ ] unit tests

What tests should be added? (if any) This is the first time i am contributing to this project.

What should be added to the changelog? Specific date/version?


Laserology:
Hi, this PR should be ready. but i would like some input if this should be allowed into the main repo.

Is this code stable? Is the functionality it adds useful/wanted? So far from my testing it works entirely as expected.

Additionally, what should be added to the change log? What should i be testing for in the suggested unit tests?

ayushpahwa (commented):
Hey, apologies for the late response. 

The experience for not allowing tax inclusive prices when automated taxes are enabled is intentional mainly due to the side effects on the woo core. The main issue would be incorrect prices since taxes can't be calculated without an address. Even if the address is present, the rate fetch might cause perf degradation. Hence, this PR alone would be insufficient and the change if needed would require a larger change.

Laserology:
Thanks for the review,

Short on time lately so i don't have time immediately but i will be looking into this.

Pardon me if i am misunderstanding, but when taxes are included in the price, the price shouldn't need a calculation, correct? Apart from the itemization of tax entries to be _subtracted_ from the final order total, item prices should remain static as entered in the admin interface. Typically by the time an order is finalized at any point, an address has been entered (for shipping/billing) which is a typical lookup that happens under normal use already.

With that logic, my concern is that the admin and mobile app interfaces aren't automatic, and need to be manually set to an address. But to me it sounds like there is a deeper issue i do not quite understand. Are there any existing issues or documentation i can read on to get more context?

Thanks, much appreciated.

PR #2927, opened 2026-02-16
Fix build workflow to fix QIT API Test fatal

allie500:
## Description

Fixes the Build workflow artifact so the produced zip has the structure QIT expects.

Previously, the workflow used an unzip/re-zip approach that accidentally produced a **nested duplicate** (plugin contents at root plus a second copy inside a `woocommerce-services/` folder). This erroneous Zip archive would pass QIT API test for months until very recently an unidentified upstream change resulted in both copies of the plugin being activated in the QIT API test. This caused the following fatal error:

```
PHP Fatal error: Cannot redeclare class Automattic\Jetpack\Autoloader\jp7109386222804879fea6838ada605690\al5_0_9\Autoloader (previously declared in wp-content/plugins/woocommerce-services/vendor/jetpack-autoloader/class-autoloader.php:15) in wp-content/plugins/vendor/jetpack-autoloader/class-autoloader.php on line 15
```

This PR updates the Build workflow to re-zip the artifact after `npm run build`: the flat zip is extracted into a directory named from `plugin_name`, then only that directory is zipped. The uploaded artifact therefore has one copy of the plugin in the Zip archive to avoid the class redeclaration fatal error.

### Related issue(s)

Fixes WOOSHIP-2009

1. Visit the [Manual Test Runner](https://github.com/Automattic/woocommerce-services/actions/workflows/manual_qit.yml).
2. Click the run workflow drop down.
   - Choose this PR branch as the branch for the workflow: `fix/WOOSHIP-2009-qit-fatal-error`
   - Leave WordPress version on 'latest'
   - Set WooCommerce version to 'nightly'
   - Leave PHP version set to 8.4
   - Set extension tests to none, we don't need to run those for this fix test.
   - Leave QIT tests set to the WooCommerce Pre-Release Tests (this includes the API Test)
   - Leave the QIT additional options field blank
   - Click run workflow
3. Once the build process has completed, click the the 'Upload plugin zip' step and download the plugin zip artifact.
4. Unzip it: the root of the archive should be a single directory named `woocommerce-services` with all plugin release related files inside it. Only a single copy.
5. Once the QIT tests finish running, click the 'Echo API results' step, and click the link to view the report. It should be a success with no fatal error logged.
6. All QIT tests should pass.

Note: E2E tests are currently failing due to a known issue with the docker-compose client being too far out of date. This issue has been added to Linear.

### Checklist

- [x] unit tests

### DIFF
--- a/.agents/skills/create-pr/SKILL.md
+++ b/.agents/skills/create-pr/SKILL.md
@@ -1,0 +1,403 @@
+---
+name: create-pr
+description: Generate PR description using repository's PR template and current branch changes, copy to clipboard
+---
+
+# Create PR
+
+Generate GitHub pull request descriptions using the PR template at `.github/PULL_REQUEST_TEMPLATE.md` and referencing changes in the current branch. Outputs markdown and copies to clipboard.
+
+## When to Use
+
+Activate this skill when the user:
+- Uses `/create-pr` command with or without a repo name
+- Says "create a PR description", "generate PR body", "make a pull request"
+- Wants a PR description for their current branch
+
+## Important
+
+**This skill does NOT:**
+- ❌ Commit changes
+- ❌ Push to remote
+- ❌ Create the actual PR on GitHub
+
+**This skill ONLY:**
+- ✅ Generates PR title and description
+- ✅ Outputs as markdown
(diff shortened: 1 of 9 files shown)


samnajian (approved):
LGTM, thanks.
The failing e2e tests is delegated to WOOTAX-263

PR #2945, opened 2026-03-25
chore: exclude AI instruction files from archive

Abdalsalaam:
Excludes `AGENTS.md` and/or `CLAUDE.md` from the Composer archive zip to keep the distributed plugin package clean. These files are for AI coding agents and are not needed by end users.


Abdalsalaam:
No need for this - AI files will not be copied

Issue #2826, opened 2024-11-04
Subscription switching ends up zeroing out existing tax rate 0.0000%


ryanr14:
When using subscriptions alongside automated rates, I've come across an issue in this ticket that an existing tax rate (added by automated rates) gets zeroed out. This only happens when you're switching your subscription. I've had a hard time replicating this on a Pressable test site, but we can easily do this on their live & staging site. If you then disable automated rates, fix the existing zeroed out rates, this never happens again.

What would be causing our automated rate service to zero out a rate just previously used to purchase a subscription? This only happens, that we can tell, when switching subscriptions.

8889343-zen

![Image](https://github.com/user-attachments/assets/782c3112-afef-49e0-9209-462ac8b65ede)

```
2024-11-04T01:12:21+00:00 Notice :::: TaxJar Plugin requested :::: (WCS Tax)  
Additional context
{
    "_legacy": true
}
2024-11-04T01:12:21+00:00 Notice :::: TaxJar API called :::: (WCS Tax)  
Additional context
{
    "_legacy": true
}
[...]

PR #2966, opened 2026-06-18
WOOTAX-25 - Add regression test for zero-amount tax calculations

Abdalsalaam:
## Description

When a subscription is switched, or when a free-trial subscription's initial cart total is `$0`, TaxJar returns a response whose top-level `rate` (and `amount_to_collect`) is `0` because `taxable_amount` is `0` — while the real per-jurisdiction rates are still present inside `breakdown.line_items`. Historically the rate builder wrote the zeroed top-level rate into the shared WooCommerce tax-rate row, which set an existing rate to `0.0000%` (WOOTAX-18) and left subscription renewals with no tax (WOOTAX-25).

The current `get_itemized_tax_rates()` already persists the real per-jurisdiction rates from `breakdown.line_items` and never writes the zeroed top-level rate (fixed by the itemized-tax-rate refactor). This PR is a **test-only** change that locks that behaviour in with a regression test, so the zeroing-out cannot silently come back.

The test feeds the exact zero-amount response captured in WOOTAX-18 into `get_itemized_tax_rates()` and asserts the real per-jurisdiction rates (city 1.5%, county 1.25%, state 4.225%) are persisted and the combined rate is 6.975%, never 0. It was verified to fail if the builder is reverted to writing the top-level rate.

### Related issue(s)

Linear: WOOTAX-25 (same root cause as WOOTAX-18 / #2826).

Original bug (no longer reproducible on trunk):
1. Enable automated tax rates (WooCommerce Tax).
2. Create a subscription product with a free trial and no sign-up fee.
3. Add it to the cart; observe the "recurring total" tax line shows $0.00 because the $0 initial calculation zeroed the shared tax rate.

Verifying the guard:
1. Run the suite: `./vendor/bin/phpunit --filter test_zero_amount_response_persists_real_itemized_rates` — passes on trunk.
2. Temporarily change the line-item rate in `get_itemized_tax_rates()` to write the top-level rate — the test fails, confirming it catches the regression.

N/A - test-only change, no visual changes.

### Checklist

- [x] unit tests

- [ ] `changelog.txt` entry added — N/A: test-only change, no user-facing impact.
- [ ] `readme.txt` entry added — N/A: test-only change, no user-facing impact.

### DIFF
--- a/tests/php/test-class-wc-connect-taxjar-integration.php
+++ b/tests/php/test-class-wc-connect-taxjar-integration.php
@@ -1146,4 +1146,84 @@ class WP_Test_WC_Connect_TaxJar_Integration extends WC_Unit_Test_Case {
 		$this->assertEmpty( $result['line_items'] );
 		$this->assertEquals( 0, $result['has_nexus'] );
 	}
+
+	/**
+	 * A tax calculation whose taxable amount is zero must not zero out the tax rate.
+	 *
+	 * When a subscription is switched, or when a free-trial subscription's initial
+	 * cart total is $0, TaxJar returns a response whose top-level `rate` (and
+	 * `amount_to_collect`) is 0 because `taxable_amount` is 0 — yet the real
+	 * per-jurisdiction rates are still present inside `breakdown.line_items`. The
+	 * itemized rate builder must persist those real rates and must never write the
+	 * zeroed top-level rate, which previously clobbered an existing tax rate to
+	 * 0.0000%.
+	 *
+	 * Regression test for WOOTAX-25 (free-trial renewal charged no tax) and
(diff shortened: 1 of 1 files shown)


bartech (approved):
✅ LGTM — approving.

The regression test correctly locks in the zero-amount tax fix (WOOTAX-25): I traced the path and it genuinely fails if production regresses to writing the top-level shared `tax_rate` = 0, and CI is green across the PHP/WP/WC matrix.

A couple of optional, non-blocking hardening ideas:
- Consider also asserting `$result['tax_rate']` is `0.0` — that's the field actually carrying the zeroed shared rate, and it's currently unasserted (the other half of the contract).
- A one-line comment on why `assertCount( 4, … )` is load-bearing would help future readers.

Nice, well-scoped test.

PR #2969, opened 2026-06-24
Fix/issue WOOSHIP-2256 - Fixes for TaxJar tax lines wiped when REST API order update includes address change

iyut:
## Description

When an external integration (e.g. UPS itembase) sends a REST API PUT request to update a WooCommerce order's `shipping_lines` and includes a `shipping` address that differs from the original order address, all TaxJar tax line items are wiped from the order — leaving `total_tax: 0` and `tax_lines: []`.

**Root cause:** `WC_Abstract_Order::calculate_taxes()` calls `WC_Tax::find_rates()` with the new shipping address. TaxJar's WC rate records are stored scoped to the original checkout address. When the address changes (even slightly — e.g. postcode `90210` → `90211`), `find_rates()` returns empty and `update_taxes()` removes all existing tax line items. TaxJar's override hooks are no-ops in REST context because `$this->response_rate_ids` is never populated outside of cart/checkout.

**Fix:** Once an order is placed, its tax is a record of what was actually charged, so we preserve it rather than recompute it against a changed address. A new handler hooked on `woocommerce_order_before_calculate_taxes` snapshots the order's existing tax state before WC recalculates, and a matching handler on `woocommerce_order_after_calculate_totals` restores it once the totals have been recalculated:

1. `preserve_order_taxes_on_recalculation()` snapshots the tax line items (with full metadata), per-item tax arrays, and order-level tax totals, keyed by order id.
2. WC recalculates the order (and, for a changed address, wipes the tax lines).
3. `restore_order_taxes_after_recalculation()` puts the snapshotted taxes back and rebases the order total on the preserved tax, while keeping any updated non-tax amounts (such as a changed shipping total).

The handler only runs in REST/programmatic context — it bails for the cart/checkout flow (`response_rate_ids` populated), admin AJAX edits (handled by `calculate_backend_totals()`), new orders (id `0`), and orders that have no existing tax lines (so a first-time calculation still runs normally). The snapshot is stashed by order id rather than in a per-call closure, so a batch can recalculate many orders without stacking one-shot callbacks.

### Related issue(s)

Fixes [WOOSHIP-2256](https://linear.app/a8c/issue/WOOSHIP-2256/order-updates-through-rest-api-requests-drop-woocommerce-tax-line)

**Before fix:**

1. Create an order with TaxJar tax lines (e.g. CA / Beverly Hills / 90210)
2. Send a REST API PUT to `/wc/v3/orders/{id}` with `shipping_lines` + a `shipping` address with a different postcode (e.g. `90211`) or different state
3. GET the order — `tax_lines` is empty, `total_tax` is `0.00`

**After fix:**

Same PUT request → the recorded tax lines are preserved instead of being wiped, and the order total is rebased on the preserved tax (accounting for any updated shipping amount).

Verified via Postman:

| Test | Payload | Before | After |
|------|---------|--------|-------|
| A | `shipping_lines` only | Taxes preserved | Taxes preserved |
| B | `shipping_lines` + same address | Taxes preserved | Taxes preserved |
| C | `shipping_lines` + different postcode | **Taxes wiped** | Taxes preserved |
| D | `shipping_lines` + different state | **Taxes wiped** | Taxes preserved |

### Checklist

- [x] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Tax Changelog ***
 
+= 3.6.7 - 2026-xx-xx =
+* Fix   - TaxJar tax lines wiped when REST API order update includes address change.
+
 = 3.6.6 - 2026-06-22 =
 * Tweak - WooCommerce 10.9 Compatibility.
 
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -78,6 +78,22 @@ class WC_Connect_TaxJar_Integration {
 	private $backend_tax_classes;
 
 	/**
+	 * Whether TaxJar API recalculation failed on the current order update.
+	 *
+	 * @since x.x.x
+	 * @var bool
+	 */
+	private $taxjar_recalculation_failed = false;
+
+	/**
+	 * Snapshotted tax state saved before a non-cart recalculation, for fallback restore.
+	 *
+	 * @since x.x.x
+	 * @var array
+	 */
+	private $pre_recalculation_tax_snapshot = array();
+
+	/**
 	 * Tracks instance.
 	 *
 	 * @var WC_Connect_Tracks
@@ -214,6 +230,9 @@ class WC_Connect_TaxJar_Integration {
 
 		// Calculate Taxes for Backend Orders (Woo 2.6+)
 		add_action( 'woocommerce_before_save_order_items', array( $this, 'calculate_backend_totals' ), 20 );
+
+		// Calculate taxes via TaxJar API for REST API / programmatic order updates.
+		add_action( 'woocommerce_order_before_calculate_taxes', array( $this, 'calculate_order_taxes_via_taxjar' ), 10, 2 );
 
 		// Set customer taxable location for local pickup
 		add_filter( 'woocommerce_customer_taxable_address', array( $this, 'append_base_address_to_customer_taxable_address' ), 10, 1 );
@@ -2159,5 +2178,19 @@ class WC_Connect_TaxJar_Integration {
 		if ( 'not set' === $to_state || 'not set' === $to_country || null === $has_nexus ) {
 			throw new Exception( sprintf( 'One or more values are not set : to_state=>%1$s, to_country=>%2$s, has_nexus=>%3$s', $to_state, $to_country, json_encode( $has_nexus ) ) );
 		}
+	}
+
+	/**
+	 * Calculate order taxes via TaxJar API for non-cart, non-admin contexts.
+	 *
+	 * Fires on woocommerce_order_before_calculate_taxes to intercept tax
+	 * recalculation triggered by REST API or direct programmatic calls to
+	 * $order->calculate_totals(), before WC's native find_rates() runs.
+	 *
+	 * @since x.x.x
+	 * @param array    $args  Args passed to calculate_taxes().
+	 * @param WC_Order $order The order being recalculated.
+	 */
+	public function calculate_order_taxes_via_taxjar( $args, $order ) {
 	}
 }
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -2192,6 +2192,9 @@ class WC_Connect_TaxJar_Integration {
 	 * @param WC_Order $order The order being recalculated.
 	 */
 	public function calculate_order_taxes_via_taxjar( $args, $order ) {
+		// Reset failure flag from any previous call.
+		$this->taxjar_recalculation_failed = false;
+
 		// Skip if TaxJar already ran for this request (cart/checkout path populated response_rate_ids).
 		if ( ! empty( $this->response_rate_ids ) ) {
 			return;
@@ -2205,6 +2208,69 @@ class WC_Connect_TaxJar_Integration {
 		// Skip if new order being created from scratch (no existing taxes to recalculate).
 		if ( ! $order->get_id() ) {
 			return;
+		}
+
+		// Snapshot existing tax state in case we need to restore on API failure.
+		$this->pre_recalculation_tax_snapshot = $this->snapshot_order_taxes( $order );
+
+		// Ensure WC customer object is available (may be null in REST/CLI context).
+		if ( is_null( WC()->customer ) ) {
+			WC()->customer = new WC_Customer( $order->get_customer_id() );
+		}
+
+		// Build address from order (reflects any address update already applied by REST request).
+		$to_country = strtoupper( $order->get_shipping_country() ?: $order->get_billing_country() );
+		$to_state   = strtoupper( $order->get_shipping_state() ?: $order->get_billing_state() );
+		$to_zip     = $order->get_shipping_postcode() ?: $order->get_billing_postcode();
+		$to_city    = $order->get_shipping_city() ?: $order->get_billing_city();
+		$to_street  = $order->get_shipping_address_1() ?: $order->get_billing_address_1();
+
+		// Get line items and shipping using the existing backend helper.
+		$line_items = $this->get_backend_line_items( $order );
+		$shipping   = $order->get_shipping_total();
+
+		// Call TaxJar API.
+		$taxes = $this->calculate_tax(
+			array(
+				'to_country'      => $to_country,
+				'to_state'        => $to_state,
+				'to_zip'          => $to_zip,
+				'to_city'         => $to_city,
+				'to_street'       => $to_street,
+				'shipping_amount' => $shipping,
+				'line_items'      => $line_items,
+			)
+		);
+
+		if ( false === $taxes ) {
+			$this->_log( 'TaxJar API failed for order ' . $order->get_id() . '. Preserving existing taxes as fallback.' );
+			$this->taxjar_recalculation_failed = true;
+			$snapshot                          = $this->pre_recalculation_tax_snapshot;
+			add_action(
+				'woocommerce_order_after_calculate_totals',
+				function ( $and_taxes, $updated_order ) use ( $order, $snapshot ) {
+					if ( $updated_order->get_id() === $order->get_id() ) {
+						$this->restore_order_taxes( $updated_order, $snapshot );
+					}
+				},
+				10,
+				2
+			);
+			return;
+		}
+
+		$this->response_rate_ids   = $taxes['rate_ids'];
+		$this->response_line_items = $taxes['line_items'];
+
+		// Adjust line_total on response_line_items to match what override_woocommerce_tax_rates expects.
+		if ( isset( $this->response_line_items ) ) {
+			foreach ( $this->response_line_items as $response_line_item_key => $response_line_item ) {
+				$line_item = $this->get_line_item( $response_line_item_key, $line_items );
+				if ( isset( $line_item ) ) {
+					$this->response_line_items[ $response_line_item_key ]->line_total =
+						( $line_item['unit_price'] * $line_item['quantity'] ) - $line_item['discount'];
+				}
+			}
 		}
 	}
 
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -66,6 +66,14 @@ class WC_Connect_TaxJar_Integration {
 	private $response_line_items;
 
 	/**
+	 * Tax snapshots captured before an out-of-cart recalculation, keyed by order id,
+	 * so they can be restored after WC recalculates the order totals.
+	 *
+	 * @var array
+	 */
+	private $pre_recalculation_tax_snapshots = array();
+
+	/**
 	 * @var bool
 	 */
 	private $is_itemized_tax_display;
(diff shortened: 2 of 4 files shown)


dustinparker (changes requested):
Hey @iyut,

Thanks for digging into this one. I went through the PR and tested it locally, and it does fix the reported problem. But before we go further, I think we should step back on the approach because I don't think we should be recalculating the taxes at all here.

My thinking is that once an order is placed, the tax on it should reflect what was actually charged at checkout. The customer or merchant already paid that amount, and the order is the record of it. So changing the tax from the back end after the fact feels wrong to me, even if we can get a fresh number from TaxJar.

The issue doesn't seem to be that "the tax is stale." It seems to be that "the tax gets wiped to 0 when a REST update changes the address." So I think this is more of a data integrity problem. We should be preserving the existing tax instead of recomputing it.

A couple of concerns I have with recalculating:

- On a successful TaxJar call, we'd actually change the recorded tax on an already-paid order. That looks like a compliance problem, since the tax on record would no longer match what was collected and remitted. It would also be confusing for the merchant to see an order's tax change after the fact.
- It adds a synchronous TaxJar API call to every back-end recalc, including `/orders/batch` and any programmatic `calculate_totals()` call. That's extra calls and timeout/rate-limit risk for paths that aren't part of the original issue.

The good news is the snapshot and restore you already wrote for the API-failure case is basically what we want. I think we can keep that, always preserve the existing tax, and drop the TaxJar recalculation, the `response_rate_ids` plumbing, and the failure flag. That makes the change a lot smaller and removes most of the risk.

One thing to watch: we should only restore when there were existing tax lines, so we don't accidentally turn a first-time tax calculation into 0. A brand new order that calculates tax for the first time hits this same hook with no prior tax, and we don't want to wipe that.

I did check Subscriptions, by the way. Renewal orders copy the tax from the subscription and don't trigger this hook, so they're not affected either way, which is good.

What do you think? Does that line up with what you're seeing, or is there a case where recalculating is actually the behavior we want?

bartech (changes requested):
## Review — 🔴 Request changes

Thanks for the thorough work here — the snapshot/restore machinery is solid. My concern is the same one @dustinparker already raised: the **approach recomputes tax on an already-placed order**, which is the opposite of what WOOSHIP-2256 asks for ("preserve the recorded tax"). A few correctness issues compound it.

### High — Approach contradicts the issue intent and widens blast radius
The PR rewrites recorded tax on a paid order whenever the TaxJar API succeeds (data-integrity/compliance concern — tax on record no longer matches what was collected) and adds a **synchronous TaxJar API call to every non-AJAX recalc path**: REST single + `/orders/batch`, cron, webhooks, bulk actions, programmatic `calculate_totals()` — with timeout/rate-limit exposure on paths unrelated to the bug.
- `classes/class-wc-connect-taxjar-integration.php` :244 (hook), :2203 (method)
- The snapshot/restore you already wrote, applied **unconditionally**, is essentially the smaller, lower-risk fix the issue wants. Drop the TaxJar recalculation + `response_rate_ids` plumbing + failure flag.

### High — Snapshot/restore loses most tax-line metadata
`snapshot_order_taxes()` captures only `rate_id`, `tax_total`, `shipping_tax_total`; restore rebuilds each `WC_Order_Item_Tax` from just those. **Rate code, label, compound flag, and rate percent are dropped** — restored lines render with a blank tax name, losing exactly the data the PR is meant to preserve. (Your own snapshot test sets `set_rate_code('US-CA-TAX-1')` / `set_label('CA Tax')` — neither is captured.)
- :2314 (snapshot), :2376 (restore loop)

### High — Destination address ignores `woocommerce_tax_based_on`
The new method hardcodes shipping-first/billing-fallback, whereas the existing `get_address()`/`get_backend_address()` honor the store's base/billing/shipping setting. On billing- or base-based stores this persists rates against the wrong address.
- :2242–2246

### Medium
- Restore only fires on `after_calculate_totals`; a programmatic `$order->calculate_taxes()` with no following `calculate_totals()` wipes taxes and the restore closure never runs (and leaks on the hook for the request). :2275–2281
- Recalculates even when the order had no prior tax lines.

### Low / hygiene
`@since x.x.x` placeholders in new props/methods (changelog declares 3.6.7) and `@since` not placed last after a blank comment line; new tests miss `@testdox`, use forbidden `// Arrange/Act/Assert` comments and `$this->integration` instead of `$sut`; over-commenting in new methods; public callback `calculate_order_taxes_via_taxjar` lacks `@internal`; minor WPCS `=` alignment at :2211–2213.

CI is green and the branch is mergeable, but I'd hold for the approach rework before merging.

Abdalsalaam (commented):

---
## Findings

None. The documented golden REST PUT path is covered: WooCommerce core's REST orders controller calls `$order->calculate_totals( true )` when `shipping_lines` are present, which both triggers the new hook and fires `woocommerce_order_after_calculate_totals` (so the restore closure works there).

**M1 — Always-on hook makes a synchronous TaxJar API call (and tax-rate writes) on every non-AJAX recalc of an existing order**
`classes/class-wc-connect-taxjar-integration.php:244` + method body
```php
add_action( 'woocommerce_order_before_calculate_taxes', array( $this, 'calculate_order_taxes_via_taxjar' ), 10, 2 );
```
The only guards are `wp_doing_ajax()`, non-empty `response_rate_ids`, and `! $order->get_id()`. Every programmatic `$order->calculate_taxes()`/`calculate_totals()` on an existing order — Subscriptions renewals, cron jobs, WP-CLI, bulk tools, other integrations, and admin REST edits — now incurs a blocking external API call and routes through `create_or_update_tax_rate()` (`:1759`), writing/updating the shared `wc_tax_rates` table. This is a meaningful latency/failure-mode and backwards-compatibility expansion well beyond the address-change PUT the issue describes. *Fix: gate on an actual address change (compare pre/post), and/or restrict to the REST update path; consider deferring recalcs that aren't user-facing.*

**M2 — Failure-fallback restore is lossy and re-derives the order total by hand**
`classes/class-wc-connect-taxjar-integration.php` `snapshot_order_taxes()` / `restore_order_taxes()` (~2300–2415)
The snapshot captures only `rate_id`, `tax_total`, `shipping_tax_total`:
```php
$snapshot['tax_lines'][ $item_id ] = array(
    'rate_id' => $tax_item->get_rate_id(),
    'tax_total' => $tax_item->get_tax_total(),
    'shipping_tax_total' => $tax_item->get_shipping_tax_total(),
);
```
Restore rebuilds `WC_Order_Item_Tax` with only those three setters, so **label, rate code, rate percent, and compound flag are lost** — after a TaxJar failure the order shows an unlabeled tax row and downstream readers of `rate_code`/`rate_percent` get defaults. Separately, the total is recomputed with a hand-rolled formula:
```php
$order->set_total( $order->get_subtotal() - $order->get_discount_total() + $order->get_shipping_total() + $order->get_total_fees() + $cart_tax + $shipping_tax );
```
This re-implements WC core's own total computation and can diverge on compound taxes, the `woocommerce_calculated_total` filter, and rounding. *Fix: capture `get_rate_code()`/`get_label()`/`get_rate_percent()`/`get_compound()` and restore them; for the total, snapshot and restore `$order->get_total()` rather than recomputing.* (Note: the earlier "double-subtracts discount" claim was checked and is **not** correct — `get_subtotal()` is pre-discount, so subtracting `get_discount_total()` once is right; the real risk is the re-implementation divergence above.)

**M3 — Restore closure only fires from `calculate_totals()`; direct `calculate_taxes()` callers leave taxes wiped**
`classes/class-wc-connect-taxjar-integration.php` (failure branch, ~2270–2285)
```php
add_action( 'woocommerce_order_after_calculate_totals', $restore_callback, 10, 2 );
```
`woocommerce_order_after_calculate_totals` fires only from `WC_Abstract_Order::calculate_totals()`. The golden REST PUT path goes through `calculate_totals( true )`, so it works there — but any caller invoking `$order->calculate_taxes()` directly triggers WC's wipe (`update_taxes()`) without ever firing the restore, so taxes stay at $0 and the orphaned closure lingers registered for the request. `taxjar_recalculation_failed` is set but never consulted to cover this. *Fix: also register restore on `woocommerce_order_after_calculate_taxes` (which fires at the end of `calculate_taxes()` itself), or have callers honor the failure flag.*

**M4 — The central fix is not exercised end-to-end by tests**
`tests/php/test-class-wc-connect-taxjar-integration.php:538` (`test_calculate_order_taxes_happy_path_sets_response_properties`)
The happy-path test mocks `calculate_tax()` and `get_backend_line_items()` and asserts only the private `response_rate_ids`/`response_line_items` props — it never runs a real `$order->calculate_taxes()` through `create_or_update_tax_rate → find_rates → override_order_item_taxes`, so the actual "recalculate taxes for the new address" behavior (cases C/D) is untested; only the plumbing is. The batch test (`:351`) uses an `id=0` order to short-circuit immediately, so the stale-state-clearing gate's real purpose (order B getting its own TaxJar call after order A) is unverified. Restore fidelity (M2) is also unasserted. *Fix: add an integration test that drives a real recalculation and asserts `get_taxes()`/`get_total_tax()`, plus a genuine two-order batch test.*

### Minors (nice to fix)
- **`taxjar_recalculation_failed` is dead state** — set on failure (`:~2270`) and reset on entry (`:~2205`) but never read in production. Either consume it (e.g. to suppress the override hooks after a failure) or remove it. *(correctness, silent-failure)*
- **No-nexus / cross-state result silently zeroes taxes** — `calculate_tax()` returns a non-`false` empty array for US cross-state (`:1633`) / no-nexus (`:1724`), so `response_rate_ids = array()`, the overrides no-op, WC wipes to $0, and **no snapshot/restore or log** occurs. Correct for a genuinely non-taxable new address, but indistinguishable from a transient empty response and emits no diagnostic. *Fix: when the snapshot had taxes but the new result is empty, log it.*
- **Failure observability is thin** — the single `_log()` omits the TaxJar error body/status (`calculate_tax()` discards it at `:1626/1649/1658`) and `_log()` no-ops entirely when logging is disabled, so the only failure signal can be suppressed. *(silent-failure)*
- **`restore_order_taxes()` swallows all write outcomes** — `set_taxes`/`save`/`remove_item`/`add_item`/the two `$order->save()` calls are unchecked and unguarded; a mid-loop failure leaves the order half-restored. *Fix: wrap in try/catch and log.* *(silent-failure)*
- **Per-item `save()` in loops + two `$order->save()`** in `restore_order_taxes()` is an N+1 write pattern on the failure path. *Fix: mutate items, then a single `$order->save()`.* *(perf)*
- **`WC()->customer` mutation not in `try/finally`** — if `calculate_tax()` throws, the restore line is skipped and global customer state leaks to later orders in a batch. *Fix: use try/finally.* *(perf/correctness)*
- **Same-request cart→programmatic reuse** — if the cart/checkout path set `response_rate_ids` (flag stays `false`) and a programmatic recalc of a *different* order then fires, the early `! empty( $this->response_rate_ids )` gate returns early and reuses the cart's rate IDs. Edge case, but possible. *(correctness)*
- **`@since x.x.x` placeholders** left in all new property/method docblocks (`:83,92,100` and method blocks); file convention uses concrete versions. *Fix: set to `@since 3.6.7`.* *(docs)*

### Nits (optional)
- Non-Yoda conditionals in new code (e.g. `$updated_order->get_id() === $order->get_id()`) vs. the WC standard. *(conventions)*
- Unused `$rate_ids = $reflection->getValue( $integration );` in the happy-path test (`:589`) — dead read; the real assertion re-fetches. *(tests)*
- `test_calculate_order_taxes_hook_registered_after_init` (`:92`) calls `init()`, which early-returns unless store country is supported and `is_enabled()` — can pass/fail on the ambient test-store country. *Fix: stub a supported country.* *(tests)*
- Comment "Adjust line_total … to match what override_woocommerce_tax_rates expects" (`:~2289`) states the what but not the why. *(docs)*

iyut:
> It seems to be that "the tax gets wiped to 0 when a REST update changes the address." So I think this is more of a data integrity problem. We should be preserving the existing tax instead of recomputing it.

@dustinparker , yeah, I agree on this one. Which I think @Abdalsalaam has worked on the changes in [this commit](https://github.com/Automattic/woocommerce-services/pull/2969/changes/c9d8153f9021562dde2d89f62f725f534fe155a7).

iyut:
> What do you think? Does that line up with what you're seeing, or is there a case where recalculating is actually the behavior we want?

I was just thinking that probably there is a possibility of shipping address changes when they are updating the order using `PUT` Rest API. That's why I consider to recalculate.

bartech (approved):
Approving — the `c9d8153` rewrite fully resolves the earlier blocking feedback (the abandoned `2f75fd1` approach that recomputed tax via a synchronous TaxJar API call on every non-AJAX recalc, which was both a compliance and a latency/rate-limit concern).

The pivot to a pure snapshot → restore is the right call:
- Snapshots and restores the recorded tax lines with full metadata (rate_code / label / compound / rate_percent).
- Drops the TaxJar call and failure-flag plumbing entirely.
- Gates out first-time calculations, and replaces the leaky per-call restore closure with an order-id-keyed map + a single named handler.

Verified the hook mechanics against WC core: `before_calculate_taxes` fires pre-wipe (clean snapshot) and `after_calculate_totals` fires last (restore is the final word); the total-rebase isolates the non-tax portion correctly.

One thing worth a release-note line (not blocking): the preservation gate is intentionally broad, so a legitimately tax-affecting REST edit (e.g. adding a taxable item) will also retain the old tax and leave the new item untaxed. That's the accepted "order tax is a record of what was charged" tradeoff — just worth documenting.

Note: GitHub still shows my prior CHANGES_REQUESTED, but that was against the pre-rewrite commit — this approval supersedes it.

dustinparker (approved):
This fixes the issue, and the change to snapshotting and restoring the recorded tax is the right call. Once an order is placed its tax is the record of what was charged, so preserving the existing lines with their full metadata (rather than recomputing them against the changed address) is exactly the behavior we want, and it resolves the concern from my first pass. First-time calculations are still left to run normally, so a new order taxes as before.

PR #2972, opened 2026-07-02
Fix/issue WOOTAX-297 - Prevent fatal error on Atomic sites from malformed API client require path

iyut:
## Description

WooCommerce Services was throwing a PHP fatal error on Atomic (WordPress.com) sites on every page load, accounting for ~3,833 errors over 7 days. The root cause is a misuse of `plugin_basename()` in `classes/class-wc-connect-api-client-live.php` — the function is passed a bare relative filename instead of an absolute path. On Atomic, Jetpack's symlink entries in WordPress's `$wp_plugin_paths` table corrupt the path, producing `jetpackclass-wc-connect-api-client.php` which PHP cannot open.

The fix replaces `plugin_basename( 'class-wc-connect-api-client.php' )` with `__DIR__ . '/class-wc-connect-api-client.php'`, using PHP's magic constant that resolves to the file's own directory at compile time — bypassing WordPress path manipulation entirely. This matches the pattern used by every other `require_once` in the codebase.

### Related issue(s)

Fixes [WOOTAX-297](https://linear.app/a8c/issue/WOOTAX-297/fatal-error-failed-opening-required-jetpackclass-wc-connect-api), [WOOTAX-300](https://linear.app/a8c/issue/WOOTAX-300/harden-bare-relative-require-once-calls-in-class-wc-connect)

**Before (Atomic sites only):**
1. Install WooCommerce Services on an Atomic (WordPress.com) site with Jetpack active
2. Activate the plugin
3. Observe: `PHP Fatal error: Uncaught Error: Failed opening required 'jetpackclass-wc-connect-api-client.php'`

**After:**
1. Install WooCommerce Services on an Atomic site with Jetpack active
2. Activate the plugin
3. Plugin loads without error

N/A — no UI changes.

### Checklist

- [ ] unit tests

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Tax Changelog ***
 
+= 3.6.7 - 2026-xx-xx =
+* Fix   - Prevent fatal error on Atomic sites caused by incorrect path resolution when loading the API client class.
+
 = 3.6.6 - 2026-06-22 =
 * Tweak - WooCommerce 10.9 Compatibility.
 
--- a/classes/class-wc-connect-api-client-live.php
+++ b/classes/class-wc-connect-api-client-live.php
@@ -10,7 +10,7 @@ if ( ! defined( 'WOOCOMMERCE_CONNECT_SERVER_URL' ) ) {
 }
 
 if ( ! class_exists( 'WC_Connect_API_Client_Live' ) ) {
-	require_once plugin_basename( 'class-wc-connect-api-client.php' );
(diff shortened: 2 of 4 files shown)


dustinparker (approved):
This fixes the fatal on Atomic sites. Resolving the require with `__DIR__` sidesteps the `plugin_basename()` path corruption that produced `jetpackclass-wc-connect-api-client.php`, and it matches how every other require in the plugin already loads its dependencies, so normal installs behave exactly as before.

Issue #2638, opened 2023-03-03
Fix PHP tests


waclawjacek:
PHP tests are currently not executing. Please make sure that PHP tests run both in GitHub Actions and when ran locally using `composer run test`.

PR #2973, opened 2026-07-02
[WOOTAX-27] Fix - Fail CI when PHPUnit runs zero tests

### LINKED ISSUE #2645: Fail CI pipeline when PHP tests don't run
A while back our PHP tests started failing and we didn't notice (#2638). I did some digging and learned that this is because when the PHPUnit execution thinks there's nothing to run, it will just end with the exit code 0 (= success).

There is a `--fail-on-warning` argument that can be used on the CLI now but it's not supported by the PHPUnit versions we are using (7 and 8, conditionally). https://github.com/sebastianbergmann/phpunit/issues/4299

We could just wait until we can use these versions or we could write a shell script that will detect this. I tried experimenting with `--list-tests`/`--list-suites` combined with `grep` and `wc -l` but I didn't achieve anything I'd be happy with.

My experiment:

```
#!/bin/bash
vendor/bin/phpunit -c phpunit.xml.dist | grep 'No tests executed'

if [ $? -eq 0 ]; then
        # No tests were executed.
        exit 1
fi
```

The problem with the above idea is that the entire test suite has to run which takes a while.

bartech:
## Description

PHPUnit exits with code `0` even when it executes **zero** tests (it prints `No tests executed!`). A broken bootstrap, a bad path in `phpunit.xml.dist`, or a suite that fails to load can therefore make CI pass green without running anything — which happened before in #2638 and went unnoticed.

This guards the direct `./vendor/bin/phpunit -c phpunit.xml.dist` invocation in both CI workflows (`php_tests.yml` and `unit_test_runner.yml`):

- Preserves PHPUnit's real exit code via `PIPESTATUS`, so genuine test failures still propagate (`tee` can't mask them).
- Additionally fails the build when the run reports `No tests executed`.

Because CI already runs the full suite, the guard adds **no** extra runtime, and it works on the pinned PHPUnit 8 — no version bump (the built-in `--fail-on-empty-test-suite` is 9.6+).

### Related issue(s)

Closes [WOOTAX-27](https://linear.app/a8c/issue/WOOTAX-27)
Fixes #2645
Context: #2638 — the incident where the suite silently stopped running yet CI stayed green.

CI-only change, no UI. To verify by intentionally breaking the suite (the ticket's own acceptance step):

1. Make zero tests match — e.g. point the `<directory>` in `phpunit.xml.dist` at an empty folder, or rename `tests/php/test-*.php` so none match the `test-` prefix.
2. **Before** (`trunk`): `./vendor/bin/phpunit -c phpunit.xml.dist` prints `No tests executed!` and exits `0` → CI passes (the bug).
3. **After** (this PR): the step prints `::error::PHPUnit executed zero tests — failing the build.` and exits `1` → CI fails (the fix).
4. Normal runs are unaffected: a green suite still passes, and real test failures still exit non-zero via `PIPESTATUS`.

The guard's exit-code logic was validated against 5 scenarios (pass, real-failure, empty, empty+non-zero, fatal) — each yields the correct exit code.

### Checklist

- [ ] unit tests

> CI-only change (workflow YAML): no user-facing behavior change, so no `changelog.txt` / `readme.txt` entry — consistent with recent CI/dev-tooling merges (e.g. WOOTAX-287, #2961). No unit-test framework covers workflow YAML; the guard logic was verified via shell simulation (see testing steps above).

### DIFF
--- a/.github/workflows/php_tests.yml
+++ b/.github/workflows/php_tests.yml
@@ -95,4 +95,18 @@ jobs:
       - name: Install Composer dependencies
         run: composer install
       - name: Run PHPUnit tests
-        run: ./vendor/bin/phpunit -c phpunit.xml.dist
+        # PHPUnit exits 0 even when it executes zero tests (it prints
+        # "No tests executed!"), so a broken bootstrap or misconfigured suite
+        # silently turned CI green before (see #2638). Preserve PHPUnit's own
+        # exit code via PIPESTATUS (tee must not mask it) and additionally fail
+        # the build on an empty run.
+        shell: bash
+        run: |
+          set +e
+          ./vendor/bin/phpunit -c phpunit.xml.dist 2>&1 | tee phpunit-output.log
+          status=${PIPESTATUS[0]}
+          if grep -q 'No tests executed' phpunit-output.log; then
+            echo "::error::PHPUnit executed zero tests — failing the build."
+            exit 1
+          fi
+          exit "$status"
--- a/.github/workflows/unit_test_runner.yml
+++ b/.github/workflows/unit_test_runner.yml
@@ -99,4 +99,18 @@ jobs:
       - name: Install Composer dependencies
         run: composer install
       - name: Run PHPUnit tests
-        run: ./vendor/bin/phpunit -c phpunit.xml.dist
+        # PHPUnit exits 0 even when it executes zero tests (it prints
+        # "No tests executed!"), so a broken bootstrap or misconfigured suite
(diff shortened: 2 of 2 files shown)


CezaryDrewniak (approved):
**Looks good**

Bash guard is correct, `shell: bash` is required for `${PIPESTATUS[0]}` (good catch), and `2>&1 | tee` covers stderr where PHPUnit writes "No tests executed!". No other phpunit invocations to guard. Agreed on skipping changelog/readme, CI-only change

Approving

PR #2974, opened 2026-07-09
Fix/issue WOOTAX-303 - Harden StoreApiExtendSchema against container resolution failure

iyut:
## Description

WOOTAX-298 (#2971) added a `class_exists( '\Automattic\WooCommerce\StoreApi\StoreApi' )` guard so the plugin no longer fatals on WooCommerce versions that don't ship the Store API. This PR closes the residual, narrower path through the same class that can still fatal even when that class **is** present: on a partial or broken WooCommerce install where the container fails to resolve `ExtendSchema`.

Previously, if `StoreApi::container()->get( ExtendSchema::class )` threw, `StoreApiExtendSchema` caught it and logged a debug line but never assigned `self::$instance`. Three gaps:

1. `$instance` was a typed static with no default, so `instance()` returned an uninitialized typed property and fataled with *"Typed static property must not be accessed before initialization."*
2. The `catch` only handled `Exception` — an `Error` or `TypeError` from the container would propagate straight up.
3. The `: ExtendSchema` return type meant the wrapper couldn't signal failure, so the caller kept going and called `extend_store()` on an unusable instance.

**The fix** (`src/StoreApi/StoreApiExtendSchema.php`):
- `private static ?ExtendSchema $instance = null;` — nullable and defaulted, so it can never be read uninitialized.
- Added `private static bool $attempted = false;`, set **before** the `try`, so container resolution (and its logging) runs at most once per request even when it throws.
- `catch ( Throwable $e )` instead of `Exception` — covers `Error`/`TypeError`.
- `public static function instance(): ?ExtendSchema` — nullable return to signal failure.

And in `woocommerce-services.php`, `register_store_api_extensions()` resolves through a `get_store_api_extend_schema()` seam and returns early when the result is not an `ExtendSchema` instance, before constructing the typed-constructor controllers. If the container can't hand us an `ExtendSchema`, registration is skipped quietly instead of fataling.

This is defensive hardening for a low-probability path, not a fix for an observed crash.

### Related issue(s)

- Fixes [WOOTAX-303](https://linear.app/a8c/issue/WOOTAX-303/storeapiextendschemainstance-can-still-fatal-if-the-container-fails-to)
- Follow-up to WOOTAX-298 (#2971)
- Follow-up: [WOOTAX-304](https://linear.app/a8c/issue/WOOTAX-304/storeapiextensioncontroller-still-catches-only-exception-not-throwable) - widen `StoreApiExtensionController` catch blocks to `Throwable` (same broken-install exposure, one step later in the same call path)

This path is hard to hit on a healthy store — you need a partial/broken WooCommerce install where the top-level `StoreApi` class exists but the container cannot resolve `ExtendSchema` (so `StoreApi::container()->get( ExtendSchema::class )` throws).

**Before:** on such an install, `StoreApiExtendSchema::instance()` fatals with *"Typed static property `$instance` must not be accessed before initialization"* (or an uncaught `Error`/`TypeError` from the container), taking down the request during `woocommerce_blocks_loaded`.

**After:** resolution failure is caught (`Throwable`), `instance()` returns `null`, and `register_store_api_extensions()` skips registration quietly — no fatal.

Covered by unit tests rather than a manual repro:
- Both new tests pass; full suite `OK (186 tests, 403 assertions)`.
- Each test was verified non-vacuous by temporarily reverting the guard and confirming it fails with the exact predicted error.

### Checklist

- [x] unit tests

### DIFF
--- a/src/StoreApi/StoreApiExtendSchema.php
+++ b/src/StoreApi/StoreApiExtendSchema.php
@@ -47,15 +47,30 @@ class StoreApiExtendSchema {
 
 	/**
 	 * ExtendSchemaService constructor.
+	 *
+	 * Protected rather than private so tests can subclass and override
+	 * resolve_extend_schema() to exercise the resolution-failure path.
 	 */
-	private function __construct() {
+	protected function __construct() {
 		self::$attempted = true;
 
 		try {
-			self::$instance = StoreApi::container()->get( ExtendSchema::class );
+			self::$instance = static::resolve_extend_schema();
 		} catch ( Throwable $e ) {
 			wc_get_logger()->debug( 'Failed to get ExtendSchema instance.', array( 'exception' => $e ) );
 		}
+	}
+
+	/**
+	 * Resolve the ExtendSchema instance from the Store API container.
+	 *
+	 * Extracted as a seam so a broken container can be simulated in tests (subclass
+	 * and override to throw) without needing a genuinely partial WooCommerce install.
+	 *
+	 * @return ExtendSchema
+	 */
+	protected static function resolve_extend_schema(): ExtendSchema {
+		return StoreApi::container()->get( ExtendSchema::class );
 	}
 
 	/**
@@ -67,7 +82,7 @@ class StoreApiExtendSchema {
 	 */
 	public static function instance(): ?ExtendSchema {
 		if ( ! self::$attempted ) {
-			new self();
+			new static();
 		}
 
 		return self::$instance;
--- a/src/StoreApi/StoreApiExtendSchema.php
+++ b/src/StoreApi/StoreApiExtendSchema.php
@@ -31,8 +31,10 @@ class StoreApiExtendSchema {
 	/**
 	 * Whether resolving the ExtendSchema instance has been attempted.
 	 *
-	 * Guards against re-running container resolution (and re-logging) on every
-	 * request when resolution fails on an install with a broken Store API.
+	 * Guards against re-running container resolution (and re-logging) more than
+	 * once within a single request when resolution fails. Note this cannot span
+	 * requests: PHP statics reset on every page load, so a broken install still
+	 * logs once per request.
 	 *
 	 * @var bool
 	 */
@@ -57,7 +59,13 @@ class StoreApiExtendSchema {
 		try {
 			self::$instance = static::resolve_extend_schema();
 		} catch ( Throwable $e ) {
-			wc_get_logger()->debug( 'Failed to get ExtendSchema instance.', array( 'exception' => $e ) );
+			wc_get_logger()->debug(
+				'Failed to get ExtendSchema instance.',
+				array(
+					'source'    => 'woocommerce-services',
+					'exception' => $e,
+				)
+			);
 		}
 	}
 
--- a/src/StoreApi/StoreApiExtendSchema.php
+++ b/src/StoreApi/StoreApiExtendSchema.php
@@ -11,7 +11,7 @@ namespace Automattic\WCServices\StoreApi;
 
 use Automattic\WooCommerce\StoreApi\Schemas\ExtendSchema;
 use Automattic\WooCommerce\StoreApi\StoreApi;
-use Exception;
+use Throwable;
 
 defined( 'ABSPATH' ) || exit;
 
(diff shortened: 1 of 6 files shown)


CezaryDrewniak (approved):
LGTM. The `Throwable` widening is the right call here, the `$attempted` latch fixes the re-resolution loop the old `! isset()` check had, and the `get_store_api_extend_schema()` seam keeps the loader test surface clean. Full suite green on my end (189/407, no regressions).

One non-blocker for the WOOTAX-304 follow-up: register_store_api_extensions() now silently returns when the container can't resolve ExtendSchema, while log_store_api_unavailable() only fires for the missing-class branch and its message is specific to that case. A throttled "container resolution failed" notice (mirroring the WOOTAX-298 pattern) would help diagnose broken installs.

bartech (approved):
Approving. The fix cleanly closes all three original fatal vectors (uninitialized typed static, uncaught Error/TypeError, non-nullable return into a typed constructor).

- `catch(Throwable)` is correctly scoped to the single container call rather than wrapping a broader block, so it doesn't swallow unrelated errors.
- Degradation is graceful: `instance()` returns null, registration skips, and the Store API simply omits this plugin's extensions — no corrupted response.
- The caller's `instanceof ExtendSchema` guard is autoload-safe and rejects both null and wrong types.
- No new PHP/WC version floor; test coverage is thorough with documented non-vacuity.

One nit (non-blocking, already tracked as WOOTAX-304): the container-failure path logs only at `debug`, while the missing-class path emits a throttled `notice` — a genuinely broken install ends up diagnostically silent. Fine to defer.

src/StoreApi/StoreApiExtendSchema.php
```diff
@@ -22,9 +22,21 @@ class StoreApiExtendSchema {
...
+	 *
+	 * @var ExtendSchema|null
+	 */
+	private static ?ExtendSchema $instance = null;
+
+	/**
+	 * Whether resolving the ExtendSchema instance has been attempted.
+	 *
+	 * Guards against re-running container resolution (and re-logging) on every
+	 * request when resolution fails on an install with a broken Store API.
```
Abdalsalaam: PHP statics reset on every request, so $attempted cannot guard 'on every request'. extend_store_api() is hooked on woocommerce_blocks_loaded (fires on nearly every request) and there is exactly one production call site of instance() (woocommerce-services.php:1064), so the latch dedupes nothing in production. On a broken install the debug() call at line 60 still writes on every page load - exactly the flooding that #2971 solved for the sibling path with a DAY_IN_SECONDS transient in log_store_api_unavailable() (woocommerce-services.php:1018-1031). Four independent lenses converged on this.

```suggestion
	 * Guards against re-running container resolution (and re-logging) more than
	 * once within a single request when resolution fails. Note this cannot span
	 * requests: PHP statics reset on every page load, so a broken install still
	 * logs once per request.
```

src/StoreApi/StoreApiExtendSchema.php
```diff
@@ -35,21 +47,42 @@ class StoreApiExtendSchema {
...
-	private function __construct() {
+	protected function __construct() {
+		self::$attempted = true;
+
 		try {
-			self::$instance = StoreApi::container()->get( ExtendSchema::class );
-		} catch ( Exception $e ) {
+			self::$instance = static::resolve_extend_schema();
+		} catch ( Throwable $e ) {
 			wc_get_logger()->debug( 'Failed to get ExtendSchema instance.', array( 'exception' => $e ) );
```
Abdalsalaam: WC_Log_Handler_File::handle() uses $context['source'] as the log handle and falls back to a generic wc-logs/log-*.log when absent. The sibling log_store_api_unavailable() (woocommerce-services.php:1027) does pass array( 'source' => 'woocommerce-services' ). This PR's purpose is making a silent container failure diagnosable, so the one breadcrumb should be findable in the plugin's own log. The line is textually unchanged but the diff rewrites its enclosing catch and it is now the only signal for the Throwable class this PR newly catches.

```suggestion
			wc_get_logger()->debug(
				'Failed to get ExtendSchema instance.',
				array(
					'source'    => 'woocommerce-services',
					'exception' => $e,
				)
			);
```

iyut: @Abdalsalaam Thanks for catching this! I've made the changes

PR #2975, opened 2026-07-14
Quality: Undefined variable used in subscription activation logic

tomaioo:
## Summary

Quality: Undefined variable used in subscription activation logic

## Problem

**Severity**: `High` | **File**: `classes/class-wc-rest-connect-subscription-activate-controller.php:L22`

The `$activation_response` variable is used in `wp_remote_retrieve_response_code` and `wp_remote_retrieve_body`, but it is never defined in the scope. It appears to be a typo, as the correct variable should be `$response`.

## Solution

Replace `$activation_response` with `$response` to correctly evaluate the API response.

## Changes

- `classes/class-wc-rest-connect-subscription-activate-controller.php` (modified)


bartech (approved):
Nice catch. Looks good. Thank you!

PR #2976, opened 2026-07-15
[WOOTAX-306] Add - AGENTS.md backward compatibility guardrail

Abdalsalaam:
## Description

Adds a Backward Compatibility section to AGENTS.md, copied from [WooCommerce Core's canonical guardrail](https://github.com/woocommerce/woocommerce/blob/trunk/AGENTS.md#backward-compatibility) and adapted to this plugin.

The core rule: any signature change to externally exposed code is high-risk and must state its BC impact in the PR description. Following review feedback, the section has been refreshed against Core trunk to pick up woocommerce/woocommerce#66572, which widened the guardrail beyond PHP signatures to cover hooks and filters, global state, multisite, and install layout, and added an agent checklist.

Adapted to this plugin:

- The global `WC_Connect_*` classes stay called out as exposed surface, and the newer `Automattic\WCServices\*` namespace is noted as no safer by default.
- The `wc_connect_*`, `wc_services_*`, `wcservices_*`, and `wcship_*` hooks are named as public contracts.
- Deprecation is tied to this repo's grandfathering rule: because shipping is compatibility-only for eligible installs, removing or narrowing an existing shipping surface is a BC change for those stores rather than cleanup.
- The global-state rule is grounded in the REST and Store API tax paths, which have regressed this way before (see the recent TaxJar tax-line fix on REST order updates).

Docs-only change; no changelog entry, matching prior AGENTS.md-only PRs. No code touched, so there is no BC impact from this PR itself.

Linear: WOOTAX-306

### Testing instructions

1. Read the Backward Compatibility section in AGENTS.md.
2. Compare it against [Core's trunk section](https://github.com/woocommerce/woocommerce/blob/trunk/AGENTS.md#backward-compatibility) and confirm every rule is carried over: the signature rule, interface additions/removals, deprecate-don't-rename, the wider surface (hooks, global state, multisite, install layout), and the agent checklist.
3. Confirm the plugin-specific details are accurate for this repo.

### Checklist

- [ ] unit tests - N/A, docs-only
- [ ] `changelog.txt` entry added - N/A, docs-only
- [ ] `readme.txt` entry added - N/A, docs-only

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -52,6 +52,16 @@ The husky pre-commit hook (`bin/wc-phpcbf.sh` via lint-staged) auto-fixes staged
 - Keep changes in source files; do not hand-edit generated build outputs in `dist/`.
 - Follow existing WordPress/WooCommerce coding standards and linting rules in this repository.
 
+## Backward Compatibility
+
+The `WC_Connect_*` classes live in the global namespace and are referenced by third-party code; treat them all as externally exposed.
+
+Any change to the signature of a public or externally exposed class, interface, function, method, or hook is high-risk and must state its backward-compatibility impact in the PR description.
+
+- Treat a symbol as externally exposed when code outside this repository (other plugins, themes, or site snippets) can call, extend, implement, or hook into it. When in doubt, assume it is exposed and state the BC impact.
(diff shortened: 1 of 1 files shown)


bartech (approved):
Approving. Verified against the codebase: `WC_Connect_*` resolves to 39 real global-namespace classes, so "referenced by third-party code; treat them all as externally exposed" holds. Faithful, well-tailored condensation of Core's `AGENTS.md#backward-compatibility`. Doc-only; CI green. 👍

PR #2977, opened 2026-07-15
[WOOTAX-294] Fix React 19 incompatibility in block cart/checkout store notices

Abdalsalaam:
## Description

The store notices bundle on the block cart and checkout created its slot fill elements with the React copy bundled in the plugin (React 16) and handed them to the WooCommerce Blocks checkout React through `registerPlugin` and `ExperimentalOrderMeta`. React 19 rejects elements created by an older React runtime, so under React 19 the fill throws inside its error boundary and the TaxJar store notices are lost.

React 19 is currently reverted in Gutenberg and planned for a later WordPress release, so this is not breaking in production today, but it needs fixing before that lands.

This builds the store notices elements with the WordPress provided React (`window.wp.element`) instead of the bundled React, so they are created by the same React that renders them. It works on React 18 (current installs) and React 19. The bundle no longer inlines a React element factory and drops from about 9 KB to about 2 KB.

It also fixes the script registration. `WooCommerceBlocksIntegration::register_script()` looked for the asset file by calling `file_exists()` on a URL, which never matched, so the script registered with no dependencies and relied on load order. On WooCommerce 11 this surfaces as a console warning that the script accessed `wc.blocksCheckout` without declaring `wc-blocks-checkout`. The script now declares its real dependencies (`wp-plugins`, `wp-element`, `wp-data`, `wc-blocks-checkout`) and the asset lookup uses the dist directory path.

### Backward compatibility

The `WooCommerceBlocksIntegration` constructor parameter changes from a base URL to a dist directory path and becomes optional. No third-party instantiation path is expected, since the class is only instantiated through the Blocks integration registry. `window.wp.element` was already read by this bundle at load time, so there is no new runtime requirement, and the dependency handles exist on all WooCommerce versions that fire the blocks integration hooks.

The admin app (tax and label screens) is out of scope on purpose. It mounts its own isolated React 16 root with `ReactDOM.render` and both creates and renders its elements with the same bundled React, so there is no cross version boundary for React 19 to reject. The issue's repro pointed at the admin UI from the initial source analysis, but the confirmed break is the cart and checkout store notices fill. The admin UI was verified to render under React 19 without this error.

### Related issue

WOOTAX-294

### Steps to test

1. WordPress + WooCommerce with this plugin active, store base country US, block based Cart and Checkout pages.
2. Add a product to the cart.
3. Force a store notice into the session. Either trigger a real TaxJar error (US address with an invalid ZIP while automated taxes are on), or use a small mu-plugin:
   ```php
   add_action( 'woocommerce_after_calculate_totals', function () {
       if ( empty( WC()->session ) ) {
           return;
       }
       WC()->session->set( 'wcservices_store_notices', array(
           'error' => array( array( 'message' => 'Test notice', 'data' => '', 'group' => 'taxjar' ) ),
       ) );
   }, 5 );
   ```
4. React 18 (no Gutenberg): open the Cart and Checkout blocks. The notice renders, no console errors.
5. React 19 (activate Gutenberg 23.3.1, the last version shipping React 19): open the Cart and Checkout blocks. Before this change the console throws the error below and the notice is lost; with this change the notice renders and the console is clean.
6. Confirm the store notices script declares `wc-blocks-checkout`, so the WooCommerce 11 dependency warning is gone.

Error before the fix:

```
Error: A React Element from an older version of React was rendered. This is not supported. It can happen if:
- Multiple copies of the "react" package is used.
- A library pre-bundled an old copy of "react" or "react/jsx-runtime".
- A compiler tries to "inline" JSX instead of using the runtime.
    at throwOnInvalidObjectTypeImpl (.../gutenberg/build/scripts/vendors/react-dom.js?ver=19.2.4)
    at reconcileChildFibersImpl (.../gutenberg/build/scripts/vendors/react-dom.js?ver=19.2.4)
    at updateFunctionComponent (.../gutenberg/build/scripts/vendors/react-dom.js?ver=19.2.4)
The above error occurred in the <render> component, caught by PluginErrorBoundary.
```

`npm run test-client` and the PHPUnit suite pass. New tests cover the store notices entry using the host React, the component guarding missing extension data, and the script registering with its dependencies.

### Checklist

- [x] unit tests


Abdalsalaam:
## Test evidence

Automated: `npm run test-client` (225 tests) and the PHPUnit suite (200 tests) pass. ESLint and PHPCS clean on the changed files.

Manual, block cart and checkout with a forced store notice:

| Env | Before (bundled React) | After (host React) |
|---|---|---|
| React 18.3.1 (no Gutenberg) | notice renders, console clean | notice renders, console clean |
| React 19.2.4 (Gutenberg 23.3.1) | "older version of React" error, notice lost | notice renders, console clean |

Also verified:
- Non-US shipping address: notice correctly withheld, no error.
- No notice in session: cart and checkout render clean.
- Classic shortcode cart: store notices script not loaded.
- Woo Tax admin UI under React 19: renders without this error (isolated React 16 root, out of scope).
- WooCommerce 11 dependency warning for `wc-blocks-checkout` on this script: gone after the fix.

Checked locally (WC 10.9) and on a live site (WordPress 7.0.1, WooCommerce 11.0-beta.1), same results.

CezaryDrewniak (approved):
LGTM

Confirmed the React 19 fix works (slot-fill elements built with `window.wp.element` only, verified the `createdByHost` invariant in the new entry test) and the `file_exists(URL)` -> filesystem-path fix holds. 

All 203 PHPUnit + 10 client tests green locally. The new ESLint guard against importing bundled React into the store-notices tree is a nice touch to keep it from regressing

bartech (commented):
Leaving this as a comment rather than an approval. The code is correct and I'd merge it happily — what's holding it is that the changelog entry now targets a version that has already shipped, so the conflict needs a decision from you rather than a mechanical resolution at merge time.

## 🟡 The changelog entry targets a version that has already shipped

The entry is added under `= 3.6.8 - 2026-xx-xx =`. Since this branch was cut, 3.6.8 has **shipped** and two more releases have gone out:

| Version | State on trunk |
|---|---|
| 3.6.8 | **released 2026-07-15** ← the PR still writes here |
| 3.6.9 | released 2026-07-20 |
| 3.6.10 | released 2026-07-27 (current plugin header) |
| **3.6.11** | `2026-xx-xx` — the unreleased block |

Why this needs a decision: the obvious "keep both sides" resolution produces

```
= 3.6.8 - 2026-07-15 =
* Fix   - TaxJar tax lines wiped when REST API order update includes address change.
* Fix   - Prevent fatal error on sites running WooCommerce versions without StoreApi support.
* Fix   - Prevent unbounded growth of WooCommerce tax rate rows when checkout city contains a semicolon.
* Fix   - React 19 compatibility for store notices on the block cart and checkout.   ← NOT in 3.6.8
```

…which files an unshipped fix under an already-released version, in both `changelog.txt` and `readme.txt`, in a public repo. The React 19 line belongs under **3.6.11**.

The rest of the merge is clean, for what it's worth — I ran a trial merge and only those two changelog surfaces conflict; `woocommerce-services.php` auto-merges correctly and the version header correctly stays 3.6.10.

**Recommend rebasing on `trunk` and moving the entry to 3.6.11.** Worth doing anyway: the branch is 16 days old and its green CI ran against a base that has since moved by 21 files and three releases.

## 🟢 "No public signatures change" isn't quite accurate

```php
- public function __construct( string $wc_connect_base_url ) {
-     $this->base_url = $wc_connect_base_url;
+ public function __construct( string $dist_dir = '' ) {
+     $this->dist_dir = '' !== $dist_dir ? $dist_dir : WCSERVICES_PLUGIN_DIST_DIR;
```

The constructor is public and its first parameter **changed meaning** — it was a base *URL*, it's now a filesystem *path*. Making it optional is BC-friendly for callers passing nothing, but a caller passing the old value hands a URL where a path is expected: `file_exists()` on a URL returns false, so it silently falls back to the default rather than erroring.

The actual risk is low — `WooCommerceBlocksIntegration implements IntegrationInterface` and is only instantiated through the Blocks integration registry, so third-party instantiation is implausible. I'd just phrase the BC section as *"the constructor parameter changes from a base URL to a dist path and becomes optional; no third-party instantiation path is expected"* rather than "no public signatures change" — this repo is public, and the BC-guardrail work landing across these repos right now makes precision here worth having.

## ℹ️ The asset-path fix is currently inert

The description says *"the asset lookup uses the dist directory path"*, which is true, but the inline comment notes:

> *"The webpack build does not emit asset files, so the defaults below are the live dependency list."*

So `file_exists( $script_asset_path )` is still always false, and the **hardcoded** dependency list (`wp-plugins`, `wp-element`, `wp-data`, `wc-blocks-checkout`) is what actually takes effect today. The path correction is right and future-proofs the lookup — it just isn't what fixes the missing dependencies. Worth phrasing so nobody later assumes an asset file is being consumed.

## What's excellent here

**This is the only PR I've reviewed across these repos recently that makes its invariant *executable* rather than merely documented.** The `.eslintrc.js` `no-restricted-imports` override bans `import React from 'react'` in `client/store-notices.js` and `client/components/store/notices/*.js`, with an error message pointing at the reason. A future contributor physically cannot reintroduce this bug without the linter stopping them.

That's worth calling out because the failure mode being fixed here is exactly the kind that comes back: a comment saying "don't do X" is invisible to someone adding a new file six months from now. This isn't.

**The guard's scope is correctly drawn, too.** The glob deliberately excludes `test/`, and I confirmed the only surviving `import React from 'react'` is `client/components/store/notices/test/store-notices.js:6` — correct, since that runs under Jest with its own React rather than inside the host React tree.

**A second real bug fixed along the way:** `register_script()` called `file_exists()` on a **URL**, which never matched, so the script registered with no declared dependencies and relied on load order — surfacing on WooCommerce 11 as a console warning about accessing `wc.blocksCheckout` without declaring `wc-blocks-checkout`. Now the real dependencies are declared.

**345 lines of new tests** across three files, including a PHP one that exercises the constructor's new dist-dir behaviour via a temp directory — which is the reason the parameter was made injectable in the first place, so the signature change earns its keep.

Scoping to the block cart/checkout notices only, and leaving the admin app to the broader toolchain work, is also the right call — that needs the webpack/dependency-extraction modernization rather than a spot fix.

## What I didn't verify

No build, no Jest/PHPUnit run, and no React 19 reproduction — that needs Gutenberg 23.3.1 (the last version shipping React 19) on a live store. The fix's correctness rests on the well-documented React 19 cross-runtime rejection, which isn't in doubt; what stays unverified is the *original* symptom in this specific plugin, which the linked issue itself notes was never directly observed here. Worth someone confirming step 6 below at some point.

Suggested check once rebased: load the block checkout under Gutenberg 23.3.1 on this branch (notice renders) and on trunk (notice lost) — that's the step that validates the original report.

bartech (approved):
## Follow-up review — delta since my 2026-07-31 review (`a633576`)

**Verdict: Approve.** Both findings addressed, nothing new.

### The changelog rebase

My note last time was "ask for a rebase, don't resolve it yourself" — the naive resolution would have filed an unshipped fix under an already-released version. The resolution is right:

```
= 3.6.11 - 2026-xx-xx =   ← unreleased
* Fix - Prevent a fatal error during cart and checkout tax calculation…
* Fix - React 19 compatibility for store notices on the block cart and checkout.   ← here

= 3.6.8 - 2026-07-15 =    ← shipped, retains its real contents, no residue
```

Applied identically in **both** `changelog.txt` and `readme.txt`, same position, same text — the two public changelogs stay in sync.

I also checked the merge didn't clobber trunk: `git diff 59cf387 39af250` (merge result vs the trunk parent) is exactly this PR's own change set — `.eslintrc.js`, the store-notices client and tests, `WooCommerceBlocksIntegration.php` and tests, the `woocommerce-services.php` constructor-arg removal, and one added changelog line each. No conflict markers.

### The BC statement

The *Backward compatibility* section now states the constructor signature change and why no third-party instantiation path is affected, instead of asserting no signatures changed. Accurate.

Version headers correctly left at `3.6.10` for the release process. CI is green across all 9 checks. One thing to glance at before merging: `gh pr view` reports `mergeable: UNKNOWN`, though the branch merged trunk cleanly.

PR #2978, opened 2026-07-15
Fix - Validate TaxJar response before applying tax rate override (WOOTAX-74)

bartech:
### Summary

Fixes [WOOTAX-74](https://linear.app/a8c/issue/WOOTAX-74/validate-taxjar-response-object).

`WC_Connect_TaxJar_Integration::maybe_override_taxjar_tax()` assumed the TaxJar tax response was always well-formed. When a rate-override filter (e.g. `woocommerce_services_override_tax_rate`) changes the rate, the method rewrote properties on `breakdown`, `breakdown->shipping` and every `line_item` **unconditionally**. If TaxJar returned an incomplete response — a `null`/non-object line item, or a missing `breakdown`/`shipping` member — this fataled during cart and checkout tax calculation:

```
PHP Fatal error: Uncaught Error: Attempt to assign property "combined_tax_rate" on null
  in .../classes/class-wc-connect-taxjar-integration.php:1042
```

A merchant hit this via a custom `woocommerce_services_override_tax_rate` filter (used to force CA tax to 0% because of a separate upstream TaxJar data issue). The malformed response took down the mini-cart / product archive pages with a white screen.

### Changes

The response is now validated at **three** levels, each covering callers the others cannot:

1. **The request boundary** — `calculate_tax()` bails on a response whose `tax` node is missing or not an object. This is what protects `get_itemized_tax_rates()`, which dereferences the tax node unconditionally; a guard inside `maybe_override_taxjar_tax()` structurally cannot reach that.
2. **The method boundary** — `maybe_override_taxjar_tax()` returns a non-object input unchanged. `calculate_tax()` already guarantees an object, so this is a safety net for third-party callers of a `public` method, preserving the behaviour `trunk` had.
3. **The inner members** — the actual reported bug. Only mutate `breakdown` and `breakdown->shipping` when present and object-typed; skip non-object line items instead of assigning properties on them; default missing `taxable_amount` / `country_taxable_amount` to `0` so the collectable math can't warn or fatal.

Behaviour is **unchanged for well-formed responses** — the override still rewrites every rate exactly as before.

### Backward compatibility

Per the AGENTS.md backward-compatibility rules: `WC_Connect_TaxJar_Integration` is a `WC_Connect_*` class in the global namespace, so it is treated as externally exposed, and `maybe_override_taxjar_tax()` is `public`.

- **Method signature: unchanged.** No parameters added, removed, reordered or type-declared. A PHP `object` type declaration was considered and deliberately rejected — it would convert a silent no-op into a hard `TypeError`, i.e. *add* a possible fatal in a PR whose purpose is removing one.
- **Behaviour for a well-formed object: unchanged.** Byte-identical output, covered by `..._applies_override_to_wellformed_response`.
- **Behaviour for a non-object input: preserved as on `trunk`.** `trunk` had `if ( ! isset( $taxjar_resp_tax ) ) { return; }`. An earlier revision of this PR removed it once `calculate_tax()` validated the node; that was a BC change to a public method and has been reverted. The guard is back, now covering `false`/array/string as well as `null`.
- **Hooks: unchanged.** `woocommerce_services_override_tax_rate` fires with the same three arguments, in the same order, at the same point. It no longer fires at all when a non-object is passed in — but on `trunk` that input returned before the filter too, so this is not a change for any real consumer.
- **Return type: documented as `mixed`.** Only the docblock changed; the method already returned its input unchanged on `trunk`'s guard path. No consumer sees a different value.
- **Globals / multisite / install layout:** not touched. This change adds only type checks inside an existing code path.

### Tests

Adds 6 test cases (9 with data-provider expansion) to `test-class-wc-connect-taxjar-integration.php`:

- `..._applies_override_to_wellformed_response` — behaviour preservation (override still rewrites all rates).
- `..._survives_null_line_item` — reproduces the exact reported fatal; fails on `trunk`, passes with this fix.
- `..._survives_missing_breakdown`
- `..._survives_missing_shipping`
- `..._returns_unchanged_without_override` — asserts every level is untouched, not just object identity.
- `..._returns_non_object_unchanged` — data provider over `null` / `false` / `array` / `string`. Each case registers a **rate-changing** filter, because the method early-returns when the rate is unchanged; without that the test would pass with no guard at all. Removing the guard fails all four with `Error: Attempt to assign property "rate" on …`.

Also clears `woocommerce_services_override_tax_rate` in `tear_down()` — the existing override tests registered filters and never removed them.

Full suite via `composer test`: **226 tests, 492 assertions green**. PHPCS clean on the changed regions.

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1658,13 +1658,8 @@ class WC_Connect_TaxJar_Integration {
 
 		// Decode Response.
 		$taxjar_response = json_decode( $response['body'] );
-		// TaxJar's `tax` node is always a JSON object; domestic and international
-		// responses differ only in the inner `breakdown` shape (US uses
-		// state_/county_/city_ fields, international uses country_ fields), not in
-		// the type of `tax` itself. Validate the type at this boundary so both
-		// maybe_override_taxjar_tax() and get_itemized_tax_rates() can rely on an
-		// object — a malformed non-object response bails here instead of fataling
-		// downstream on the first property access.
+		// Bail on a malformed response: maybe_override_taxjar_tax() and
+		// get_itemized_tax_rates() both access `tax` as an object.
 		if ( empty( $taxjar_response->tax ) || ! is_object( $taxjar_response->tax ) ) {
 			return false;
 		}
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1330,16 +1330,17 @@ class WC_Connect_TaxJar_Integration {
 	/**
 	 * This method is used to override the TaxJar result.
 	 *
+	 * The caller (calculate_tax) guarantees a non-empty object here, so this
+	 * method always receives — and returns — the tax object. Malformed *inner*
+	 * members (a missing/null breakdown or shipping, a non-object line item) are
+	 * still tolerated below; only those, not the tax node itself, are unreliable.
+	 *
 	 * @param object $taxjar_resp_tax TaxJar response object.
 	 * @param array  $body            Body of TaxJar request.
 	 *
 	 * @return object
 	 */
 	public function maybe_override_taxjar_tax( $taxjar_resp_tax, $body ) {
-		if ( ! isset( $taxjar_resp_tax ) || ! is_object( $taxjar_resp_tax ) ) {
-			return $taxjar_resp_tax;
-		}
-
 		$original_rate = isset( $taxjar_resp_tax->rate ) ? floatval( $taxjar_resp_tax->rate ) : 0.0;
 		$new_tax_rate  = floatval( apply_filters( 'woocommerce_services_override_tax_rate', $taxjar_resp_tax->rate ?? 0, $taxjar_resp_tax, $body ) );
 
@@ -1657,7 +1658,14 @@ class WC_Connect_TaxJar_Integration {
 
 		// Decode Response.
 		$taxjar_response = json_decode( $response['body'] );
-		if ( empty( $taxjar_response->tax ) ) {
+		// TaxJar's `tax` node is always a JSON object; domestic and international
+		// responses differ only in the inner `breakdown` shape (US uses
+		// state_/county_/city_ fields, international uses country_ fields), not in
+		// the type of `tax` itself. Validate the type at this boundary so both
+		// maybe_override_taxjar_tax() and get_itemized_tax_rates() can rely on an
+		// object — a malformed non-object response bails here instead of fataling
+		// downstream on the first property access.
+		if ( empty( $taxjar_response->tax ) || ! is_object( $taxjar_response->tax ) ) {
 			return false;
 		}
 		$taxjar_taxes = $this->maybe_override_taxjar_tax( $taxjar_response->tax, $body );
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1330,17 +1330,24 @@ class WC_Connect_TaxJar_Integration {
 	/**
 	 * This method is used to override the TaxJar result.
 	 *
-	 * The caller (calculate_tax) guarantees a non-empty object here, so this
-	 * method always receives — and returns — the tax object. Malformed *inner*
-	 * members (a missing/null breakdown or shipping, a non-object line item) are
-	 * still tolerated below; only those, not the tax node itself, are unreliable.
+	 * The tax node is validated by calculate_tax() before this is called, so
+	 * in-plugin callers always pass an object. The type check below is a net for
+	 * third-party callers of this public method: overriding a rate is meaningless
+	 * without a tax object, and a property write on a non-object is the very fatal
+	 * this method exists to prevent. Malformed *inner* members (a missing/null
+	 * breakdown or shipping, a non-object line item) are tolerated further down.
 	 *
-	 * @param object $taxjar_resp_tax TaxJar response object.
-	 * @param array  $body            Body of TaxJar request.
+	 * @param mixed $taxjar_resp_tax TaxJar response tax node. Expected to be an object.
+	 * @param array $body            Body of TaxJar request.
 	 *
-	 * @return object
+	 * @return mixed The tax object with its rates overridden, or the input returned
+	 *               unchanged when it is not an object.
 	 */
 	public function maybe_override_taxjar_tax( $taxjar_resp_tax, $body ) {
+		if ( ! is_object( $taxjar_resp_tax ) ) {
+			return $taxjar_resp_tax;
+		}
+
 		$original_rate = isset( $taxjar_resp_tax->rate ) ? floatval( $taxjar_resp_tax->rate ) : 0.0;
 		$new_tax_rate  = floatval( apply_filters( 'woocommerce_services_override_tax_rate', $taxjar_resp_tax->rate ?? 0, $taxjar_resp_tax, $body ) );
 
@@ -1663,6 +1670,7 @@ class WC_Connect_TaxJar_Integration {
 		if ( empty( $taxjar_response->tax ) || ! is_object( $taxjar_response->tax ) ) {
 			return false;
 		}
+
 		$taxjar_taxes = $this->maybe_override_taxjar_tax( $taxjar_response->tax, $body );
 		$taxes        = $this->get_itemized_tax_rates( $taxes, $taxjar_taxes, $options );
 
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1336,34 +1336,52 @@ class WC_Connect_TaxJar_Integration {
 	 * @return object
 	 */
 	public function maybe_override_taxjar_tax( $taxjar_resp_tax, $body ) {
-		if ( ! isset( $taxjar_resp_tax ) ) {
-			return;
-		}
-
-		$new_tax_rate = floatval( apply_filters( 'woocommerce_services_override_tax_rate', $taxjar_resp_tax->rate, $taxjar_resp_tax, $body ) );
-
-		if ( $new_tax_rate === floatval( $taxjar_resp_tax->rate ) ) {
+		if ( ! isset( $taxjar_resp_tax ) || ! is_object( $taxjar_resp_tax ) ) {
 			return $taxjar_resp_tax;
 		}
 
-		if ( ! empty( $taxjar_resp_tax->breakdown->line_items ) ) {
-			$taxjar_resp_tax->breakdown->line_items = array_map(
-				function ( $line_item ) use ( $new_tax_rate ) {
-					$line_item->combined_tax_rate       = $new_tax_rate;
-					$line_item->country_tax_rate        = $new_tax_rate;
-					$line_item->country_tax_collectable = $line_item->country_taxable_amount * $new_tax_rate;
-					$line_item->tax_collectable         = $line_item->taxable_amount * $new_tax_rate;
+		$original_rate = isset( $taxjar_resp_tax->rate ) ? floatval( $taxjar_resp_tax->rate ) : 0.0;
+		$new_tax_rate  = floatval( apply_filters( 'woocommerce_services_override_tax_rate', $taxjar_resp_tax->rate ?? 0, $taxjar_resp_tax, $body ) );
 
-					return $line_item;
-				},
-				$taxjar_resp_tax->breakdown->line_items
-			);
+		if ( $new_tax_rate === $original_rate ) {
+			return $taxjar_resp_tax;
 		}
 
-		$taxjar_resp_tax->breakdown->combined_tax_rate           = $new_tax_rate;
-		$taxjar_resp_tax->breakdown->country_tax_rate            = $new_tax_rate;
-		$taxjar_resp_tax->breakdown->shipping->combined_tax_rate = $new_tax_rate;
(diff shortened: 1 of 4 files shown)


tests/php/test-class-wc-connect-taxjar-integration.php
```diff
@@ -1865,4 +1865,117 @@ public function test_create_or_update_tax_rate_does_not_duplicate_rows_for_semic
...
+	/**
+	 * With no override filter registered the response is returned unchanged.
+	 */
+	public function test_maybe_override_taxjar_tax_returns_unchanged_without_override() {
+		$resp = $this->build_taxjar_tax_response();
+
+		$result = $this->integration->maybe_override_taxjar_tax( $resp, array() );
+
+		$this->assertSame( $resp, $result );
+		$this->assertSame( 0.08, $result->rate );
```
Abdalsalaam, line 1982: assertSame( $resp, $result ) compares object identity, and maybe_override_taxjar_tax() mutates $taxjar_resp_tax in place and returns that same handle — so the assertion passes even if the method rewrote every field, and cannot detect the 'unchanged' property the test name claims. Compounding it: with no filter registered, $new_tax_rate === $original_rate === 0.08, so the object is byte-identical whether or not the early return at class-wc-connect-taxjar-integration.php:1346 fires — the test cannot distinguish the two paths. Only the ->rate assertion carries signal. Asserting the breakdown and line-item fields still hold 0.08 would give the test real teeth.

```suggestion
		$this->assertSame( $resp, $result );
		$this->assertSame( 0.08, $result->rate );
		$this->assertSame( 0.08, $result->breakdown->combined_tax_rate );
		$this->assertSame( 0.08, $result->breakdown->shipping->combined_tax_rate );
		$this->assertSame( 0.08, $result->breakdown->line_items[0]->combined_tax_rate );
```

bartech: Good catch — you're right that object identity alone proved nothing since the method mutates in place. Strengthened the test to assert `rate`, `breakdown`, `shipping` and the line-item rate are all untouched in `8a907df`.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1336,34 +1336,52 @@ public function append_base_address_to_customer_taxable_address( $address ) {
 	 * @return object
```
Abdalsalaam: The new guard at 1339-1340 returns $taxjar_resp_tax unchanged for any non-object input (null, array, string), so the documented @return object contract no longer holds. @param object $taxjar_resp_tax on line 1333 is equally contradicted — the guard exists precisely because the argument may not be an object (line 1333 is outside this diff hunk, so update it in the same pass).

```suggestion
	 * @return object|mixed The tax object with rates overridden, or the input returned unchanged when it is not an object.
```

bartech: Good catch — the docblock did stop matching the code. Resolved in 4b0e1fd1, but by fixing the contract rather than weakening the type.

**Why not `object|mixed` / `mixed` / `object|null`:**
- `object|mixed` collapses to `mixed`, and `mixed` is "no type" — it documents the smell instead of removing it.
- `object|null` is tempting but doesn't match reality either. The sole caller (`calculate_tax`) decodes `$taxjar_response` with `json_decode()` and already bails on `empty( ->tax )`, so `null` is filtered *before* the call and no path returns null. And the non-object passthrough would hand back a **string/array**, not null — so `object|null` is inaccurate in both directions.

**The deeper issue:** `get_itemized_tax_rates()` dereferences the return as an object unconditionally (`->freight_taxable`, `->has_nexus`, `->rate`, …). A non-object `tax` would fatal one line downstream regardless — the top-level `! is_object` guard only **relocated** the crash, it didn't prevent it. The honest contract is `object`; the clean way to earn it is to validate the type at the boundary where the `empty()` guard already lives.

**So this commit:**
- Adds `! is_object( $taxjar_response->tax )` to `calculate_tax()`'s existing guard → a malformed non-object response bails early (`return false`, consistent with the other bad-response branches).
- Removes the now-redundant non-object early return from `maybe_override_taxjar_tax()`, restoring a truthful `@param object` / `@return object`. Every return path is now the tax object — so line 1333's `@param object` no longer contradicts the body either.

**Domestic vs international (checked explicitly):** TaxJar's `tax` node is always a JSON object for both locales; only the inner `breakdown` shape differs (US `state_`/`county_`/`city_` fields vs international `country_` fields). That inner variance is already handled by the existing member guards (`breakdown`/`shipping` `is_object`, `line_items` `is_array`, per-item `is_object`) — the real WOOTAX-74 "assign property on null" fix, retained untouched. No response-shape recovery is needed and `@return object` holds for both.

Full TaxJar integration test class green (72 tests, 184 assertions), including the five `maybe_override_taxjar_tax` cases.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1330,40 +1330,59 @@ public function append_base_address_to_customer_taxable_address( $address ) {
...
+			$breakdown->combined_tax_rate = $new_tax_rate;
+			$breakdown->country_tax_rate  = $new_tax_rate;
+
+			if ( isset( $breakdown->shipping ) && is_object( $breakdown->shipping ) ) {
+				$breakdown->shipping->combined_tax_rate = $new_tax_rate;
+				$breakdown->shipping->country_tax_rate  = $new_tax_rate;
+			}
+		}
 
 		$taxjar_resp_tax->rate = $new_tax_rate;
```
Abdalsalaam, line 1394: The PR deletes `if ( ! isset( $taxjar_resp_tax ) ) { return; }` and adds no replacement inside the method, while the PR description claims 'Return the response unchanged for a non-object input rather than fataling.' For $taxjar_resp_tax = null: line 1344 gives $original_rate = 0.0, line 1345 gives null->rate ?? 0, and the method returns unchanged ONLY if the filter also yields 0.0. With any woocommerce_services_override_tax_rate filter returning a non-zero rate, the isset() at 1354 is skipped and execution reaches line 1387 — a property write on null (Error on PHP 8; on PHP 7.4 it autovivifies and returns a fabricated stdClass where the old code returned null). Reachability: calculate_tax() at line 1671 is the only in-repo caller and it now validates is_object() at line 1668, and nothing hooks this method — so this is a latent hazard for third-party callers of a public, untyped method, not an internally reachable bug. The docblock at 1333-1336 asserts a caller invariant that nothing enforces, and the PR's own tests (lines 1907/1934) bypass that caller.

```suggestion
		if ( ! is_object( $taxjar_resp_tax ) ) {
			return $taxjar_resp_tax;
		}

		$original_rate = isset( $taxjar_resp_tax->rate ) ? floatval( $taxjar_resp_tax->rate ) : 0.0;
```

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1639,7 +1658,14 @@ public function calculate_tax( $options = array() ) {
...
 		// Decode Response.
 		$taxjar_response = json_decode( $response['body'] );
-		if ( empty( $taxjar_response->tax ) ) {
+		// TaxJar's `tax` node is always a JSON object; domestic and international
+		// responses differ only in the inner `breakdown` shape (US uses
+		// state_/county_/city_ fields, international uses country_ fields), not in
+		// the type of `tax` itself. Validate the type at this boundary so both
+		// maybe_override_taxjar_tax() and get_itemized_tax_rates() can rely on an
+		// object — a malformed non-object response bails here instead of fataling
+		// downstream on the first property access.
```
Abdalsalaam: Seven lines of comment justify a one-token change and argue against themselves: the block opens with 'TaxJar's `tax` node is always a JSON object' and closes with 'a malformed non-object response bails here.' If it were always an object the check at line 1668 would be dead code — the check exists precisely because the first sentence is not reliable, and a reader trusting it could later delete the guard this PR just added. The US-vs-international breakdown claim (US uses state_/county_/city_ fields, international uses country_ fields) is an assertion about a third-party API contract with no fixture, schema, or vendored spec anywhere in the checkout to verify it against, and is irrelevant to the is_object check — nothing at this line branches on US vs international. It will rot silently with no test to catch it. The comment is also wedged between the existing `// Decode Response.` comment and the statement it describes.

```suggestion
		// Bail on a malformed response: maybe_override_taxjar_tax() and
		// get_itemized_tax_rates() both access `tax` as an object.
```

CezaryDrewniak:
LGTM, with two nits to address before merge, both already raised by @Abdalsalaam:

1. **`classes/class-wc-connect-taxjar-integration.php:1661-1667`** -comment is self-contradictory. *"TaxJar's `tax` node is always a JSON object"* vs *"a malformed non-object response bails here"* — the check exists because the first sentence isn't reliable, and a future reader could delete the guard trusting it. The US-vs-international breakdown claim is irrelevant to the `is_object` check and will rot without a test. Suggest Abdalsalaam's one-liner: `// Bail on a malformed response: maybe_override_taxjar_tax() and get_itemized_tax_rates() both access 'tax' as an object.`

2. **`classes/class-wc-connect-taxjar-integration.php:1343`** - `public function maybe_override_taxjar_tax()` lost its `if ( ! isset( $taxjar_resp_tax ) )` guard with no in-method replacement. The in-repo caller (`calculate_tax`) validates at line 1668, so internal calls are safe, but this is a public method - a third-party caller passing `null` fatals on line 1387 (`$taxjar_resp_tax->rate = $new_tax_rate;` on PHP 8). Suggest re-adding an `is_object()` guard at the top, e.g. `if ( ! is_object( $taxjar_resp_tax ) ) { return $taxjar_resp_tax; }` - cheap insurance that doesn't weaken the `@return object` contract.

Fix, 5 tests, and changelog all look correct. Behavior preservation is pinned by `test_maybe_override_taxjar_tax_applies_override_to_wellformed_response`, fatal regression covered by `test_maybe_override_taxjar_tax_survives_null_line_item`. PHPCS clean on changed files (verified 144 → 140 errors vs trunk).

approve pending the two nits above.

classes/class-wc-connect-taxjar-integration.php
bartech: You're right, and I've restored the guard in `0f108cd6`.

The mistake in `4b0e1fd1` was treating boundary validation as a *replacement* for the method-level guard rather than an *addition*. They protect different callers and neither subsumes the other:

- The `calculate_tax()` check is the one that matters for the relocated-crash problem — it protects `get_itemized_tax_rates()`, which dereferences the tax node unconditionally. A guard inside `maybe_override_taxjar_tax()` can never reach that. It stays.
- The method-level guard protects callers I don't control. The AGENTS.md backward-compatibility rules that landed on trunk since I opened this settle it independently of my own judgement: `WC_Connect_*` classes are to be treated as externally exposed, and dropping a null guard from a `public` method on one of them is a BC change, not cleanup. The PR description now carries a BC impact section.

I also dropped the `object` type-declaration idea (PHP 7.4 supports it) for the same reason: it converts a silent no-op into a hard `TypeError`, i.e. *adds* a possible fatal in a fix-a-fatal PR.

On your reachability analysis — the filter dependency is exactly right, and it's now encoded in the test rather than left implicit. `..._returns_non_object_unchanged` covers `null` / `false` / `array` / `string`, and each case registers a filter returning a **different** rate, because with an unchanged rate the method early-returns and the test would pass with no guard at all. Removing the guard fails all four:

```
Error: Attempt to assign property "rate" on null
Error: Attempt to assign property "rate" on false
Error: Attempt to assign property "rate" on array
Error: Attempt to assign property "rate" on string
```

The stale PR-description bullet you quoted has been rewritten too — it described `e77f4a7`, not `4b0e1fd1`.

classes/class-wc-connect-taxjar-integration.php
bartech: Taking the `mixed` half of this back — you were right, I just needed the code underneath it to be right first.

With the guard restored in `0f108cd6`, the docblock now reads `@param mixed` / `@return mixed` with prose describing the expected type. My earlier objection was that `mixed` documents a smell instead of removing it; that held while the passthrough was the *only* protection. Now that `calculate_tax()` establishes the real contract, `mixed` is just an accurate statement about what a `public` method can enforce about its own callers — which is nothing.

The seven-line comment at `calculate_tax()` is gone too (that one landed in `c88b1563`). You were right that the US-vs-international claim had no fixture behind it and would rot silently.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1639,9 +1665,12 @@ public function calculate_tax( $options = array() ) {
 
 		// Decode Response.
 		$taxjar_response = json_decode( $response['body'] );
-		if ( empty( $taxjar_response->tax ) ) {
+		// Bail on a malformed response: maybe_override_taxjar_tax() and
+		// get_itemized_tax_rates() both access `tax` as an object.
+		if ( empty( $taxjar_response->tax ) || ! is_object( $taxjar_response->tax ) ) {
 			return false;
 		}
```
Abdalsalaam, line 1672: None of the new guards (1347, 1361, 1364, 1388, 1670) logs anything. A partially-applied override is indistinguishable from a clean one, and a whole-cart tax failure records no reason — the merchant's symptom is a wrong tax total, not a visible error. Every other bail in this code path logs (':1636' 'From country is missing. Aborting.', ':1658' 'Received: none.'). The only trace left is the raw response body at :1664, gated on logging being enabled.

```suggestion
		if ( empty( $taxjar_response->tax ) || ! is_object( $taxjar_response->tax ) ) {
			$this->_log( 'TaxJar response has no usable `tax` object. Aborting.' );
			return false;
		}
```

Abdalsalaam (approved):
LGTM! tested and it work as expected

PR #2979, opened 2026-07-15
Fix - Non-taxable product no longer zeroes shared standard tax rate (WOOTAX-240)

bartech:
### Summary

Fixes [WOOTAX-240](https://linear.app/a8c/issue/WOOTAX-240/mixed-carts-of-taxable-and-non-taxable-products-dont-calculate-tax).

When a cart mixes a **taxable** product and a **non-taxable** one (Tax Status = *None*, Tax Class = *Standard*), automated taxes stop calculating for the whole cart — the "Standard Rates" tax rate row is overwritten to **0%**, and it stays zeroed (adding another taxable product doesn't restore it; removing the non-taxable product does).

### Root cause

`get_itemized_tax_rates()` writes one WooCommerce tax rate row **per TaxJar breakdown line item**, and selects which row by the product's **Tax Class** — not its Tax Status:

```php
$tax_class = $product->get_tax_class();   // 'Standard' ('') even when Tax Status = None
...
$this->create_or_update_tax_rate( $location, round( $tax_rate * 100, 4 ), $tax_class, ... );
```

The non-taxable product is sent to TaxJar as exempt (code `99999`), so its breakdown line returns **0%** — but it still maps back to the **Standard** class. `create_or_update_tax_rate()` therefore updates the *same* rate row the taxable product just populated (same class + location) down to 0%. The zeroed row persists in `wp_woocommerce_tax_rates`, which is why the problem sticks.

> Note: the ticket's proposed change to `get_line_items()` (`! $product->is_taxable()` → `'none' === $tax_status`) is a **no-op** — `WC_Product::is_taxable()` already returns false for `tax_status = 'none'`, so the exempt product is already sent as `99999`. The defect is downstream, in how the itemized rate rows are written.

### Fix

Skip rate-row writes for line items whose product has Tax Status **None**. WooCommerce already applies no tax to such products (`override_cart_item_tax_rates()` falls back to core, which taxes nothing for a non-taxable item), so the shared class row is left intact for genuinely taxable items. Zero-rate **class** products are unaffected — they own a separate `zero-rate` row.

### Tests

Adds `test_get_itemized_tax_rates_non_taxable_line_does_not_zero_standard_rate`, which drives `get_itemized_tax_rates()` with a mixed taxable/exempt breakdown and asserts (a) the exempt line writes no rate row and (b) the taxable line's Standard row stays at 7.25%. It **fails on trunk** (exempt line writes/zeroes the shared row) and passes with this fix.

Full `WP_Test_WC_Connect_TaxJar_Integration` suite: **68 tests, 166 assertions green**. PHPCS clean on the changed files.

### Update — guard widened to "Shipping only" (review follow-up)

The guard now skips rate-row writes for **any** non-taxable Tax Status (`'taxable' !== $status`), not just `'None'`. `get_backend_line_items()` sends every non-taxable status to TaxJar as exempt (`99999`), so a **"Shipping only"** product returned the same 0% breakdown line and reproduced the zeroing bug on the admin order screen. `ProductTaxStatus` defines exactly three statuses and `set_tax_status()` normalises empty → `taxable`, so the new condition is precisely "none or shipping".

**Backward compatibility / scope:** this is a behaviour change on **both** paths, not just the backend one where the bug reproduces. A "Shipping only" product on the **cart** previously had a rate row written with a real (non-zero) rate; it is now skipped. This is safe — `WC_Cart_Totals` gates item tax on `is_taxable()` and `WC_Order_Item::calculate_taxes()` on `ProductTaxStatus::TAXABLE`, so core applies no item tax to either status, and the shipping tax rate row is written separately after this loop, so shipping tax is unaffected. No hook, signature, or option contract changes; no new global or `WC()` state reads; nothing site- or path-scoped, so multisite and non-root installs are unaffected.

Additional coverage: `get_backend_line_items()` sends a "Shipping only" product as exempt (input side), and its 0% line leaves the Standard row at 7.25% (output side). The latter was verified to fail against the previous `'none' ===` guard.

A pre-existing parent/variation Tax Class divergence found during this review is tracked separately in [WOOTAX-310](https://linear.app/a8c/issue/WOOTAX-310/itemized-tax-rate-rows-use-the-parents-tax-class-for-variations-on-the).

### Update — taxability recorded at emit time, not re-derived (review follow-up)

Review surfaced a third route to the same zeroing: `woocommerce_product_is_taxable` can filter a product to non-taxable while its Tax Status still reads `taxable`. `get_line_items()` sets the exempt code from the **filtered** `is_taxable()`, while the guard read the **raw** status — so the 0% line passed through and zeroed the shared Standard row. Confirmed against live TaxJar by the reviewer.

Rather than widen the guard a third time, the request side now records which line items WooCommerce will not tax (`$non_taxable_line_items`, keyed by TaxJar line item id) and the response side skips exactly those. The two halves share one decision instead of each re-deriving one from a different predicate.

The two paths deliberately record **different** predicates, because core gates them differently:

| Path | Core's item-tax gate | Location |
|---|---|---|
| Cart | `$item->product->is_taxable()` — filtered | `class-wc-cart-totals.php` 677 / 735 / 796 |
| Order | `ProductTaxStatus::TAXABLE === $this->get_tax_status()` — raw | `class-wc-order-item.php` 251 |

`WC_Order_Item::calculate_taxes()` never consults `is_taxable()`, so the order path has no filter divergence to close and recording the raw status is the accurate mirror there.

This is **not** implemented as "skip anything sent as `99999`" — zero-rate **class** products are emitted as exempt for an unrelated reason and must keep updating their own `zero-rate` row, so only the non-taxability reason is recorded.

Two earlier review points fall out as a side effect: the deleted-product case is now moot (the lookup never dereferences a possibly-gone product), as is any variation/parent-id mismatch at this seam, since the key is the emitted composite rather than a re-resolved product id.

**Backward compatibility:** the new property is private; no signature, hook, or option contract changes; no new global or `WC()` state reads; nothing site- or path-scoped, so multisite and non-root installs are unaffected. The only behaviour change is which rate rows are written, and only for line items core applies no item tax to.

**Tests:** the two output-side tests previously fabricated breakdown keys and called `get_itemized_tax_rates()` with no matching request — a shape that structurally cannot catch an emit/response divergence, which is why this survived two review rounds with a green suite. Both now drive the request side first and read the real emitted key back. The "Shipping only" case also moved to the **backend order path**, which is where that status is actually emitted as exempt; it had been asserting a cart scenario that cannot occur. New: `test_get_itemized_tax_rates_filtered_non_taxable_line_does_not_zero_standard_rate`, reproducing the filter case and asserting both the input side (`99999` emitted despite Tax Status "taxable") and the output side (shared row stays 7.25%) — verified red against the previous guard.

One gotcha for anyone re-running these: `is_taxable()` also requires `wc_tax_enabled()`, which is off in the bare WC test bootstrap, so the cart tests set `woocommerce_calc_taxes`.

`WP_Test_WC_Connect_TaxJar_Integration`: **80 tests, 206 assertions green**. Full suite: **235 tests, 521 assertions green**. PHPCS `--report=source` byte-identical to baseline on both changed files.

**Changelog retargeted:** 3.6.11 shipped from trunk while this PR was open, so the entry moved to a new unreleased `= 3.6.12 - 2026-xx-xx =` section in both `changelog.txt` and `readme.txt` rather than being merged into the released 3.6.11 block.

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1791,6 +1791,17 @@ class WC_Connect_TaxJar_Integration {
 					$tax_class = $this->backend_tax_classes[ $product_id ];
 				}
 
+				// A product with Tax Status = "None" is exempt from tax but can still
+				// carry the standard Tax Class. TaxJar returns a 0% breakdown line for
+				// it, and because tax rate rows are keyed by tax class, writing that 0%
+				// overwrites the shared class rate used by genuinely taxable products in
+				// the same class — zeroing tax for the whole cart. Skip rate-row writes
+				// for non-taxable products; WooCommerce already applies no tax to them.
+				// See WOOTAX-240.
+				if ( $product && 'none' === $product->get_tax_status() ) {
+					continue;
+				}
+
 				$_tax_rates = (array) $line_item;
 				$priority   = 1;
 				foreach ( $_tax_rates as $tax_rate_name => $tax_rate ) {
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -87,6 +87,17 @@ class WC_Connect_TaxJar_Integration {
 	private $backend_tax_classes;
 
 	/**
+	 * Line items WooCommerce will not apply item tax to.
+	 *
+	 * Recorded while the TaxJar request is built and read back when the response is
+	 * turned into tax rate rows, so both halves share a single taxability decision
+	 * instead of each re-deriving one. Keyed by TaxJar line item id, values unused.
+	 *
+	 * @var array<string, bool>
+	 */
+	private $non_taxable_line_items = array();
+
+	/**
 	 * Tracks instance.
 	 *
 	 * @var WC_Connect_Tracks
@@ -962,8 +973,9 @@ class WC_Connect_TaxJar_Integration {
 	 * @return array
 	 */
 	protected function get_line_items( $wc_cart_object ) {
-		$line_items       = array();
-		$default_location = get_option( 'woocommerce_tax_based_on', 'shipping' );
+		$line_items                   = array();
+		$this->non_taxable_line_items = array();
+		$default_location             = get_option( 'woocommerce_tax_based_on', 'shipping' );
 
 		foreach ( $wc_cart_object->get_cart() as $cart_item_key => $cart_item ) {
 			$product       = $cart_item['data'];
@@ -981,6 +993,14 @@ class WC_Connect_TaxJar_Integration {
 
 			if ( 'shipping' !== $product->get_tax_status() && ( ! $product->is_taxable() || 'zero-rate' == sanitize_title( $product->get_tax_class() ) ) ) {
 				$tax_code = '99999';
+			}
+
+			// WC_Cart_Totals gates item tax on is_taxable(), which is filterable, so a
+			// product can be untaxed while still reporting Tax Status "taxable". Record
+			// the decision here rather than re-deriving it from the response, where the
+			// two can disagree and a 0% line ends up overwriting a shared rate row.
+			if ( ! $product->is_taxable() ) {
+				$this->non_taxable_line_items[ $id . '-' . $cart_item_key ] = true;
 			}
 
 			// Get WC Subscription sign-up fees for calculations
@@ -1039,9 +1059,10 @@ class WC_Connect_TaxJar_Integration {
 	 * @return array
 	 */
 	protected function get_backend_line_items( $order ) {
-		$line_items                = array();
-		$this->backend_tax_classes = array();
-		$default_location          = get_option( 'woocommerce_tax_based_on', 'shipping' );
+		$line_items                   = array();
+		$this->backend_tax_classes    = array();
+		$this->non_taxable_line_items = array();
+		$default_location             = get_option( 'woocommerce_tax_based_on', 'shipping' );
 
 		foreach ( $order->get_items() as $item_key => $item ) {
 			if ( is_object( $item ) ) { // Woo 3.0+
@@ -1069,6 +1090,11 @@ class WC_Connect_TaxJar_Integration {
 			}
 			if ( 'taxable' !== $tax_status ) {
 				$tax_code = '99999';
+
+				// WC_Order_Item::calculate_taxes() gates item tax on the raw tax status,
+				// so the order path records exactly what it emitted as exempt. See the
+				// matching note in get_line_items() for why this is recorded, not derived.
+				$this->non_taxable_line_items[ $id . '-' . $item_key ] = true;
 			}
 
 			/** This filter is documented in get_line_items() */
@@ -1780,6 +1806,15 @@ class WC_Connect_TaxJar_Integration {
 
 			// Add line item tax rates.
 			foreach ( $taxes['line_items'] as $line_item_key => $line_item ) {
+				// A line item WooCommerce will not tax was sent to TaxJar as exempt, so its
+				// breakdown line comes back at 0% — yet it keeps its Tax Class. Since rate
+				// rows are keyed by tax class, writing that 0% overwrites the shared row
+				// used by genuinely taxable products in the same class, zeroing tax for the
+				// whole cart. Skip the write; WooCommerce applies no item tax to it anyway.
+				if ( isset( $this->non_taxable_line_items[ $line_item_key ] ) ) {
+					continue;
+				}
+
 				$line_item_key_chunks = explode( '-', $line_item_key );
 				$product_id           = $line_item_key_chunks[0];
 				$product              = wc_get_product( $product_id );
@@ -1789,16 +1824,6 @@ class WC_Connect_TaxJar_Integration {
 					$tax_class = $product->get_tax_class();
 				} elseif ( isset( $this->backend_tax_classes[ $product_id ] ) ) {
 					$tax_class = $this->backend_tax_classes[ $product_id ];
-				}
-
-				// Products that are not fully taxable (Tax Status "None" or "Shipping
-				// only") are sent to TaxJar as exempt, so their breakdown line returns 0% —
-				// yet they keep their Tax Class. Since rate rows are keyed by tax class,
-				// writing that 0% overwrites the shared row used by genuinely taxable
-				// products in the same class, zeroing tax for the whole cart. Skip the
-				// write; WooCommerce applies no item tax to either status anyway.
-				if ( $product && 'taxable' !== $product->get_tax_status() ) {
-					continue;
 				}
 
 				$_tax_rates = (array) $line_item;
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -999,7 +999,22 @@ class WC_Connect_TaxJar_Integration {
 			// product can be untaxed while still reporting Tax Status "taxable". Record
 			// the decision here rather than re-deriving it from the response, where the
 			// two can disagree and a 0% line ends up overwriting a shared rate row.
-			if ( ! $product->is_taxable() ) {
+			//
+			// The 'shipping' exclusion mirrors the exempt branch directly above, and has
+			// to: a "Shipping only" product is excluded there, so this path sends it to
+			// TaxJar as taxable and its breakdown line comes back with a real, non-zero
+			// rate. Recording it would make get_itemized_tax_rates() skip that write --
+			// the one case where skipping discards good data instead of preventing a 0%
+			// clobber. WooCommerce applies no *item* tax to such a product, but the row
+			// still matters: it carries `tax_rate_shipping`, and with the default
+			// `woocommerce_shipping_tax_class = 'inherit'` WooCommerce resolves the
+			// shipping tax class to this product's class and looks the rate up there.
+			// TaxJar's own shipping write only ever covers the standard class.
+			//
+			// The backend order path records on `'taxable' !== $tax_status` instead, and
+			// that is correct rather than inconsistent: there the same condition is what
+			// sets the 99999 exempt code, so it too records exactly what it emitted.
+			if ( 'shipping' !== $product->get_tax_status() && ! $product->is_taxable() ) {
 				$this->non_taxable_line_items[ $id . '-' . $cart_item_key ] = true;
 			}
 
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1791,14 +1791,13 @@ class WC_Connect_TaxJar_Integration {
 					$tax_class = $this->backend_tax_classes[ $product_id ];
 				}
 
-				// A product with Tax Status = "None" is exempt from tax but can still
-				// carry the standard Tax Class. TaxJar returns a 0% breakdown line for
-				// it, and because tax rate rows are keyed by tax class, writing that 0%
-				// overwrites the shared class rate used by genuinely taxable products in
-				// the same class — zeroing tax for the whole cart. Skip rate-row writes
-				// for non-taxable products; WooCommerce already applies no tax to them.
-				// See WOOTAX-240.
-				if ( $product && 'none' === $product->get_tax_status() ) {
+				// Products that are not fully taxable (Tax Status "None" or "Shipping
+				// only") are sent to TaxJar as exempt, so their breakdown line returns 0% —
+				// yet they keep their Tax Class. Since rate rows are keyed by tax class,
+				// writing that 0% overwrites the shared row used by genuinely taxable
+				// products in the same class, zeroing tax for the whole cart. Skip the
+				// write; WooCommerce applies no item tax to either status anyway.
+				if ( $product && 'taxable' !== $product->get_tax_status() ) {
 					continue;
 				}
 
--- a/tests/php/test-class-wc-connect-taxjar-integration.php
+++ b/tests/php/test-class-wc-connect-taxjar-integration.php
@@ -3588,4 +3588,88 @@ class WP_Test_WC_Connect_TaxJar_Integration extends WC_Unit_Test_Case {
 			);
 		}
 	}
+
+	/**
+	 * A non-taxable product (Tax Status = "None") that shares the standard Tax
+	 * Class with a taxable product must not zero out the shared Standard rate row.
+	 *
+	 * TaxJar returns a 0% breakdown line for the exempt product (sent as code
+	 * 99999). Because rate rows are keyed by tax class — and Tax Status "None"
+	 * does not change the Tax Class — that 0% would overwrite the same Standard
+	 * row the taxable product just populated, zeroing tax for the whole cart.
+	 * Regression for WOOTAX-240.
+	 */
+	public function test_get_itemized_tax_rates_non_taxable_line_does_not_zero_standard_rate() {
+		$taxable_product = WC_Helper_Product::create_simple_product();
+		$taxable_product->set_tax_status( 'taxable' );
+		$taxable_product->set_tax_class( '' );
+		$taxable_product->save();
+
+		$exempt_product = WC_Helper_Product::create_simple_product();
+		$exempt_product->set_tax_status( 'none' );
+		$exempt_product->set_tax_class( '' );
+		$exempt_product->save();
+
+		$taxable_id  = $taxable_product->get_id();
+		$exempt_id   = $exempt_product->get_id();
+		$taxable_key = $taxable_id . '-taxable_item';
+		$exempt_key  = $exempt_id . '-exempt_item';
+
+		// Taxable line is listed first, so without the fix the exempt line would
+		// overwrite the shared Standard rate row to 0% afterwards.
+		$taxjar_taxes = (object) array(
+			'freight_taxable' => 0,
+			'has_nexus'       => 1,
+			'rate'            => 0.0725,
+			'jurisdictions'   => (object) array(
+				'county' => '',
+				'city'   => '',
+			),
+			'breakdown'       => (object) array(
+				'line_items' => array(
+					(object) array(
+						'id'                => $taxable_key,
+						'combined_tax_rate' => 0.0725,
+						'state_tax_rate'    => 0.0725,
+					),
+					(object) array(
+						'id'                => $exempt_key,
+						'combined_tax_rate' => 0.0,
+						'state_tax_rate'    => 0.0,
+					),
+				),
+			),
+		);
+
+		$options = array(
+			'to_country' => 'US',
+			'to_state'   => 'CA',
+			'to_zip'     => '90210',
+			'to_city'    => 'Beverly Hills',
+		);
+
+		$result = $this->invoke_protected_method(
+			'get_itemized_tax_rates',
+			array(
+				array(
+					'rate_ids'   => array(),
+					'line_items' => array(),
+				),
+				$taxjar_taxes,
+				$options,
+			)
+		);
+
+		// The exempt line must not create or update any tax rate row.
+		$this->assertArrayNotHasKey( $exempt_key, $result['rate_ids'], 'Non-taxable line item should not write a tax rate row.' );
+
+		// The taxable line's Standard rate row must remain at 7.25%, not clobbered to 0.
+		$this->assertArrayHasKey( $taxable_key, $result['rate_ids'] );
+		$taxable_rate_id = $result['rate_ids'][ $taxable_key ][0];
+		$rate_data       = WC_Tax::_get_tax_rate( $taxable_rate_id );
+		$this->assertSame( 7.25, (float) $rate_data['tax_rate'], 'Standard rate row was zeroed by the non-taxable line item (WOOTAX-240).' );
+
+		WC_Helper_Product::delete_product( $taxable_id );
+		WC_Helper_Product::delete_product( $exempt_id );
+	}
 }
--- a/tests/php/test-class-wc-connect-taxjar-integration.php
+++ b/tests/php/test-class-wc-connect-taxjar-integration.php
@@ -3828,12 +3828,6 @@ class WP_Test_WC_Connect_TaxJar_Integration extends WC_Unit_Test_Case {
 	}
 
 	/**
-	 * Tax Status "Shipping only" is exempt from item tax, and the backend order path
-	 * sends it to TaxJar as exempt (code 99999) — unlike the cart path, which excludes
-	 * that status from the exempt branch. This is the input side of the zeroing bug:
-	 * an exempt code is what makes TaxJar return the 0% breakdown line.
-	 */
-	/**
 	 * Cart path: a "Shipping only" product must NOT be recorded as non-taxable.
 	 *
 	 * The exempt branch in `get_line_items()` deliberately excludes Tax Status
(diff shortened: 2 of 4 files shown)


CezaryDrewniak (approved):
## LGTM

**Test file PHPCS: 0 errors, 5 warnings, but all 5 are on pre-existing lines (197, 339, 885), none on the new test's lines (1869–1951).**

| PR claim | Verified |
|---|---|
| New test passes | ✅ 1 test, 3 assertions, green |
| Full suite green | ✅ 68 tests, 166 assertions (matches PR description exactly) |
| PHPCS clean on changed files | ✅ Production class: 0 violations on lines 1735–1750. Test file: 0 violations on lines 1869–1951. The 5 pre-existing warnings are at lines 197, 339, 885 (unrelated, in older tests) |
| Root cause analysis correct | ✅ Cursor's `get_line_items()` change is a no-op (`is_taxable()` already false for `tax_status='none'`); real defect is downstream in `get_itemized_tax_rates()` row-write path |
| Fix placement correct | ✅ `continue` placed after `$tax_class` is determined, before `create_or_update_tax_rate()`, leaving the shared Standard row intact for genuinely taxable items |
| No regression for zero-rate class products | ✅ They own a separate rate row keyed by `tax_class='zero-rate'`; the skip only triggers on `tax_status='none'` |
| Changelog/readme updated | ✅ Both `changelog.txt:7` and `readme.txt:77` add the new entry in the right format under `= 3.6.8 - 2026-xx-xx =` |

The fix is minimal, correctly targeted at the actual defect (not the ticket's misdiagnosed surface), well-tested, and clean

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1735,6 +1735,17 @@ private function get_itemized_tax_rates( $taxes, $taxjar_taxes, $options ): arra
...
+				// A product with Tax Status = "None" is exempt from tax but can still
+				// carry the standard Tax Class. TaxJar returns a 0% breakdown line for
+				// it, and because tax rate rows are keyed by tax class, writing that 0%
+				// overwrites the shared class rate used by genuinely taxable products in
+				// the same class — zeroing tax for the whole cart. Skip rate-row writes
+				// for non-taxable products; WooCommerce already applies no tax to them.
+				// See WOOTAX-240.
+				if ( $product && 'none' === $product->get_tax_status() ) {
+					continue;
+				}
```
Abdalsalaam: get_backend_line_items() emits the exempt code for ANY non-taxable status (line 1060: `if ( 'taxable' !== $tax_status ) { $tax_code = '99999'; }`) using `$item->get_tax_status()` (line 1043), while keying the line item by `$item->get_product_id()` (line 1038). The new guard instead re-resolves the product via `wc_get_product( $product_id )` (line 1729) and only matches 'none'. Three divergences each reproduce the original bug on the admin/backend order path: (1) Tax Status 'shipping' (Shipping only) is emitted as 99999 but not skipped, so its 0% line still clobbers the shared Standard row; (2) for variations $product_id is the PARENT id, so a variation with Tax Status None under a taxable parent emits 99999 while the guard reads the parent's 'taxable'; (3) when wc_get_product() returns false (deleted/trashed product still referenced by an order item) the `$product &&` short-circuit skips the guard entirely and the old unconditional write runs via the backend_tax_classes fallback at lines 1734-1736, which stores only the tax class (line 1054), never the status. The cart path is unaffected by case (1) because line 972 explicitly excludes 'shipping' from the 99999 branch. Minimal fix closes case (1) safely — core does not tax 'shipping'-status items either, so their rate rows go unused. Complete fix: record the emitted status alongside the tax class in get_backend_line_items() and skip on that.

```suggestion
				// A product that is not fully taxable (Tax Status "None" or "Shipping only")
				// is exempt from item tax but can still carry the standard Tax Class. TaxJar
				// returns a 0% breakdown line for it, and because tax rate rows are keyed by
				// tax class, writing that 0% overwrites the shared class rate used by genuinely
				// taxable products in the same class — zeroing tax for the whole cart. Skip
				// rate-row writes for these products; WooCommerce already applies no item tax
				// to them. See WOOTAX-240.
				if ( $product && 'taxable' !== $product->get_tax_status() ) {
					continue;
				}
```

bartech:
Thanks for the careful read @Abdalsalaam — I verified all three cases against WooCommerce core. **Case 1 is real and I've implemented your suggestion** (8232331e). Cases 2 and 3 don't hold up; details below.

## ✅ Implemented — case 1, "Shipping only" status

Confirmed, and it's the whole reason to take the change. `get_backend_line_items()` sends *any* non-taxable status as `99999`, so a **Tax Status = "Shipping only"** product returns a 0% breakdown line that the `'none' ===` guard let straight through — the identical zeroing failure, on the admin order screen. The path is ungated: `calculate_backend_totals()` → `calculate_tax()` → `get_itemized_tax_rates()`. You're also right that the cart path is immune, since line 972 excludes `'shipping'` from the exempt branch.

Your safety argument holds on both paths — core never taxes these items, so the skipped row goes unused:

| Path | Gate | Location |
|---|---|---|
| Cart | `$item->product->is_taxable()` (≡ `'taxable' === get_tax_status()`) | `WC_Cart_Totals` 677 / 735 / 796 |
| Order | `ProductTaxStatus::TAXABLE === $this->get_tax_status()` | `WC_Order_Item::calculate_taxes()` 251 |

And `'taxable' !==` is *exactly* "none or shipping": `ProductTaxStatus` defines precisely three constants, and `set_tax_status()` normalises empty → `taxable` and rejects anything else. No fourth-value risk.

Two deviations from the suggestion as written:

1. **Trimmed the comment** — dropped the ticket reference, since Linear IDs don't belong in runtime source (they live in the branch/commit/PR, where Linear resolves them anyway).
2. **Added tests.** The existing test only covered `'none'`, so it passed either way and would not have caught a revert. There are now two: one asserting the backend path sends a "Shipping only" product as exempt (input side), one asserting its 0% line leaves the Standard row at 7.25% (output side). I verified the second fails against the old `'none' ===` guard and passes with the fix. Suite: **70 tests, 171 assertions green**; PHPCS unchanged vs. baseline on both files.

**Scope note for reviewers:** you framed this as backend-only, but the guard runs on both paths — so a "Shipping only" product on the **cart** now also has its rate row skipped (previously written with a real rate). Safe, because core's `is_taxable()` gate discards it and the shipping tax row is written separately after this loop, but it is a cart-path behaviour change and I've noted it in the PR description.

## ❌ Rejected — case 2, variations

Not reproducible. `WC_Product_Variation_Data_Store_CPT::read_product_data()` line 481:

```php
// Pull data from the parent when there is no user-facing way to set props.
$product->set_tax_status( $parent_data['tax_status'] );
```

Variations **always** inherit the parent's tax status — there is no per-variation Tax Status field. So "a variation with Tax Status None under a taxable parent" cannot exist, and `wc_get_product( $product_id )` in the guard reads the same value `$item->get_tax_status()` did.

There *is* a genuine parent/variation divergence next door, but it's the `$tax_class` resolution on the line above — for a variation, `$product_id` is the parent, so a variation with its own Tax Class gets the wrong rate row. That predates this PR and is out of scope here; filed separately as [WOOTAX-310](https://linear.app/a8c/issue/WOOTAX-310/itemized-tax-rate-rows-use-the-parents-tax-class-for-variations-on-the).

## ❌ Rejected — case 3, deleted product

The `$product &&` short-circuit does structurally bypass the guard, but the state that would make it matter is unreachable. `WC_Order_Item_Product::get_tax_status()` (core, line 482):

```php
$product = $this->get_product();
return $product ? $product->get_tax_status() : ProductTaxStatus::TAXABLE;
```

If the product is gone, the item is reported **taxable** → no `99999` → TaxJar returns a real rate → there is no 0% line to clobber anything. The Woo 2.6 branch has the same `$product ? … : 'taxable'` fallback at line 1052. Reaching the described state would require a `99999` emission alongside a failed `wc_get_product()` — i.e. a live variation with a dead parent, which can't happen since the variation read pulls parent post meta.

The `$product &&` check stays, though: it's what prevents a fatal on `false->get_tax_status()`.

## ❌ Rejected — the proposed "complete fix"

Recording the emitted status in a `backend_tax_statuses` array alongside `backend_tax_classes` adds a second request-scoped mutable state array to a class that already has that smell, and it only buys the unreachable case above. Sticking with the minimal guard.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1764,6 +1764,16 @@ private function get_itemized_tax_rates( $taxes, $taxjar_taxes, $options ): arra
 					$tax_class = $this->backend_tax_classes[ $product_id ];
 				}
 
+				// Products that are not fully taxable (Tax Status "None" or "Shipping
+				// only") are sent to TaxJar as exempt, so their breakdown line returns 0% —
+				// yet they keep their Tax Class. Since rate rows are keyed by tax class,
+				// writing that 0% overwrites the shared row used by genuinely taxable
+				// products in the same class, zeroing tax for the whole cart. Skip the
+				// write; WooCommerce applies no item tax to either status anyway.
+				if ( $product && 'taxable' !== $product->get_tax_status() ) {
```
dustinparker: The fix works for the case in the ticket, but there's one more route to the same zeroing that I think we should close here.

Taxability is filterable, so a product can be sent to TaxJar as exempt while still reporting Tax Status "taxable". The guard checks `get_tax_status()`, while `get_line_items()` sets the exempt code from `! $product->is_taxable()`. When those disagree, the 0% line still wipes the shared Standard row.

Repro, with two Standard class products both set to Taxable:

```php
add_filter( 'woocommerce_product_is_taxable', function ( $taxable, $product ) {
	return 123 === $product->get_id() ? false : $taxable;
}, 10, 2 );
```

Add the first product, note the rate under Standard Rates, then add the filtered one. Cart tax drops to 0 and every Standard row goes to 0%. I confirmed this against live TaxJar.

I think the fix is to stop re-deriving taxability on the response side. If `get_line_items()` and `get_backend_line_items()` record the keys they treat as non-taxable, `get_itemized_tax_rates()` can skip exactly those. WDYT?

dustinparker (changes requested):

classes/class-wc-connect-taxjar-integration.php
bartech: Confirmed, and thanks — this is a real hole and the diagnosis is right. Implemented your
suggested design.

## The divergence

`WC_Product::is_taxable()` (abstract-wc-product.php:1832) is filtered:

```php
return apply_filters( 'woocommerce_product_is_taxable', $this->get_tax_status() === ProductTaxStatus::TAXABLE && wc_tax_enabled(), $this );
```

So the request side (`get_line_items()`, exempt code from `! is_taxable()`) and the
response side (the guard, reading raw `get_tax_status()`) can disagree, exactly as you
describe — 0% line, shared Standard row zeroed. The previous guard was a third patch to
a predicate that shouldn't have been re-derived in the first place.

## What changed

Request side records, response side reads:

- `private $non_taxable_line_items` — keyed by TaxJar line item id
  (`<product_id>-<cart_or_order_item_key>`), the same key the breakdown echoes back.
- `get_line_items()` records `! $product->is_taxable()`.
- `get_backend_line_items()` records inside its existing `'taxable' !== $tax_status`
  branch — i.e. exactly what it emitted as exempt.
- `get_itemized_tax_rates()` skips on `isset( $this->non_taxable_line_items[ $key ] )`.
  The product lookup there no longer participates in the decision.

**The two paths deliberately record different predicates**, because core gates them
differently:

| Path | Core's item-tax gate | Location |
|---|---|---|
| Cart | `$item->product->is_taxable()` — filtered | `class-wc-cart-totals.php` 677 / 735 / 796 |
| Order | `ProductTaxStatus::TAXABLE === $this->get_tax_status()` — raw | `class-wc-order-item.php` 251 |

`WC_Order_Item::calculate_taxes()` never consults `is_taxable()`, so the order path has no
filter divergence to close and recording the raw status there is the accurate mirror.
Each side records what its own core gate will do.

One thing worth calling out: I did **not** implement this as "skip anything sent as
99999". Zero-rate-class products are emitted as `99999` for an unrelated reason and must
keep updating their own `zero-rate` row — so only the non-taxability reason is recorded.

Two earlier review points fall out for free: the deleted-product case @Abdalsalaam raised
is now moot (the lookup never dereferences a possibly-gone product), as is any
variation/parent-id mismatch at this seam, since the key is the emitted composite rather
than a re-resolved product id.

## Tests

The two output-side tests fabricated breakdown keys and called `get_itemized_tax_rates()`
with no matching request — a shape that structurally cannot catch an emit/response
divergence, which is why this survived two rounds. Both now drive the request side first
and read the real key back. The "Shipping only" one also moved to the **backend order
path**, which is where that status is actually emitted as exempt; it had been asserting a
cart scenario that can't occur.

New: `test_get_itemized_tax_rates_filtered_non_taxable_line_does_not_zero_standard_rate`,
reproducing your filter case, asserting both the input side (`99999` emitted despite Tax
Status "taxable") and the output side (shared row stays 7.25%). Verified red against the
previous guard — it writes the row — while the "None" and "Shipping only" tests stay
green.

Gotcha for anyone re-running these: `is_taxable()` also requires `wc_tax_enabled()`, which
is off in the bare WC test bootstrap, so the cart tests set `woocommerce_calc_taxes`.

TaxJar class **80 tests / 206 assertions**, full suite **230 / 510** green. PHPCS
`--report=source` byte-identical to baseline on both files.

## BC

Nothing public moved: new property is private, no signature/hook/option changes, no new
global or `WC()` state reads, nothing site- or path-scoped. The only behaviour change is
which rate rows get written, and only for line items core does not tax.


Pushed as `834527da`.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -973,6 +985,14 @@ protected function get_line_items( $wc_cart_object ) {
 				$tax_code = '99999';
 			}
 
+			// WC_Cart_Totals gates item tax on is_taxable(), which is filterable, so a
+			// product can be untaxed while still reporting Tax Status "taxable". Record
+			// the decision here rather than re-deriving it from the response, where the
+			// two can disagree and a 0% line ends up overwriting a shared rate row.
+			if ( ! $product->is_taxable() ) {
+				$this->non_taxable_line_items[ $id . '-' . $cart_item_key ] = true;
+			}
```
Abdalsalaam: The exempt branch at line 984 deliberately excludes Tax Status 'shipping' from the 99999 code, so on the cart path a "Shipping only" product is sent to TaxJar as genuinely taxable and its breakdown line returns a real, non-zero rate. But the new record at line 992 uses the unconditional `! $product->is_taxable()`, which is false for any status other than 'taxable' — including 'shipping'. That line is therefore added to $non_taxable_line_items and its correct, non-zero rate is discarded at line 1787, where pre-fix it was written. This is the one case where the skip is not preventing a 0% clobber but discarding good data. The order path has no such asymmetry ('taxable' !== $tax_status at line 1081 is exactly the condition that sets 99999), so this cart/order divergence is unintended rather than the deliberate filtered-vs-raw split the PR documents. Concretely: for a "Shipping only" product in a NON-standard tax class with woocommerce_shipping_tax_class = 'inherit' (WC default), WC_Cart::get_cart_item_tax_classes_for_shipping() includes it because is_shipping_taxable() is true for 'shipping' status, so WC resolves the shipping tax class to that non-standard class and calls WC_Tax::get_shipping_tax_rates() against it. Pre-fix the item write kept that row current (it carries tax_rate_shipping, line 1923); post-fix nothing writes it, because TaxJar's separate shipping write at line 1828 uses tax class '' (standard) only. The PR body's safety argument covers item tax and the standard class but not this path.

```suggestion
			if ( 'shipping' !== $product->get_tax_status() && ! $product->is_taxable() ) {
				$this->non_taxable_line_items[ $id . '-' . $cart_item_key ] = true;
			}
```

bartech: Confirmed and applied verbatim — `5af84040`. This is a regression my own fix introduced, and your framing of it is the part that made it land: **this is the one case where the skip discards good data instead of preventing a 0% clobber.** I had been reasoning about the guard purely as a clobber-preventer, so a false positive looked free. It isn't.

I checked each supporting claim against the code rather than taking the chain on trust, since the harm path is several hops long:

| Claim | Verified |
|---|---|
| Exempt branch excludes `'shipping'`, so the line is emitted as taxable | `'shipping' !== $product->get_tax_status() && ( ! $product->is_taxable() \|\| 'zero-rate' == … )` |
| `! $product->is_taxable()` is true for `'shipping'` | `is_taxable()` is `'taxable' === get_tax_status() && wc_tax_enabled()` |
| The item write is what carries `tax_rate_shipping` | `create_or_update_tax_rate()`, `'tax_rate_shipping' => $freight_taxable` |
| TaxJar's own shipping write is standard-class only | the shipping-breakdown loop passes `''` as the tax class |

### The divergence is in the recording, not the emission

Worth stating explicitly because it changes what the fix should be. The two paths *do* emit differently for this status — the cart path sends "Shipping only" as taxable, the order path sends it as `99999` — and that difference is pre-existing and deliberate. What diverged was that the order path records on `'taxable' !== $tax_status`, which is *exactly the condition that sets its 99999*, so it records what it emitted; the cart path recorded on a condition that had drifted from what *it* emitted. Aligning the cart path restores the same rule on both sides rather than making the two paths identical.

### One alternative I considered and rejected

`'99999' === $tax_code && ! $product->is_taxable()` is semantically identical here and has the appeal of reading the emitted value directly, so it cannot drift if the exempt branch changes. I did not take it: `$tax_code` is derived from `end( explode( '-', $tax_class ) )`, so a tax class whose slug ends in `-99999` would collide. Vanishingly unlikely, but your version has no exotic failure mode and mirrors the branch above in the same idiom, so it is the better line.

### Tests — both sides of the boundary

* **`test_get_line_items_does_not_record_shipping_only_status_as_non_taxable`** — the regression test. Fails against the previous condition with `'10-<cart_key>' => true` recorded.
* **`test_get_line_items_still_records_none_status_as_non_taxable`** — the boundary test. Passes in *both* states by design; it pins that the exclusion narrows by exactly one status, so a line genuinely emitted as exempt stays recorded and its 0% cannot come back.

The reason this got through is worth recording: **the cart path had no coverage for this status at all.** Both existing "Shipping only" tests are on the order path, where the recording happened to be correct — so the suite was green on the path where the rule held and silent on the path where it didn't.

Full suite 237 tests / 527 assertions green.

tests/php/test-class-wc-connect-taxjar-integration.php
```diff
@@ -2022,4 +2037,403 @@ public function provide_non_object_tax_nodes() {
...
+			$line_items[0]['id'],
+			$recorded,
+			'A line emitted as exempt must stay recorded, or its 0% rate overwrites the shared Standard row.'
+		);
+
+		WC()->cart->empty_cart();
+		WC_Helper_Product::delete_product( $product->get_id() );
+	}
+
+	public function test_get_backend_line_items_sends_shipping_only_status_as_exempt() {
```
Abdalsalaam, line 3916: phpcs.xml.dist imports .phpcs.php.xml, which loads WordPress-Docs globally (.phpcs.php.xml:11). Its tests/ exclusions cover only Generic.Commenting, WordPress.Files.FileName, WordPress.WP.GlobalVariablesOverride, WooCommerce.Commenting.CommentHooks and the two Squiz.Commenting.FileComment.* sniffs, so Squiz.Commenting.FunctionComment.Missing still applies to tests/. This is the only method of ~70 in the file without a docblock, so `composer run check-all` should error, contradicting the PR body's claim that PHPCS --report=source is byte-identical to baseline. The intended docblock already exists but is stranded at lines 2264-2269 above the wrong method; once this suggestion is applied, delete that stranded block.

```suggestion
	/**
	 * Tax Status "Shipping only" is exempt from item tax, and the backend order path
	 * sends it to TaxJar as exempt (code 99999) — unlike the cart path, which excludes
	 * that status from the exempt branch. This is the input side of the zeroing bug:
	 * an exempt code is what makes TaxJar return the 0% breakdown line.
	 */
	public function test_get_backend_line_items_sends_shipping_only_status_as_exempt() {
```

bartech: Confirmed, and the reasoning holds at every step — I checked each of the three claims separately rather than taking the conclusion.

**1. The ruleset claim.** `.phpcs.php.xml` does load `WordPress-Docs` globally, and its `tests/` exclusions are exactly the five you listed (`Generic.Commenting`, `WordPress.Files.FileName`, `WordPress.WP.GlobalVariablesOverride`, `WooCommerce.Commenting.CommentHooks`, and the two `Squiz.Commenting.FileComment.*`). `Squiz.Commenting.FunctionComment` is not among them, so it does apply to `tests/`. Config-reading alone doesn't settle it though — whether the sniff is actually pulled in is decided inside `WordPress-Docs` — so I ran it, and it fires:

```
2350 | ERROR | Missing doc comment for function
     |       | test_get_backend_line_items_sends_shipping_only_status_as_exempt()
     |       | (Squiz.Commenting.FunctionComment.Missing)
```

**2. The stranded docblock.** Lines 2264-2269 held the backend test's docblock sitting immediately above the *cart* test's own docblock. PHPCS binds a docblock to the immediately following token, so the orphan never attached to anything — the cart test kept its own comment and looked fine, and the method 60 lines below silently lost its docblock. Exactly as you described.

**3. The PR body claim.** You're right that it contradicts it, and this is the part I'd have missed with a sloppier check. Measured `--report=source` across both changed PHP files, against merge-base `590070bf`:

| | violations | sources | `FunctionComment.Missing` |
|---|---|---|---|
| baseline `590070bf` | 192 | 31 | 3 |
| PR head `5af84040` | 193 | 31 | **4** |
| after `70b5d287` | 192 | 31 | 3 — byte-identical to baseline |

Worth noting for anyone re-running it: the **source count stays 31 on all three rows**. The class file already carries 3 pre-existing `FunctionComment.Missing`, so the regression showed up only as a count bump, not a new source line. A "did a new sniff appear?" check would have come back clean — which is plausibly how the original claim was made in good faith. It needs a count-level diff.

**Changed in `70b5d287`:** moved the docblock onto the method it documents and deleted the stranded copy. Pure 6-line move, no code touched; `phpcbf` was a no-op on it. Full suite green — 237 tests, 527 assertions.

**On the usual regression test:** deliberately none here, rather than an oversight. This is a comment-only relocation with no runtime behaviour to pin, so a PHPUnit test would assert nothing real, and nothing narrows a condition so there's no boundary case either. The two-directional check is the PHPCS measurement above: the violation is present without the change and absent with it, against a fixed baseline. Happy to add something if you see a behavioural angle I don't.

One flag unrelated to this change: `npm run test-client` currently fails in my environment on missing `wpcom` / `react-pure-render` modules, which is local `node_modules` breakage rather than anything on this branch (this commit touches one PHP file). PHP side is clean.

tests/php/test-class-wc-connect-taxjar-integration.php
```diff
@@ -2022,4 +2037,403 @@ public function provide_non_object_tax_nodes() {
...
+	/**
+	 * The other side of the same boundary: excluding "Shipping only" must not stop a
+	 * genuinely untaxed product from being recorded.
+	 *
+	 * Tax Status "None" is emitted as exempt (99999) and comes back at 0%, which is the
+	 * rate that would overwrite the shared Standard row. Pins that the `'shipping' !==`
+	 * clause narrows the condition by exactly one status and no more.
+	 */
+	public function test_get_line_items_still_records_none_status_as_non_taxable() {
+		$product = WC_Helper_Product::create_simple_product();
```
Abdalsalaam, line 3884: Unlike its siblings at :2121 and :2191, this test never calls update_option('woocommerce_calc_taxes','yes'), and set_up() (:44) does not enable taxes. WC_Product::is_taxable() is "'taxable' === get_tax_status() && wc_tax_enabled()", so with taxes off it is false for every product. Both assertions ('99999' emitted at :2329, key recorded at :2334) therefore hold identically for a plain 'taxable'-status product, so the test cannot pin what its docblock claims at :2314 — that the "'shipping' !==" clause narrows the condition by exactly one status and no more.

```suggestion
		// is_taxable() is false for every product while taxes are off, so without this the
		// assertions below hold for any Tax Status and cannot pin the 'shipping' exclusion.
		update_option( 'woocommerce_calc_taxes', 'yes' );

		$product = WC_Helper_Product::create_simple_product();
```

dustinparker (approved):
This fixes the bug. A cart mixing a taxable product with a Tax Status "None" product keeps calculating tax now, the shared Standard rate rows stay intact instead of being overwritten to 0%, and stores already carrying a zeroed row recover on their next calculation rather than staying stuck on it.

One thing before merge: I agree with @Abdalsalaam about `test_get_line_items_still_records_none_status_as_non_taxable`. With `woocommerce_calc_taxes` off, `is_taxable()` is false for every product, so those assertions hold for any Tax Status and the test can't pin the exclusion its docblock describes. His suggested one-liner covers it.

PR #2984, opened 2026-07-29
[WOOTAX-313] Refactor - Introduce the TaxJar Address value object

bartech:
## Description

`classes/class-wc-connect-taxjar-integration.php` handles addresses in **10 distinct shapes** — positional 5-tuple, `to_`/`from_`-prefixed, unprefixed nexus, `WC_Tax::find_rates` args, store settings, the `$location` rate-CRUD array, `address_parts`, TaxJar `jurisdictions`, and the WC order address. Normalisation (uppercasing, `wc_clean`, `wp_unslash`, comma-split postcode, semicolon-strip in city, space-strip in state) is re-derived at every call site.

A census of the file found **nine write/read asymmetries**, four of which are latent bugs rather than style — most notably `create_or_update_tax_rate()` *writing* the postcode through `wc_normalize_postcode( wc_clean() )` but *looking it up raw*, so a postcode that normalises differently never matches and a fresh rate row is inserted on every checkout. WOOTAX-19 fixed one instance of this class of bug by hand, one call site at a time. A single immutable value object makes the asymmetries structurally impossible.

This is **PR 1 of 4** in that refactor. It lands the value object **unwired** — nothing calls it on the calculation path yet. Each following PR migrates one seam and is independently revertable:

1. **R1.1 — this PR.** Land the value object + tests. No behaviour change.
2. **R1.2 read seam** — `get_taxable_address()`, `get_address()`, `get_backend_address()`, `append_base_address_to_customer_taxable_address()`.
3. **R1.3 request seam** — `calculate_tax()` body assembly, the nexus filter, `is_nexus_address_valid()`, `get_address_parts()`.
4. **R1.4 rate-table seam** — `create_or_update_tax_rate()` and `allow_street_address_for_matched_rates()` derive find-args and write-args from one object. Characterization tests land *before* that PR.

### What's in this PR

**`Automattic\WCServices\Tax\Address` (`src/Tax/Address.php`)** — immutable, with named constructors for each input shape (`from_taxable_tuple`, `from_options`, `from_nexus`, `from_store_settings`, `from_customer_billing`/`_shipping`, `from_post_request`, `from_jurisdictions`), normalised accessors, `validate()`, and one output method per consumer shape (`to_taxable_tuple`, `to_taxjar_body`, `to_nexus_array`, `to_find_rates_args`, `to_legacy_options`).

Placed in `src/` under an existing namespace rather than as a new global `WC_Connect_*` class **on purpose**: per `AGENTS.md` every `WC_Connect_*` symbol is a permanent external contract, and we do not want to publish a name we can never rename. The object is internal — every boundary converts back to the array shapes third-party code expects.

**`Address::jurisdiction_key()`** — the single place a jurisdiction cache key is derived, for both the write and the read side. It is added now because the follow-up rate-table work (WOOTAX-314) depends on it, and because deriving the key differently at each end is precisely what made an earlier attempt at jurisdiction-keyed rate rows (PR #2906) miss on every lookup; a reviewer on that PR asked for exactly this method.

Its field set is **measured against the live TaxJar API**, not assumed. An earlier `country + state + postcode` hypothesis was falsified by that run:

| Field | In the key? | Why |
|---|---|---|
| country, state | yes | — |
| postcode **as entered, `+4` included** | yes | `81323-2100` resolves to the county-only `US\|CO\|MONTEZUMA\|` even with an in-town city *and* street supplied. Truncating to 5 digits would merge genuinely different jurisdictions. |
| city (normalised) | yes | ZIP `81323` CO with city `Dolores` resolves to `US\|CO\|MONTEZUMA\|DOLORES`; with the city absent or misspelled it resolves to `US\|CO\|DOLORES\|RICO` — a different, real county. |
| street | **no** | Across three ZIPs and every field pairing it never moved `jurisdictions`, not even as a tiebreaker when the city was absent. |
| store nexus | **no** | Measured nexus-independent, so a resolved jurisdiction survives a store-address change. |

Eight tests pin these, anchored to the actual probed addresses so the suite doubles as the evidence record.

> **Invariant for the follow-up work, stated here so it is on the record:** nexus does *not* change the jurisdiction, but it *does* change the rate breakdown within it. A jurisdiction is therefore a sufficient rate-row identity **only while the store nexus is fixed** — a store-address change invalidates stored rates even though every jurisdiction key stays the same. WOOTAX-314 should carry a regression test for this rather than an assumption.

**`WC_Connect_TaxJar_Integration::normalize_city()` deprecated** (not removed) — see Backward compatibility below.

**`.phpcs.php.xml`** — `WordPress.Files.FileName` excluded for `src/`. `composer.json` maps `Automattic\WCServices\` => `src/` via PSR-4, which requires the file name to match the class name exactly, so the WordPress `class-my-thing.php` convention is unsatisfiable there. The sniff was reporting two unfixable errors on *every* file under `src/` (20 in total); this clears them. `classes/` is unaffected and keeps the WordPress convention.

### Reviewing this efficiently

The first commit is a verbatim `git cherry-pick -x` of the original extraction. **Diff `9e364a9d` against `f5e8f37e` to see exactly what relocation changed** — otherwise the value object reads as 1,000 lines of new code. The relocation commit contains: the namespace move, the `normalize_city()` reconciliation, per-method docblocks, and `jurisdiction_key()`.

### Backward compatibility

**No public surface is added, changed, or removed.** The value object is new, internal, namespaced, and unreferenced outside this repository. No hook is added, renamed, retimed, or has its arguments changed. No global state or `WC()->…` dereference is introduced. Nothing reads or writes options, so multisite and install-layout behaviour are unchanged.

One existing symbol is touched:

- **`WC_Connect_TaxJar_Integration::normalize_city()` is deprecated, not removed.** It now delegates to `Address::normalize_city()` so the normalisation policy has one implementation rather than two that can drift while the remaining seams are migrated. It is `protected static`, so a subclass outside this repository may be calling it and deleting it would break that subclass on load — per `AGENTS.md`, deprecate rather than remove. Its non-string guard is preserved verbatim, since callers may rely on a non-string argument coming back untouched. **No `_deprecated_function()` notice is raised**: four in-repo call sites still use the method and a notice today would fire on every checkout. The notice can be added in R1.4 once those call sites move.

  Behaviour is unchanged, and a new data-provider test asserts the deprecated method and the value object agree on every input the existing provider covers.

**One caveat worth flagging explicitly:** this PR is otherwise "unwired", but that delegate is a real runtime dependency from `class-wc-connect-taxjar-integration.php` on the value object. It is the only behavioural-surface touch in the PR. The alternative — duplicating the implementation for three PRs — is the exact drift the value object exists to prevent, so the trade was made deliberately.

### Related issue(s)

Part of [WOOTAX-313](https://linear.app/a8c/issue/WOOTAX-313) (R1.1 of 4 — the ticket stays open until R1.4 lands).

Unblocks [WOOTAX-314](https://linear.app/a8c/issue/WOOTAX-314) (jurisdiction-keyed tax rate rows), which consumes `jurisdiction_key()`.

Context: supersedes the approach in closed PR #2906, which was re-implemented rather than rebased.

No user-visible change, so there is nothing to screenshot. The claim to verify is **"zero behaviour change"**:

1. `composer test` — expect green. On this branch: **OK (280 tests, 602 assertions)**, of which 42 come from the new file (`./vendor/bin/phpunit --testdox --filter WP_Test_WCServices_Tax_Address`).
2. Confirm the existing shape-pinning tests in `tests/php/test-class-wc-connect-taxjar-integration.php` still pass unchanged — in particular the WOOTAX-19 `normalize_city` provider, which now also runs against the delegate.
3. On a store with Automated Taxes enabled, run a checkout in a multi-jurisdiction US destination and confirm the calculated tax and the `wp_woocommerce_tax_rates` rows are identical to trunk. Nothing on the calculation path calls the value object yet.
4. Lint: `./vendor/bin/phpcs --standard=./phpcs.xml.dist --report=source`.

   ⚠️ Note for the reviewer: `composer check-all` is **not** green on trunk today — the baseline is ~3,841 violations across 106 sources, 2,223 of them in `i18n/strings.php` alone. The meaningful check is therefore *no new violations*, verified by diffing `--report=source` before and after. This branch introduces **zero** new violations and removes 20 (the `src/` file-naming errors described above).

### Checklist

- [x] unit tests

### DIFF
--- a/src/Tax/Address.php
+++ b/src/Tax/Address.php
@@ -442,14 +442,33 @@ final class Address {
 	 * ignores, do not fragment the key. The separator is stripped from each
 	 * component first so a value containing it cannot forge a different key.
 	 *
+	 * The case folding is `wc_strtoupper()`, not `strtoupper()`, and the difference
+	 * is load-bearing. `strtoupper()` is byte-wise ASCII: it leaves every non-ASCII
+	 * byte alone, so `Zürich` folds to `ZüRICH` while `ZÜRICH` folds to itself —
+	 * two keys for one jurisdiction, which is precisely the fragmentation this
+	 * method exists to prevent. `normalize_city()` preserves multibyte input, so a
+	 * non-ASCII city reaches here intact from any international checkout.
+	 *
+	 * Note this deliberately diverges from the city sent to `to_find_rates_args()`,
+	 * which must stay on `strtoupper()`. That value's contract is byte-parity with
+	 * WooCommerce core, which upper-cases with plain `strtoupper()` on both the write
+	 * (`WC_Tax::format_tax_rate_city()`) and the read (the `location_code`
+	 * comparison inside `WC_Tax::find_rates()`). Folding wider than core there would
+	 * store `ZÜRICH` while core searched for `ZüRICH` — manufacturing the unbounded
+	 * rate-row growth that normalisation exists to close. Same operation, opposite
+	 * correct answers, because this key is ours and that one is core's.
+	 *
+	 * `wc_strtoupper()` falls back to `strtoupper()` when ext-mbstring is absent, so
+	 * this is no worse than the previous behavior on a minimal PHP build.
+	 *
 	 * @return string Key of the form `COUNTRY|STATE|POSTCODE|CITY`.
 	 */
 	public function jurisdiction_key(): string {
 		$components = array(
-			$this->country,
-			$this->state,
-			strtoupper( $this->postcode ),
-			strtoupper( $this->city ),
+			wc_strtoupper( $this->country ),
+			wc_strtoupper( $this->state ),
+			wc_strtoupper( $this->postcode ),
+			wc_strtoupper( $this->city ),
 		);
 
 		$components = array_map(
--- a/src/Tax/Address.php
+++ b/src/Tax/Address.php
@@ -16,31 +16,35 @@
  *
  * Each input shape used to be hand-extracted at the call site, and each output
  * shape was hand-built. That spread normalisation rules (uppercasing country/state,
- * trimming city semicolons per WOOTAX-19, taking the first segment of a comma-list
- * postcode) across many call sites and made it easy to forget one.
+ * trimming city semicolons, taking the first segment of a comma-list postcode)
+ * across many call sites and made it easy to forget one.
  *
  * This class centralises the policy:
  *
- *   $address = WC_Connect_TaxJar_Address::from_taxable_tuple( $tuple );
+ *   $address = Address::from_taxable_tuple( $tuple );
  *   $body    = $address->to_taxjar_body( 'to_' );
  *   $rates   = WC_Tax::find_rates( $address->to_find_rates_args( $tax_class ) );
  *
  * Construction normalises once; every accessor and `to_*()` method is a pure
  * function of the stored fields.
  *
- * @package WooCommerce_Services
+ * This object is **internal**. Every boundary that faces third-party code — filters,
+ * public method signatures — converts back to the array shapes those consumers expect.
+ *
+ * @package Automattic/WCServices
  */
 
-// phpcs:disable Squiz.Commenting.FunctionComment.MissingParamTag
+namespace Automattic\WCServices\Tax;
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit;
-}
+use WC_Customer;
+use WP_Error;
+
+defined( 'ABSPATH' ) || exit;
 
 /**
  * Address value object for the TaxJar integration. Immutable.
  */
-final class WC_Connect_TaxJar_Address {
+final class Address {
 
 	/**
 	 * Two-letter ISO country code, uppercase. Empty string if unknown.
@@ -85,8 +89,17 @@ final class WC_Connect_TaxJar_Address {
 	private $id = null;
 
 	/**
+	 * Separator joining the components of `jurisdiction_key()`.
+	 *
+	 * @var string
+	 */
+	private const KEY_SEPARATOR = '|';
+
+	/**
 	 * Validation schema used by `validate()`. Mirrors the rules previously inlined
-	 * in `is_nexus_address_valid()`.
+	 * in `WC_Connect_TaxJar_Integration::is_nexus_address_valid()`.
+	 *
+	 * @var array
 	 */
 	private const VALIDATION_SCHEMA = array(
 		'id'       => array(
@@ -119,6 +132,13 @@ final class WC_Connect_TaxJar_Address {
 
 	/**
 	 * Private constructor — use the named static factories below.
+	 *
+	 * @param string      $country  Country code, uppercased on construction.
+	 * @param string      $state    State / province code, uppercased on construction.
+	 * @param string      $postcode Postal code; only the first comma-separated segment is kept.
+	 * @param string      $city     City name; semicolons stripped and whitespace collapsed.
+	 * @param string      $street   Street address line, trimmed.
+	 * @param string|null $id       Optional nexus-address identifier.
 	 */
 	private function __construct( string $country = '', string $state = '', string $postcode = '', string $city = '', string $street = '', ?string $id = null ) {
 		$this->country  = self::upper( $country );
@@ -133,6 +153,8 @@ final class WC_Connect_TaxJar_Address {
 
 	/**
 	 * Empty address. Useful for early-return paths (e.g. when `WC()->customer` is null).
+	 *
+	 * @return self
 	 */
 	public static function empty(): self {
 		return new self();
@@ -144,6 +166,7 @@ final class WC_Connect_TaxJar_Address {
 	 * Matches the contract of the `woocommerce_customer_taxable_address` filter.
 	 *
 	 * @param array $tuple Indexed 5-element array. Missing positions become empty strings.
+	 * @return self
 	 */
 	public static function from_taxable_tuple( array $tuple ): self {
 		return new self(
@@ -160,6 +183,7 @@ final class WC_Connect_TaxJar_Address {
 	 *
 	 * @param array  $options Source array. Missing keys default to empty.
 	 * @param string $prefix  Either 'to_' (default) or 'from_'.
+	 * @return self
 	 */
 	public static function from_options( array $options, string $prefix = 'to_' ): self {
 		return new self(
@@ -176,6 +200,9 @@ final class WC_Connect_TaxJar_Address {
 	 *
 	 * Note: `zip` (not `postcode`) — matches both the TaxJar nexus_addresses[] payload
 	 * and the `woocommerce_taxjar_nexus_address` filter contract.
+	 *
+	 * @param array $nexus Nexus address array, optionally carrying an `id`.
+	 * @return self
 	 */
 	public static function from_nexus( array $nexus ): self {
 		return new self(
@@ -192,6 +219,9 @@ final class WC_Connect_TaxJar_Address {
 	 * Build from the store-settings shape `{street, city, state, country, postcode}`.
 	 *
 	 * Matches the array returned by `WC_Connect_TaxJar_Integration::get_store_settings()`.
+	 *
+	 * @param array $settings Store settings array.
+	 * @return self
 	 */
 	public static function from_store_settings( array $settings ): self {
 		return new self(
@@ -207,6 +237,7 @@ final class WC_Connect_TaxJar_Address {
 	 * Build from a WooCommerce customer's billing address.
 	 *
 	 * @param WC_Customer $customer Live customer object — typically `WC()->customer`.
+	 * @return self
 	 */
 	public static function from_customer_billing( WC_Customer $customer ): self {
 		return new self(
@@ -222,6 +253,7 @@ final class WC_Connect_TaxJar_Address {
 	 * Build from a WooCommerce customer's shipping address.
 	 *
 	 * @param WC_Customer $customer Live customer object — typically `WC()->customer`.
+	 * @return self
 	 */
 	public static function from_customer_shipping( WC_Customer $customer ): self {
 		return new self(
@@ -240,6 +272,7 @@ final class WC_Connect_TaxJar_Address {
 	 * nonce / capability — this method does not check authorization.
 	 *
 	 * @param array|null $post Override source for testing. Defaults to `$_POST`.
+	 * @return self
 	 */
 	public static function from_post_request( ?array $post = null ): self {
 		// phpcs:disable WordPress.Security.NonceVerification.Missing
@@ -266,6 +299,7 @@ final class WC_Connect_TaxJar_Address {
 	 * postcode and street default to empty.
 	 *
 	 * @param object|array|null $jurisdictions A `$response_body->tax->jurisdictions` snippet.
+	 * @return self
 	 */
 	public static function from_jurisdictions( $jurisdictions ): self {
 		if ( null === $jurisdictions ) {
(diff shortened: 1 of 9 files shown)


src/Tax/Address.php
```diff
@@ -0,0 +1,672 @@
...
+	 *
+	 * Applies `wc_clean()` to each field. Caller must have already verified the
+	 * nonce / capability — this method does not check authorization.
+	 *
+	 * @param array|null $post Override source for testing. Defaults to `$_POST`.
+	 * @return self
+	 */
+	public static function from_post_request( ?array $post = null ): self {
+		// phpcs:disable WordPress.Security.NonceVerification.Missing
+		$source = is_array( $post ) ? $post : $_POST;
```
Abdalsalaam: WordPress slashes $_POST and wc_clean()/sanitize_text_field() does not strip slashes, so a city or street such as O'Brien is carried through as O\'Brien into the TaxJar request body and the wp_woocommerce_tax_rates tables. Legacy get_backend_address() (classes/class-wc-connect-taxjar-integration.php:931-935) has the same omission, so this is not a regression today, but it is a defect being carried into new code while it is still unwired and free to fix. PHPCS cannot flag it because the superglobal is aliased into $source one line before use, so the PR's zero-new-violations claim rests on the sniff being blinded rather than satisfied.

```suggestion
		$source = is_array( $post ) ? $post : wp_unslash( $_POST ); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- Each field is passed through wc_clean() below.
```

iyut (approved):
This is a well-built foundation PR. The `Address` value object is genuinely immutable, `normalize_city()` is deprecated correctly (kept + `@deprecated` + delegates, with no `_deprecated_function()` notice since live in-repo callers would fire it every checkout), the PHPCS exemption is narrowly scoped to the file-naming sniff, and it's genuinely unwired — zero behavior change on the calc path today. The namespacing rationale (keeping it out of the permanent `WC_Connect_*` contract) is the right call.

Two things addressed before approval:

- **Fixed the missing `wp_unslash()`** that @Abdalsalaam caught in `from_post_request()` — the real `$_POST` path now unslashes (the `$post` test override stays unslashed), replacing the aliased-superglobal pattern that was blinding the input-sanitization sniff. Added `test_from_post_request_unslashes_superglobal` to pin it (`O'Brien` no longer stored as `O\'Brien`). PHPCS is green on `src/Tax/Address.php`.
- **Logged three normalization divergences** as migration-time watch-items (see the separate comment): backend-address uppercasing, from-side over-normalization, and `validate()` vs `is_nexus_address_valid()` strictness. None are bugs today; each seam migration (R1.2–R1.4) should either preserve legacy behavior or state the divergence is an intended fix per `AGENTS.md`, and land golden-parity tests alongside.

Good to merge once the follow-up seam PRs pick up those parity tests. 🚀

iyut:
**Migration-time watch items — normalization divergences from the legacy call sites**

This foundation PR is clean and unwired, so none of these are bugs today. But since the whole point of `Address` is that later "no behavior change" PRs swap call sites into it, I want to pin three places where the object does **not** byte-for-byte reproduce the legacy normalization. When each seam is migrated (R1.2–R1.4), the PR should either preserve the legacy behavior or, if the divergence is the intended fix, **say so explicitly** — per `AGENTS.md`, an inherited-behavior change makes it a bug-fix PR, not "no behavior change."

1. **Backend address loses uppercasing (and `''` → `false`).** `get_backend_address()` (`classes/class-wc-connect-taxjar-integration.php:930-935`) uppercases *every* field (country/state/postcode/city/street) and maps present-but-empty via `isset()` only. `from_post_request()->to_legacy_options()` uppercases only country/state, leaves postcode/city/street mixed-case, and converts `'' → false`. Migrating this seam changes the admin-order-recalc TaxJar body from UPPERCASE to mixed-case. Add a parity test before wiring it.

2. **From-side (store) over-normalizes.** In `calculate_tax()` the store fields are passed through raw — `$from_zip = $store_settings['postcode']` (`:1552`), `$from_city = $store_settings['city']` (`:1553`), and `get_address_parts()` (`:1690-1694`) only uppercases, never `explode(',')` or `normalize_city()`. Only `to_zip` is comma-split (`:1546-1547`). The value object applies `first_postcode_segment()` + `normalize_city()` to *every* address regardless of side, so `from_store_settings()->to_taxjar_body('from_')` would comma-split the store ZIP and strip semicolons from the store city. Low real-world probability, but a real divergence.

3. **`validate()` is stricter than `is_nexus_address_valid()`.** `is_nexus_address_valid()` (`:1407`) only errors a required field when the **key is absent**; a present-but-empty `state`/`country` passes (e.g. a store in a stateless country carries `state => ''` and currently validates). `Address::validate(['country','state'])` fails any empty required field. When migrating the nexus check, pass a `$required` set that matches the old key-presence semantics (or `array()`), not the strict default — otherwise you change which addresses get their `from_*` fields replaced by `nexus_addresses`.

Recommend landing **golden-parity tests** (same input → legacy method vs object → assert equal) alongside each seam migration; those are what actually protect the "no behavior change" claim and would catch all three mechanically.

bartech:

src/Tax/Address.php
```diff
@@ -0,0 +1,672 @@
...
+	 * component first so a value containing it cannot forge a different key.
+	 *
+	 * @return string Key of the form `COUNTRY|STATE|POSTCODE|CITY`.
+	 */
+	public function jurisdiction_key(): string {
+		$components = array(
+			$this->country,
+			$this->state,
+			strtoupper( $this->postcode ),
+			strtoupper( $this->city ),
```
Abdalsalaam: The docblock at src/Tax/Address.php:441 states "Components are upper-cased so that differences of letter case, which TaxJar ignores, do not fragment the key." strtoupper() is locale-independent ASCII-only, so 'Zurich' with an umlaut folds to 'ZuRICH' while the already-uppercase spelling folds to 'ZURICH' - two keys for one TaxJar jurisdiction. The same applies to upper() at :648 for country/state. TaxJar supports international addresses, so non-ASCII cities are reachable from checkout input. Tests cover ASCII only (tests/php/Tax/test-address.php:404). WooCommerce ships wc_strtoupper(), which is mbstring-aware. to_find_rates_args() at :577 should stay on strtoupper() because it must match WC core's stored city format.

```suggestion
			wc_strtoupper( $this->postcode ),
			wc_strtoupper( $this->city ),
```

bartech: Confirmed and taken — `8eb99c8c`. Verified empirically before changing anything:

| input | `strtoupper` | `mb_strtoupper` |
|---|---|---|
| `Zürich` | `ZüRICH` | `ZÜRICH` |
| `ZÜRICH` | `ZÜRICH` | `ZÜRICH` |
| `São Paulo` | `SãO PAULO` | `SÃO PAULO` |

Two keys, one jurisdiction — which contradicts the method's own docblock ("upper-cased so that differences of letter case ... do not fragment the key"). And the path is genuinely reachable: `normalize_city()` deliberately preserves multibyte input (pinned by `test_normalize_city_preserves_multibyte_characters`), so a non-ASCII city arrives at the key intact.

### Your carve-out is right, and it is load-bearing

I checked it against core rather than assuming, because it is the part that would have been easy to "clean up" later. WooCommerce uses plain `strtoupper()` on **both** sides:

* write — `WC_Tax::format_tax_rate_city()`, `class-wc-tax.php:998`
* read — the `location_code` comparison inside `WC_Tax::find_rates()`, `class-wc-tax.php:403` and `:412`

So the contract there is *byte-parity with core*, not Unicode correctness. Folding wider would store `ZÜRICH` while core searched for `ZüRICH` — manufacturing exactly the write/read asymmetry that normalisation exists to close, i.e. a fresh rate row per calculation. Same operation, opposite correct answers, because one key is ours and the other is core's.

Since that is a genuinely counter-intuitive divergence, I added `test_find_rates_city_matches_core_ascii_folding_not_multibyte` to pin it, so a later consistency pass cannot quietly collapse the two call sites onto one function. (That one is a characterization test — it passes with and without the fix, which is correct for its job.)

### Went slightly wider than the suggestion, deliberately

I applied `wc_strtoupper()` to all four key components rather than just postcode and city, which picks up your `upper()` note — but **without** touching `upper()` itself.

The reasoning: `upper()` is applied to country and state in the constructor (`:144-145`), so its output feeds *every* consumer of the object — TaxJar request bodies, legacy option shapes, `find_rates()` args — and core also `strtoupper()`s state and country on both sides (`:371-372`, `:1008`, `:1019`). Changing it is not the same low-risk edit as changing the key. But your underlying point stands for the key itself: a state carried as free text — which is what WooCommerce stores for countries with no predefined state list — reaches `jurisdiction_key()` half-folded. Folding inside the key finishes the job with the blast radius contained to the key.

Left alone on purpose: `state_compact()` (its `sanitize_key()` strips non-ASCII anyway, so folding cannot help) and `upper()`.

`wc_strtoupper()` falls back to `strtoupper()` when ext-mbstring is absent, so this is no worse than the previous behaviour on a minimal PHP build.

### Severity, for the record

`jurisdiction_key()` currently has **no callers** anywhere in the stack — grep across all four branches turns up only `src/Tax/Address.php` and its own tests. So this is latent rather than live. Same posture as the `wp_unslash()` finding on this PR: cheap to fix while unwired, considerably less cheap once it keys stored rate rows. No changelog entry for that reason — no user-visible behaviour changed.

### Verification

Both new key tests fail against the previous implementation with exactly the predicted diff:

```
-'CH|ZH|8001|ZÜRICH'
+'CH|ZH|8001|ZüRICH'
```

and pass with it. Full suite 362 tests / 881 assertions at the top of the stack (was 359/872 — +3 tests, +9 assertions).

The whole stack was restacked onto this commit; all four branches are `MERGEABLE`, both approvals survived, and this PR took the change as a fast-forward rather than a history rewrite.

src/Tax/Address.php
bartech: Good catch, and agreed on both halves — the missing `wp_unslash()` and the fact that aliasing the superglobal into `$source` is what hid it from the sniff. Sorry this reply took so long; the fix went in well before the write-up did.

That second half is the part worth keeping: a zero-new-violations claim is only as good as what the sniffs can still see. The violation count was honest and the code was still wrong, because the alias moved the superglobal out of the sniff's line of sight one line before use.

This landed as `28b2b7e6` (thanks @iyut), taking the `wp_unslash( $_POST )` form on `$source` with the sniff exemptions spelled out, rather than unslashing per field. Note the SHA if you had the old one — the stack has been rebased onto trunk twice since (3.6.11 shipped mid-review), so the commit you may have seen as `c3f052b5` is now `28b2b7e6`.

### One thing worth flagging for the rest of the stack, because it bit us

**#2985 had already fixed the same defect independently**, per field inside `$pluck`. When #2985 was rebased onto this branch, both fixes survived — they sit on different lines of the same method, so git merged them with no conflict — and the result was a double `wp_unslash()`.

`wp_unslash()` is `stripslashes_deep()` and is not idempotent. `O'Brien` still came out right, which is exactly why nothing went red: the existing apostrophe test passes under one pass or two. A street entered as `A\B` was being reduced to `AB`.

Removed on #2985 in `d8995a5e`, with two tests that pin it:

* `test_from_post_request_unslashes_only_once` — `A\B` survives as `A\B`
* `test_from_post_request_does_not_unslash_the_override` — the `$post` test seam is *not* unslashed, since it is never slashed to begin with

The general shape is worth remembering for stacked PRs: two branches fixing the same defect on different lines of one method is a **silent** merge, not a conflict, and the fix that is not idempotent is the one that hurts.

No change needed on this PR.

PR #2985, opened 2026-07-29
[WOOTAX-313] Refactor - Migrate the TaxJar read seam onto the Address value object

bartech:
### Stacked PR

**Base is `chore/wootax-313` (#2984), not `trunk`.** This is R1.2 of the address/rate-table
refactor; #2984 is R1.1 and is still open. Review #2984 first — this diff only makes sense on
top of the value object it lands. If #2984 merges, this will be rebased onto `trunk` and the
base updated.

R1.1 landed `Automattic\WCServices\Tax\Address` unwired. R1.2 wires the **read seam**: the four
places that build an address for a tax calculation now construct exactly one `Address` each and
convert back at the boundary, so normalisation policy has a single home instead of being
re-derived per call site.

| Call site | Now |
|---|---|
| `get_taxable_address()` | `Address::from_store_settings()` / `from_customer_billing()` / `from_customer_shipping()` → `to_taxable_tuple()` |
| `get_address()` | `Address::from_taxable_tuple( … )->to_legacy_options()` |
| `get_backend_address()` | `Address::from_post_request()->to_legacy_options()` |
| `append_base_address_to_customer_taxable_address()` | arity-preserving rebuild through the value object |

Three defects fall out of unifying the seam. Each has a regression test that **fails against the
base branch**.

---

## 1. The taxable-address tuple changed length depending on its data

This is the one worth reading closely, because the fix is the opposite of what it looks like.

`woocommerce_customer_taxable_address` is fired with **two different arities**:

- **WooCommerce core** — `WC_Customer::get_taxable_address()` fires it with **four** elements:
  country, state, postcode, city. No street.
- **This plugin** — `WC_Connect_TaxJar_Integration::get_taxable_address()` fires the *same*
  filter with **five**, the fifth being the street, which `get_address()` reads off index `[4]`.

And core's consumer checks that length strictly:

```php
// WC_Tax::get_rates_from_location()
if ( count( $location ) === 4 ) {
    list( $country, $state, $postcode, $city ) = $location;
    $matched_tax_rates = self::find_rates( … );
}
return apply_filters( 'woocommerce_matched_rates', $matched_tax_rates, $tax_class, $customer );
```

Five elements ⇒ `find_rates()` is never called and core returns an **empty rate set**.

`append_base_address_to_customer_taxable_address()` used to size its return by whether the
street happened to be empty:

```php
if ( '' != $street ) {
    return array( $country, $state, $postcode, $city, $street );
}
return array( $country, $state, $postcode, $city );
```

So it returned **five elements into core's four-element pipeline** on any local-pickup checkout
where the store has a street address configured — because the `'base'` branch substitutes the
store street into the previously-empty slot. And it **dropped the fifth slot from this plugin's
own pipeline** whenever the street was empty, which is the originally-reported symptom.

The reason this has not been reported as "local pickup collects no tax" is that
`allow_street_address_for_matched_rates()` is unconditionally hooked on
`woocommerce_matched_rates` and replaces the (empty) result wholesale with its own
`find_rates()` call. So the damage is contained to callers of `WC_Tax::get_rates()` /
`get_rates_from_location()` that read the return value before that filter — and to anything that
removes or reorders our filter.

**Fix:** return a tuple of exactly the length we were handed, clamped to the two lengths this
filter is ever fired with.

Both directions are covered: `test_append_base_address_preserves_four_element_core_tuple()` and
`test_append_base_address_preserves_five_element_plugin_tuple()`.

> **Note for R2.** The `woocommerce_get_tax_location` override planned for the jurisdiction work
> returns `$location` straight back into that same strict `count( $location ) === 4` check. It
> must not grow a street element either.

## 2. The two getters disagreed on how to represent an empty field

`get_address()` returned `false` for an empty field. `get_backend_address()` returned `''` for a
`$_POST` key that was present but blank. Both feed `calculate_tax()`, which serialises the values
straight into the TaxJar request body — so the cart path and the admin-recalculate path put
different JSON on the wire for the same empty input.

`Address::to_legacy_options()` is now the single place that decision is made, and
`test_read_seam_uses_one_empty_value_representation()` asserts the two paths agree.

## 3. `get_backend_address()` never unslashed `$_POST`

WordPress slashes every superglobal on load, and `wc_clean()` sanitises without unslashing. So an
apostrophe in an admin order address reached TaxJar — and, via `create_or_update_tax_rate()`, the
`wp_woocommerce_tax_rates` city column — as the literal `O\'Brien`. That value can never match
the unslashed city the cart path writes for the same address, so recalculating an order from the
order-edit screen could mint a second rate row for a city the store already has one for.

`wp_unslash()` is now applied ahead of `wc_clean()` in `Address::from_post_request()`.
Covered by `test_get_backend_address_unslashes_post_values()`.

---

## Backward compatibility

Per AGENTS.md, this section covers signatures, hooks, global-state assumptions, multisite and
install layout.

**Hooks**

- `woocommerce_customer_taxable_address` keeps receiving and returning WooCommerce's positional
  tuple. No argument is added, removed or reordered. Its **arity is now preserved** rather than
  data-dependent, which is a fix in both directions: core callbacks stop seeing an unexpected
  fifth element, and this plugin's callbacks stop losing the fifth slot. A third-party callback
  registered after ours that previously branched on `count( $address )` will now see a stable
  length — that is the contract core documents, and the previous behaviour was not something a
  consumer could have depended on deliberately.
- **The values passed to that filter are now normalised before it fires, rather than after.**
  Country and state upper-cased, city semicolon-stripped and whitespace-collapsed, a comma-list
  postcode reduced to its first segment. Every one of those transformations was already applied
  by this plugin *downstream* of the filter, so our own pipeline is unchanged; the difference is
  that third-party callbacks now receive the canonical value instead of the raw one. Called out
  explicitly because it is the only change here that a consumer could observe without asking for
  it.
- No other hook is added, removed, renamed or re-timed. `taxjar_store_settings` and
  `woocommerce_taxjar_nexus_address` are untouched.

**Public / externally exposed methods**

- `get_taxable_address()` — public. Signature and return shape unchanged. The
  `array( '', '', '', '', '' )` early return for a null `WC()->customer` is untouched and still
  bypasses the filter; two existing tests pin it.
- `append_base_address_to_customer_taxable_address()` — public. Signature unchanged; return shape
  unchanged apart from the arity fix described above. The `@param` tag it was missing is now
  documented.
- `get_address()` and `get_backend_address()` are `protected`. A subclass outside this repo could
  be calling them, so both keep their array shape and key names.
- `get_backend_address()` **keeps its historic `strtoupper()` over postcode, city and street**,
  applied after the value object builds them. This looks like a leftover and is not: the city
  casing feeds the rate-table write path and is pinned by
  `test_get_backend_address_normalizes_semicolon_city()`, while `get_address()` does *not*
  upper-case. Reconciling that asymmetry means moving both ends of the write/read round trip
  together, which is R1.4's job. There is a comment on it in the code so it does not get
  "cleaned up" in isolation.
- `WC_Connect_TaxJar_Integration::normalize_city()` keeps its deprecated delegate and still raises
  **no** `_deprecated_function()` notice. R1.2 moved two of its five in-repo call sites; the three
  that remain are all on the rate-table seam, and a notice would fire on every checkout until
  R1.4 moves them.

**Two behaviour changes worth a reviewer's eye**

1. `'0'` is no longer treated as an empty field. The old `! empty()` extraction turned a postcode
   or city of `"0"` into `false`; `Address::to_legacy_options()` only maps `''`. Strictly a fix,
   but it is a change.
2. A `$_POST` field that is present but blank now yields `false` rather than `''` from
   `get_backend_address()`. That is defect 2 above, and it makes the admin path match the cart
   path.

**Global state, multisite, install layout**

- No new read of a global or of `WC()->…` state is introduced. `get_taxable_address()` already
  null-guarded `WC()->customer` and still does.
  `append_base_address_to_customer_taxable_address()`'s pre-existing `WC()->session` dereference
  is in the WC < 2.6.2 fallback branch only and is unchanged.
- No option, table or upload path is read or written, so there is no site-scoped vs
  network-scoped question. Behaviour under multisite is unchanged; not separately tested, because
  nothing here touches storage.
- No path or URL construction. Install layout is not a factor.

---

## Follow-up this PR deliberately does not action

**`append_base_address_to_customer_taxable_address()` looks redundant against modern
WooCommerce core, not just buggy.** Core's `WC_Customer::get_taxable_address()` already performs
the local-pickup → base substitution itself, for all four fields, reading the same
`WC()->countries->get_base_*` values that `get_store_settings()` reads:

```php
if ( true === apply_filters( 'woocommerce_apply_base_tax_for_local_pickup', true ) && count( array_intersect( wc_get_chosen_shipping_method_ids(), … ) ) > 0 ) {
    $tax_based_on = TaxBasedOn::BASE;
}
if ( TaxBasedOn::BASE === $tax_based_on ) {
    $country  = WC()->countries->get_base_country();
    …
}
```

So on core's invocation of the filter, our callback re-derives postcode / city / street to the
values core already put there, and its only *observable* effect was the arity change that broke
the rate lookup. On this plugin's own invocation it is likewise near-redundant, since
`get_taxable_address()` has already resolved `'base'` before firing the filter — and it is
arguably incoherent, because it overrides postcode / city / street to the store's while leaving
country and state as the customer's.

Retiring it is a backward-compatibility change (a public method and a registered callback on a
public filter), so per AGENTS.md it wants its own ticket, a deprecation window, and a check on
whether any grandfathered local-pickup behaviour depends on the street it appends. Filing that
separately rather than bundling it here — bundling unrelated concerns is one of the four reasons
the earlier PR #2906 was closed.

---

## Verification

- `composer test` → **OK (285 tests, 613 assertions)**. 280 on the base branch, +5 new, zero
  regressions.
- PHPCS → **zero new violations, net −2** against the base branch (3821 → 3819): one
  `Squiz.Commenting.FunctionComment.MissingParamTag` fixed by documenting `$address`, one
  `Universal.Operators.StrictComparisons.LooseNotEqual` removed with `'' != $street`.
  Measured by diffing `phpcs --standard=./phpcs.xml.dist --report=source` before and after —
  `check-all` is **not** green on this repo's trunk (~3841 pre-existing violations, 2223 of them
  in `i18n/strings.php`), so "zero new" is the achievable bar, not "green".
- `php -l` clean on all three changed files, matching the CI lint step.

## Testing instructions

1. Configure a US store with a **street address** set in WooCommerce → Settings → General.
2. Enable Automated Taxes.
3. Add a local pickup shipping method and choose it in the cart. Confirm tax is calculated and,
   with `WC_Tax::get_rates()` called directly (e.g. via WP-CLI `wp eval`), that it now returns
   rates instead of an empty array.
4. Place an order to a US destination, open it in the admin, change the city to one containing an
   apostrophe (e.g. `O'Brien`), and recalculate. Confirm the city stored in
   `wp_woocommerce_tax_rates` / `wp_woocommerce_tax_rate_locations` has no backslash.
5. Confirm a second recalculation of the same order does not add a new rate row.

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -446,6 +446,13 @@ class WC_Connect_TaxJar_Integration {
 	 * Modified version of TaxJar's plugin.
 	 * See: https://github.com/taxjar/taxjar-woocommerce-plugin/blob/4b481f5/includes/class-wc-taxjar-integration.php#L910
 	 *
+	 * The `taxjar_store_settings` return value is guaranteed to be an array, matching
+	 * this method's documented contract. The filter is public and unguarded, so a
+	 * callback returning null / false / a string would otherwise flow into callers that
+	 * index it — and, since the address value object declares `array $settings`, into an
+	 * uncaught TypeError on every cart and checkout render that resolves a base address.
+	 * A non-array return is treated as "no opinion" and the unfiltered settings are used.
+	 *
 	 * @return array
 	 */
 	public function get_store_settings() {
@@ -457,7 +464,15 @@ class WC_Connect_TaxJar_Integration {
 			'postcode' => WC()->countries->get_base_postcode(),
 		);
 
-		return apply_filters( 'taxjar_store_settings', $store_settings, array() );
+		$filtered = apply_filters( 'taxjar_store_settings', $store_settings, array() );
+
+		if ( ! is_array( $filtered ) ) {
+			$this->_log( 'Warning: the taxjar_store_settings filter returned a non-array value; ignoring it and using the store address.' );
+
+			return $store_settings;
+		}
+
+		return $filtered;
 	}
 
 	/**
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1328,8 +1328,8 @@ class WC_Connect_TaxJar_Integration {
 	 * plugin's own pipeline whenever the street was empty.
 	 *
 	 * @param array $address Positional taxable-address tuple, four or five elements.
-	 * @return array The same tuple, same length, with the base address substituted for
-	 *               local pickup.
+	 * @return array The tuple with the base address substituted for local pickup,
+	 *               its length clamped to four or five elements to match the input.
 	 */
 	public function append_base_address_to_customer_taxable_address( $address ) {
 		$tax_based_on = '';
@@ -1357,16 +1357,11 @@ class WC_Connect_TaxJar_Integration {
 		$tuple = Address::from_taxable_tuple( array( $country, $state, $postcode, $city, $street ) )->to_taxable_tuple();
 
 		/*
-		 * Return the arity we were handed, rather than a length of our own choosing.
-		 * Deliberately expressed as "preserve the input" and not as "core sends four,
-		 * we send five": this stays correct if a WooCommerce version we did not test
-		 * against fires the filter with a different shape, and it does not depend on
-		 * the plugin's declared minimum WooCommerce version holding.
-		 *
-		 * `$tuple` always has five elements, so `array_slice()` bounds the upper end on
-		 * its own. The only clamp needed is the lower one, which keeps a malformed
-		 * short tuple from being propagated below the four fields every consumer of
-		 * this filter reads.
+		 * Return the arity we were handed, clamped to the two shapes this filter is
+		 * known to carry: four elements (core) or five (this plugin). `$tuple` is
+		 * rebuilt with exactly five elements, so a longer input loses anything past
+		 * the street slot, and a shorter one is padded up to the four fields every
+		 * consumer of this filter reads.
 		 */
 		return array_slice( $tuple, 0, max( 4, $tuple_length ) );
 	}
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1338,9 +1338,19 @@ class WC_Connect_TaxJar_Integration {
 
 		$tuple = Address::from_taxable_tuple( array( $country, $state, $postcode, $city, $street ) )->to_taxable_tuple();
 
-		// Return the arity we were handed. Clamped to the two lengths this filter is
-		// ever fired with, so a malformed input tuple cannot propagate its own shape.
-		return array_slice( $tuple, 0, max( 4, min( 5, $tuple_length ) ) );
+		/*
+		 * Return the arity we were handed, rather than a length of our own choosing.
+		 * Deliberately expressed as "preserve the input" and not as "core sends four,
+		 * we send five": this stays correct if a WooCommerce version we did not test
+		 * against fires the filter with a different shape, and it does not depend on
+		 * the plugin's declared minimum WooCommerce version holding.
+		 *
+		 * `$tuple` always has five elements, so `array_slice()` bounds the upper end on
+		 * its own. The only clamp needed is the lower one, which keeps a malformed
+		 * short tuple from being propagated below the four fields every consumer of
+		 * this filter reads.
+		 */
+		return array_slice( $tuple, 0, max( 4, $tuple_length ) );
 	}
 
 	/**
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -446,12 +446,15 @@ class WC_Connect_TaxJar_Integration {
 	 * Modified version of TaxJar's plugin.
 	 * See: https://github.com/taxjar/taxjar-woocommerce-plugin/blob/4b481f5/includes/class-wc-taxjar-integration.php#L910
 	 *
-	 * The `taxjar_store_settings` return value is guaranteed to be an array, matching
-	 * this method's documented contract. The filter is public and unguarded, so a
-	 * callback returning null / false / a string would otherwise flow into callers that
-	 * index it — and, since the address value object declares `array $settings`, into an
-	 * uncaught TypeError on every cart and checkout render that resolves a base address.
-	 * A non-array return is treated as "no opinion" and the unfiltered settings are used.
+	 * The `taxjar_store_settings` return value is guaranteed to be an array carrying
+	 * all five address keys, matching this method's documented contract. The filter is
+	 * public and unguarded, so a callback returning null / false / a string would
+	 * otherwise flow into callers that index it — and, since the address value object
+	 * declares `array $settings`, into an uncaught TypeError on every cart and checkout
+	 * render that resolves a base address. A non-array return is treated as "no
+	 * opinion" and the unfiltered settings are used; keys missing from an array return
+	 * are backfilled from the unfiltered settings, so callers can keep indexing the
+	 * result directly.
 	 *
 	 * @return array
 	 */
@@ -472,7 +475,7 @@ class WC_Connect_TaxJar_Integration {
 			return $store_settings;
 		}
 
-		return $filtered;
+		return array_merge( $store_settings, $filtered );
 	}
 
 	/**
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -673,6 +673,10 @@ class WC_Connect_TaxJar_Integration {
 	/**
 	 * Get formatted address for tax calculation.
 	 *
+	 * Empty fields come back as `false` rather than `''`. That is the historic shape of
+	 * this array and `Address::to_legacy_options()` is the single place the choice is
+	 * made, so `get_backend_address()` cannot drift away from it.
+	 *
 	 * @param string|null $location_type Location type: 'base', 'shipping', or 'billing'.
 	 *                                   If null, uses default behavior from get_taxable_address().
 	 * @return array Address array with to_country, to_state, etc.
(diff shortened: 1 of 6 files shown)


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -892,56 +889,55 @@ public function get_taxable_address( $location_type = null ) {
 		}
 
 		if ( 'base' === $tax_based_on ) {
-			$store_settings = $this->get_store_settings();
-			$country        = $store_settings['country'];
-			$state          = $store_settings['state'];
-			$postcode       = $store_settings['postcode'];
-			$city           = $store_settings['city'];
-			$street         = $store_settings['street'];
+			$address = Address::from_store_settings( $this->get_store_settings() );
```
Abdalsalaam, line 910: `get_store_settings()` returns `apply_filters( 'taxjar_store_settings', $store_settings, array() )` (line 460) — a public filter with no return-type guard — and `Address::from_store_settings()` declares `array $settings`. A callback returning null/false/an object/a string previously produced only a PHP warning on `$store_settings['country']` and tax calculation continued with empty values; it is now an uncaught TypeError on every cart/checkout render where tax_based_on === 'base' or local pickup is chosen. Lines 1333 and 1574 still read the same filtered value with raw array access and still degrade, so the file is now internally inconsistent about it.

```suggestion
			$store_settings = $this->get_store_settings();
			$address        = Address::from_store_settings( is_array( $store_settings ) ? $store_settings : array() );
```

bartech:
## Pushed — review fixes + rebase onto the updated base

Rebased onto `chore/wootax-313` (which picked up `c3f052b5`) and addressed @Abdalsalaam's review. Two changes.

### 1. `taxjar_store_settings` returning a non-array is no longer fatal

@Abdalsalaam was right that this is a real fatal, not a theoretical one: `get_store_settings()` returned an unguarded `apply_filters()` result straight into `Address::from_store_settings( array $settings )`. A callback returning `null` / `false` / a string / an object previously produced a PHP warning and calculation continued with empty values; against the typed parameter it becomes an uncaught `TypeError` on **every cart and checkout render** that resolves a base address.

Fixed in `get_store_settings()` rather than at the call site, for two reasons:

- The method already documents `@return array`, so a non-array return is it breaking its own contract. Enforcing it there is the honest fix.
- @Abdalsalaam's point about the file being internally inconsistent is the deciding one — a call-site ternary fixes one line and leaves the other raw array reads still warning. Guarding the source covers every consumer, including future ones.

A non-array return is now treated as "no opinion": it's logged and the unfiltered store settings are used, so a broken callback yields a *correct* calculation rather than a silently empty one.

Pinned by `test_get_taxable_address_survives_non_array_store_settings` (provider: null / false / string / object). All four raise an uncaught `TypeError` without the guard.

### 2. Removed a double `wp_unslash()` that the rebase created

Worth flagging because nothing went red. This branch and its base fixed the *same* unslashing defect independently — the base on `$source`, this branch per field inside `$pluck`. They sit on different lines of the same method, so the rebase merged both with **no conflict**, leaving `wp_unslash()` applied twice.

`wp_unslash()` is `stripslashes_deep()` and is not idempotent. Apostrophes still came out right — which is exactly why the existing test kept passing — but a street entered as `A\B` was being reduced to `AB`.

Kept the base branch's version, removed the per-field one, and added two tests: one pinning unslash-exactly-once, one pinning that the `$post` test override is *not* unslashed (it is never slashed to begin with).

### Backward compatibility

`taxjar_store_settings` is a public filter and its contract is unchanged — same input shape, same expected return. What changed is that a **non-conforming return is now ignored** instead of propagated. Any callback returning an array is unaffected; a callback returning a non-array was already producing warnings and empty tax data, so nothing that previously worked stops working. No signature, hook name, argument count or firing order changed. Site-scoped, so no multisite implication.

### Verification

- `composer test` — 292 tests, all green. All 6 new tests fail without their fixes.
- PHPCS `--report=source` diff — zero new violations, net −1.
- `php -l` clean; merges clean with `trunk`.

bartech:

tests/php/test-class-wc-connect-taxjar-integration.php
```diff
@@ -2053,4 +2100,171 @@ public function provide_non_object_tax_nodes() {
...
+			);
+		} finally {
+			$this->reset_local_pickup_with_store_street();
+		}
+
+		$this->assertCount(
+			4,
+			$address,
+			'WC_Tax::get_rates_from_location() requires exactly 4 elements; returning 5 makes core return an empty rate set.'
+		);
```
Abdalsalaam, line 2208: test_append_base_address_preserves_four_element_core_tuple() forces the local-pickup branch and then asserts only assertCount(4, $address). Nothing pins that the base substitution actually ran. If force_local_pickup_with_store_street() ever stops forcing the 'base' branch (session not started under a different bootstrap, another test leaving woocommerce_apply_base_tax_for_local_pickup filtered false, or a change in wc_get_chosen_shipping_method_ids() semantics), $street stays '' and the OLD unfixed code also returns 4 elements — so the test would pass against the base branch and stop being a regression test for the PR's headline defect.

```suggestion
		$this->assertCount(
			4,
			$address,
			'WC_Tax::get_rates_from_location() requires exactly 4 elements; returning 5 makes core return an empty rate set.'
		);
		$this->assertSame(
			array( 'US', 'FL', '33033', 'HOMESTEAD' ),
			$address,
			'The local-pickup base substitution did not run, so the arity assertion above is vacuous.'
		);
```

bartech: Confirmed and taken — `cfb92ab4`. The mechanism is exactly as you describe, and I verified it in both directions rather than taking the reasoning on trust.

### The defect

The pre-fix implementation was:

```php
if ( '' != $street ) {
	return array( $country, $state, $postcode, $city, $street );  // 5
}

return array( $country, $state, $postcode, $city );               // 4
```

The test feeds a four-element tuple, so `$street` starts empty and only becomes non-empty because the store street is substituted in. With the substitution running, the old code returns 5 and `assertCount( 4, ... )` fails — the test works. Without it, `$street` stays `''`, the old code returns 4, and the assertion passes against the very defect it exists to catch. Its power rests entirely on a precondition nothing pinned.

### One nuance worth recording: fragile, not vacuous *today*

Before changing anything I injected your `assertSame` against the current branch to find out which situation we were actually in. It **passes** — so the substitution does fire today, and the test is a genuine regression test right now, not a dead one. Your wording is correctly conditional ("if ... ever stops forcing"), and that is what the fix buys: it converts a *silent* degradation into a loud one. Worth being explicit that nothing is currently unprotected.

### Verified the strengthened test is still a regression test

The failure mode I wanted to avoid was making the assertion stricter without it still catching the original bug. So I reverted the production return to the pre-fix form and re-ran:

```
✘ Append base address preserves four element core tuple
  WC_Tax::get_rates_from_location() requires exactly 4 elements; returning 5 makes core return an empty rate set.
  Failed asserting that actual size 5 matches expected size 4.
```

Still fails against the old code. So it stayed a regression test rather than becoming merely a stricter one.

I kept your assertion and added a comment above it recording *why* it is there — the store postcode and city (`33033` / `HOMESTEAD`) are observable proof the branch was taken, and neither matches the customer values fed in (`33030` / `Miami`). Without that note the next reader sees a redundant-looking assertion sitting under an `assertCount` and is fairly likely to delete it.

Test-only change. Full suite green — 362 tests / 881 assertions at the top of the stack.

Abdalsalaam (approved):
LGTM!
I pushed changes to address some findings, mostly fixing docs and tests

PR #2986, opened 2026-07-29
[WOOTAX-313] Refactor - Migrate the TaxJar request seam onto the Address value object

bartech:
### Stacked PR

**Base is `chore/wootax-313-r1-2` (#2985), not `trunk`.** This is R1.3 of the address/rate-table refactor. #2984 (R1.1, the value object) and #2985 (R1.2, the read seam) are both still open — review them first; this diff only makes sense on top of them. If they merge, this will be rebased down the stack and the base updated.

R1.1 landed `Automattic\WCServices\Tax\Address`. R1.2 wired the **read** seam. R1.3 wires the **request** seam: `calculate_tax()`'s body assembly, the nexus address, `is_nexus_address_valid()`, and `get_address_parts()`.

The invariant this PR establishes: **the address that gets validated is the address that gets sent.**

---

## 1. The `from_zip` divergence is two bugs, and the documented one is the harmless half

The refactor plan describes this as "`from_zip` is never upper-cased where it's sent but is upper-cased where it's validated, so the value validated differs from the value sent." That is true, and on its own it is **unreachable**:

`get_address_parts()` upper-cased both postcodes before handing them to `validate_taxjar_request()`. But the only postcode check there is `WC_Validation::is_postcode( …, 'US' )`, and core's US rule is digits-only:

```php
// WC core, includes/class-wc-validation.php
case 'US':
    $valid = (bool) preg_match( '/^([0-9]{5})(-[0-9]{4})?$/i', $postcode );
```

Case cannot change that verdict, and non-US postcodes never reach the branch at all. Real divergence, no symptom.

**The live bug is sitting next to it.** `calculate_tax()` reduced the *destination* postcode to its first comma-separated segment and never did the same for the store's:

```php
$to_zip = explode( ',', $to_zip );
$to_zip = array_shift( $to_zip );   // to_zip only

$from_zip = $store_settings['postcode'];   // verbatim
```

So a store postcode entered as `90210, 90211` reached `is_postcode()` intact, failed it, and `smartcalcs_request()` returned `false` **before any request was made** — no tax, no TaxJar call, and nothing in the response log to explain it. The identical value supplied by a *customer* was handled.

Both halves are fixed at the **source** rather than the reader: both ends of the body are now built from one `Address`, which comma-splits and normalises by construction.

```php
$destination   = Address::from_options( $options, 'to_' );
$store_address = Address::from_store_settings( $this->get_store_settings() );

$body = array_merge(
    $store_address->to_taxjar_body( 'from_' ),
    $destination->to_taxjar_body( 'to_' ),
    array( 'shipping' => $shipping_amount, 'plugin' => 'woo' )
);
```

Key order and field names are unchanged, so the request shape on the wire is the same.

## 2. `get_address_parts()` is now a projection, not a second normaliser

It returns the postcodes **verbatim**. Country and state stay upper-cased there, and that asymmetry is deliberate — it is on the method docblock so it does not get "finished" later:

- country and state are compared against the literal `'US'` and against each other;
- `validate_taxjar_request()` is **public**, so it can be handed a body this class never built.

Re-deriving anything else there is precisely what recreates a validated-≠-sent gap for a hand-built body.

## 3. `is_nexus_address_valid()` delegates to `Address::validate()`

The inlined 40-line schema is replaced by the value object's, which carries the same rules (`/^[A-Z]{2}$/` country, `/^[A-Z0-9\-]{1,100}$/` state, the same max lengths, country + state required).

Because the value object normalises on construction, validating a normalised value and then sending the raw one would only relocate the divergence. So the accepted nexus address is written back normalised:

```php
$normalized    = Address::from_nexus( $nexus_address )->to_nexus_array();
$nexus_address = array_merge( $nexus_address, array_intersect_key( $normalized, $nexus_address ) );
```

`array_intersect_key` is load-bearing: it rewrites **only the keys the filter actually supplied**. Without it, `woocommerce_taxjar_nexus_address`'s documented array-in / array-or-`false`-out contract would quietly become a whitelist, and a filter that omitted `street` would start sending `"street": ""`. Pinned by a test.

---

## Behaviour deltas a reviewer should look at

1. **A present-but-blank required nexus field now falls back instead of aborting.** The old loop did `if ( ! $exists || '' === $value ) continue;`, so `array( 'country' => '' )` *passed* validation, replaced the `from_*` fields, and then tripped the "From country is missing. Aborting." gate — killing a calculation that would have succeeded had the nexus simply been dropped. It is now rejected and the store address is used.
2. **Lower-case country/state codes from the filter are accepted.** The schema pattern is upper-case-only, so `'us'` used to be silently discarded and logged. Normalising before validating accepts it and sends `'US'`.
3. **A numeric postcode from the filter is accepted** rather than rejected as a non-string (the value object casts). Non-scalar values are still rejected — an explicit guard, because the cast would otherwise turn an array into the literal `"Array"`.
4. **Empty `to_city` / `to_street` reach the body as `""` rather than `false`**, and the store postcode may now differ from the raw setting. Both change the JSON, so `smartcalcs_cache_request()`'s `md5( $json )` key changes: **one cache generation is invalidated**, nothing else.
5. **Validation error text names the field `postcode` where it used to say `zip`** — log output only, no contract.

## Backward-compatibility impact (per AGENTS.md)

- **No public signature changes.** `calculate_tax()`, `smartcalcs_request()`, `smartcalcs_cache_request()` and `validate_taxjar_request()` are untouched. The three methods actually modified — `is_nexus_address_valid()` and `get_address_parts()` — are `private`.
- **Hooks.** `woocommerce_taxjar_nexus_address` keeps its `(array $nexus_address, array $body)` signature and its array-in / array-or-`false`-out contract, and still fires at the same point. `taxjar_store_settings` is unchanged. Nothing is added, removed, reordered or re-timed. What a filter *returns* is now normalised for the keys it supplied — deltas 2 and 3 above — which is a change in leniency, not in shape.
- **Version assumptions.** None introduced. `Address` is internal, every boundary converts back to arrays, and no new WooCommerce or WordPress symbol is referenced. No `class_exists` / `function_exists` guard was added or removed.
- **Globals.** No new global or `WC()->…` read. The pre-existing `WC()->countries->get_vat_countries()` and `WC()->customer->is_vat_exempt()` calls in the early-return gate are unchanged and unmoved.
- **Multisite / install layout.** No option or table access, no path or URL construction. Neither is a factor.
- **Grandfathered shipping installs.** Not touched — this is the tax request path only.

---

## Follow-up this PR deliberately does not action

`$this->notifier` is `?StoreNoticesNotifier = null` on the constructor but is dereferenced unguarded in `smartcalcs_cache_request()`:

```php
$this->notifier->clear_notices( 'taxjar' );
```

So constructing the integration with four arguments fatals on the first tax request. It is pre-existing and unrelated to this seam, and it belongs with the defensive-guard work in [WOOTAX-318](https://linear.app/a8c/issue/WOOTAX-318) rather than being bundled here — bundling unrelated concerns is one of the four reasons the earlier PR #2906 was closed. Surfaced because the new tests must pass a **real** `StoreNoticesNotifier`: `clear_notices()` is `static`, which PHPUnit cannot mock.

---

## Verification

- `composer test` → **OK (291 tests, 634 assertions)**. 285 on the base branch, +6 new, zero regressions.
- **Fail-first confirmed** by stashing only the production file and re-running: 4 of the 6 fail against the base branch. The other two (`test_nexus_address_passes_through_unknown_keys`, `test_non_scalar_nexus_field_is_rejected`) pass before *and* after by design — they pin behaviour across the migration rather than demonstrate a fix, and their docblocks say so.
- PHPCS → **zero new violations, net −6** against the base branch (3819 → 3813). Every changed sniff row decreased; no sniff appears that was not already present. Measured by diffing `phpcs --standard=./phpcs.xml.dist --report=source` before and after — `check-all` is **not** green on this repo's trunk (~3841 pre-existing violations, 2223 of them in `i18n/strings.php`), so "zero new" is the achievable bar, not "green". One violation *was* introduced and caught this way: an added log line used `print_r()` and pushed `WordPress.PHP.DevelopmentFunctions.error_log_print_r` from 3 to 4; swapped for `wp_json_encode()`.
- `php -l` clean on both changed files, matching the CI lint step.

### Note on the test harness, for whoever writes the R1.4 tests

`woocommerce_taxjar_nexus_address` is the *wrong* seam for asserting on a request body — it fires **before** `$body['nexus_addresses']` is assigned, so anything captured there is missing the key you want. The right seam is the `WC_Connect_API_Client` mock the test class already injects: stub `proxy_request()` with a callback that records `$args['body']` and returns a `WP_Error`. That yields the exact JSON that would go on the wire, ends the call cleanly, and writes no transient, so tests do not poison each other's cache.

### Related issue(s)

Part of [WOOTAX-313](https://linear.app/a8c/issue/WOOTAX-313) (R1.3 of four). Does not close it — R1.4, the rate-table seam, is still outstanding.

Stacked on #2985, which is stacked on #2984.

**Reproducing the live bug on the base branch:**

1. Configure a US store with Automated Taxes enabled.
2. In WooCommerce → Settings → General, set the store postcode to a comma-separated value, e.g. `90210, 90211`.
3. Add a product to the cart with a US destination in the same state and calculate totals.
4. **Before:** no tax is applied. The WCS Tax log shows `API request is stopped. Country store is set to US but the zip code has incorrect format.` and no request is ever sent.
5. **After:** the request is sent with `from_zip` (or `nexus_addresses[0].zip`) reduced to `90210`, and tax is calculated.

**Checking the nexus changes:**

6. Add a `woocommerce_taxjar_nexus_address` filter returning a lower-case `country` (e.g. `'us'`). **Before:** the nexus address is silently dropped and the log records `[country] field format is invalid`. **After:** it is normalised to `'US'` and sent.
7. Add a filter returning `array( 'country' => '' )`. **Before:** the whole calculation aborts with `From country is missing. Aborting.` **After:** the nexus is rejected and the store address is used, so tax still calculates.
8. Add a filter that sets an extra key on the array. Confirm it still reaches TaxJar untouched.

### Checklist

- [x] unit tests

### DIFF
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1559,13 +1559,17 @@ class WC_Connect_TaxJar_Integration {
 			'tax_rate'        => 0,
 		);
 
-		// Strict conditions to be met before API call can be conducted.
+		// Strict conditions to be met before API call can be conducted. The postcode
+		// check runs on the normalised value, so a postcode that collapses to an
+		// empty string (for example a lone comma) aborts here; log the abort so the
+		// stop is diagnosable.
 		if (
 			empty( $destination->country() ) ||
 			( empty( $destination->postcode() ) && ! in_array( $destination->country(), WC()->countries->get_vat_countries() ) ) ||
 			( empty( $line_items ) && ( empty( $shipping_amount ) ) ) ||
 			WC()->customer->is_vat_exempt()
 		) {
+			$this->_log( 'Destination address or cart data is incomplete, or the customer is VAT exempt. Aborting.' );
 			return false;
 		}
 
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1720,28 +1720,13 @@ class WC_Connect_TaxJar_Integration {
 	 */
 	private function get_address_parts( $body ) {
 		return array(
-			'from_country' => strtoupper( $this->scalar_to_string( $body['nexus_addresses'][0]['country'] ?? $body['from_country'] ?? '' ) ),
-			'from_state'   => strtoupper( $this->scalar_to_string( $body['nexus_addresses'][0]['state'] ?? $body['from_state'] ?? '' ) ),
-			'from_zip'     => $this->scalar_to_string( $body['nexus_addresses'][0]['zip'] ?? $body['from_zip'] ?? '' ),
-			'to_country'   => strtoupper( $this->scalar_to_string( $body['to_country'] ?? '' ) ),
-			'to_state'     => strtoupper( $this->scalar_to_string( $body['to_state'] ?? '' ) ),
-			'to_zip'       => $this->scalar_to_string( $body['to_zip'] ?? '' ),
+			'from_country' => strtoupper( (string) ( $body['nexus_addresses'][0]['country'] ?? $body['from_country'] ?? '' ) ),
+			'from_state'   => strtoupper( (string) ( $body['nexus_addresses'][0]['state'] ?? $body['from_state'] ?? '' ) ),
+			'from_zip'     => (string) ( $body['nexus_addresses'][0]['zip'] ?? $body['from_zip'] ?? '' ),
+			'to_country'   => strtoupper( (string) ( $body['to_country'] ?? '' ) ),
+			'to_state'     => strtoupper( (string) ( $body['to_state'] ?? '' ) ),
+			'to_zip'       => (string) ( $body['to_zip'] ?? '' ),
 		);
-	}
-
-	/**
-	 * Cast a request body value to string, treating non-scalars as absent.
-	 *
-	 * A hand-built body handed to `validate_taxjar_request()` can carry non-scalar
-	 * values; casting those to string would yield 'Array' and slip past the country
-	 * checks.
-	 *
-	 * @param mixed $value Raw body value.
-	 *
-	 * @return string
-	 */
-	private function scalar_to_string( $value ): string {
-		return is_scalar( $value ) ? (string) $value : '';
 	}
 
 	/**
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1470,6 +1470,8 @@ class WC_Connect_TaxJar_Integration {
 
 		// The value object casts these fields to string, so reject anything that cannot
 		// survive that cast rather than triggering an array-to-string conversion.
+		// Booleans survive the cast but silently change the sent value (false becomes
+		// "" and true becomes "1"), so they are rejected too, as the old schema did.
 		//
 		// Only the fields the value object reads are checked. Any other key a filter
 		// added is passed through untouched below and is never cast, so it carries no
@@ -1482,8 +1484,11 @@ class WC_Connect_TaxJar_Integration {
 
 			$value = $address[ $field ];
 
-			if ( null !== $value && ! is_scalar( $value ) ) {
-				$this->logger->error( 'Nexus Address ERRORS: [' . $field . '] field must be a string' . PHP_EOL . 'Nexus address removed from request body.' . PHP_EOL . wp_json_encode( $address ), 'WCS Tax' );
+			if ( null !== $value && ( ! is_scalar( $value ) || is_bool( $value ) ) ) {
+				// wp_json_encode() returns false for unencodable payloads, which is
+				// what this branch catches; fall back so the log keeps the address.
+				$encoded = wp_json_encode( $address );
+				$this->logger->error( 'Nexus Address ERRORS: [' . $field . '] field must be a string' . PHP_EOL . 'Nexus address removed from request body.' . PHP_EOL . ( false === $encoded ? print_r( $address, true ) : $encoded ), 'WCS Tax' );
 
 				return false;
 			}
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1714,13 +1714,20 @@ class WC_Connect_TaxJar_Integration {
 	 * @return array
 	 */
 	private function get_address_parts( $body ) {
+		// A hand-built body handed to validate_taxjar_request() can carry non-scalar
+		// values; casting those to string would yield 'Array' and slip past the
+		// country checks, so treat them as absent instead.
+		$scalar = static function ( $value ) {
+			return is_scalar( $value ) ? (string) $value : '';
+		};
+
 		return array(
-			'from_country' => strtoupper( (string) ( $body['nexus_addresses'][0]['country'] ?? $body['from_country'] ?? '' ) ),
-			'from_state'   => strtoupper( (string) ( $body['nexus_addresses'][0]['state'] ?? $body['from_state'] ?? '' ) ),
-			'from_zip'     => (string) ( $body['nexus_addresses'][0]['zip'] ?? $body['from_zip'] ?? '' ),
-			'to_country'   => strtoupper( (string) ( $body['to_country'] ?? '' ) ),
-			'to_state'     => strtoupper( (string) ( $body['to_state'] ?? '' ) ),
-			'to_zip'       => (string) ( $body['to_zip'] ?? '' ),
+			'from_country' => strtoupper( $scalar( $body['nexus_addresses'][0]['country'] ?? $body['from_country'] ?? '' ) ),
+			'from_state'   => strtoupper( $scalar( $body['nexus_addresses'][0]['state'] ?? $body['from_state'] ?? '' ) ),
+			'from_zip'     => $scalar( $body['nexus_addresses'][0]['zip'] ?? $body['from_zip'] ?? '' ),
+			'to_country'   => strtoupper( $scalar( $body['to_country'] ?? '' ) ),
+			'to_state'     => strtoupper( $scalar( $body['to_state'] ?? '' ) ),
+			'to_zip'       => $scalar( $body['to_zip'] ?? '' ),
 		);
 	}
 
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -1520,6 +1520,10 @@ class WC_Connect_TaxJar_Integration {
 		$errors = $nexus->validate( $required )->get_error_messages();
 
 		if ( ! empty( $errors ) ) {
+			// The schema names the field postcode; the nexus array key is zip. Report
+			// the key the filter author actually used.
+			$errors = str_replace( '[postcode]', '[zip]', $errors );
+
 			$this->logger->error( 'Nexus Address ERRORS: ' . implode( ', ', $errors ) . PHP_EOL . 'Nexus address removed from request body.' . PHP_EOL . print_r( $address, true ), 'WCS Tax' );
 
 			return false;
@@ -1542,11 +1546,8 @@ class WC_Connect_TaxJar_Integration {
 		// Normalize options to an array and safely map to local variables.
 		$options = is_array( $options ) ? $options : array();
 
-		// Both ends of the request go through the same normalisation. That symmetry is
-		// the point: the destination postcode used to be reduced to its first
-		// comma-separated segment while the store's was sent verbatim, so a store
-		// postcode entered as a list reached `validate_taxjar_request()` intact and
-		// aborted the calculation.
+		// Both ends of the request go through the same normalisation, so the address
+		// that is validated is the address that is sent.
 		$destination     = Address::from_options( $options, 'to_' );
 		$shipping_amount = $options['shipping_amount'] ?? 0;
 		$line_items      = $options['line_items'] ?? null;
@@ -1565,7 +1566,7 @@ class WC_Connect_TaxJar_Integration {
 		// stop is diagnosable.
 		if (
 			empty( $destination->country() ) ||
-			( empty( $destination->postcode() ) && ! in_array( $destination->country(), WC()->countries->get_vat_countries() ) ) ||
+			( empty( $destination->postcode() ) && ! in_array( $destination->country(), WC()->countries->get_vat_countries(), true ) ) ||
 			( empty( $line_items ) && ( empty( $shipping_amount ) ) ) ||
 			WC()->customer->is_vat_exempt()
 		) {
@@ -1708,10 +1709,10 @@ class WC_Connect_TaxJar_Integration {
 	 * This must report what the body actually carries. Country and state are
 	 * upper-cased because they are compared against the literal `'US'` and against
 	 * each other, and this runs from the public `validate_taxjar_request()`, which
-	 * any caller can hand a hand-built body. The **postcodes are returned verbatim**:
-	 * re-deriving them here is what previously made the value that was validated
-	 * differ from the value that was sent. Postcode normalisation belongs where the
-	 * body is assembled, in `calculate_tax()`.
+	 * any caller can hand a hand-built body. Postcodes are returned verbatim:
+	 * normalising them here would make the value that is validated differ from the
+	 * value that is sent. Postcode normalisation belongs where the body is
+	 * assembled, in `calculate_tax()`.
 	 *
 	 * @param array $body Request body.
 	 *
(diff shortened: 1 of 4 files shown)


classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1486,39 +1448,18 @@ private function is_nexus_address_valid( $address ): bool {
...
-
-			if ( isset( $rules['pattern'] ) && is_string( $value ) ) {
-				if ( ! preg_match( $rules['pattern'], $value ) ) {
-					$errors[] = "[$field] field format is invalid";
-				}
+				return false;
 			}
 		}
 
+		$errors = Address::from_nexus( $address )->validate()->get_error_messages();
```
Abdalsalaam: Address::validate() defaults to required = country+state and errors on present-but-blank (src/Tax/Address.php:486,501). The old inline loop only errored when the key was ABSENT and skipped present-but-blank values. The plugin builds the nexus address from its own store settings at line 1527 via to_nexus_array(), which always emits a 'state' key, and WC()->countries->get_base_state() returns '' for GB, DE, FR, NL, BE, DK, SE and most of the supported EU list in get_supported_countries() (line 429). Result: on every tax calculation the nexus block is skipped, the body reverts from nexus_addresses to from_*, and $this->logger->error('Nexus Address ERRORS: [state] field is required') fires. WC_Connect_Logger::error() (classes/class-wc-connect-logger.php:98) writes the error_notice option and calls log() with force=true, bypassing the merchant's logging-enabled setting. The PR frames this delta as applying only to a filter-supplied incomplete nexus; it also hits the plugin's own auto-generated nexus, is undocumented, and is untested (no non-US store test).

```suggestion
		$nexus_object = Address::from_nexus( $address );
		$has_states   = ! empty( WC()->countries->get_states( $nexus_object->country() ) );
		$errors       = $nexus_object->validate( $has_states ? array( 'country', 'state' ) : array( 'country' ) )->get_error_messages();
```

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1486,39 +1448,18 @@ private function is_nexus_address_valid( $address ): bool {
...
-				}
-			}
-
-			if ( isset( $rules['pattern'] ) && is_string( $value ) ) {
-				if ( ! preg_match( $rules['pattern'], $value ) ) {
-					$errors[] = "[$field] field format is invalid";
-				}
+				return false;
 			}
 		}
```
Abdalsalaam: The new guard uses `foreach ( $address as $field => $value )`, iterating ALL keys of the filter-supplied nexus address. The old code used `foreach ( $schema as $field => $rules )` and never inspected custom keys. Unknown keys are also never string-cast by the value object (line 1583 passes them through untouched via array_intersect_key), so there is no array-to-string-conversion risk for them. A filter attaching a non-scalar under a custom key therefore causes the ENTIRE nexus address to be rejected, silently falling back to the store from_* address and changing which nexus TaxJar uses. This contradicts the code comment at lines 1579-1582 in the same diff ('any other key a filter added is passed through untouched, so the documented array-in / array-out contract does not become a whitelist') — the whole-address rejection happens before that pass-through is reached. The pass-through test at tests/php/test-class-wc-connect-taxjar-integration.php:2467 uses a scalar custom key and misses this. The log message is also misleading: it reports '[meta] field must be a string' for a field the code never treats as a string.

```suggestion
		foreach ( array( 'id', 'country', 'state', 'zip', 'city', 'street' ) as $field ) {
			if ( ! array_key_exists( $field, $address ) ) {
				continue;
			}

			$value = $address[ $field ];

			if ( null !== $value && ! is_scalar( $value ) ) {
				$this->logger->error( 'Nexus Address ERRORS: [' . $field . '] field must be a string' . PHP_EOL . 'Nexus address removed from request body.' . PHP_EOL . wp_json_encode( $address ), 'WCS Tax' );

				return false;
			}
		}
```

classes/class-wc-connect-taxjar-integration.php
iyut: Credit to @Abdalsalaam for catching this one first — I'm just writing up the detail.

Could you sanity-check my reasoning here? `validate()` defaults to requiring both `country` and `state`, and it counts a present-but-empty value as an error. For a store in a stateless country (GB and a lot of the EU), `WC()->countries->get_base_state()` comes back as `''`, so the store's own nexus reaches this line with an empty `state` and gets rejected — which force-logs through `logger->error()` on line 1464, on every calc. The previous loop only errored when the key was *absent*, so I think this is an unintended behavior change rather than a deliberate tightening.

Since `from_country` seems to be the only field with a real downstream gate (the abort further down around line 1592), what do you think about requiring just `country` here? Something like:

```suggestion
		$errors = Address::from_nexus( $address )->validate( array( 'country' ) )->get_error_messages();
```

If there's a reason `state` needs to stay required that I'm missing, I'm very happy to be corrected. Either way it might be worth a small test with a `state => ''` store (the current `default_store_settings()` is US/CA only) just so we pin the behavior down — what do you reckon?

iyut (dismissed):
Thanks for this one, @bartech — pulling all the address shaping into a single value object reads really cleanly, and it's nice that it quietly fixes a few longstanding rough edges on the way through (the `wp_unslash()` in `from_post_request()` especially).

I went through it with a correctness-first lens and also ran a second pass from a fresh perspective, and both landed on the same single thing I'd love your take on before merge — which @Abdalsalaam was the first to spot. I left the detail inline on line 1461, but the short version:

- **Stateless-country stores (GB and much of the EU) look like they'd force-log a nexus error on every tax calc.** `validate()` now treats a present-but-empty required `state` as invalid, and those stores' base state is `''`, so the store's own nexus gets rejected each call and force-logs via `logger->error()`. The tax result is fine (it falls back to `from_*`), so it's log/DB noise rather than wrong tax — but it'd be every request for that group of stores. There's a one-line idea + a test suggestion inline; genuinely curious whether I've read it right.

A few smaller, non-blocking notes:
- The `wp_unslash()` addition changes what's sent for admin-recalculated orders containing apostrophes (for the better) — might be worth a line in the PR description's BC note.
- `get_backend_address()` now returns `false` where it used to return `''` for present-but-empty POST fields, but `from_options()` casts it back to `''` downstream, so no real impact — just flagging it.
- The non-scalar nexus guard iterates all keys now, so it's a touch stricter/looser than before at the very edges (filter-supplied values only) — reads fine to me.

Really nice work overall, and happy to talk through any of it.

bartech:
## Pushed — both nexus findings fixed, and the state rule is now measured against the live API

Thanks @Abdalsalaam and @iyut — you independently found the same thing and you were both right. It would have shipped. Rebased onto the updated stack and fixed.

### 1. The origin state is now required for the US, and only the US

@Abdalsalaam's diagnosis was accurate and, if anything, understated. I checked the premise rather than taking it on trust — loaded WooCommerce's `i18n/states.php` and intersected it with `get_supported_countries()`: **19 of the 31 supported countries have no states at all** (AT, BE, CY, CZ, DK, EE, FI, FR, GB, LT, LU, LV, MT, NL, PL, PT, SE, SI, SK). One correction to the original list — DE is *not* among them, it has 16 states — but that doesn't soften the conclusion.

And @iyut is right that this was never filter-only: `calculate_tax()` builds the nexus from the store's own settings via `to_nexus_array()`, which always emits a `state` key, and runs it through this gate. So every GB/FR/NL/DK store would have had its own nexus rejected on **every** calculation — and because `WC_Connect_Logger::error()` calls `enable_notice()` and logs with `force = true`, that's a persistent admin notice, not just log noise.

**@iyut suggested `validate( array( 'country' ) )`, and that was my first instinct too — but it turns out to be unsafe for exactly one country.** Before committing to it I probed the live API through the normal proxy path (scripts and full output are in the ticket's `work/` folder):

| Country | nexus `state: ""` | Result |
|---|---|---|
| GB, FR, NL, DK | HTTP 200 | identical rate (20 / 20 / 21 / 25%) |
| DE, ES, IT, IE | HTTP 200 | identical rate (19 / 21 / 22 / 23%) |
| CA | HTTP 200 | identical rate (13%) |
| AU | HTTP 200 | identical rate (10%) |
| **US** | HTTP 200 | **`has_nexus: false`, rate 0, tax 0** |

A US nexus with a blank state isn't rejected by TaxJar — it comes back **200 with `has_nexus: false` and zero tax**. So requiring only `country` would convert a loud local rejection into a silent under-collection on US stores.

I think that's worth internalising beyond this line: **TaxJar's failure mode for an under-specified address is a successful-looking response with no tax, never a 4xx.** A gate like this can only be judged on `has_nexus` / `rate`, not on the HTTP status.

The rule is therefore `'US' === $nexus->country() ? array( 'country', 'state' ) : array( 'country' )` — @iyut's relaxation everywhere it's safe, the strict check kept where it's load-bearing.

This is also narrower than @Abdalsalaam's suggested `WC()->countries->get_states()` gate, for the same measured reason: DE, ES, IT, IE, CA and AU all *have* WooCommerce states and none of them need one for TaxJar, so that gate would have kept wrongly rejecting them — plus GR, HU, HR, RO and BG. That matters because TaxJar documents US/CA only but serves those markets for legacy accounts, which is much of who is still on this integration. Side benefit: no `WC()->countries` read at all, so the AGENTS.md "do not assume global state" question doesn't arise.

@iyut — you were also right that there was no test. There are ten now: nine non-US store countries, deliberately mixing WC-stateless (GB, FR, NL, DK) with WC-stateful (DE, ES, IT, CA, AU) so the mix itself asserts that the distinction is irrelevant, plus a US blank-state case pinning the strict check. `default_store_settings()` stays US/CA; the provider supplies the store per case.

One thing your suggestion shook out that I'd have missed: `validate_taxjar_request()` requires a **destination** state when `to_country` is US **or CA** — a rule about where the customer is, nothing to do with the nexus. My first CA data-set set `to_state = ''` and the request was never attempted at all. The provider now carries `'ON'` for that row.

### 2. The non-scalar guard no longer inspects unknown keys

@Abdalsalaam's sharpest point: the guard contradicted the code comment four lines below it. The whole-address rejection happened *before* the pass-through was reached, so the "does not become a whitelist" claim wasn't true as written.

The risk it was guarding against was also imaginary for those keys — `from_nexus()` reads only `country`, `state`, `zip`, `city`, `street`, `id`, and `array_intersect_key` merges everything else back untouched, so an unknown key is never cast and can't trigger an array-to-string conversion. Meanwhile a filter attaching a non-scalar under a custom key discarded the entire nexus, silently changing which origin TaxJar rated against. And the log line claiming "[meta] field must be a string" for a field we never treat as a string was actively misleading.

Fixed exactly as suggested — the loop walks the six known fields. The existing pass-through test used a *scalar* custom key and couldn't have caught this, so `test_non_scalar_unknown_nexus_key_is_passed_through` puts an array under a custom key and asserts both that the nexus survives and that the key arrives intact.

### On the non-blocking notes

- **`wp_unslash()` in the BC note** — added on #2985, where the change lives. There's a wrinkle: this stack and its base fixed that defect independently, on different lines of the same method, so the rebase merged both with no conflict and produced a *double* `wp_unslash()`. Not idempotent, so `A\B` was being reduced to `AB` while the apostrophe test passed either way. Removed, with tests pinning single-application.
- **`get_backend_address()` returning `false` where it returned `''`** — matches your reading, and it's listed as a behaviour delta. `to_legacy_options()` only maps `''`, so `'0'` is no longer swallowed as empty either, which is strictly a fix.
- **The non-scalar guard reading fine** — @Abdalsalaam was less comfortable with it and I think he was right; see above.

### Backward compatibility

`woocommerce_taxjar_nexus_address` acceptance is **widened, never narrowed**. Relative to trunk, three classes of address that were previously *rejected* are now accepted: a blank state outside the US, a lower-case country or state code, and a numeric postcode. One class previously accepted is now rejected — a present-but-blank `country`, which used to pass validation and then abort the whole calculation at the "From country is missing" gate, so the store address is used instead. Unknown keys added by a callback pass through untouched, including non-scalar ones. No hook name, argument count or firing order changed. Site-scoped, so no multisite implication.

### Verification

- `composer test` — 309 tests, all green. 10 of the 11 new tests fail without their fixes; the US blank-state case bounds the relaxation and passes either way (its docblock says so).
- PHPCS `--report=source` diff — identical, 184 → 184.
- `php -l` clean; merges clean with `trunk`.

### One gap, recorded

Every probed row is a **domestic** sale except the four US cross-state ones (which are 0 in both columns — the control confirming the in-state zero comes from the missing origin state, not the destination). EU cross-border / OSS was **not** probed. It doesn't bear on this change, which only decides whether the nexus block is sent at all, but it's noted in the probe log for whoever works on *which* nexus to send.

bartech:

Abdalsalaam (approved):
LGTM! 
I pushed changes to address some findings, mostly fixing docs

PR #2988, opened 2026-07-29
[WOOTAX-313] Refactor - Migrate the TaxJar rate-table seam onto the Address value object

bartech:
### Changes proposed in this Pull Request

Fourth and final PR of the R1 series (WOOTAX-313). Migrates the **rate-table seam** — `create_or_update_tax_rate()` and `allow_street_address_for_matched_rates()` — onto `Automattic\WCServices\Tax\Address`, so the arguments used to *find* a `wp_woocommerce_tax_rates` row and the values written *into* it come from one address through a matched pair of projections the value object owns: `to_find_rates_args()` and `to_rate_table_locations()`.

**Stacked on [#2986](https://github.com/Automattic/woocommerce-services/pull/2986)** (`chore/wootax-313-r1-3`), which is stacked on [#2985](https://github.com/Automattic/woocommerce-services/pull/2985) → [#2984](https://github.com/Automattic/woocommerce-services/pull/2984). Review the base PRs first; this diff is only R1.4.

This is the one PR in the series that can silently duplicate rate rows, so **the characterization tests are a separate first commit** (`ad4ca528`) and the refactor lands under them (`4a0c966e`). The behaviour changes are visible as deliberate edits to that first commit's expectations.

### Measuring the seam corrected two of the three defects the plan attributed to it

The refactor plan lists three write/read asymmetries here. Two are real but **not reachable**, and this PR says so rather than shipping a fix for a bug that does not exist:

| Plan item | Measured |
|---|---|
| **A1** — postcode written through `wc_normalize_postcode()`, read raw | **Not live.** `WC_Tax::find_rates()` applies the identical `wc_normalize_postcode( wc_clean( … ) )` to its own argument before querying. ZIP+4 always round-tripped. |
| **A2** — city written un-uppercased, read uppercased | **Casing half not live.** Core upper-cases the city on both ends (`format_tax_rate_city()` on the write, `strtoupper()` inside the lookup SQL). A *different* half of A2 is live — see below. |
| **A3** — state stripped for `find_rates`, stored un-stripped | **Backwards.** `WC_Tax::prepare_tax_rate()` singles `tax_rate_state` out for `sanitize_key()` before storing, so the column holds a *more* normalized value than either reader asked for. |

### Defects fixed

1. **A state carrying anything outside `[a-z0-9_-]` duplicated a rate row on every calculation.** The lookup stripped spaces; core's storage runs `sanitize_key()`. Those agree for `'N Y'` and disagree for `'N.Y.'` and for anything non-ASCII, so the row inserted could never be found again. `Address::state_compact()` now mirrors the core function rather than approximating it.
2. **A city carrying anything `sanitize_text_field()` removes did the same** — the write applied `wc_clean()`, the lookup did not. Sanitisation now happens once, inside the shared projection.
3. **`allow_street_address_for_matched_rates()` could not see rows `create_or_update_tax_rate()` had just written.** It normalised the city but not the state. The rows existed; the `woocommerce_matched_rates` path — price display, shipping tax, coupon math — returned **no rates** for those addresses, while itemized cart totals stayed correct. That asymmetry is the reason the test suite checks the two methods against each other rather than each in isolation.

### Two observations for R2, recorded but not acted on

- **`allow_street_address_for_matched_rates()` passes no street to `WC_Tax::find_rates()`** — that function takes no street argument, and the value the method unpacked from the tuple was never read. What it actually does is accept a location of *four or more* elements where `WC_Tax::get_rates_from_location()` accepts exactly four, supplying the rates core skips when this plugin's five-element taxable address is in play. It is an arity workaround, not a street feature. This independently corroborates the July measurement that street never moves TaxJar's jurisdiction. It is **public** and stays; the docblock now says what it does.
- **Rate rows are matched by ordinal position, not by priority value** (`array_keys( $wc_rates )[ $priority - 1 ]`, against a `find_rates()` that returns one rate per priority). The lookup for priority N only lands on the intended row when priorities 1..N-1 already exist. True of our own caller, so it works — but it is the constraint any re-keying has to keep or replace. Pinned by a test.

### Behaviour changes

- **A comma-list postcode is stored as its first segment** rather than whole. The row is now keyed by the segment the rate was quoted for; the request body has carried only that segment since #2986.
- **`get_backend_address()` no longer upper-cases postcode, city and street.** This was the admin path's own historic behaviour, unmatched by `get_address()`, kept only because the city reaches the rate-table city column — which is now upper-cased at both ends. The admin recalculate and the cart put identically-cased values on the wire. **This changes the request JSON, so one generation of the `md5( $json )` request cache is invalidated** (nothing else; the same input recalculates to the same result).
- `tax_rate_country` is trimmed before storage, so a whitespace-padded country can no longer produce a row nothing matches.
- `create_or_update_tax_rate()` tolerates a missing or non-scalar location field instead of emitting a notice or storing the literal `"Array"`.
- **`WC_Connect_TaxJar_Integration::normalize_city()` now raises its deprecation notice.** It could not before — the last three in-repo callers were all on this seam and the notice would have fired on every checkout.

### Backward compatibility

**Public and externally exposed surface touched. No signature changes.**

- `create_or_update_tax_rate()`, `allow_street_address_for_matched_rates()` and `get_taxable_address()` keep their signatures and return shapes. `get_backend_address()` is `protected` and keeps its array shape; only the casing of three values changes.
- **`normalize_city()` is deprecated, not removed.** It is `protected static`, so an out-of-repo subclass may call it, and removing it would break that subclass on load. It keeps delegating to the value object and keeps returning non-string input untouched. Per AGENTS.md: deprecate, don't rename or delete.
- **No hook is added, removed, reordered or retimed.** `woocommerce_taxjar_enable_florida_shipping_tax` still fires with the same argument at the same point.
- **The Florida shipping-tax predicate still reads `$location` raw**, deliberately. Routing it through the value object would make it fire for a lower-cased `'us'`/`'fl'` where it does not today — a change to what tax is charged, not to how a row is keyed. Out of this PR's scope; flagged for R2.
- **Stored data changes shape for exactly the addresses that were already broken.** A space- or punctuation-bearing state stored `'NY'` before and after — only the lookup changed. A comma-list postcode is the one case where a *new* row shape is written, and the old row was unreachable from the TaxJar request anyway.
- **No new global state is read.** No `WC()->…` dereference, no `$post`, no `$wp_query`. `WC_Tax` and `wc_clean()`/`wc_normalize_postcode()`/`sanitize_key()` are all already required by these methods.
- **Multisite:** unchanged. `wp_woocommerce_tax_rates` is a per-site table and no option access was added or moved.
- **Install layout:** no paths or URLs are constructed.

### Testing instructions

`composer test` → **331 tests / 793 assertions green** (291 on the base branch, +40, zero regressions).

**Fail-first confirmed** by restoring only the two production files and re-running: 44 tests fail — the 11 value-object round-trip cases, the `sanitize_key()` state case, the 25 deprecation-notice expectations, both `get_backend_address()` casing tests, and the 6 rate-table cases whose behaviour this PR changes.

PHPCS: **zero new violations, net −3** (branch baseline 3813 → 3810), measured by diffing `--report=source` before and after. `php -l` clean on both changed production files.

Manual check worth doing on a store: enable automated taxes, check out twice to the same US address, and confirm `wp_woocommerce_tax_rates` gains one row per priority on the first checkout and none on the second.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -9,7 +9,10 @@
 * Fix   - Prevent tax calculation from failing when the store address postcode holds more than one comma-separated value.
 * Fix   - Accept lower-case country and state codes from the woocommerce_taxjar_nexus_address filter.
 * Fix   - Fall back to the store address when a custom nexus address is incomplete instead of abandoning the calculation.
+* Fix   - Prevent unbounded growth of WooCommerce tax rate rows when the customer state contains a period, a space, or a non-Latin character.
+* Fix   - Restore tax rates on the price display, shipping tax and coupon paths for addresses whose state was stored in a normalized form.
 * Tweak - Centralize TaxJar address handling in an internal value object. No change to tax calculation.
+* Tweak - Store tax rate rows against the postcode the rate was quoted for when the address postcode holds more than one comma-separated value.
 
 = 3.6.11 - 2026-08-05 =
 * Fix   - Prevent a fatal error during cart and checkout tax calculation when TaxJar returns an incomplete tax response.
--- a/changelog.txt
+++ b/changelog.txt
@@ -9,7 +9,7 @@
 * Fix   - Prevent tax calculation from failing when the store address postcode holds more than one comma-separated value.
 * Fix   - Accept lower-case country and state codes from the woocommerce_taxjar_nexus_address filter.
 * Fix   - Fall back to the store address when a custom nexus address is incomplete instead of abandoning the calculation.
-* Fix   - Prevent unbounded growth of WooCommerce tax rate rows when the customer state contains a period, a space, or a non-Latin character.
+* Fix   - Prevent unbounded growth of WooCommerce tax rate rows when the customer state contains a character WooCommerce strips before storing it, such as a period or an accented letter.
 * Fix   - Restore tax rates on the price display, shipping tax and coupon paths for addresses whose state was stored in a normalized form.
 * Tweak - Centralize TaxJar address handling in an internal value object. No change to tax calculation.
 * Tweak - Store tax rate rows against the postcode the rate was quoted for when the address postcode holds more than one comma-separated value.
--- a/classes/class-wc-connect-taxjar-integration.php
+++ b/classes/class-wc-connect-taxjar-integration.php
@@ -709,27 +709,32 @@ class WC_Connect_TaxJar_Integration {
 	/**
 	 * Allow street address to be passed when finding rates
 	 *
-	 * @param array  $matched_tax_rates
-	 * @param string $tax_class
+	 * Despite the name, no street ever reaches `WC_Tax::find_rates()` — it takes no
+	 * street argument, and the value this method used to unpack from the tuple was
+	 * never read. What the callback actually does is accept a location of *four or
+	 * more* elements where `WC_Tax::get_rates_from_location()` accepts exactly four,
+	 * so it supplies the rates core skips when this plugin's five-element taxable
+	 * address is in play. It is public and hooked on `woocommerce_matched_rates`,
+	 * which core fires from the price-display, shipping-tax and coupon paths.
+	 *
+	 * The lookup arguments now come from the same value object `create_or_update_tax_rate()`
+	 * writes with. Before, this method normalised the city but not the state, so it
+	 * could not see rows that method had just written: the rate rows existed and
+	 * these paths returned nothing for them.
+	 *
+	 * @param array  $matched_tax_rates Rates core matched; replaced wholesale.
+	 * @param string $tax_class         Tax class slug.
 	 * @return array
 	 */
 	public function allow_street_address_for_matched_rates( $matched_tax_rates, $tax_class = '' ) {
-		$tax_class         = sanitize_title( $tax_class );
-		$location          = WC_Tax::get_tax_location( $tax_class );
-		$matched_tax_rates = array();
-		if ( sizeof( $location ) >= 4 ) {
-			list( $country, $state, $postcode, $city, $street ) = array_pad( $location, 5, '' );
-			$matched_tax_rates                                  = WC_Tax::find_rates(
-				array(
-					'country'   => $country,
-					'state'     => $state,
-					'postcode'  => $postcode,
-					'city'      => strtoupper( self::normalize_city( $city ) ),
-					'tax_class' => $tax_class,
-				)
-			);
+		$tax_class = sanitize_title( $tax_class );
+		$location  = WC_Tax::get_tax_location( $tax_class );
+
+		if ( ! is_array( $location ) || count( $location ) < 4 ) {
+			return array();
 		}
-		return $matched_tax_rates;
+
+		return WC_Tax::find_rates( Address::from_taxable_tuple( $location )->to_find_rates_args( $tax_class ) );
 	}
 
 	public function cleanup_tax_label( $rate_name ) {
@@ -935,27 +940,17 @@ class WC_Connect_TaxJar_Integration {
 	 * Security: WooCommerce has already verified the nonce and capability by the time
 	 * `woocommerce_before_save_order_items` fires.
 	 *
+	 * This method used to `strtoupper()` postcode, city and street on top of that,
+	 * which `get_address()` did not — the same order recalculated from the admin and
+	 * from the cart put differently-cased values on the wire. The casing was kept only
+	 * because the city feeds the `wp_woocommerce_tax_rates` city column; now that the
+	 * rate-table seam upper-cases the city at both the write and the lookup, nothing
+	 * downstream depends on it and the two paths agree.
+	 *
 	 * @return array
 	 */
 	protected function get_backend_address() {
-		$address = Address::from_post_request()->to_legacy_options();
-
-		/*
-		 * Historic behaviour of this method, preserved deliberately: the admin path
-		 * upper-cases every field, not only country and state (which `Address` already
-		 * normalises). It is load-bearing for the city, which reaches the
-		 * `wp_woocommerce_tax_rates` city column, and pinned by
-		 * `test_get_backend_address_normalizes_semicolon_city()`. Reconciling the casing
-		 * asymmetry between this path and `get_address()` belongs with the rate-table
-		 * seam, where both ends of the write/read round trip can move together.
-		 */
-		foreach ( array( 'to_zip', 'to_city', 'to_street' ) as $key ) {
-			if ( is_string( $address[ $key ] ) ) {
-				$address[ $key ] = strtoupper( $address[ $key ] );
-			}
-		}
-
-		return $address;
+		return Address::from_post_request()->to_legacy_options();
 	}
 
 	/**
@@ -1859,9 +1854,11 @@ class WC_Connect_TaxJar_Integration {
 	 * normalisation policy has one home rather than one copy per consumer. This
 	 * method is retained as a delegate — it is `protected static`, so a subclass
 	 * outside this repository may be calling it, and removing it would break that
-	 * subclass on load. The remaining in-repo call sites move to the value object
-	 * when their seams are migrated; no `_deprecated_function()` notice is raised
-	 * because those call sites are still live and would fire it on every checkout.
+	 * subclass on load.
+	 *
+	 * The `_deprecated_function()` notice can be raised as of this release because
+	 * the last in-repo caller has moved to the value object. Until then the notice
+	 * would have fired on every checkout.
 	 *
 	 * The non-string guard is preserved: this method has always returned its
 	 * argument untouched when handed a non-string, and callers may rely on that.
@@ -1870,6 +1867,8 @@ class WC_Connect_TaxJar_Integration {
 	 * @return string Normalized city, safe for `_update_tax_rate_cities` and `find_rates`.
 	 */
 	protected static function normalize_city( $city ) {
+		wc_deprecated_function( __METHOD__, '3.6.11', '\Automattic\WCServices\Tax\Address::normalize_city()' );
+
 		if ( ! is_string( $city ) ) {
 			return $city;
 		}
@@ -1890,10 +1889,33 @@ class WC_Connect_TaxJar_Integration {
 	 * @return int
 	 */
 	public function create_or_update_tax_rate( $location, $rate, $tax_class = '', $freight_taxable = 1, $rate_priority = 1, $tax_rate_name = 'Tax' ) {
-		// Prevent filling "State code" column for countries with VAT tax.
-		// VAT tax is country wide.
-		$to_state      = 'VAT' === $tax_rate_name ? '' : strtoupper( $location['to_state'] );
 		$rate_priority = absint( $rate_priority );
+
+		/*
+		 * One address, two consumers: the `find_rates()` lookup below and the location
+		 * rows written beside the rate. Deriving them separately is what let this
+		 * method insert a fresh row on every calculation for some addresses — it
+		 * looked up values it had never stored. The value object now owns both
+		 * projections, so they cannot drift apart.
+		 *
+		 * Non-scalars collapse to an empty string rather than reaching `(string)` and
+		 * becoming the literal "Array". This method is public and takes the location
+		 * it is handed.
+		 */
+		$field = static function ( $value ) {
+			return is_scalar( $value ) ? (string) $value : '';
+		};
+
+		$address = Address::from_options(
+			array(
+				'to_country' => $field( $location['to_country'] ?? '' ),
+				// Prevent filling "State code" column for countries with VAT tax.
+				// VAT tax is country wide.
+				'to_state'   => 'VAT' === $tax_rate_name ? '' : $field( $location['to_state'] ?? '' ),
+				'to_zip'     => $field( $location['to_zip'] ?? '' ),
+				'to_city'    => $field( $location['to_city'] ?? '' ),
+			)
+		);
 
 		/**
 		 * @see https://github.com/Automattic/woocommerce-services/issues/2531
@@ -1918,8 +1940,8 @@ class WC_Connect_TaxJar_Integration {
 		}
 
 		$tax_rate = array(
-			'tax_rate_country'  => $location['to_country'],
-			'tax_rate_state'    => $to_state,
+			'tax_rate_country'  => $address->country(),
+			'tax_rate_state'    => $address->state_compact(),
 			// For the US, we're going to modify the name of the tax rate to simplify the reporting and distinguish between the tax rates at the counties level.
 			// I would love to do this for other locations, but it looks like that would create issues.
 			// For example, for the UK it would continuously rename the rate name with an updated `state` "piece", each time a request is made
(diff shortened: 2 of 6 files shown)


iyut (approved):
Approving. Nice piece of work Bart — the "one address, two projections" framing makes the write/read symmetry easy to follow, and shipping the characterization tests as a separate first commit made the behavior changes genuinely reviewable rather than something to reverse-engineer from the diff.

I checked `state_compact()`, the postcode/city round-trip, and the `allow_street_address_for_matched_rates()` early-return against WC core and they line up — the `sanitize_key()` mirroring holds on both ends, and the idempotency argument for the postcode/city normalization checks out. No regression for normal single-postcode addresses; the projections come out byte-identical to the old code there.

One small optional thing, not blocking: the changelog calls out a non-Latin state as a motivating case, but I don't see a provider row exercising a state that `sanitize_key()` collapses to `''`. The behavior's correct, it just currently rests on reasoning rather than a test — might be worth one row to pin it. Happy to leave it for R2 if you'd rather.

The Florida-gate note in the description reads right to me too — agree that's an R2 concern.

CezaryDrewniak (approved):
Approving. The "one address, two projections" framing is the right abstraction and it makes the seam's symmetry obvious in a way the previous hand-rolled normalizations never were. Shipping the characterization tests as a separate first commit was the right call. The behavior changes read as deliberate edits to a known baseline rather than something you have to reverse-engineer from the diff.

I re-ran the projection logic against a normal US address end-to-end and the to_find_rates_args() / to_rate_table_locations() pair comes out byte-identical to what the old code produced for that case, so the seam change is invisible to the 95% path. The matched-rates fix is the one I'm most glad landed, silent disappearance of tax from price display / shipping / coupons is the kind of bug nobody files a ticket for.

Only thing I'd add, also non-blocking: the same sanitize_key() empty-collapse case the prior reviewer flagged would be nice to pin in state_compact_provider, but I agree it's R2.

Confirmed `composer test` locally: 331 tests, 793 assertions, all green.

bartech:
## Pushed — the `sanitize_key()` collapse cases are pinned

Thanks @iyut and @CezaryDrewniak. You both independently asked for the same thing and you were right that it rested on reasoning rather than a test: the changelog cites a non-Latin state as the motivating case, but no provider row exercised one.

Added five rows to `state_compact_provider`, not one:

- **Three that collapse to empty outright** — Cyrillic (`ЛЕН`), CJK (`东京`), and a bare diacritic (`Ö`).
- **Two that collapse only partially** — `ÎF` → `F`, `ÅL` → `L`.

The partial pair is the one I'd have missed by only adding the obvious case, and it's the more dangerous shape: the result is a shorter but entirely plausible state code rather than an obviously broken empty string. That's the duplication path exactly — the row is stored under the collapsed value, looked up under the raw one, never found again, and a fresh row inserted on every calculation.

Expected values were read off WordPress's own `sanitize_key()` rather than worked out by hand, and the test's existing second assertion re-checks that agreement on every row. All five fail against the pre-migration `str_replace( ' ', '' )` implementation.

Happy this didn't get deferred to R2 after all.

### Also in this push

This branch was rebased onto the updated stack, which brings in the review fixes from #2985 and #2986 — the `taxjar_store_settings` array guard, and the nexus validation scoping (state required for the US only, based on a live-API probe; details on #2986). No changes to this PR's own seam beyond the provider rows.

### Verification

- `composer test` — **354 tests, 861 assertions**, all green at the top of the stack.
- PHPCS `--report=source` diff across all four changed files — identical, 181 → 181.

One correction worth flagging, since it affects how much earlier "zero new violations" claims are worth: my diff method was filtering `--report=source` on `[ ]` rows, which silently skips auto-fixable sniffs marked `[x]`. The new provider rows introduced 12 `MultipleStatementAlignment` violations and the check reported clean. Fixed with `phpcbf` and re-verified with a pattern that catches both. The earlier claims do hold — but only because nothing else happened to introduce an `[x]` row.

bartech:

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1860,6 +1857,8 @@ private function get_itemized_tax_rates( $taxes, $taxjar_taxes, $options ): arra
 	 * @return string Normalized city, safe for `_update_tax_rate_cities` and `find_rates`.
 	 */
 	protected static function normalize_city( $city ) {
+		wc_deprecated_function( __METHOD__, '3.6.11', '\Automattic\WCServices\Tax\Address::normalize_city()' );
```
Abdalsalaam: 3.6.11 shipped on 2026-08-05 (changelog.txt:14, woocommerce-services.php:11 `Version: 3.6.11`, readme.txt:10 `Stable tag: 3.6.11`). This PR's changelog entries land in the unreleased `= 3.6.12 - 2026-xx-xx =` section at changelog.txt:3. The runtime notice will tell out-of-repo subclasses — the exact audience `normalize_city()` is retained for — that the method was deprecated in a release where it was not.

```suggestion
		wc_deprecated_function( __METHOD__, '3.6.12', '\Automattic\WCServices\Tax\Address::normalize_city()' );
```

bartech: Confirmed and fixed, and thanks — the reasoning about *who* reads this string is the part that made it worth splitting the fix across two PRs rather than patching only the line you flagged.

The stack was opened before 3.6.11 was cut, and both version stamps were written against what was then the next unreleased block. 3.6.11 shipped on 2026-08-05 mid-stack; the rebase moved the changelog entries to `= 3.6.12 - 2026-xx-xx =` but left the two literals behind, which is the drift you caught.

### There were two of them, in two different PRs

The deprecation was staged deliberately, so the two version stamps live on different branches:

| String | Added by | PR | Fixed by |
|---|---|---|---|
| `@deprecated 3.6.11` (docblock tag) | `2437cb8a` | #2984 | `403b3f95` |
| `wc_deprecated_function( ..., '3.6.11', ... )` | `498ba8dc` | **#2988 (here)** | `100c5e3e` |

`2437cb8a` added the tag but deliberately *no* runtime notice, because four in-repo call sites were still live at that point and a notice would have fired on every checkout. The runtime call only appears here in #2988, once the last caller moved to the value object. Your suggestion is applied verbatim in `100c5e3e`.

### Why the base branch too, and not just this line

Fixing only the line in this PR would have left `@deprecated 3.6.11` on the base branch, and #2986 is still `CHANGES_REQUESTED` — so if #2984/#2985 merged and 3.6.12 were cut before the tail of the stack landed, the release would have shipped the exact wrong claim.

The docblock is also the surface your argument applies to hardest: `normalize_city()` is `protected static` and retained solely for out-of-repo subclasses, and the tag is what their IDEs, PHPStan, and generated API docs read.

### Restack

| Branch | old tip | new tip |
|---|---|---|
| `chore/wootax-313` | `28b2b7e6` | `403b3f95` |
| `chore/wootax-313-r1-2` | `a81fac15` | `500e6ff4` |
| `chore/wootax-313-r1-3` | `8b8d6627` | `6765d408` |
| `chore/wootax-313-r1-4` | `662605d8` | `100c5e3e` |

No conflicts. The total tree delta from the old stack tip to the new one is exactly the two one-line changes, and every pre-existing commit is preserved. **#2984 was a fast-forward, not a force-update** — the fix went on top rather than amending `2437cb8a` — so its approval was never at risk. All four are `MERGEABLE`, both approvals survived, and the PR bases are unchanged.

### Verification

`composer test` — **359 tests / 872 assertions**, unchanged from the pre-rebase baseline.

Worth noting that the suite could never have caught this: `setExpectedDeprecated( 'WC_Connect_TaxJar_Integration::normalize_city' )` matches on the function name, and the version argument is unasserted by construction. Version stamps in `@deprecated` / `wc_deprecated_function()` are review-caught or not caught at all — so this one is entirely down to you spotting it.

No BC-impact statement change is needed on either PR description: the deprecation surface itself is unchanged (same method, same signature, same retained-delegate rationale). Only the release the notice *names* was corrected, which makes the existing statement accurate rather than altering it.

changelog.txt
```diff
@@ -7,6 +7,9 @@
 * Fix   - Prevent a fatal error during cart and checkout tax calculation when a customization supplies an invalid store address.
 * Fix   - Prevent tax calculation from failing when the store address postcode holds more than one comma-separated value.
 * Fix   - Accept lower-case country and state codes from the woocommerce_taxjar_nexus_address filter, and fall back to the store address when a custom nexus address is incomplete instead of abandoning the calculation.
+* Fix   - Prevent unbounded growth of WooCommerce tax rate rows when the customer state contains a period, a space, or a non-Latin character.
```
Abdalsalaam: The old lookup was str_replace(' ','',$state), so 'N Y' became 'NY' and core stored strtoupper(sanitize_key('N Y')) = 'NY' — they matched and no row was duplicated. The PR's own provider note at tests/php/test-class-wc-connect-taxjar-integration.php:2985 says the old code 'happened to agree with sanitize_key() for a space, and only for a space'. Separately, a wholly non-Latin state collapses to '' under sanitize_key(), and core's criterion is `tax_rate_state IN ( %s, '' )`, so the stored '' already matched the second arm of the IN clause. Only punctuation (e.g. 'N.Y.') and partially-collapsing accented Latin (e.g. 'ÎF' -> 'F') actually diverged — the exact distinction the PR's own test provider draws at tests/php/Tax/test-address.php:512-519 and the changelog erases.

```suggestion
* Fix   - Prevent unbounded growth of WooCommerce tax rate rows when the customer state contains a character WooCommerce strips before storing it, such as a period or an accented letter.
```

readme.txt
```diff
@@ -77,6 +77,9 @@ This plugin relies on the following external services:
 * Fix   - Prevent a fatal error during cart and checkout tax calculation when a customization supplies an invalid store address.
 * Fix   - Prevent tax calculation from failing when the store address postcode holds more than one comma-separated value.
 * Fix   - Accept lower-case country and state codes from the woocommerce_taxjar_nexus_address filter, and fall back to the store address when a custom nexus address is incomplete instead of abandoning the calculation.
+* Fix   - Prevent unbounded growth of WooCommerce tax rate rows when the customer state contains a period, a space, or a non-Latin character.
```
Abdalsalaam: readme.txt carries a verbatim copy of the changelog.txt:10 entry and is the copy published to wordpress.org. The same two of three named triggers ('a space', 'a non-Latin character') never produced the unbounded row growth the entry claims to prevent, per the PR's own tests.

```suggestion
* Fix   - Prevent unbounded growth of WooCommerce tax rate rows when the customer state contains a character WooCommerce strips before storing it, such as a period or an accented letter.
```

changelog.txt
bartech: Applied verbatim, to both copies — `86eca793`.

You're right, and the mechanism is sharper than the entry was. I checked it against core rather than taking the reasoning as read, so here's what I confirmed and where it landed.

Core writes and reads that column with two different functions, and they are not symmetric:

- **Write** — `WC_Tax::prepare_tax_rate()` singles `tax_rate_state` out for `sanitize_key()` before formatting, i.e. `strtoupper( sanitize_key( $value ) )`. (`includes/class-wc-tax.php:1073`, WC 10.7.0.)
- **Read** — `WC_Tax::get_matched_tax_rates()` builds `tax_rate_state IN ( %s, '' )`. (`includes/class-wc-tax.php:370`.)

So a row is unfindable, and duplicated per calculation, only when the stored value is **non-empty and different** from what the lookup asked for. Your `''` arm is the part the entry missed:

| State | Stored | Old lookup asked for | Duplicated? |
|---|---|---|---|
| `'N Y'` | `NY` | `NY` | no — agreed by accident |
| `'ЛЕН'` | `''` | `ЛЕН` | no — `''` arm matches |
| `'N.Y.'` | `NY` | `N.Y.` | **yes** |
| `'ÎF'` | `F` | `ÎF` | **yes** |

Two things I'd push back on, neither of which changes the outcome:

1. **The `test-address.php:512-519` citation points at the wrong test.** That range is `test_..._city_matches_core_ascii_folding_not_multibyte` — the `jurisdiction_key()` / city byte-parity contract, a different concern. The distinction you mean is drawn by `state_compact_provider()` at `test-address.php:930-949`, which already separates *"Collapse to empty"* from *"Partial collapse — the dangerous shape"*. Right evidence, wrong line.

2. **"a non-Latin character never produced the growth" is too strong the other way.** A *partially* collapsing state does duplicate — `'ÎF'` → `'F'`, `'ÅL'` → `'L'`, `'ЖNY'` → `'NY'`. Only a *wholly* collapsing one is safe. Your replacement wording sidesteps this correctly by keying on "a character WooCommerce strips" instead of naming scripts, which is why I took it as written rather than editing it.

**The same overbroad claim was in two more places**, so `86eca793` fixes those too:

- the `'state containing punctuation'` provider note — *"Any state value carrying a character outside `[a-z0-9_-]` behaved this way, non-ASCII included"*;
- the `state_compact()` docblock, which implied duplication for any non-ASCII state.

**And the gap that let it drift: nothing pinned the collapse at the rate table.** `state_compact_provider()` pins what the *projection* returns; the duplication claim itself was unpinned. `4b07593c` adds two rows to `provide_rate_table_round_trips()`, which feeds both the round-trip test and the matched-rates lookup, so two rows are four cases:

- **`'state partially collapsing'`** (RO / `'ÎF'` → stored `'F'`) — regression test.
- **`'state wholly non-Latin'`** (RU / `'ЛЕН'` → stored `''`) — boundary characterization. It also pins that such an address yields a country-wide rate.

Both cities are plain ASCII on purpose, so the state dimension isn't confounded with the byte-parity contract `to_find_rates_args()` keeps with core.

Verified in both directions by reverting `state_compact()` to the old `strtoupper( str_replace( ' ', '', … ) )` — core sanitizes the write regardless, so that one line reproduces the old seam:

- `partially collapsing` — **both cases fail.** Round trip inserts a second row (`3 !== 2`); `allow_street_address_for_matched_rates()` cannot see the row `create_or_update_tax_rate()` had just written.
- `wholly non-Latin` — **both cases pass.** So it is a characterization, not a regression test, and that is the direct measurement of your claim.

Suite **366 tests / 899 assertions** green (was 362/881 — exactly +4/+18). PHPCS zero new: 5 violations on the two changed PHP files before and after.

Note the PR description's "331 tests / 793 assertions" is now stale — I've left it rather than editing the body mid-review.

readme.txt
bartech: Applied — `86eca793` changes both copies together, since `readme.txt` is the published verbatim mirror of `changelog.txt:10`.

Full reasoning on the [`changelog.txt` thread](https://github.com/Automattic/woocommerce-services/pull/2988#discussion_r3721895857): short version is that I confirmed it against core (`prepare_tax_rate()` stores `strtoupper( sanitize_key() )`, `get_matched_tax_rates()` matches `tax_rate_state IN ( %s, '' )`), the space and wholly-collapsing cases genuinely never duplicated, and `4b07593c` now pins that boundary with a regression case (`'ÎF'` → `'F'`) and a characterization case (`'ЛЕН'` → `''`) that behave differently against the pre-fix code.

Abdalsalaam (approved):
LGTM

PR #2990, opened 2026-08-07
[WOOTAX-322] Refactor - Hoist the canonical-id exemption remap

bartech:
## Description

Follow-up to #2983, filed from @Abdalsalaam's review there.

`assign_canonical_line_item_ids()` rewrites each TaxJar line item's `id`, and that rewrite is
exactly what invalidates `$non_taxable_line_items`: callers record that map while ids are still
context-specific, but `get_itemized_tax_rates()` reads it back under the canonical id. Landing
WOOTAX-258 and WOOTAX-240 together left the map keyed one way and the lookup the other, so the
non-taxable guard stopped firing and WOOTAX-240 silently regressed.

The fix shipped in #2983 repairs the map in the caller — an identical eight-line block duplicated
verbatim in `get_line_items()` and `get_backend_line_items()`. That leaves a coupling invariant
living one level below where it belongs: a third caller of `assign_canonical_line_item_ids()` that
omits the block reintroduces the same regression, silently, with nothing failing. That is the same
failure signature as the original bug.

This PR moves the remap **into** `assign_canonical_line_item_ids()`, which already holds both
`$key` and `$product_id` in scope:

```php
$canonical_id = $product_id . '-' . substr( $fingerprint, 0, 12 ) . '-' . $occurrences[ $fingerprint ];

if ( isset( $this->non_taxable_line_items[ $product_id . '-' . $key ] ) ) {
    $non_taxable[ $canonical_id ] = true;
}

$line_items[ $key ]['id'] = $canonical_id;
```

The map is no longer observable in the stale keying, and a future caller inherits the invariant
instead of having to remember it. Each call site keeps a one-line comment noting the side effect.

### Why this is behaviour-preserving

- Both callers already reset the map to `array()` before building (`:983` cart, `:1098` order), so
  replacing it wholesale inside the helper cannot clobber the other path's state.
- The old block dropped entries whose item key was absent from `$line_items`
  (`isset( $line_items[ $item_key ] )`). The new form iterates `$line_items`, so such an entry
  simply never appears — same result. This case is real, not theoretical: `get_backend_line_items()`
  records the exempt status *before* the `if ( $unit_price )` guard decides whether the line is
  emitted at all, so a zero-priced exempt item is recorded and then omitted.
- `$product_id . '-' . $key` reconstructs the caller's key exactly, so the lookup is equivalent to
  the old parse for every input the two callers can produce.

### One strict difference, unreachable today

The removed `substr( $legacy_key, strpos( $legacy_key, '-' ) + 1 )` mis-parses any key containing a
`-`. The new form does not parse at all, so it is robust where the old one was not. No current
caller can produce such a key — cart item keys are md5 hashes, order item IDs and product IDs are
integers — so this is unobservable in practice.

I deliberately did **not** add a test pinning it: asserting it would assert a behaviour change in a
refactor that is meant not to have one. Flagging it here instead, since it is the undocumented
assumption the ticket set out to remove.

### Backward compatibility

**No BC impact.** `assign_canonical_line_item_ids()` is `private` and `$non_taxable_line_items` is a
`private` property, both introduced in the current unreleased 3.6.12 — neither has ever been
reachable from outside this class. No hook is added, removed, reordered or retimed; no new read of a
global or of `WC()` state is introduced; no site-scoped option or path/URL construction is touched,
so multisite and non-root install layouts are unaffected.

### On the changelog checkboxes

Left unchecked deliberately, rather than overlooked. `assign_canonical_line_item_ids()` is
`@since 3.6.12` and `$non_taxable_line_items` came from the WOOTAX-240 fix, also in the 3.6.12
block — which is still unreleased (`2026-xx-xx` placeholder; `Stable tag: 3.6.11`). Both the code
being changed and the state it replaces are unreleased, so there is no user-visible delta to report
and an entry would document internal churn in a version that never shipped. Happy to add one if you
would rather the block record it.

### Related issue(s)

Closes [WOOTAX-322](https://linear.app/a8c/issue/WOOTAX-322)

Follow-up to #2983 (WOOTAX-258). Guards the fix for [WOOTAX-240](https://linear.app/a8c/issue/WOOTAX-240).

No user-visible change, so there is nothing to screenshot — this is a behaviour-preserving refactor
and the test suite is the evidence. To verify:

1. `composer test` — expect **401 tests / 978 assertions** green. Baseline on `trunk` is 399/975;
   the delta is exactly the two new tests added here.
2. Confirm the five behavioural guards from WOOTAX-240 / WOOTAX-258 pass **unmodified** — they are
   untouched in the diff, which is the point:
   ```
   ./vendor/bin/phpunit --filter '(test_get_itemized_tax_rates_non_taxable_line_does_not_zero_standard_rate|test_get_itemized_tax_rates_filtered_non_taxable_line_does_not_zero_standard_rate|test_get_itemized_tax_rates_shipping_only_line_does_not_zero_standard_rate|test_get_line_items_still_records_none_status_as_non_taxable|test_get_backend_line_items_sends_shipping_only_status_as_exempt)'
   ```
3. Optional end-to-end check of the underlying WOOTAX-240 behaviour: build a cart mixing a taxable
   Standard-class product with a non-taxable (Tax status "None") one and confirm the Standard rate
   row is not zeroed.

**PHPCS:** zero new violations. `--report=source` over the two changed files is identical to `trunk`
at 30 sniff types / 179 violations.

### Checklist

- [x] unit tests

#### On the two new tests

Both call the `private` helper **directly via reflection, with no caller involved**. That is
deliberate: a test entering through `get_line_items()` / `get_backend_line_items()` would stay green
if someone re-duplicated the block back into both callers, so it would pin the two known paths
rather than the contract. Testing at the seam is what makes a hypothetical third caller safe.

- `test_assign_canonical_line_item_ids_rekeys_non_taxable_map` — seeds the map in the legacy
  `<product_id>-<key>` shape and asserts it comes back keyed by canonical id, with the legacy key gone.
- `test_assign_canonical_line_item_ids_drops_non_taxable_entries_without_a_line_item` — pins the drop
  semantics that make replacing the map wholesale safe.

### DIFF
--- a/tests/php/test-class-wc-connect-taxjar-integration.php
+++ b/tests/php/test-class-wc-connect-taxjar-integration.php
@@ -4140,6 +4140,87 @@ class WP_Test_WC_Connect_TaxJar_Integration extends WC_Unit_Test_Case {
 	}
 
 	/**
+	 * Rewriting the ids is what invalidates `$non_taxable_line_items`, so re-keying it
+	 * belongs to the method that does the rewriting — not to each caller afterwards.
+	 *
+	 * Called directly here, with no caller involved, precisely so a future third caller
+	 * of `assign_canonical_line_item_ids()` inherits the remap instead of having to
+	 * remember it. Forgetting it has already caused one silent regression: the map
+	 * stayed keyed by the pre-canonical id while `get_itemized_tax_rates()` looked up
+	 * the canonical one, so the non-taxable guard never fired and no test failed.
+	 */
+	public function test_assign_canonical_line_item_ids_rekeys_non_taxable_map() {
+		$line_items = array(
+			'cart_key_exempt'  => array(
+				'id'               => 101,
+				'product_tax_code' => '99999',
+				'quantity'         => 1,
+				'unit_price'       => '10.00',
+				'discount'         => '0.00',
+				'tax_location'     => 'shipping',
+			),
+			'cart_key_taxable' => array(
+				'id'               => 202,
+				'product_tax_code' => '',
+				'quantity'         => 1,
+				'unit_price'       => '20.00',
+				'discount'         => '0.00',
+				'tax_location'     => 'shipping',
+			),
+		);
+
+		// Keyed the way both callers record it: "<product_id>-<context-specific key>".
+		$this->set_private_property( 'non_taxable_line_items', array( '101-cart_key_exempt' => true ) );
+
+		$result   = $this->invoke_protected_method( 'assign_canonical_line_item_ids', array( $line_items ) );
+		$recorded = $this->get_private_property( 'non_taxable_line_items' );
+
+		$this->assertSame(
+			array( $result['cart_key_exempt']['id'] => true ),
+			$recorded,
+			'The helper must leave the map keyed by canonical id, since that is what get_itemized_tax_rates() looks up.'
+		);
+		$this->assertArrayNotHasKey(
+			'101-cart_key_exempt',
+			$recorded,
+			'The pre-canonical key must not survive, or the stale entry outlives the id it described.'
+		);
+	}
+
+	/**
+	 * Replacing the map wholesale is only safe because an entry with no surviving line
+	 * item is meant to be dropped.
+	 *
+	 * `get_backend_line_items()` records the exempt status before the `if ( $unit_price )`
+	 * guard decides whether the line is sent at all, so a zero-priced exempt item is
+	 * recorded and then omitted. It has no canonical id to be re-keyed onto, and TaxJar
+	 * never echoes a breakdown line for it, so carrying it forward would be dead state.
+	 */
+	public function test_assign_canonical_line_item_ids_drops_non_taxable_entries_without_a_line_item() {
+		$line_items = array(
+			'item_1' => array(
+				'id'               => 101,
+				'product_tax_code' => '',
+				'quantity'         => 1,
+				'unit_price'       => '10.00',
+				'discount'         => '0.00',
+				'tax_location'     => 'shipping',
+			),
+		);
+
+		// 303 was recorded as exempt but never made it into $line_items.
+		$this->set_private_property( 'non_taxable_line_items', array( '303-item_2' => true ) );
+
+		$this->invoke_protected_method( 'assign_canonical_line_item_ids', array( $line_items ) );
+
+		$this->assertSame(
+			array(),
+			$this->get_private_property( 'non_taxable_line_items' ),
+			'An exempt record with no line item to re-key onto must be dropped, not carried forward.'
+		);
+	}
+
+	/**
 	 * Build a representative TaxJar request body for cache-signature tests.
 	 *
 	 * @param array $overrides Values to replace.
--- a/tests/php/test-class-wc-connect-taxjar-integration.php
+++ b/tests/php/test-class-wc-connect-taxjar-integration.php
@@ -4148,10 +4148,17 @@ class WP_Test_WC_Connect_TaxJar_Integration extends WC_Unit_Test_Case {
 	 * remember it. Forgetting it has already caused one silent regression: the map
 	 * stayed keyed by the pre-canonical id while `get_itemized_tax_rates()` looked up
 	 * the canonical one, so the non-taxable guard never fired and no test failed.
+	 *
+	 * Two exempt lines, not one, because the re-key accumulates: the assignment sits
+	 * inside an `isset()` guard, so with a single exempt line "add to the map" and
+	 * "replace the map" are indistinguishable and a suite of one-exempt-item fixtures
+	 * stays green either way. A cart with two exempt products would then leave only the
+	 * last one guarded, and the first one's 0% breakdown line would overwrite the shared
+	 * tax-class rate row — the exact WOOTAX-240 regression this seam exists to prevent.
 	 */
 	public function test_assign_canonical_line_item_ids_rekeys_non_taxable_map() {
 		$line_items = array(
-			'cart_key_exempt'  => array(
+			'cart_key_exempt'   => array(
 				'id'               => 101,
 				'product_tax_code' => '99999',
 				'quantity'         => 1,
@@ -4159,7 +4166,15 @@ class WP_Test_WC_Connect_TaxJar_Integration extends WC_Unit_Test_Case {
 				'discount'         => '0.00',
 				'tax_location'     => 'shipping',
 			),
-			'cart_key_taxable' => array(
+			'cart_key_exempt_2' => array(
+				'id'               => 303,
+				'product_tax_code' => '99999',
+				'quantity'         => 1,
+				'unit_price'       => '30.00',
+				'discount'         => '0.00',
+				'tax_location'     => 'shipping',
+			),
+			'cart_key_taxable'  => array(
 				'id'               => 202,
 				'product_tax_code' => '',
 				'quantity'         => 1,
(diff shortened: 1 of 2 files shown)


tests/php/test-class-wc-connect-taxjar-integration.php
```diff
@@ -4139,6 +4139,87 @@ public function test_canonical_line_item_id_keeps_product_id_prefix() {
...
+		$this->assertSame(
+			array( $result['cart_key_exempt']['id'] => true ),
+			$recorded,
+			'The helper must leave the map keyed by canonical id, since that is what get_itemized_tax_rates() looks up.'
+		);
+		$this->assertArrayNotHasKey(
+			'101-cart_key_exempt',
+			$recorded,
+			'The pre-canonical key must not survive, or the stale entry outlives the id it described.'
+		);
```
Abdalsalaam, line 4211: Every test touching non_taxable_line_items has exactly one exempt line item: the two new tests (:4152, :4199), test_get_line_items_still_records_none_status_as_non_taxable (:3892), and the end-to-end guard test_get_itemized_tax_rates_shipping_only_line_does_not_zero_standard_rate (:3963). Changing classes/class-wc-connect-taxjar-integration.php:1216 from '$non_taxable[ $canonical_id ] = true;' to '$non_taxable = array( $canonical_id => true );' passes the entire suite, because the enclosing if-guard means only matching iterations reset the accumulator and no test produces two matches. In production a cart with two exempt products would leave only the last guarded, and the first one's 0% TaxJar breakdown line would overwrite the shared tax-class rate row at :2069 - the exact WOOTAX-240 regression this PR exists to guard. The PR's stated thesis is that testing at the seam is what makes a hypothetical third caller safe, so the seam test should pin accumulation, not just re-keying.

```suggestion
		$line_items = array(
			'cart_key_exempt'   => array(
				'id'               => 101,
				'product_tax_code' => '99999',
				'quantity'         => 1,
				'unit_price'       => '10.00',
				'discount'         => '0.00',
				'tax_location'     => 'shipping',
			),
			'cart_key_exempt_2' => array(
				'id'               => 303,
				'product_tax_code' => '99999',
				'quantity'         => 1,
				'unit_price'       => '30.00',
				'discount'         => '0.00',
				'tax_location'     => 'shipping',
			),
			'cart_key_taxable'  => array(
				'id'               => 202,
				'product_tax_code' => '',
				'quantity'         => 1,
				'unit_price'       => '20.00',
				'discount'         => '0.00',
				'tax_location'     => 'shipping',
			),
		);

		// Keyed the way both callers record it: "<product_id>-<context-specific key>".
		$this->set_private_property(
			'non_taxable_line_items',
			array(
				'101-cart_key_exempt'   => true,
				'303-cart_key_exempt_2' => true,
			)
		);

		$result   = $this->invoke_protected_method( 'assign_canonical_line_item_ids', array( $line_items ) );
		$recorded = $this->get_private_property( 'non_taxable_line_items' );

		$this->assertSame(
			array(
				$result['cart_key_exempt']['id']   => true,
				$result['cart_key_exempt_2']['id'] => true,
			),
			$recorded,
			'Every exempt line must survive the re-key, accumulated onto its own canonical id - not just the last one.'
		);
		$this->assertArrayNotHasKey(
			'101-cart_key_exempt',
			$recorded,
			'The pre-canonical key must not survive, or the stale entry outlives the id it described.'
		);
```

CezaryDrewniak (commented):
Reviewed the full change and traced every consumer of the re-keyed map (`get_itemized_tax_rates`, both callers, the cache seam) — the refactor is behavior-preserving as described, the key reconstruction is exactly equivalent, and the drop semantics match the old caller-side remap. Tests pass (401/978), and no PHPCS violation lands on a changed line.

Agree with Abdalsalaam's comment, though: the seam test only pins re-keying, not accumulation. If `$non_taxable[ $canonical_id ] = true` on line 1216 were ever rewritten as a reset, the suite would stay green (no test produces two exempt line items), and a cart with two exempt products in one tax class would regress straight back to WOOTAX-240. I'd fold a second exempt line item into `test_assign_canonical_line_item_ids_rekeys_non_taxable_map` (or add a third test) and assert both canonical keys survive — that's exactly the invariant the "test at the seam" thesis is supposed to pin.

Small note for a follow-up (not blocking): while checking for similar single-element blind spots I found three pre-existing ones outside this PR — `override_woocommerce_tax_rates` has no test at all, the `$product_id-` prefix matcher has no collision test (product 1 vs 10), and `pre_recalculation_tax_snapshots` is only ever tested with one order. Worth separate tickets, not this PR.

tests/php/test-class-wc-connect-taxjar-integration.php
bartech: Applied in 1f01aa14 — you're right, and I reproduced the mutation before taking the fix, because the claim is precise enough to be checkable.

Swapping `:1216` to `$non_taxable = array( $canonical_id => true );` passes the **entire suite: 401 tests, 978 assertions, zero failures**. So the seam test was pinning re-keying and nothing else, exactly as you said. Root cause is the one you identified: the record sits inside `if ( isset( ... ) )`, so only matching iterations touch the accumulator, and with one exempt line per fixture "add to the map" and "replace the map" are indistinguishable. Every fixture in the file had exactly one.

I took your suggestion. With `cart_key_exempt_2` folded in, the same mutation now dies, and it names the entry it lost:

```
Every exempt line must survive the re-key, accumulated onto its own canonical id — not just the last one.
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
 Array &0 (
-    '101-e9bb84901207-0' => true
     '303-b3451431437e-0' => true
 )
```

Two notes on what I changed around the suggestion:

- I extended `test_assign_canonical_line_item_ids_rekeys_non_taxable_map` rather than adding a third test. It's one contract at one seam, and a separate two-item test would duplicate the fixture to assert a strictly weaker version of what this one now asserts.
- I added the reasoning to the docblock, not just the assertion message — specifically *why two lines and not one*. The next person to trim a fixture down to one item is the person who needs to read it, and an assertion message only surfaces after the test has already failed.

Worth stating plainly: **no production change was needed.** `class-wc-connect-taxjar-integration.php` is byte-identical to `1dc1d31e`. The helper was already accumulating correctly — the test just couldn't prove it. That's the more uncomfortable version of the finding, since a green suite was the only evidence the PR offered for a refactor whose entire premise is "test at the seam so a third caller is safe." A seam test that survives breaking the seam pins one instance of the contract, not the contract. Good catch.

bartech:
Thanks — the accumulation gap is fixed in 1f01aa14, and I reproduced your mutation before accepting it rather than taking it on trust. `$non_taxable = array( $canonical_id => true )` at `:1216` passes **401/401, 978 assertions**. So the criticism lands squarely: the PR argued "test at the seam so a third caller is safe" while shipping a seam test that survived breaking the seam. Pinning one instance of a contract isn't pinning the contract.

`cart_key_exempt_2` is folded into `test_assign_canonical_line_item_ids_rekeys_non_taxable_map` and both canonical keys are now asserted, so that mutation dies. Details and the failure output are in [my reply to @Abdalsalaam](https://github.com/Automattic/woocommerce-services/pull/2990#discussion_r3740449541). Worth saying explicitly: **no production change was needed** — `class-wc-connect-taxjar-integration.php` is byte-identical to `1dc1d31e`.

## Your three follow-ups — filed, and all three verified first

Agreed they're out of scope here. I checked each against the code before filing, since a ticket asserting an unverified gap is its own kind of debt. All three are real; one needed its framing corrected.

- **WOOTAX-324** — `override_woocommerce_tax_rates` has no test at all. Confirmed exactly as you described: **0** test references, against 15 for `preserve_order_taxes_on_recalculation`, 8 for `snapshot_order_taxes`, 3 for `restore_order_taxes_after_recalculation`. Worth noting the `woocommerce_calc_taxes` matches in the test file are the *option*, not the `woocommerce_calc_tax` filter, so there's no indirect coverage either — on a filter that produces the actual per-line tax amounts. Medium.

- **WOOTAX-326** — `pre_recalculation_tax_snapshots` with one order. This is the closest analogue to what this PR fixed, and it fails the same way: mutating `:2626` to `$this->pre_recalculation_tax_snapshots = array( (int) $order->get_id() => $snapshot )` also passes **401/401**. Consequence is concrete — snapshotting order B discards order A's snapshot, A hits the `! isset` early return at `:2649`, and A's pre-recalculation taxes are silently not restored. Bulk order actions, Store API, WP-CLI and cron all put two orders in one request. Medium.

- **WOOTAX-325** — the `$product_id-` prefix matcher. One correction here: **this isn't a collision bug.** The dash-terminated prefix already prevents it — product `1` searches for `"1-"`, which doesn't match `"10-abc123def456-0"` because position 0-1 is `"10"`. Both sites (`:1399`, `:1465`) even carry a comment stating that case verbatim. The real finding is the one your framing implies: nothing *pins* the delimiter. Dropping it to `strpos( $line_item_key, (string) $product_id )` at both sites passes **401/401** — so the plausible "simplification" of a concatenation that looks like noise reintroduces cross-product rate bleeding with CI green. A comment is not a test. Filed as Low, since the behaviour is correct today; I also flagged the second unpinned assumption in the same comment block ("first-match-wins is safe because same product ⇒ same tax_location").

Three for three surviving the full suite is the more interesting result than any individual one — the single-element fixture is a pattern in this file, not an isolated miss. Happy to pick these up next, or leave them for whoever grabs the queue.

CezaryDrewniak (approved):
Verified the follow-up end to end — good to merge.

The seam test now does what the PR's thesis claims: I applied the reset mutation (`$non_taxable = array( $canonical_id => true )`) to :1216 and it fails with the first exempt key missing from the map — exactly the WOOTAX-240 failure shape (first exempt line loses its guard, its 0% breakdown would clobber the shared rate row). Reverted, suite is green at 401/978, both seam tests pass, and PHPCS shows zero new violations (test file report identical to baseline). Production code confirmed byte-identical across the two commits, and I traced every consumer of the map and the canonical ids — nothing else is affected.

PR #2995, opened 2026-08-22
Modernize the version requirements maintenance workflow

Abdalsalaam:
### Changes proposed in this Pull Request:

Replaces the outdated `update-requires-headers.yml` workflow with `maintenance-update-version-requirements.yml` ("Maintenance: Update version requirements"):

* Bumps `Tested up to` when a WordPress RC is available, and `WC tested up to` once beta 2 of the next WooCommerce minor is out.
* Sets `Requires at least` to WP stable - 1 and `WC requires at least` to WC stable - 2. `Requires PHP` mirrors WooCommerce core. All updates are upgrade-only, never downgrades.
* Updates the headers in the main plugin file, `readme.txt`, and `README.md` wherever they exist, and adds matching changelog entries to `changelog.txt` and the readme changelogs, reusing the unreleased section when one exists (without duplicating entries) or creating one for the next patch version.
* Auto-detects the main plugin file and the default branch, so the same workflow file is shared across all extension repos with no per-repo configuration.
* Skips the run when a version update PR is already open (including PRs from the old workflow).
* No more WooCommerce zip download; versions come from the WordPress.org and GitHub APIs.

The generated PRs request reviews from the team and serve as the reminder to test the plugin against the upcoming WordPress/WooCommerce versions before merging.

Same change as woocommerce/woocommerce-shipping-usps#585, verified end-to-end with manual runs before rollout.

### Backward compatibility impact:

No plugin code changes, but the generated bumps do change who keeps receiving updates from WordPress.org:

* `Requires at least` used to mirror WooCommerce core's own WordPress minimum. It is now computed as WP stable - 1, which can sit above core's minimum. Today that means asserting WP 7.0 while WooCommerce core still supports 6.9, and since writes are upgrade-only the header never walks back. A store on the older WordPress stops being offered plugin updates until it upgrades.
* `Requires PHP` keeps mirroring WooCommerce core, so it will propose 8.0 when core moves.

Both are intentional policy choices, not side effects of the rewrite.

### How to test the changes in this Pull Request:

1. Merge, then trigger "Maintenance: Update version requirements" from the Actions tab.
2. Check the PR it opens: bumped headers and changelog entries, reviewers requested.
3. Trigger the workflow again and confirm the run skips because an update PR is already open.

Note: PRs created by the workflow use the default `GITHUB_TOKEN`, so CI does not start on them automatically. Close and reopen such a PR to trigger CI.

### Changelog entry

> None, developer tooling change only.


CezaryDrewniak (approved):
Tested this locally, extracted the script blocks from the workflow YAML and ran them against real copies of our files with stubbed `curl`/`gh`, plus a few live API checks. Everything checks out:

- Version math is correct at the roll-over edges (7.9→7.8, 8.0→7.9, 11.0→10.8, 12.0→11.8) and upgrade-only is respected, computed `10.8` for WC requires did *not* overwrite our current `10.9`.
- Tested the changelog logic against our actual unreleased `3.6.14` sections: entries land in the right format (`* Tweak - WordPress X.Y Compatibility.`), the existing WooCommerce 11.1 entry isn't duplicated, and a second run is a clean no-op. New-section creation works for both changelog styles, and the released-top-greater-than-next case re-bumps correctly.
- README.md is update-only as claimed (missing headers were not injected), and our bare README.md is left untouched.
- Plugin file detection, the open-PR guard (old + new branch names), PR body/reviewers, and the auto-skip all behave as described. With today's live data (WP 7.1, WC 11.0.1 + 11.1.0-beta.2) the run correctly comes out as "already up to date" and opens nothing.
- Graceful degradation checks out too: unrecognizable changelog or a readme without `Stable tag` → warning, headers still bumped, no corruption.

Two tiny notes, neither blocking: `releases/latest` resolves by publish time, so on a day where WC ships a new minor and older-minor patches together, `WC requires at least` could compute one minor low for a run (next run fixes it, and upgrade-only keeps it safe). And the RC probe only checks RC1–RC3, so if we ever get RC4 (I don't think we had one so far) it would delay the Tested up to bump until stable day.

Good to merge from my side

PR #3000, opened 2026-09-01
[WOOTAX-332] Fix - Require the labels capability for every address normalization request

Abdalsalaam:
## Description

`POST /wc/v1/connect/normalize-address` only checked the labels capability when `type` was exactly `origin`. Send any other value, or no `type` at all, and the route was public: the body went to the Connect server signed with the site's credentials. The only caller is the shipping label form in wp-admin, which is authenticated for both address types.

The route now requires `WC_Connect_Functions::user_can_manage_labels()` for every request. A body with a missing or non-object `address` gets a 400 rather than a warning or a fatal, and a missing `phone` no longer warns.

BC note: anonymous destination normalization is gone. Nothing in this plugin used it, and the `wcship_user_can_manage_labels` filter still decides access.

### Related issue(s)

WOOTAX-332

Before: with the plugin active and TOS accepted, run this without logging in. It returns 200 or a Connect error, not 401.

```
curl -X POST https://example.test/wp-json/wc/v1/connect/normalize-address \
  -H 'Content-Type: application/json' \
  -d '{"type":"destination","address":{"address":"1 Main St","city":"New York","state":"NY","postcode":"10001","country":"US"}}'
```

After: the same request returns 401. As a shop manager, validating the origin and destination address in the shipping label modal still works.

### Checklist

- [x] unit tests


bartech (approved):
**Approve with nits.**

The fix is correct and complete. I tested it end to end on a live install and ran the suite. All of the acceptance criteria pass. No other route in the plugin has the same gap.

### One request before merge

`readme.txt` goes to wordpress.org and to every store. Please make the entry non-disclosing:

```diff
-* Fix   - Require the shipping label permission for every address normalization request, including the destination address.
+* Fix   - Security update.
```

Please change `changelog.txt` in the same way, to keep the two files equal. Keep the detail in WOOTAX-332.

### Nits, none of them blocking

1. `tests/php/classes/test-class-wc-rest-connect-address-normalization-controller.php:50` — the `assertArrayNotHasKey` guard runs before line 54 replaces the server, so it checks a server that no test uses. Please move it after line 55.
2. `classes/class-wc-rest-connect-address-normalization-controller.php:40` — `$phone` is checked for presence but not for type. A non-scalar phone goes back into the response.
3. The 11 new test methods have no `@testdox` line. This repo uses them, and `.phpcs.php.xml` excludes `Generic.Commenting` under `tests/` for that reason.
4. `classes/class-wc-rest-connect-address-normalization-controller.php:92-94` — `check_permission()` is now the same as the base class method. Keep it for the docblock, or delete it, but please make it a decision.
5. `malformed_body_provider` has no case for a body that is a JSON scalar, such as `"1 Main St"`.

The `Spy_REST_Server` swap on lines 52-55 is correct, and so is the comment above it. I checked it because it looked redundant, and it is not. Please keep it.

<details>
<summary>Nit 2 — example</summary>

A body of `{"address":{"phone":{"x":1}, ...}}` makes `$response->normalized->phone` an object in the response. The capability check now limits who can do this, and the behaviour is the same as before this PR. But this PR hardens the same shape, so you can close it on the same line:

```php
$phone = isset( $address['phone'] ) && is_scalar( $address['phone'] ) ? (string) $address['phone'] : '';
```
</details>

<details>
<summary>Why the <code>Spy_REST_Server</code> swap is needed, for the record</summary>

`WC_REST_Unit_Test_Case::setUp()` installs `WP_Test_Spy_REST_Server`. That class overrides `dispatch()` to add the filter back, in `tests/legacy/framework/vendor/class-wp-test-spy-rest-server.php:31-41` of WooCommerce:

```php
/**
 * Call dispatch() with the rest_post_dispatch filter
 */
public function dispatch( $request ) {
    $result = parent::dispatch( $request );
    ...
    return apply_filters( 'rest_post_dispatch', rest_ensure_response( $result ), $this, $request );
}
```

So `WC_REST_Connect_Base_Controller::send_nocache_header()` runs on every `dispatch()` in these tests. That class has no `send_header()` method of its own, so the call goes to `WP_REST_Server::send_header()`, which calls `header()`.

WordPress core's `Spy_REST_Server` overrides `send_header()` and stores the value in `$sent_headers` instead, in `wordpress-tests-lib/includes/spy-rest-server.php:41-43`. The swap also gives each test a server that holds only the route under test, as lines 48-49 say.

The `setUp(): void` comment is also correct. The parent creates `$this->server` after the polyfilled `set_up()` returns, so an override of `set_up()` would not work.
</details>

<details>
<summary>What I verified</summary>

| Check | Result |
|---|---|
| PHPCS (`phpcs.xml.dist`) on both changed PHP files | 0 errors, 0 warnings |
| New test file alone | OK — 20 tests, 81 assertions. Repeated 3 times, in two environments. |
| Full suite, this branch against its base commit | The same. 165 errors and 2 failures on both. All of them are in `TaxJar_Integration`, `WCServices_Tax_Address` and `Blocks_Integration`, and all of them are earlier problems. |
| Release package | Clean. `tasks/release.js` copies a fixed list, so the new test file cannot ship. |
| Scope of the fix | The only other `check_permission()` overrides are the TOS controller, which is stricter, and the base class. No `permission_callback` in the plugin returns `true` without a condition. |

I ran a live matrix on WP 7.1 / WC 11.0.1 / PHP 8.4. I registered the controller on the REST server with a stub API client, so no request left the machine. 14 body shapes, 3 roles, on `trunk` and on this branch.

On this branch:

| Body shape | anonymous | subscriber | shop manager |
|---|---|---|---|
| All 5 `type` values, with a valid address | 401 | 403 | 200, sent to the server |
| Scalar, null, empty or absent `address`; `{}`; a JSON array | 401 | 403 | 400 `bad_form_data`, not sent |
| An address with no phone | 401 | 403 | 200, and no PHP notice |
| A body that is not JSON | 400 `rest_invalid_json` | 400 | 400 |

On `trunk` the same matrix returns 200 for an anonymous caller in 10 of the 14 shapes, and one shape gives a PHP 8 `TypeError`. This branch removes all of that, and it removes every PHP warning on the path.
</details>

<details>
<summary>Out of scope, for information</summary>

An authenticated label manager can still send a 100 KB `city`. WOOTAX-332 notes that `city` has no length limit. The anonymous path is now closed, so this needs a separate issue only if you want a limit at all.
</details>

bartech:
Correction to my review: **please ignore nit 1.** The `Spy_REST_Server` swap is necessary, and the comment above it is accurate. Keep both. My other points stand, but nit 2 needs a different fix, below.

I checked `WP_REST_Server::dispatch()` in WordPress core. The tests do not dispatch on that class. They dispatch on the server that `WC_REST_Unit_Test_Case::setUp()` installs, and that subclass adds the filter back.

<details>
<summary>Why the swap is necessary</summary>

`WP_Test_Spy_REST_Server` overrides `dispatch()` to apply `rest_post_dispatch`, in `tests/legacy/framework/vendor/class-wp-test-spy-rest-server.php:31-41` of WooCommerce:

```php
/**
 * Call dispatch() with the rest_post_dispatch filter
 */
public function dispatch( $request ) {
    $result = parent::dispatch( $request );
    ...
    return apply_filters( 'rest_post_dispatch', rest_ensure_response( $result ), $this, $request );
}
```

So `WC_REST_Connect_Base_Controller::send_nocache_header()` does run on every `dispatch()` in these tests. That class has no `send_header()` method of its own, so the call goes to `WP_REST_Server::send_header()`, which calls `header()`.

WordPress core's `Spy_REST_Server` overrides `send_header()` and stores the value in `$sent_headers` instead, in `wordpress-tests-lib/includes/spy-rest-server.php:41-43`. That is what the comment says, and it is right.

The swap also gives each test a server that holds only the route under test. The comment on lines 48-49 states that, and it is a further reason to keep it.
</details>

**Nit 2, restated.** The `assertArrayNotHasKey` on line 50 still runs before the swap, so it checks a server that no test dispatches to. Please move it after line 55, where it will check the server the tests use.

bartech:

PR #3001, opened 2026-09-03
[WOOTAX-345] Fix - Teach PHPCS the custom capabilities this plugin uses

bartech:
Closes WOOTAX-345 — https://linear.app/a8c/issue/WOOTAX-345/teach-phpcs-the-custom-capabilities-this-plugin-uses

## What

Declares `manage_woocommerce` and `wcship_manage_labels` in the `custom_capabilities` property of the `WordPress.WP.Capabilities` sniff in `.phpcs.php.xml`.

One file changed. No PHP touched.

## Why

The sniff only knows WordPress core capabilities, so it flagged all 5 uses of these two capabilities as unknown. They are correctly spelled and genuinely registered — `manage_woocommerce` by WooCommerce core for the shop_manager role, `wcship_manage_labels` by this plugin for shipping label access.

Warnings that are always false train reviewers to skip the sniff, which defeats its purpose: a misspelled capability in a `current_user_can()` call raises no runtime error, it just silently grants or denies access, and this sniff is what catches it.

It also charged a cost to doing the right thing. WOOTAX-336 adds two correct `manage_woocommerce` checks to close a security issue and is penalised with 2 new warnings for it.

## Verification

Measured with the `phpcs.xml.dist` aggregator, before and after, **at the same path** — PHPCS output is path-sensitive, so a scratch copy at a different location reports different sniffs.

| | Before | After |
|---|---|---|
| `WordPress.WP.Capabilities.Unknown` | 5 | **0** |
| Total violations | 3801 | 3796 |
| Sources | 104 | 103 |

A `diff` of the two `--report=source` outputs shows only those lines changing. No other sniff moved.

**The sniff is narrowed, not disabled.** Against a fixture containing all three forms:

```
current_user_can( 'manage_woocommerce' )    -> silent
current_user_can( 'wcship_manage_labels' )  -> silent
current_user_can( 'manage_woocommrce' )     -> still WARNING (typo caught)
```

Note the pre-push hook was skipped deliberately (`HUSKY=0`): it runs the full `npm test`, and this change contains no PHP and no JS. PHPUnit is unaffected at 414 tests.

## Backward compatibility

None. Lint configuration only — no PHP is touched, so there is no change to any signature, hook, filter, script or style handle, capability, global-state assumption, multisite behaviour or install-layout assumption. Nothing ships to users, so there is no changelog entry.

### DIFF
--- a/.phpcs.php.xml
+++ b/.phpcs.php.xml
@@ -10,6 +10,22 @@
 	<rule ref="WordPress-Extra"/>
 	<rule ref="WordPress-Docs"/>
 
+	<!--
+		The Capabilities sniff only knows WordPress core capabilities, so it reports every
+		correctly-spelled capability that WooCommerce or this plugin registers as "unknown".
+		Declaring the two we actually use keeps the sniff useful for genuine typos:
+		  - manage_woocommerce   is registered by WooCommerce core for the shop_manager role.
+		  - wcship_manage_labels is registered by this plugin for shipping label access.
+	-->
+	<rule ref="WordPress.WP.Capabilities">
+		<properties>
+			<property name="custom_capabilities" type="array">
+				<element value="manage_woocommerce"/>
+				<element value="wcship_manage_labels"/>
+			</property>
+		</properties>
+	</rule>
+
 	<rule ref="PHPCompatibility">
 		<exclude-pattern>tests/</exclude-pattern>
 	</rule>


Abdalsalaam (approved):
Checked against trunk with the aggregator at the same path. `WordPress.WP.Capabilities.Unknown` drops from 5 to 0 and nothing else moves. A fixture with `manage_woocommrce` still warns, so typos are still caught.

PR #3003, opened 2026-09-03
[WOOTAX-336] Fix - Require a nonce and capability to dismiss admin notices

bartech:
## Description

Two `admin_notices` callbacks wrote persistent, site-wide state straight from a query parameter, with no nonce and no capability check:

- `WC_Connect_Error_Notice::render_notice()` — `?wc-connect-error-notice=disable` set the `error_notice` option to `false`, hiding shipping-rate diagnostics.
- `WC_Connect_Loader::render_schema_notices()` — `?wc-connect-dismiss-server-notice=<id>` set the `wcc_notice_dismissed_<id>` transient for a month, hiding a Connect-server notice.

**Attack precondition.** Not unauthenticated. Either (a) a logged-in user who can load any wp-admin page — WooCommerce's `prevent_admin_access()` only gates on `edit_posts`, so **Author and above** qualifies with no WooCommerce capability at all — or (b) an administrator clicking a crafted link, since both are plain GETs with no nonce (classic CSRF). Impact is low: operational notices get hidden. No data disclosure.

**Fix.** Both dismiss links are now built with `wp_nonce_url()`, and both handlers require a valid nonce **and** `current_user_can( 'manage_woocommerce' )` before writing. The server-notice nonce action embeds the notice id, so a nonce only dismisses the notice it was issued for.

Two design choices worth a reviewer's attention:

- **`wp_verify_nonce()`, not `check_admin_referer()`.** The latter calls `wp_die()`. A failed check here degrades gracefully instead — the notice simply keeps rendering. An admin notice should never be able to fatal the page it renders on.
- **The nonce travels in `_wc_connect_notice_nonce`, not `_wpnonce`.** `add_query_arg()` with no URL argument builds off the current `REQUEST_URI`. If the admin screen showing the notice already carries its own `_wpnonce`, `wp_nonce_url()` would overwrite it and clicking dismiss would re-fire that screen's action with the wrong nonce. WooCommerce core avoids this identically with `_wc_notice_nonce`.

`manage_woocommerce` was chosen over `manage_options` because it is this plugin's existing convention — 3 existing call sites use it, `manage_options` appears once.

### Related issue(s)

Linear: [WOOTAX-336 — Global admin notices are dismissed by nonce-free, capability-free GET](https://linear.app/a8c/issue/WOOTAX-336/global-admin-notices-are-dismissed-by-nonce-free-capability-free-get)

Part of the *Security review of Escargot extensions* project.

**Before this change:**

1. On a store where a Connect server notice or the shipping-rate error notice is showing, log in as a user with the **Author** role (or any role that can reach wp-admin).
2. Visit any admin URL with `?wc-connect-error-notice=disable` appended.
3. The site-wide option flips and the notice disappears for every user — no nonce, no WooCommerce capability required.

**After this change:**

- The same bare URL does nothing; the notice keeps rendering.
- Clicking the notice's own **Dismiss** link still works normally as an administrator or shop manager — the link now carries a nonce.
- A shop manager (`manage_woocommerce`) can dismiss; an Author cannot.

### Backward compatibility

**No signature changed.** Additions only: two public class constants on each class. Nothing public was removed, renamed or reordered. No hook, filter, or script/style handle touched. A `private` helper added in the first round was folded back into its only caller during review; private methods are not part of the compatibility surface.

**The URL contract does change.** The bare-GET forms `?wc-connect-error-notice=disable` and `?wc-connect-dismiss-server-notice=<id>` no longer dismiss anything on their own. A full-repo grep (excluding `dist/`, `node_modules/` and `.po` files) found the **only** emitters are the two `add_query_arg()` calls changed here — no JS, no build artifact, no documentation. Both were one-click affordances rendered inside the notice itself, never a documented API, and the plugin regenerates both links with the nonce attached. Anyone hand-crafting those URLs was exploiting the bug. Assessed as safe.

**No new global-state assumption.** `current_user_can()` returns false rather than fataling when no user is initialised, which is the graceful degradation wanted on a path reachable outside a standard request.

**Install layout:** `add_query_arg()` / `wp_nonce_url()` work off the relative current `REQUEST_URI`; no path is built from the domain root.

**Multisite: reasoned correct, but not tested on a multisite install.** Both writes were already site-scoped and stay that way — `update_option()` and `set_transient()`, with no `get_site_option` / `set_site_transient` anywhere. One thing to be aware of: `manage_woocommerce` is a per-site role capability, so a Super Admin lacking it on a given site will not pass the check. That is identical to every other `manage_woocommerce` gate already in this plugin, so no new inconsistency is introduced.

### Verification

| Check | Result |
|---|---|
| `php -l` | No syntax errors detected |
| `composer test` (PHPUnit) | **441 tests / 1108 assertions, OK** |
| PHPCS, changed file, before vs after | **12 → 12** violations, none of them nonce- or capability-related |

No new PHPCS violations. #3001 (WOOTAX-345) has since landed on trunk and teaches the sniff `manage_woocommerce`, so the `WordPress.WP.Capabilities.Unknown` warnings this change would previously have raised no longer happen.

<details>
<summary>How PHPCS was measured</summary>

Before and after were run **at an identical path**, because PHPCS output is path-sensitive. The ruleset also excludes any directory whose name starts with a dot, so a scan run from a worktree under `.claude/` checks zero files and reports clean. Each run here confirmed `1 / 1 (100%)` files actually checked.

</details>

### Checklist

- [x] unit tests — 3 rejection-path regression tests (no nonce / invalid nonce / Author role)
- [x] `changelog.txt` — covered by the existing `Fix - Security update.` line
- [x] `readme.txt` — same

**No dedicated changelog line, deliberately.** An earlier revision of this PR added `Fix - Require a valid nonce and store management permission before an admin notice can be dismissed.` That sentence describes the flaw precisely enough to be a recipe against stores that have not updated yet, so it has been dropped in favour of the generic `Fix - Security update.` line already in the 3.6.15 section — the same treatment #3000 chose for the sibling security fix in this release. This branch's 3.6.15 section is now identical to trunk's.

**The tests did not work as first pushed, and now do.** `render_notice()` read the query parameter with `filter_input( INPUT_GET, ... )`, which reads the request data PHP captured at startup rather than the `$_GET` array. Under PHPUnit there is none, so it always returned `null` and the three tests never reached the guard — deleting the guard outright left all three passing. Fixed in 70fc51f0 by reading `$_GET` directly, the way the nonce already did and the way `render_schema_notices()` does. Confirmed by removing the guard again afterwards: the run now aborts instead of passing.

<details>
<summary>Remaining test gaps, called out rather than hidden</summary>

The *success* path still is not directly assertable, because `render_notice()` calls `exit` on a successful dismissal. A regression therefore shows up as an aborted run rather than a clean failure. Covering it properly needs that `exit` moved behind a seam.

`render_schema_notices()` has no test; it needs a `WC_Connect_Service_Schemas_Store` stub returning a `notices` payload.

Both are follow-up candidates.

</details>

### DIFF
--- a/classes/class-wc-connect-error-notice.php
+++ b/classes/class-wc-connect-error-notice.php
@@ -13,6 +13,22 @@ if ( ! class_exists( 'WC_Connect_Error_Notice' ) ) {
 
 	class WC_Connect_Error_Notice {
 
+		/**
+		 * Nonce action guarding the "dismiss this notice" link.
+		 *
+		 * @var string
+		 */
+		const DISMISS_NONCE_ACTION = 'wc_connect_dismiss_error_notice';
+
+		/**
+		 * Query argument carrying the dismissal nonce.
+		 *
+		 * A dedicated name (rather than `_wpnonce`) so the dismiss link cannot overwrite a
+		 * nonce the surrounding admin screen already put on the current URL.
+		 *
+		 * @var string
+		 */
+		const DISMISS_NONCE_NAME = '_wc_connect_notice_nonce';
 
 		private static $inst = null;
 
@@ -34,9 +50,11 @@ if ( ! class_exists( 'WC_Connect_Error_Notice' ) ) {
 
 		public function render_notice() {
 			$error_notice = filter_input( INPUT_GET, 'wc-connect-error-notice', FILTER_SANITIZE_ENCODED );
-			if ( 'disable' === $error_notice ) {
+			// Only act on a dismissal that carries a valid nonce and comes from a user who may
+			// manage the store. Anything else is ignored, so the notice simply stays visible.
+			if ( 'disable' === $error_notice && $this->is_dismissal_authorized() ) {
 				WC_Connect_Options::update_option( 'error_notice', false );
-				$url = remove_query_arg( 'wc-connect-error-notice' );
+				$url = remove_query_arg( array( 'wc-connect-error-notice', self::DISMISS_NONCE_NAME ) );
 				wp_safe_redirect( $url );
 				exit;
 			}
@@ -46,13 +64,37 @@ if ( ! class_exists( 'WC_Connect_Error_Notice' ) ) {
 			}
 		}
 
+		/**
+		 * Whether the current request is allowed to dismiss the error notice.
+		 *
+		 * Fails closed: a missing, stale or forged nonce, or a user without the
+		 * `manage_woocommerce` capability, leaves the stored option untouched.
+		 *
+		 * @return bool
+		 */
+		private function is_dismissal_authorized() {
+			if ( ! current_user_can( 'manage_woocommerce' ) ) {
+				return false;
+			}
+
+			$nonce = isset( $_GET[ self::DISMISS_NONCE_NAME ] )
+				? sanitize_text_field( wp_unslash( $_GET[ self::DISMISS_NONCE_NAME ] ) )
+				: '';
+
+			return (bool) wp_verify_nonce( $nonce, self::DISMISS_NONCE_ACTION );
+		}
+
 		private function notice_enabled() {
 			 return WC_Connect_Options::get_option( 'error_notice', false );
 		}
 
 		private function show_notice() {
 			$link_status  = admin_url( 'admin.php?page=wc-status&tab=connect' );
-			$link_dismiss = add_query_arg( array( 'wc-connect-error-notice' => 'disable' ) );
+			$link_dismiss = wp_nonce_url(
+				add_query_arg( array( 'wc-connect-error-notice' => 'disable' ) ),
+				self::DISMISS_NONCE_ACTION,
+				self::DISMISS_NONCE_NAME
+			);
 			$error        = $this->notice_enabled();
 
 			if ( ! is_wp_error( $error ) ) {
--- a/classes/class-wc-connect-error-notice.php
+++ b/classes/class-wc-connect-error-notice.php
@@ -49,10 +49,22 @@ if ( ! class_exists( 'WC_Connect_Error_Notice' ) ) {
 		}
 
 		public function render_notice() {
-			$error_notice = filter_input( INPUT_GET, 'wc-connect-error-notice', FILTER_SANITIZE_ENCODED );
+			$error_notice = isset( $_GET['wc-connect-error-notice'] )
+				? sanitize_text_field( wp_unslash( $_GET['wc-connect-error-notice'] ) )
+				: '';
+			$nonce        = isset( $_GET[ self::DISMISS_NONCE_NAME ] )
+				? sanitize_text_field( wp_unslash( $_GET[ self::DISMISS_NONCE_NAME ] ) )
+				: '';
+
 			// Only act on a dismissal that carries a valid nonce and comes from a user who may
 			// manage the store. Anything else is ignored, so the notice simply stays visible.
-			if ( 'disable' === $error_notice && $this->is_dismissal_authorized() ) {
+			// The nonce is verified inline rather than in a helper so the read of $_GET and its
+			// verification stay in one scope, which is also what render_schema_notices() does.
+			if (
+				'disable' === $error_notice
+				&& current_user_can( 'manage_woocommerce' )
+				&& wp_verify_nonce( $nonce, self::DISMISS_NONCE_ACTION )
+			) {
 				WC_Connect_Options::update_option( 'error_notice', false );
 				$url = remove_query_arg( array( 'wc-connect-error-notice', self::DISMISS_NONCE_NAME ) );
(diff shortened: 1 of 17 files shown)


bartech:

classes/class-wc-connect-error-notice.php
```diff
@@ -34,9 +50,11 @@ public function disable_notice() {
 
 		public function render_notice() {
 			$error_notice = filter_input( INPUT_GET, 'wc-connect-error-notice', FILTER_SANITIZE_ENCODED );
-			if ( 'disable' === $error_notice ) {
+			// Only act on a dismissal that carries a valid nonce and comes from a user who may
+			// manage the store. Anything else is ignored, so the notice simply stays visible.
+			if ( 'disable' === $error_notice && $this->is_dismissal_authorized() ) {
```
Abdalsalaam, line 55: `filter_input()` never sees `$_GET`, so the tests can't drive this branch. Reading the parameter the same way as the nonce fixes that, and keeping the nonce check in this function keeps PHPCS quiet.

```suggestion
			$error_notice = isset( $_GET['wc-connect-error-notice'] )
				? sanitize_text_field( wp_unslash( $_GET['wc-connect-error-notice'] ) )
				: '';
			$nonce        = isset( $_GET[ self::DISMISS_NONCE_NAME ] )
				? sanitize_text_field( wp_unslash( $_GET[ self::DISMISS_NONCE_NAME ] ) )
				: '';

			// Only act on a dismissal that carries a valid nonce and comes from a user who may
			// manage the store. Anything else is ignored, so the notice simply stays visible.
			if (
				'disable' === $error_notice
				&& current_user_can( 'manage_woocommerce' )
				&& wp_verify_nonce( $nonce, self::DISMISS_NONCE_ACTION )
			) {
```

classes/class-wc-connect-error-notice.php
```diff
@@ -46,13 +64,37 @@ public function render_notice() {
...
+				return false;
+			}
+
+			$nonce = isset( $_GET[ self::DISMISS_NONCE_NAME ] )
+				? sanitize_text_field( wp_unslash( $_GET[ self::DISMISS_NONCE_NAME ] ) )
+				: '';
+
+			return (bool) wp_verify_nonce( $nonce, self::DISMISS_NONCE_ACTION );
+		}
+
```
Abdalsalaam, line 86: Unused once the check is inlined above.

```suggestion
```

Abdalsalaam (changes requested):
The fix is right, but the three new tests never get to it. `filter_input( INPUT_GET, ... )` reads the request's own GET data rather than the `$_GET` array, so under PHPUnit it always returns null, `$error_notice` is never `'disable'`, and the guard is skipped. I removed the guard locally and all three still passed.

Reading the parameter from `$_GET`, the way the nonce already is, fixes that. With the suggestion below the tests pass with the guard in place, and with the guard disabled the run aborts at the `exit`, so a regression would show. I inlined the nonce check because keeping it in the helper makes the nonce sniff flag the `$_GET` read in `render_notice()`. In the same function it stays quiet.

Also needs a trunk merge, the 3.6.15 changelog section now exists there.

bartech:
Good catch, and it was worse than it looked: the tests were not weak, they were testing nothing at all.

`filter_input( INPUT_GET, ... )` reads the request data PHP captured at startup, not the `$_GET` array. Under PHPUnit there is none, so it always returned `null`. To be sure the tests were really empty rather than just fragile, I put the guard back exactly as you reviewed it and then deleted it outright, leaving `filter_input` in place. All three still passed.

Applied both suggestions. Now that the parameter comes from `$_GET`, removing the guard aborts the run at the `exit`, so a regression cannot slip through green.

Also merged trunk. Full suite: 441 tests, 1108 assertions, OK.

<details>
<summary>One small correction on the PHPCS part</summary>

I measured all three variants at the same non-dot path, since the ruleset skips dot-directories and a scan from a worktree under `.claude/` checks zero files and looks clean:

| variant | `NonceVerification.Recommended` |
|---|---|
| what you reviewed (`filter_input` + helper) | 0 |
| `$_GET` in `render_notice()`, nonce still in the helper | 2 |
| both inline, as suggested | 0 |

So the sniff was not firing on the reviewed code. It fires only if you move the read to `$_GET` and leave the check in the helper. Inlining is still needed, just as a consequence of the `$_GET` move rather than as a fix to an existing violation. File total is unchanged at 12.

</details>

<details>
<summary>What the trunk merge picked up</summary>

Trunk had moved past the 3.6.15 section you mentioned. The merge also brought in #3001, which teaches PHPCS `manage_woocommerce`, so the two `Capabilities.Unknown` warnings the description called out are gone. I have updated the description to match, along with the test counts and a note about the vacuous tests.

</details>

Abdalsalaam (approved):

PR #3004, opened 2026-09-03
[WOOTAX-338] Fix - Write the tax rate backup as real CSV instead of concatenated strings

bartech:
## Description

`WC_Connect_Functions::backup_existing_tax_rates()` built the backup file by echoing each database column through `esc_attr()` and joining with commas.

`esc_attr()` is an **HTML** escaper. It does not quote a CSV field, and it does not neutralise a leading `=`, `+`, `-` or `@`. City and postcode location rows in the tax rate tables are written from customer checkout addresses, and tax rate names come from TaxJar response property names — so a shopper can plant a cell that a spreadsheet evaluates when the merchant downloads and opens the backup.

**The same defect also made the file wrong as a backup, which is the only reason it exists.** A city containing a comma split into two columns, and `O'Brien & Sons` was written out as HTML entities — so the file no longer restored the values it was taken from.

**Fix.** Rows are serialised with `fputcsv()` on a `php://temp` stream, which quotes commas, quotes and newlines correctly, and `esc_attr()` is gone. Every cell — not only the ones that look attacker-controlled — passes through a new `escape_csv_value()` helper, which prefixes a single quote when the value starts with `=`, `+`, `-`, `@`, TAB (`\t`) or CR (`\r`).

One behaviour addition: if the stream cannot be opened the function returns `false`. That matters, because `WC_Connect_TaxJar_Integration::backup_existing_tax_rates()` only TRUNCATEs the tax tables on success.

### The round-trip trade-off, chosen deliberately

A leading apostrophe is lossy — WooCommerce's importer does not strip it. So **numeric cells are exempt**: `is_numeric()` values are returned verbatim, because a number cannot open a formula and prefixing one would corrupt it on re-import. The realistic case is a negative tax rate: `-5.0000` round-trips byte-for-byte, while `=1+1` becomes `'=1+1`.

For a non-numeric formula cell there is no lossless option — quoting alone does not stop a spreadsheet evaluating `"=1+1"` — so safety wins, and the loss applies only to values that were already hostile or malformed. Everything else now round-trips **better** than trunk, since the entity-encoding is gone.

One cosmetic change: `fputcsv` quotes fields containing a space, so header cells are now quoted. Valid CSV, parses identically.

### Related issue(s)

Linear: [WOOTAX-338 — Tax-rate backup CSV built by concatenation (formula injection)](https://linear.app/a8c/issue/WOOTAX-338)

Part of the *Security review of Escargot extensions* project. **Sibling: WOOTBRT-44** is the identical finding in Table Rate Shipping — see *Portability* below.

**Before this change:**

1. Add a tax rate whose **City** is `=1+1` (or, more realistically, place an order from a city with a name containing a comma or an apostrophe).
2. Go to **WooCommerce → Status → Tools** and run the CA tax-rate deletion tool, which takes a backup CSV first.
3. Download the backup from the plugin's help view and open it in Excel, Numbers or Google Sheets.
4. The `=1+1` cell evaluates to `2`. A city containing a comma has split across two columns, and `O'Brien & Sons` reads as `O&#039;Brien &amp; Sons`.

**After this change:**

- The formula cell is stored as `'=1+1` and displays literally instead of evaluating.
- Commas, quotes and apostrophes round-trip exactly, so the file restores what it was taken from.
- Numeric values including negative rates are unchanged — `-5.0000` stays `-5.0000`.
- The file name, the `wp_upload_dir()`-derived directory and the download handler's filename regex are all unchanged, so existing backups still download normally.

### Backward compatibility

**Two new public static methods** on the globally-namespaced `WC_Connect_Functions`: `rows_to_csv( $rows )` and `escape_csv_value( $value )`. Purely additive — no existing signature, hook, filter, script handle, file name or path changes, so no third-party consumer can break on load. The exposure to state plainly: they are now part of the published surface and cannot be renamed later without a deprecation window.

**The content of the backup file changes** for anyone parsing it: fields are properly quoted, HTML entities are gone, and hostile cells carry a leading `'`. A real CSV parser is unaffected. A hand-rolled comma split was already broken by any city containing a comma.

**No new global state** — both helpers are pure and need no WordPress bootstrap, so they are safe on every path that reaches tax calculation (Store API, REST, cron, WP-CLI, webhooks). **No multisite or install-layout assumptions introduced**; paths still come from `wp_upload_dir()`.

### Portability

`escape_csv_value()` is pure PHP with zero WordPress dependency, and `rows_to_csv()` depends only on it. Both copy to **WOOTBRT-44** verbatim; the only edit is the class reference inside `array( __CLASS__, 'escape_csv_value' )` if it lands in a differently-named class.

### Verification

| Check | Result |
|---|---|
| `php -l` (2 files) | No syntax errors detected |
| `composer test` (PHPUnit) | **430 tests / 1037 assertions, OK** (trunk baseline 414) |
| PHPCS, same path before/after | 26 violations / 12 sources → **26 / 12**, identical per-sniff distribution |
| New test file | 0 violations |

PHPCS was measured before and after **at an identical path** — output is path-sensitive, and the aggregator also excludes dot-directories, so an in-worktree run silently scans nothing.

To run just this suite:

```bash
composer test -- --filter WP_Test_WC_Connect_Functions
```

### Checklist

- [x] unit tests — 16 tests / 29 assertions over both helpers, covering each formula leader, the numeric exemption, quoting of commas/quotes/newlines, and round-trip parsing

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,8 +1,10 @@
 *** WooCommerce Tax Changelog ***
 
+= 3.6.15 - 2026-xx-xx =
+* Fix   - Write the tax rate backup file as proper CSV, so cities and tax names containing a comma, a quote or an apostrophe are saved exactly as stored and no cell can be treated as a formula when the file is opened in a spreadsheet.
+
 = 3.6.14 - 2026-09-03 =
 * Fix   - Prevent a fatal error at checkout when a cart line's price is not numeric.
-* Fix   - Write the tax rate backup file as proper CSV, so cities and tax names containing a comma, a quote or an apostrophe are saved exactly as stored and no cell can be treated as a formula when the file is opened in a spreadsheet.
 * Tweak - Limit the WordPress.com connection banner to the Tax settings and Plugins pages, so it no longer appears on unrelated admin screens.
 * Tweak - Stop registering a redundant copy of the continents REST endpoint on requests that cannot reach it.
 * Tweak - WooCommerce 11.1 Compatibility.
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 3.6.14 - 2026-xx-xx =
 * Fix   - Prevent a fatal error at checkout when a cart line's price is not numeric.
+* Fix   - Write the tax rate backup file as proper CSV, so cities and tax names containing a comma, a quote or an apostrophe are saved exactly as stored and no cell can be treated as a formula when the file is opened in a spreadsheet.
 * Tweak - Limit the WordPress.com connection banner to the Tax settings and Plugins pages, so it no longer appears on unrelated admin screens.
 * Tweak - Stop registering a redundant copy of the continents REST endpoint on requests that cannot reach it.
 * Tweak - WooCommerce 11.1 Compatibility.
--- a/classes/class-wc-connect-functions.php
+++ b/classes/class-wc-connect-functions.php
@@ -167,101 +167,46 @@ if ( ! class_exists( 'WC_Connect_Functions' ) ) {
 				return false;
 			}
 
-			ob_start();
-			$header =
-				__( 'Country Code', 'woocommerce' ) . ',' .
-				__( 'State Code', 'woocommerce' ) . ',' .
-				__( 'ZIP/Postcode', 'woocommerce' ) . ',' .
-				__( 'City', 'woocommerce' ) . ',' .
-				__( 'Rate %', 'woocommerce' ) . ',' .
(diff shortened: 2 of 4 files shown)


bartech:

Abdalsalaam (approved):
Ran the real path against the database with cities `Foo, Bar`, `O'Brien & Sons` and `"Quoted" City`, a rate named `=1+1`, and a `-5.0000` rate. The file parses back to the exact values, `=1+1` comes out as `'=1+1`, the negative rate is untouched, and feeding the parsed city cell back through `WC_Tax` gives the same three cities. The 16 unit tests pass and PHPCS on the file is unchanged.

PR #3005, opened 2026-09-03
[WOOTAX-340] Fix - Skip the legacy DHL Inbox Note during admin AJAX

bartech:
## Description

WooCommerce Tax can evaluate and create its legacy DHL Inbox Note while serving `wp-admin/admin-ajax.php`.

The path guards on `is_admin()`, which is **true during admin AJAX** and is therefore not an AJAX guard. On affected legacy DHL stores that means one note-name lookup per admin-AJAX request after bootstrap, plus an insert on the first run.

This adds `! wp_doing_ajax() &&` to the note's condition in `load_admin_dependencies()`:

```php
if ( ! wp_doing_ajax() && ! self::is_wc_shipping_activated() && self::can_add_wc_admin_notice() ) {
```

Skipped, and **only** while `admin-ajax.php` is being served: the `require_once` of the note class, `WC_Connect_Note_DHL_Live_Rates_Available::init()`, the `$schemas->get_all_shipping_method_ids()` read, and `NoteTraits::possibly_add_note()` — which is a `get_notes_with_name()` DB lookup plus, on first run, the insert.

Still runs on admin AJAX, unchanged: `WC_Connect_Debug_Tools`, `WC_Connect_Help_View`, both `admin_notices` registrations, the `should_load_shipping_features()` gate, and `WC_Connect_Settings_Pages`.

**Only the note block is guarded, not the method.** `AGENTS.md` warns that a skip which bypasses an overridable method silently disables third-party overrides even though no signature changed. `WC_Connect_Note_DHL_Live_Rates_Available::init()` is also left untouched, so its public contract is identical for any external direct caller.

### Design decision, resolved with the ticket author

The ticket's Technical Notes prefer evaluating the note from `wc_admin_daily` or a versioned upgrade/schema-transition event. **This PR deliberately does not add that hook**, and the author has [confirmed that call is ours to make](https://linear.app/a8c/issue/WOOTAX-340#comment-73095615): *"I don't think it's my call tbh. I think you understand better what works best for this particular plugin. The main concern is to not make the calls when they aren't needed."*

Not making the calls when they aren't needed is exactly what this guard does. Adding a `wc_admin_daily` hook on top would mean writing **new shipping code**, which is compatibility-only surface area for grandfathered stores under `AGENTS.md § Product Scope`, to reach an outcome the existing non-AJAX admin page-load path already delivers. See *Backward compatibility* below for that path.

On acceptance criterion 3 (a test using an eligible legacy DHL schema fixture): the author noted the criteria come from a shared WCCOM issue template and said to ignore what does not fit here. The two behavioural tests below cover AJAX exclusion and non-AJAX registration, which is the behaviour that actually changed; the schema fixture is not built.

### Related issue(s)

Linear: [WOOTAX-340 — Guard the legacy DHL Inbox Note path during admin AJAX](https://linear.app/a8c/issue/WOOTAX-340/guard-the-legacy-dhl-inbox-note-path-during-admin-ajax)

Upstream coordination: WCCOM-2823 (submitting AJAX guards for Inbox Note registration upstream).

This needs an eligible legacy store — connected and past TOS bootstrap, shipping features enabled, WooCommerce Shipping **inactive**, and the legacy `wc_services_dhlexpress` service schema present. Stores without that schema stop before the note API either way.

**Before this change:** with query logging on, every `admin-ajax.php` request after bootstrap issues a `get_notes_with_name()` lookup for `wc-services-dhl-live-rates-available`.

**After this change:**

1. Trigger any admin-AJAX request (e.g. Heartbeat, or the order-list autosave). No note lookup and no insert occurs.
2. Load any ordinary wp-admin page. The lookup runs exactly as before, and the note is created if absent.
3. Confirm the note still appears in **WooCommerce → Home → Inbox** on an eligible legacy store.

### Backward compatibility

**The non-AJAX path that still delivers the note is any ordinary wp-admin page view.** `attach_hooks()` calls `load_admin_dependencies()` under plain `is_admin()`, which is true for every admin screen, not just admin-ajax. On such a request `wp_doing_ajax()` is false, so the block runs exactly as it did before.

Confirmed by reading the full call chain in this checkout: `woocommerce_init` → `after_wc_init()` → `attach_hooks()` → `is_admin()` → `load_admin_dependencies()`. The only new term in the condition is `wp_doing_ajax()`, which WordPress sets solely for admin-ajax requests.

Since the note is an Inbox note that is only ever *rendered* inside wp-admin, a store that can see it must load a non-AJAX admin page anyway. **No eligible grandfathered store loses the note** — at worst its creation moves from an AJAX request to the admin page load that follows. Narrowing an existing shipping surface would be a BC change for those stores; this does not narrow it.

**Hooks:** nothing in the skipped region is a plugin `do_action` / `apply_filters`. WooCommerce's `NoteTraits::possibly_add_note()` is `get_note()` → `can_be_added()` → `$note->save()`; the only downstream hook is WC core's `woocommerce_note_created`, fired once on creation, which still fires on the first eligible non-AJAX admin request.

**Also unaffected:** `possibly_delete_note()` is reachable only from `plugin_uninstall()`, not per-request, so it is left alone.

**Multisite:** unchanged. The guard adds no option or global read; site-scoped behaviour is identical on every path that still runs. Not tested under multisite.

### Verification

| Check | Result |
|---|---|
| `php -l` (2 files) | No syntax errors detected |
| `composer test` (PHPUnit) | **416 tests / 1015 assertions, OK** (trunk baseline 414) |
| PHPCS, same path before/after | 206 violations / 29 sources → **206 / 29**, diff only in the timing line |

PHPCS was measured before and after **at an identical path** — output is path-sensitive, and the aggregator also excludes dot-directories, so an in-worktree run silently scans nothing.

Both new tests were confirmed to **actually execute** rather than hit their `markTestSkipped` guards, which is the real risk with this pair:

```
✔ load_admin_dependencies() still registers the legacy DHL note outside admin AJAX.
✔ load_admin_dependencies() skips the legacy DHL note during admin AJAX.
```

### Checklist

- [x] unit tests — 2 tests covering AJAX exclusion and non-AJAX registration

Note: both tests `markTestSkipped` when WooCommerce Shipping is active or the `admin-note` data store is unavailable, so a green run can never silently mean the path was simply inert. If they skip rather than pass in CI, that is why and it needs a look.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,10 +1,12 @@
 *** WooCommerce Tax Changelog ***
 
+= 3.6.15 - 2026-xx-xx =
+* Tweak - Stop looking up the legacy DHL live rates inbox note on admin AJAX requests, which cannot display it.
+
 = 3.6.14 - 2026-09-03 =
 * Fix   - Prevent a fatal error at checkout when a cart line's price is not numeric.
 * Tweak - Limit the WordPress.com connection banner to the Tax settings and Plugins pages, so it no longer appears on unrelated admin screens.
 * Tweak - Stop registering a redundant copy of the continents REST endpoint on requests that cannot reach it.
-* Tweak - Stop looking up the legacy DHL live rates inbox note on admin AJAX requests, which cannot display it.
 * Tweak - WooCommerce 11.1 Compatibility.
 
 = 3.6.13 - 2026-08-24 =
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,6 +4,7 @@
 * Fix   - Prevent a fatal error at checkout when a cart line's price is not numeric.
 * Tweak - Limit the WordPress.com connection banner to the Tax settings and Plugins pages, so it no longer appears on unrelated admin screens.
 * Tweak - Stop registering a redundant copy of the continents REST endpoint on requests that cannot reach it.
+* Tweak - Stop looking up the legacy DHL live rates inbox note on admin AJAX requests, which cannot display it.
 * Tweak - WooCommerce 11.1 Compatibility.
 
 = 3.6.13 - 2026-08-24 =
--- a/readme.txt
+++ b/readme.txt
@@ -74,6 +74,7 @@ This plugin relies on the following external services:
 * Fix   - Prevent a fatal error at checkout when a cart line's price is not numeric.
 * Tweak - Limit the WordPress.com connection banner to the Tax settings and Plugins pages, so it no longer appears on unrelated admin screens.
 * Tweak - Stop registering a redundant copy of the continents REST endpoint on requests that cannot reach it.
+* Tweak - Stop looking up the legacy DHL live rates inbox note on admin AJAX requests, which cannot display it.
 * Tweak - WooCommerce 11.1 Compatibility.
 
 = 3.6.13 - 2026-08-24 =
--- a/readme.txt
+++ b/readme.txt
@@ -70,11 +70,13 @@ This plugin relies on the following external services:
 
 == Changelog ==
 
+= 3.6.15 - 2026-xx-xx =
+* Tweak - Stop looking up the legacy DHL live rates inbox note on admin AJAX requests, which cannot display it.
+
 = 3.6.14 - 2026-09-03 =
 * Fix   - Prevent a fatal error at checkout when a cart line's price is not numeric.
 * Tweak - Limit the WordPress.com connection banner to the Tax settings and Plugins pages, so it no longer appears on unrelated admin screens.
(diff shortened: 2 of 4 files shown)


bartech:

Abdalsalaam (approved):
Full suite passes here and both new tests run rather than skip. With the `wp_doing_ajax()` term removed, the AJAX test fails on the unexpected `get_all_shipping_method_ids()` call, so it guards the right thing.

One thing worth knowing if you run only this class with `--filter`: nothing loads the real API client class first, PHPUnit fabricates it for the mock, and the setter type checks fail. That is pre-existing and hits three older tests in the same file too, so not something for this PR.

PR #3006, opened 2026-09-03
Surface an incomplete WordPress.com connection in Tax settings

bartech:
## Description

A WordPress.com connection has two parts: a token for the site, and a WordPress.com account linked to it. A store can end up with the first and not the second.

When that happens, the plugin reports the store as "not connected" — the same as a store that never connected at all. Two very different situations look identical, and the merchant gets two confusing symptoms at once:

- WooCommerce → Settings → Tax has **no Automated taxes setting**, with nothing explaining why.
- A banner asks them to **connect a site that is already connected**, so clicking it looks like it does nothing.

Nothing names the one step that is actually missing: signing in with a WordPress.com account.

This PR adds a banner for that state. It says the site is connected but no account is linked, and the button takes the merchant to the sign-in. Once they finish, Automated taxes appears on its own.

<details>
<summary>Why both symptoms happen (same cause)</summary>

`WC_Connect_Jetpack::is_connected()` requires *both* a site token and a connected owner, so a site-token-only store gets `false`.

That flows downhill:

1. `WC_Connect_Nux::get_jetpack_install_status()` returns `JETPACK_NOT_CONNECTED`.
2. `WC_Services::pre_wc_init()` returns early on that value.
3. So `attach_hooks()` never runs.
4. So `WC_Connect_TaxJar_Integration::init()` never runs.
5. So its `woocommerce_tax_settings` filter is never added — and the **Automated taxes** row is never added to the Tax settings screen.

The NUX banner is registered *before* that early return, so it still renders — which is why the store gets the "connect your site" prompt while the setting is missing.

On the support case behind this issue, that combination sent two sessions chasing a REST API block and a WooCommerce.com marketplace reconnect instead of the unfinished sign-in.
</details>

<details>
<summary>What I deliberately did not change</summary>

**The bootstrap gate stays as it is.** `JETPACK_NOT_CONNECTED` is the *right* answer for this store, because tax features genuinely cannot work without a user token. Only what the merchant is told was wrong, so only that changed. `get_jetpack_install_status()` still returns its three existing values.

**No new connection flow.** `WC_Connect_Jetpack::connect_site()` already does the right thing here — it calls `try_registration()` only when the site is *not* registered, and otherwise goes straight to `get_authorization_url()`. The new banner reuses the existing `admin_post_register_woocommerce_services_jetpack` handler, so the button already lands on the sign-in.

**The Status page health item.** On this kind of store, WooCommerce → Status → WooCommerce Tax does not exist at all: `WC_Connect_Help_View` is only constructed in `load_admin_dependencies()`, which is downstream of the same early return. Editing its "Not connected to WordPress.com" message would have been unreachable code that looked like coverage. Worth its own issue — the plugin's own diagnostic surface disappears in exactly the state support is trying to diagnose.
</details>

### Related issue(s)

WOOTAX-329. The docs half of the same problem is WOODOCS-3744 (Done).

This needs a store whose WordPress.com connection has a site token and no connected owner (report card: `is_connected: true`, `has_connected_owner: false`). You cannot get there by running the normal connect flow to the end, which is why the original issue was not reproduced in a sandbox either. The decision table is covered by unit tests instead of a screenshot.

**Before:** Tax settings shows *"Connect your site to activate WooCommerce Tax"* and no Automated taxes setting.

**After:** the same screen shows *"Finish connecting your WordPress.com account — Your site is connected to WordPress.com, but no WordPress.com account is linked to it yet. Sign in to finish connecting and unlock automated tax calculation."* with a **Finish connecting** button. Completing the sign-in makes Automated taxes appear.

A store that never connected is unaffected and still gets the original banner.

### Backward compatibility

Additive. Existing callers of the one shared method get identical results, and that is covered by a test.

<details>
<summary>Surface-by-surface detail</summary>

| Symbol | Change | Impact |
|---|---|---|
| `WC_Connect_Jetpack::is_site_only_connection()` | New `public static` method. | Additive. None. |
| `WC_Connect_Nux::show_banner_site_only_connection()` | New `public` method. | Additive. None. |
| `WC_Connect_Nux::get_banner_type_to_display()` | New optional `$status` key `is_site_only_connection`; new possible return value `'site_only_connection'`. | This is `public static` on a globally-named class, so third-party code may call it. **The new return value is reachable only when the caller passes the new key**, so any existing caller gets byte-identical results — there is a dedicated regression test for that. A caller that opts in and `switch`es on the result could hit an unhandled value and fall to its `default` (no banner). No fatal, no signature change. |
| `WC_Connect_Nux::get_jetpack_install_status()` | Unchanged. | None. A fourth status value would have changed what `pre_wc_init()`'s `===` comparison and any external consumer sees, for no benefit. |
| `admin_post_register_woocommerce_services_jetpack` | Unchanged, reused as-is. | None. |
| Script/style handles | `wc_connect_banner` reused. No handle added, renamed or removed. | None. |
| Hooks | No `do_action` / `apply_filters` added, removed, renamed or reordered. | None. |

**Multisite:** reads only Jetpack connection state, which is site-scoped, exactly as the code beside it already did. No new option reads, no `get_site_option`. Not tested on a multisite install.

**Install layout:** the banner image uses `plugins_url()`, matching the sibling banners. No path concatenation.

**Global state:** `WC()->countries` is dereferenced only inside the `admin_notices` callback, after `should_display_nux_notice_for_current_store_locale()` has already done so — same as every other banner in this class. Nothing new runs on a REST, cron or CLI path.
</details>

### Checklist

- [x] unit tests

<details>
<summary>Verification results</summary>

- `composer test` → **OK (418 tests, 1015 assertions)**. Trunk baseline is 414 / 1008, so the delta is exactly the 4 tests and 7 assertions added here. No test reported as skipped.
- PHPCS on the three touched files, `--standard=phpcs.xml.dist --report=source`, at the same repo-relative paths on trunk and on this branch: **82 violations / 14 sources before and after — zero new.**
- Run in the main checkout rather than a worktree, so the known dot-directory false-pass does not apply; the progress line read `3 / 3`, confirming the files were actually scanned.
</details>

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Tax Changelog ***
 
+= 3.6.15 - 2026-xx-xx =
+* Add   - Prompt merchants to finish signing in to WordPress.com when the site is connected but no account is linked to it, so the missing step is named instead of hidden.
+
 = 3.6.14 - 2026-09-03 =
 * Fix   - Prevent a fatal error at checkout when a cart line's price is not numeric.
 * Tweak - Limit the WordPress.com connection banner to the Tax settings and Plugins pages, so it no longer appears on unrelated admin screens.
--- a/classes/class-wc-connect-jetpack.php
+++ b/classes/class-wc-connect-jetpack.php
@@ -120,6 +120,23 @@ if ( ! class_exists( 'WC_Connect_Jetpack' ) ) {
 		}
 
 		/**
+		 * Determines whether the site holds a WordPress.com token but has no connected owner.
+		 *
+		 * This is the "half connected" state: registration completed and left a blog token
+		 * behind, but nobody finished signing in with a WordPress.com account. Everything
+		 * that needs a user token - automated taxes included - stays unavailable, while
+		 * self::is_connected() reports false exactly as it does for a site that was never
+		 * connected at all. Callers that want to tell those two apart use this method.
+		 *
+		 * @return bool Whether the connection is site-level only.
+		 */
+		public static function is_site_only_connection() {
+			$connection_manager = self::get_connection_manager();
+
+			return $connection_manager->is_connected() && ! $connection_manager->has_connected_owner();
+		}
+
+		/**
 		 * Connects the site to Jetpack.
 		 * This code performs a redirection, so anything executed after it will be ignored.
 		 *
--- a/classes/class-wc-connect-nux.php
+++ b/classes/class-wc-connect-nux.php
@@ -231,6 +231,12 @@ if ( ! class_exists( 'WC_Connect_Nux' ) ) {
 				1. show_banner_before_connection()
 				2. connect to JP
 				3. show_banner_after_connection(), which sets the TOS acceptance in options
+			- Case 1b: the site registered but nobody finished signing in, so there is a blog
+				token and no connection owner. Jetpack reports this as "not connected", which is
+				what gates the plugin, but the merchant needs to be told something different:
+				there is nothing left to connect, only a sign-in to finish.
+				1. show_banner_site_only_connection()
+				2. finish the WordPress.com sign-in, which lands on the same flow as Case 1
 			- Case 2: Jetpack connected, no TOS
 				1. show_tos_only_banner(), which accepts TOS on button click
 			- Case 3: Jetpack connected, and TOS accepted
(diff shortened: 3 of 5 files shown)


Abdalsalaam (approved):
Reproduced the state on a fresh site by setting `jetpack_options.id` and a `jetpack_private_options.blog_token` with no `master_user`. The Tax tab shows the new banner, the button skips registration and lands on the WordPress.com authorize URL with the redirect back to the Tax tab, and once those options are cleared the original connect banner comes back. Unit tests pass and PHPCS is unchanged on the touched files.

PR #3011, opened 2026-09-21
[WOOTAX-320] Fix - Keep TaxJar tax code for multi-word tax classes

bartech:
## Description

Stores can link a tax class to a TaxJar category by putting the TaxJar code at the end of the class name, for example "Digital Goods 31000".

Checkout reads that code correctly. Recalculating an order in wp-admin did not when the class name had more than one word. It sent TaxJar no code at all. TaxJar then charged the normal rate, so the order's tax could differ from what the customer paid at checkout.

This affects WooCommerce's own "Reduced rate" and "Zero rate" classes too, once a code is added to them.

This PR makes the order screen read the code the same way checkout does. Both now use one shared helper.

**Backward compatibility:**
- Checkout sends exactly the same data as before.
- Order recalculation now sends the TaxJar code for multi-word classes. That changes the tax on those orders, on purpose: it now matches checkout.
- No public method, hook or filter changed. The new helper is private.
- Cached TaxJar answers for affected orders get a new cache key. That is harmless.
- Multisite: no change. The fix reads no settings.

**Not changed:** a class whose name simply ends in a number, like "Rate 2", is still read as TaxJar code `2`. Checkout already did this before the PR. Changing it would change checkout tax too, so it is left for a separate ticket. The new test pins this behaviour so any later change is deliberate.

### Related issue(s)

Closes [WOOTAX-320](https://linear.app/a8c/issue/WOOTAX-320/order-path-drops-the-taxjar-product-tax-code-for-any-multi-word-tax)

**Setup**
1. Use a store with a US address and automated taxes turned on. These steps use Colorado.
2. Go to **WooCommerce → Status → WooCommerce Tax** and turn **Logging** on.
3. Go to **WooCommerce → Settings → Tax**. Add a tax class named `Exempt Goods 99999`. (`99999` is TaxJar's code for "exempt", so the right answer is $0 tax.)
4. Create a product priced $100 and set its **Tax class** to `Exempt Goods 99999`.

**Test**
1. Create a new order in **WooCommerce → Orders → Add order**.
2. Set the billing and shipping address to a Colorado address, for example `1437 Bannock St, Denver, CO 80202`.
3. Add the product and click **Recalculate**.
4. Open **WooCommerce → Status → WooCommerce Tax** → **Taxes Log** and find the latest `Requesting:` line.

**Expected**
- On `trunk`: the log shows `"product_tax_code":""`. The `Received:` line after it shows tax of about $4.00.
- On this branch: the log shows `"product_tax_code":"99999"`. The `Received:` line shows `"tax_collectable":0.0`.

To check that checkout did not change, put the same product in the cart and go to checkout on both `trunk` and this branch. The log shows `"product_tax_code":"99999"` both times.

**Automated:** `composer test -- --filter test_cart_and_order_paths_extract_the_same_product_tax_code` runs 8 tax class names through both checkout and order recalculation. On `trunk`, the 3 multi-word names fail on the order side.

### Checklist

- [x] unit tests


CezaryDrewniak (approved):
LGTM
Reviewed and verified locally

- Traced both paths: the cart side is unchanged, the order side now matches checkout, and the `99999` overrides / canonical-ID fingerprints are untouched.
- The new test fails on trunk with exactly the 3 multi-word rows (`Expected '31000', Actual ''` on the order side) and passes 8/8 here. Full suite green (472 tests).
- PHPCS clean on the added lines.

Abdalsalaam:

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -1113,11 +1127,7 @@ protected function get_backend_line_items( $order ) {
 				$tax_status     = $product ? $product->get_tax_status() : 'taxable';
 			}
 			$this->backend_tax_classes[ $id ] = $tax_class_name;
-			$tax_class                        = explode( '-', $tax_class_name );
-			$tax_code                         = '';
-			if ( isset( $tax_class[1] ) && is_numeric( $tax_class[1] ) ) {
-				$tax_code = $tax_class[1];
-			}
+			$tax_code                         = self::get_product_tax_code_from_tax_class( $tax_class_name );
```
Abdalsalaam: Pre-existing, not from this PR: a taxable product in the built-in "Zero rate" class still gets different codes on the two paths. Checkout sends 99999 through the zero-rate check at line 1014, Recalculate sends an empty code and the order gets taxed. Worth a follow-up ticket.

Abdalsalaam (approved):

PR #3012, opened 2026-09-21
[WOOTAX-342] Fix - Stop tax lookups from overwriting a merchant's state-wide tax rate

bartech:
## Description

**The problem.** Some merchants add their own tax rate that covers the whole state, for example "Michigan, 6%", with no ZIP code or city. When a customer checked out to another town, our automated tax code saved the tax for that town and overwrote the merchant's rate. The rate was renamed (for example `MARQUETTE GWINN : City Tax`) and set to 0%, but it still covered the whole state.

After that, in-person, POS and REST API orders anywhere else in the state, including at the store's own address, were charged **no tax**. Website checkout still looked fine, so the problem was easy to miss. Fixing the rate by hand didn't help, because the next checkout broke it again.

**What this PR does.** We no longer overwrite a tax rate that covers the whole state. We save the tax for the looked-up town as new rows next to it, limited to that town's ZIP code and city.

- The merchant's own rate stays exactly as they set it, and it keeps covering every other address.
- The looked-up town uses its own new rows. A second checkout to the same town reuses them, so no extra rows are added.
- VAT rates and addresses without a ZIP code or city work as before.

**Backward compatibility.** Only how we write to the tax rate table changes. No function signatures or hooks change. Stores will see up to four extra rows per town that customers check out to, next to their own rates. This PR does **not** repair rates that were already overwritten. Merchants affected in the past still need to fix those rows (we will add an admin warning separately). A merchant rate limited by a ZIP wildcard, like `49*`, can still be overwritten; that is left for a follow-up. The tax rate table belongs to each site, so on multisite each site behaves on its own (not tested on multisite).

### Related issue(s)

Closes [WOOTAX-342](https://linear.app/a8c/issue/WOOTAX-342)

1. Set the store address to Marquette, MI 49855, and turn on **Automated taxes**.
2. Go to **WooCommerce → Settings → Tax → Standard rates**. Delete all rows, then add one row: Country `US`, State `MI`, ZIP and City left blank, Rate `6`, Priority `1`. Save.
3. On the website, add a product to the cart and check out to Gwinn, MI 49841.
4. Go back to **Standard rates**.
   - **Before this PR:** your `6%` row has been renamed `MARQUETTE GWINN : City Tax` and changed to `0%`.
   - **With this PR:** your `6%` row is unchanged. There are four new rows for `49841` / `GWINN`.
5. Check out to Gwinn again. No new rows are added.
6. Create an order to the store's address without going through checkout, for example with the REST API (`POST /wp-json/wc/v3/orders` with a Marquette, MI 49855 billing address and one product). The order should have 6% tax, not $0.00.

Automated tests: `composer test -- --filter create_or_update_tax_rate`

### Checklist

- [x] unit tests


CezaryDrewniak (approved):
Looks good to me

Tested order flow to **Gwinn 49841**

#### Results:

* **PR Behavior:** The catch-all is left untouched, we get four scoped rows per town, repeating the lookup (5 -> 5) works correctly, `find_rates()` properly returns 6% at the store, an unlooked-up town, and Gwinn, and REST-shaped orders calculate a consistent **$6.00** tax everywhere.
* **Base Commit:** Successfully reproduced the original ticket 1:1. The catch-all was renamed in-place to 0% with no locations, keeping store/Detroit orders at **$0.00**.
* **Wildcard (`49*`) Case:** Still overwritten (matching the base behavior and documented follow-up with no regressions).
* **WC 11.1 Core Premise Confirmed:** Unscoped rows match everything, `get_matched_tax_rates()` correctly keeps one row per priority while favoring more location matches, and write helpers properly invalidate the taxes cache. "Stable beside it" is officially confirmed as real core behavior.
* **Test Suite:** Full `composer test` is completely green (**470 tests passing**), and PHPCS remains unchanged compared to the base branch.

#### A Quick Non-Blocking Nit:

`create_or_update_tax_rate()` is public on an externally exposed class. It's probably worth adding a quick explicit note in the PR description mentioning that this method's behavior changes (it now does an insert instead of an in-place rewrite), rather than just generally saying "how we write to the table."

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -2237,6 +2237,33 @@ public function create_or_update_tax_rate( $location, $rate, $tax_class = '', $f
...
+		 *
+		 * VAT rows are country-wide by design and addresses without a postcode or city
+		 * write no scope, so both keep reusing the matched row. Otherwise they would
+		 * insert a new row on every calculation.
+		 */
+		if ( ! empty( $wc_rate ) && 'VAT' !== $tax_rate_name ) {
+			$wanted_locations = $address->to_rate_table_locations();
+			$wants_scope      = '' !== $wanted_locations['postcode'] || '' !== $wanted_locations['city'];
+
+			if ( $wants_scope && ! $this->tax_rate_has_locations( key( $wc_rate ) ) ) {
```
Abdalsalaam: If the merchant's state-wide row has priority 5 or higher, it now stays next to the four scoped rows and WooCommerce adds both. With a 6% catch-all at priority 5, a REST order to Gwinn got $12 tax on $100 instead of $6. Checkout is fine since it uses the returned rate ids. Same thing anywhere TaxJar returns fewer components than the merchant's priority. Needs handling here, or a note in the PR that priority 5+ rows are out of scope.

Abdalsalaam: A merchant row limited to a ZIP list like 49841;49855 has locations, so it still gets rewritten in place. After one checkout to Gwinn, a REST order at the store (49855) gets $0 tax. The wildcard follow-up should cover ZIP lists and city-only rows too.

Abdalsalaam:

PR #3014, opened 2026-09-21
[WOOTAX-347] Fix - Send the street when an order is recalculated from the admin

bartech:
## Description

**The problem.** When you click **Recalculate** on an order in the admin, WooCommerce sends the customer's country, state, ZIP and city, but not the street. We already had a small script that added the street, but it had two bugs:

- **Orders with no shipping address** (common for in-person and POS orders). WooCommerce then sends the billing address, but our script still took the street from the empty shipping address. TaxJar got the ZIP without a street and could return a different rate. Example: 18680 Highway 145, Dolores, CO 81323 is **3.3%** with the street and **7.3%** without it.
- **A double quote (`"`) in the city, state or ZIP.** The script crashed, the request was never sent, and the Recalculate spinner never stopped.

**What this PR does.**

- The script now uses the hook WooCommerce provides for this (`woocommerce_order_meta_box_recalculate_ajax_data`, WooCommerce 6.8+). It takes the street from the same address (billing or shipping) that WooCommerce uses for the other fields, as currently typed in the form. It no longer watches every admin AJAX request. jQuery is used only to listen to that hook, because WooCommerce fires it with jQuery's `triggerHandler()`. Everything else is plain JavaScript.
- A PHP fallback for when the script doesn't run (not loaded, broken by another script, or another plugin replaced the request data). If the request has no street, we use the street saved on the order, but **only** if the posted address matches the saved one. If the merchant changed the address without saving, we send no street, as before, so a new ZIP is never mixed with an old street. A street that was sent empty is kept empty.

**Known limitation.** The PHP fallback cannot see a street change that wasn't saved, because the street being edited is never sent. If the merchant changes **only** the street, doesn't save, **and** the script doesn't run, the rest of the address still matches the saved one, so we send the old saved street. This needs two things to go wrong at once. Sending no street instead would not be safer: it can give the wrong rate too (7.3% instead of 3.3% in the example above).

**Backward compatibility.**

- The script handle `wc-taxjar-order` and its file name do not change.
- The old global `ajaxSend` listener is removed. Nothing else depended on it.
- New public method `Automattic\WCServices\Tax\Address::from_order()`. It is additive, and `Address` is `final`, so no subclass can break.
- `get_backend_address()` (protected) keeps its signature and is still called, so overrides keep working. The fallback is a new private method called after it.
- No hooks are added or removed.
- The street sent to TaxJar changes in two cases: orders with no shipping country, and stores that calculate tax from the store address. In both, it used to be empty and is now the billing street, matching the billing fields WooCommerce already sends.
- Multisite: only order data and the site's own `woocommerce_tax_based_on` option are read, so each site behaves on its own (not tested on multisite).

### Related issue(s)

Closes [WOOTAX-347](https://linear.app/a8c/issue/WOOTAX-347)

Setup: store in Colorado (for example Littleton, CO 80120), **Automated taxes** on, **Calculate tax based on** = Customer shipping address.

**Order with no shipping address**

1. Create or open an order in **Pending payment** with one taxable product.
2. Leave the shipping address empty. Set the billing address to `18680 Highway 145`, `Dolores`, `CO`, `81323`.
3. Click **Recalculate** and confirm.
   - **Before this PR:** City Tax is added (about 7.3% in total).
   - **With this PR:** State and Special tax only, no City Tax (about 3.3%).
4. Optional: in the browser dev tools Network tab, the `woocommerce_calc_line_taxes` request now contains `street=18680 Highway 145`.

**Double quote in the address**

5. On the same order, set the billing city to `Do"lores` and click **Recalculate**.
   - **Before this PR:** the items box keeps spinning and nothing is sent.
   - **With this PR:** the request is sent and the box unblocks.

**PHP fallback (script not running)**

6. Put back the address from step 2 and click **Update** to save it.
7. Block the script: in dev tools, block requests matching `woocommerce-services-new-order-taxjar`, then reload the order.
8. Click **Recalculate**. The request has no `street`, but the result is still about 3.3%. With logging enabled, the WooCommerce Tax log shows `Recalculate request had no street; using the saved billing street.`
9. Change the billing ZIP in the form to `81321` **without saving** and click **Recalculate**. The log shows `... does not match the saved billing address; calculating without a street.`

Automated tests:

- `npm run test-client -- client/test/new-order-taxjar.js`
- `composer test -- --filter backend_totals`

### Checklist

- [x] unit tests


CezaryDrewniak (approved):
Verified full trace and ran everything, looks good

`getTaxableStreet()` matches core's side-pick rule exactly (incl. `base` → billing), the hook binding is right for `triggerHandler()`, and the PHP fallback no-ops on the Save/Update paths (no bare `country` in those `$_POST`s). No hook or BC concerns, `from_order()` is additive on a final class, `get_backend_address()` unchanged.

- New JS tests 7/7, backend_totals 6/6, full suites green, eslint + PHPCS clean vs base
- Mutation checks pass (side pick, `isset` guard, postcode compare, JS fallback)
- Stood up a fake Connect server and ran the real AJAX code path (TaxesController::calc_line_taxes) with the live API client pointed at it. All five scenarios behaved as described: street sent when the script runs, saved billing street used when it's blocked, no street on a changed ZIP, posted-empty kept empty, and Do"lores in the city goes through fine (3.3% vs 7.3% on 81323 in both the matching and blocked cases). Log lines match the ones in the PR.

Two optional nits: a `data || {}` guard in `addStreetToRecalculateData`, and one clause in the changelog about `base` stores now sending the billing street.

client/test/new-order-taxjar.js
```diff
@@ -0,0 +1,98 @@
...
+	);
+
+	return filtered ? filtered : data;
+};
+
+// Loads the script. The test must use the same jQuery instance as the script,
+// because jQuery keeps handlers per instance and resetModules() makes a new one.
+const loadScript = ( taxBasedOn ) => {
+	global.woocommerce_admin_meta_boxes = { tax_based_on: taxBasedOn };
+	jQuery = require( 'jquery' );
```
Abdalsalaam: Once this is merged with trunk, all 7 tests here fail with "Cannot find module 'jquery'". jquery used to resolve from the wp-calypso submodule, which #3017 dropped. The build is fine since jquery is a webpack external, only jest breaks, so CI goes red as soon as the branch is updated. Add jquery as a devDependency.

Abdalsalaam:

PR #3015, opened 2026-09-21
[WOOTAX-341] Fix - Update an order's tax when its amounts change outside the cart

bartech:
## Description

**The problem.** Since 3.6.8, when an existing order is changed through the REST API (the WooCommerce app, POS, integrations) or by code, we put back the tax **amount** the order had before. That was added so an address change would not wipe the tax to $0. But it also freezes the tax when anything else changes:

- The quantity goes from 1 to 2: the total goes up, the tax stays the same (under-collected).
- An item is removed: the total goes down, the tax still includes the removed item (over-collected).
- The same happens for fees, shipping and coupons.

**The rule this PR follows.** The **rate** is fixed when the order is placed. The **amount** follows the order.

**What this PR does.**

- **Nothing taxable changed** (for example, only the phone number, or the app sends the same items again): the tax is put back exactly as before. No rounding changes, and a tax amount typed by hand is kept.
- **Something changed:** we re-apply the rates saved on the order's tax lines to the items that changed:
  - A changed item uses the rates it already had.
  - Items that did not change keep their tax as it is.
  - A new item or fee uses the rates of an existing item in the same tax class. A new shipping line uses the rates of the existing shipping.
  - A new item that already has tax set keeps it. The Store API does this when it rebuilds a checkout draft from the cart.
  - An item added with `WC_Order::add_product()` counts as new too. That method saves the item straight away, so without this it looked like an unchanged item and stayed untaxed.
  - If the order has no rate for a new item, the item is left untaxed and an order note says so. Looking up a new rate is the next step (WOOTAX-341 part B / WOOTAX-346).
- **Every tax change adds an order note**, for example `Tax updated because the order changed: $2.10 → $2.70.` No notes are added to checkout drafts.
- **We never read the tax rate table here.** Percent, label and code come from the order itself. So a rate row that changed after the sale (WOOTAX-342) cannot leak into the order.
- **Old orders** (from before WooCommerce 3.7) do not save the rate percent. If one of them changes, we keep its tax as it is and add a note asking the merchant to check it.
- **VAT exempt:** an order made VAT exempt after the sale no longer gets its tax put back.

How we detect a change: the REST API saves an edited item **before** it recalculates the order. So we remember the item's old amount when it is saved (`woocommerce_before_order_item_object_save`). Removed items are found by comparing the saved items with the ones still on the order. The same logic works for code that edits an order and calls `calculate_totals()`.

This does **not** change what happens when only the address changes. The tax is still kept as it is (part B will handle that). There is also no special rule for paid orders: the tax follows the order, paid or not.

**Backward compatibility.**

- **Behaviour change:** when an existing, taxed order is recalculated through REST or code (`calculate_totals( true )`), its tax now moves if its item, fee or shipping amounts changed. Before, it never moved. Batch jobs that re-save old orders do not move tax unless they change an amount.
- **Behaviour change:** applying or removing a coupon on an order (`apply_coupon()` / `remove_coupon()`) now updates its tax. Subscriptions and gift card plugins use this.
- **Behaviour change:** orders that were made VAT exempt after the sale stay at $0 tax.
- New order notes on tax changes.
- New public methods `WC_Connect_TaxJar_Integration::remember_order_item_base_before_save()` (hooked to `woocommerce_before_order_item_object_save`) and `remember_order_item_created()` (hooked to `woocommerce_after_order_item_object_save`). Both are additive.
- `preserve_order_taxes_on_recalculation()` and `restore_order_taxes_after_recalculation()` keep their signatures and are still called on the same hooks, so overrides keep working. The other new methods are private.
- We now call the core filter `woocommerce_order_is_vat_exempt` with the same arguments core uses. Callbacks on it may run one extra time per recalculation.
- No hooks of our own are added or removed. No options are added.
- Multisite: no options are read or written. All the data is on the order, so each site behaves on its own (not tested on multisite).
- Not tested yet: partial refunds after the tax moves, Subscriptions renewals, and stores with prices entered including tax. For the last one we use the same calculation as WooCommerce core (`WC_Tax::calc_tax( $total, $rates, false )`).

### Related issue(s)

Part A of [WOOTAX-341](https://linear.app/a8c/issue/WOOTAX-341). Part B (a new rate lookup when the address changes) will come in a separate PR.

Setup: **Automated taxes** on, a store in a US state where you have nexus, and a REST API key with read/write access.

1. Place an order through checkout with one taxable product and a shipping address in your nexus state. Note the order ID, the line item ID and the tax (for example, $10.00 product, $0.60 tax).
2. Raise the quantity through the REST API. Send the new totals too; WooCommerce does not change the line total when only `quantity` is sent:
   ```
   curl -X PUT https://example.test/wp-json/wc/v3/orders/<order_id> \
     -u ck_xxx:cs_xxx -H 'Content-Type: application/json' \
     -d '{"line_items":[{"id":<item_id>,"quantity":2,"subtotal":"20.00","total":"20.00"}]}'
   ```
   - **Before this PR:** the total goes up, the tax stays at $0.60.
   - **With this PR:** the tax is $1.20 (same rate) and the order has the note `Tax updated because the order changed: $0.60 → $1.20.`
3. Remove the item (`"quantity":0`) from an order with two items.
   - **Before:** the tax still includes the removed item.
   - **With this PR:** its tax is removed too.
4. Send only an unchanged field, for example `{"billing":{"phone":"555-0100"}}`.
   - The tax does not change and no note is added (same as before).
5. Optional: in **WooCommerce > Settings > Tax > Standard rates**, change the rate the order used to 0%, then repeat step 2 on a new order. The tax is still calculated at the rate from the sale, not 0%.

Automated tests: `tests/php/test-class-wc-connect-taxjar-order-recalculation.php` (20 tests, real REST requests). Before this change, 4 passed on trunk and 12 failed, each because the tax did not move. All 20 pass now, and the full suite is green. I broke the new code in 14 different ways (for example, ignoring removals or recalculating untouched items), and each time at least one test failed.

### Checklist

- [x] unit tests


CezaryDrewniak (approved):
Looks good,

- Ran the new tests on trunk first: 15 of 19 fail there (all "tax didn't move"), all 19 pass with this PR, full suite green.
- Ran the real flows against the dev environment with a stub Connect server: order creation, qty raise, item removal, coupon, checkout → POS-style edit, and the 0%-overwritten rate row. All behave as described, tax follows the amounts at the order's recorded rate, note lands, rate table is never read, and amount edits make zero requests to the tax service.

One irrelevant nit:
Description says 4 passed / 12 failed on trunk, I got 4/15

Abdalsalaam:
This changes the tax on orders edited through REST, the app, POS or code. Since 3.6.8 the amount stayed fixed. Now it moves at the rate recorded on the order, and orders made VAT exempt after the sale stop getting their tax back. Was that rule agreed anywhere? I only see it in the PR description. If it's on the P2 thread #3018 mentions, link it here.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -2772,4 +2902,288 @@ private function restore_order_taxes( $order, $snapshot ) {
...
+	private function find_order_tax_base_changes( $order ) {
+		$types          = array( 'line_item', 'fee', 'shipping' );
+		$known_ids      = array();
+		$changed_ids    = array();
+		$new_item_taxes = array();
+		$added          = false;
+
+		foreach ( $order->get_items( $types ) as $item ) {
+			$item_id = (int) $item->get_id();
+			$is_new  = ! $item_id || isset( $this->order_items_created_in_request['ids'][ $item_id ] );
```
Abdalsalaam: This only catches items added in the same request. If code calls add_product() and save() in one request and calculate_totals() in a later one (cron, a second API call), the new item looks known and unchanged, its empty tax gets restored and there's no note. Reproduced: the new item stayed at $0. A total changed and saved in an earlier request has the same problem.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -2772,4 +2902,288 @@ private function restore_order_taxes( $order, $snapshot ) {
...
+	/**
+	 * Whether a saved order item's taxable amount differs from what it was taxed on.
+	 *
+	 * Compared as numbers: a caller that re-sends "10.00" for a stored "10" has not
+	 * changed anything.
+	 *
+	 * @param WC_Order_Item $item A saved line item, fee or shipping item.
+	 * @return bool
+	 */
+	private function order_item_base_moved( $item ) {
```
Abdalsalaam: If a REST edit only changes tax_class on a line item, or tax_status on a fee, and the total stays the same, it isn't treated as a change and the old tax is restored. Should those count as changes too?

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -2772,4 +2902,288 @@ private function restore_order_taxes( $order, $snapshot ) {
...
+		$order->save();
+
+		$old_tax = (float) $snapshot['cart_tax'] + (float) $snapshot['shipping_tax'];
+		$new_tax = (float) $order->get_cart_tax() + (float) $order->get_shipping_tax();
+		$notes   = array();
+
+		if ( abs( $old_tax - $new_tax ) > 0.000001 ) {
+			$notes[] = sprintf(
+				/* translators: 1: tax before the order changed, 2: tax after. */
+				__( 'Tax updated because the order changed: %1$s → %2$s. The tax rates recorded when the order was placed were used.', 'woocommerce-services' ),
```
Abdalsalaam: The arrow in this translatable string reads backwards in RTL locales. "from %1$s to %2$s" avoids that.

Abdalsalaam:
Found it, George's reply on the P2 covers the rule and the paid-order case: https://fulfillcodeable.wordpress.com/2026/09/07/in-person-and-mobile-pos-app-orders-are-taxed-incorrectly/ . Withdrawn.

Abdalsalaam (approved):

dustinparker (changes requested):
Hey @bartech, I think this one should be sorted before we merge, because it breaks the VAT exempt behaviour this PR adds.

The preserve and restore halves are on hooks with different firing conditions. `preserve_order_taxes_on_recalculation()` runs on `woocommerce_order_before_calculate_taxes`, which only fires from `calculate_taxes()`. `restore_order_taxes_after_recalculation()` runs on `woocommerce_order_after_calculate_totals`, which fires from `calculate_totals()` whether or not taxes were recalculated. Those two methods are independently callable, so the pair is not guaranteed:

- `calculate_totals( true )`: both fire, works as intended.
- `calculate_totals( false )`: only the restore half fires.
- `calculate_taxes()` on its own: only the preserve half fires, and the snapshot is never consumed.

Four of the five gates live in the preserve half, so on the second shape the restore half can run with none of them evaluated. A snapshot left over from an earlier bare `calculate_taxes()` then gets applied to an order that should have been skipped.

Repro against WC 11.2.0-beta.1, order placed at $18.00 plus $1.78 tax:

```php
$item->set_quantity( 2 ); $item->set_subtotal( '36' ); $item->set_total( '36' );
$order->calculate_taxes();                 // snapshot stashed, never consumed
add_filter( 'woocommerce_order_is_vat_exempt', '__return_true' );
$order->calculate_totals( true );
// total=39.56 tax=3.56, expected 36.00 / 0.00
// note: "Tax updated because the order changed: $1.78 -> $3.56."
```

Same result with `calculate_totals( false )`.

On reach, core always pairs its own calls, so a leftover snapshot has to come from third party code calling `calculate_taxes()` on its own, or from an exception between the two calls. On the consuming side there is one site in core that calls `calculate_totals( false )` without a preceding `calculate_taxes()` and is reachable outside AJAX: `wc_save_order_items()` in `wc-admin-functions.php:490`. That is what `TaxesController::calc_line_taxes()` calls, and core documents that method as a programmatic entry point, so `wp_doing_ajax()` does not cover it. The other consumer only sites are AJAX gated.

There is no `woocommerce_order_after_calculate_taxes` hook to pair with, so any fix is working around a gap in core. The smallest one I got working pulls the gates into a private helper and calls it from both halves:

```php
/**
 * Whether an out-of-cart recalculation of this order should have its taxes preserved.
 *
 * Both halves of the preserve/restore pair consult this, because they hang off hooks
 * with different firing conditions: preserve_order_taxes_on_recalculation() runs from
 * calculate_taxes(), while restore_order_taxes_after_recalculation() runs from
 * calculate_totals() whether or not taxes were recalculated. A snapshot can therefore
 * reach the restore half without this having been evaluated on the way in.
 *
 * @since 3.7.1
 * @param WC_Order $order The order being recalculated.
 * @return bool
 */
private function should_preserve_order_taxes( $order ) {
    // The cart/checkout flow populates response_rate_ids and manages its own taxes.
    if ( ! empty( $this->response_rate_ids ) ) {
        return false;
    }

    // Admin order edits recalculate over AJAX and are handled by calculate_backend_totals().
    if ( wp_doing_ajax() ) {
        return false;
    }

    // A new order that does not exist yet has no recorded taxes to preserve.
    if ( ! $order->get_id() ) {
        return false;
    }

    // WC zeroes the taxes of an exempt order; restoring them would undo the exemption.
    if ( apply_filters( 'woocommerce_order_is_vat_exempt', 'yes' === $order->get_meta( 'is_vat_exempt' ), $order ) ) {
        return false;
    }

    return true;
}
```

The preserve half clears any leftover snapshot first, so one cannot survive a gate that rejected it:

```php
public function preserve_order_taxes_on_recalculation( $args, $order ) {
    // A snapshot from a recalculation that never reached calculate_totals() must not
    // outlive it, or a later one would restore amounts read before it. Cleared ahead
    // of the gates, so a snapshot cannot survive a rejected one either.
    unset( $this->pre_recalculation_tax_snapshots[ (int) $order->get_id() ] );

    if ( ! $this->should_preserve_order_taxes( $order ) ) {
        return;
    }

    $snapshot = $this->snapshot_order_taxes( $order );
    // ...
}
```

And the restore half checks the same thing before applying anything:

```php
$snapshot = $this->pre_recalculation_tax_snapshots[ $order_id ];
unset( $this->pre_recalculation_tax_snapshots[ $order_id ] );

// This hook fires from calculate_totals() whether or not taxes were recalculated,
// so a snapshot can arrive here without the gates having run. Re-check them: an
// order made VAT exempt since the snapshot was taken must not get its tax back.
if ( ! $this->should_preserve_order_taxes( $order ) ) {
    return;
}
```

Worth noting `if ( ! $and_taxes ) { return; }` looks like the obvious guard in the restore half and it breaks `TaxesController::calc_line_taxes()`, which does `calculate_taxes()` then `calculate_totals( false )` and expects the snapshot to be applied. I tried it to be sure, and the preserved cart tax comes back as 0.00 instead of 1.80.

This does mean touching two tests in `test-class-wc-connect-taxjar-integration.php`. They currently assert `get_id()` is never called when an early gate fires, which the snapshot cleanup breaks. I think asserting the outcome is better anyway:

```php
$this->integration->preserve_order_taxes_on_recalculation( array(), $order );

// The first gate returns without snapshotting or arming the restore handler.
$this->assertSame( array(), $this->get_private_property( 'pre_recalculation_tax_snapshots' ) );
$this->assertFalse( has_action( 'woocommerce_order_after_calculate_totals', array( $this->integration, 'restore_order_taxes_after_recalculation' ) ) );
```

I ran this locally and the suite is green at 201 tests, with four new regression tests covering the orphan case, the `calculate_totals( false )` case, the rejected gate case, and the `calculate_taxes()` then `calculate_totals( false )` case that has to keep preserving.

What do you think?

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -2772,4 +2902,288 @@ private function restore_order_taxes( $order, $snapshot ) {
...
+
+		$known_ids   = $snapshot['base_changes']['known_ids'];
+		$changed_ids = $snapshot['base_changes']['changed_ids'];
+
+		// Rates already on the order, per tax class and for shipping, for new items.
+		$rate_ids_by_class = array();
+		$shipping_rate_ids = array();
+		foreach ( $order->get_items( array( 'line_item', 'fee', 'shipping' ) ) as $item_id => $item ) {
+			$key = 'shipping' === $item->get_type() ? 'shipping_' . $item_id : $item_id;
+			if ( ! in_array( (int) $item_id, $known_ids, true ) || empty( $snapshot['item_taxes'][ $key ]['total'] ) ) {
```
dustinparker: This check means a new item without tax of its own takes its rates from an item that survived the edit. Replace the last item in a tax class and there is nothing left to take them from, so the new item is left untaxed, even though the order's tax lines still record the rate.

One REST call on a US/CA order at 9.875%, $18.00 plus $1.78 tax, removing the item and adding another at the same price:

| | After the swap |
| --- | --- |
| trunk (`f16c2f7f`) | $19.78, tax 1.78, four tax lines intact |
| this branch (`4907b625`) | $18.00, tax 0.00, all tax lines removed |

Trunk is correct there, and never further off in the other swaps I worked through, because it keeps some tax where this keeps none. Swapping one item for another on a single item order is a normal POS action, so it is the same kind of edit the ticket is about.

The note warns the merchant but contradicts itself. It says "the tax rates recorded when the order was placed were used" and "the order has no recorded tax rate for it" in the same breath, and the order had four.

Falling back to the snapshot's tax lines when no item in the same tax class survives would cover it, at least for the standard class. That is the new rate lookup WOOTAX-346 is meant to handle, so it may be a deliberate deferral. If it is, could the note say so and the changelog mention this shape of item change?

PR #3016, opened 2026-09-21
[WOOTAX-321] Add - Check the store address and warn when its state or ZIP looks wrong

bartech:
## Description

> [!IMPORTANT]
> **Do not merge until [APIWOO-1957](https://linear.app/a8c/issue/APIWOO-1957/tax-analytics-record-taxjar-address-checks-separately-from-tax) is deployed on the Connect server.** Until then, every address check adds an empty "tax request" event to our tax analytics.

**The problem.** Automated taxes use the store's state and ZIP code to work out the tax on every US order. Nothing checks that they are right:

- A wrong **state**: orders from the store's own state are treated as out-of-state and get no tax.
- A wrong **ZIP**: every in-state order is charged the wrong local rate.
- No ZIP at all: tax can't be worked out (see WOOTAX-261).

**What this PR does.**

- When the store address changes, we check it with TaxJar's address validation (free, US only).
- A store manager sees a notice on WooCommerce admin screens only when something is wrong:
  - **No ZIP code.** Checked on our side, no request needed.
  - **TaxJar can't find the address.**
  - **The address matches more than one place.**
  - **TaxJar puts the address in a different state** (error) **or ZIP** (warning). The notice has a button to fix it: **Update state and ZIP code** or **Update ZIP code**, and **Keep my address**.
- The button changes only the state and the 5-digit ZIP. The street and city stay as the merchant typed them. They show on invoices and emails, and they don't change the tax.
- Nothing changes without that click. Saving settings is never blocked. If the check fails, we log it, show nothing, and try again an hour later.
- Dismissing a notice hides it until the address changes again.

**When the check runs.**

- Right after saving **WooCommerce → Settings → General**, with a 5-second limit, so the result shows on the same page.
- Otherwise in the background, on the next admin page load. This also covers stores that already had an address before this change, and addresses changed through onboarding, the REST API or WP-CLI.
- Only when Automated taxes is on. Stores outside the US are skipped.

**Backward compatibility.**

- No public class, method, hook or filter signature changes. The two new classes (`Automattic\WCServices\Tax\StoreAddressVerifier` and `StoreAddressNotice`) are `final` and marked `@internal`.
- `WC_Connect_TaxJar_Integration::init()` now also sets these two up. `is_nexus_address_valid()` and the `woocommerce_taxjar_nexus_address` filter are not touched.
- We check the address that is actually sent for tax, so a `taxjar_store_settings` filter is respected. If that filter changes the address, the notice does not offer the button (changing the settings would do nothing) and asks the merchant to update the custom code instead.
- New site data: the option `wc_connect_store_address_verification` (not autoloaded), the WP-Cron hook `wc_connect_verify_store_address`, and the admin-post actions `wc_connect_apply_store_address` and `wc_connect_dismiss_store_address` (both need a nonce and `manage_woocommerce`).
- New outgoing request: `taxjar/v2/addresses/validate` through the Connect server, once per address change.
- Multisite: the result is saved per site with `get_option`, like the rest of our settings. Not tested on a multisite install.

**Tests.** 46 new tests. Full suite green. I broke the code on purpose in 20 ways (mutation check) and the tests caught 19. The one they miss is a "don't schedule the check twice" guard, which WordPress core also enforces.

I also ran it on a local site against real TaxJar (through WP-CLI, not the browser). A real address came back verified. A wrong ZIP (80202 for a Littleton, CO address) came back with the right one (80120) and showed the warning. That run was before the button was limited to state and ZIP. That part is covered by the tests.

### Related issue(s)

Fixes [WOOTAX-321](https://linear.app/a8c/issue/WOOTAX-321). Blocked by [APIWOO-1957](https://linear.app/a8c/issue/APIWOO-1957). Related: [WOOTAX-261](https://linear.app/a8c/issue/WOOTAX-261) (this PR warns about a missing ZIP, but does not fix the checkout error).

Set up: Automated taxes on, store in the US with a real address.

1. Go to **WooCommerce → Settings → General** and click **Save changes**. No notice.
2. Change the ZIP to one from another town in the same state (for example 80202 for a Littleton, CO address). Save. A yellow notice says the ZIP may be wrong and names the right one.
3. Click **Update ZIP code**. The ZIP changes, the street and city stay as typed, and the notice goes away.
4. Change the state to a wrong one and save. The notice is red and the button says **Update state and ZIP code**.
5. Click **Keep my address** (or the ×). The notice stays hidden until you change the address again.
6. Clear the ZIP and save. A red notice says the store address has no ZIP code.
7. Turn Automated taxes off. No notices show.

Before this PR, none of these cases showed anything.

### Checklist

- [x] unit tests


CezaryDrewniak (approved):
Looks good, tested locally (unit + fake Connect server e2e in the docker WP, including real browser flows), nothing blocking. All PR claims reproduce: settings-save check, ZIP/state suggestions with apply, dismiss semantics, error retry via cron, 5s save timeout, filter respect. PHPCS clean, suite green.

Nit, not blocking: unvalidated state written to `woocommerce_default_country` (simple regex would close it), and a pending cron event can duplicate one request after a sync verify.

Abdalsalaam:

Abdalsalaam:
WOOTAX-321 says to use the shipping address normalisation path so this has no TaxJar dependency and survives a backend change. This uses TaxJar's validate endpoint instead. Was that switch deliberate? Not blocking, but worth a line in the description since the ticket calls it out for the Fulcrum move.

Abdalsalaam (approved):

PR #3018, opened 2026-09-22
[WOOTAX-341] Fix - Look up tax for new and re-addressed API orders

bartech:
## Description

> Stacked on #3015 (WOOTAX-341 part A). Review that one first; this PR's base is its branch.

**The problem.** Orders created or edited outside the cart (the WooCommerce app, POS, 3PL and ERP tools, any REST client, or code) never ask TaxJar for a rate:

- **A new order** is taxed with whatever rate rows happen to be stored for that area, left behind by earlier website customers. If no one from that town has bought online, the order gets **$0 tax**, with no error and no log (WOOTAX-346).
- **An address change** on an existing order keeps the old tax (since 3.6.8), even when the new address is in a different tax area (WOOTAX-341 part B).

**The rule this PR follows** (agreed on the P2 thread): when the address changes, look up the tax again. Always, with no setting. Keep the cost down by calling only when an address field really changed. If the lookup fails, keep the existing tax, never set it to zero, and write an order note.

**What this PR does.**

- **We ask TaxJar in two cases only:**
  - The address the order is taxed on changed. That is the one WooCommerce uses: shipping, or billing when there is no shipping country, or the store address for local pickup or when **Calculate tax based on** is set to the store address. The street counts too, because two streets in the same ZIP can have different rates.
  - The order has no tax yet and its items changed. This covers every new order (its items are saved for the first time), and items added to an untaxed draft.
- **Everything else makes no request.** Quantity, fee, shipping and coupon changes re-use the rates saved on the order (#3015). Saving the same address again, or changing only a phone number, makes no request. A bulk tool that re-saves thousands of orders without changing their address makes no calls.
- **Answers are cached for an hour.** We use the same request cache as checkout.
- **When TaxJar answers:** every line is taxed at the rates it returned. A taxable fee uses the rates of a product in the same tax class. Shipping uses the shipping rates TaxJar returned. When the address changed, an order note records the address and the old and new tax, for example `Tax recalculated with today's rates for 1777 Broadway, Boulder, CO, 80302, US: $4.00 → $2.67.` A new order gets no note.
- **When TaxJar does not answer:**
  - An existing order keeps its tax (or has its saved rates re-applied if amounts changed too), with the note `Tax could not be updated for the new address, because the tax service did not answer. The tax this order already had was kept. Check the tax on this order.`
  - A new order keeps what WooCommerce calculated (the same as today), with a note asking the merchant to check the tax.
- **Not touched:** checkout and the Store API (the cart already asks TaxJar), admin order edits (Recalculate already asks TaxJar), and VAT-exempt orders.
- **Moving an order out of the state you collect in** removes its tax. The order is taxed $0 because there is no nexus there. No request is made.

**Backward compatibility.**

- **Behaviour change:** a REST or programmatic `calculate_totals( true )` now makes a TaxJar request when the taxed address changed, or when an order with no tax has new or changed items. This is a new network call on those paths, the cost agreed on the P2 thread. It is reachable from cron and WP-CLI too; `calculate_tax()` no longer assumes a customer session (`WC()->customer` can be null there).
- **Behaviour change:** new orders created through REST or code are taxed from TaxJar instead of from the stored rate rows.
- **Behaviour change:** each lookup writes rate rows to the tax table, the same way checkout does. Until #3012 (WOOTAX-342) is merged, that also means more chances for the bug #3012 fixes. Please merge #3012 first.
- New order notes for address changes and failed lookups.
- New public methods, both hooked and additive: `remember_order_before_save()` (`woocommerce_before_order_object_save`) and `remember_order_created()` (`woocommerce_after_order_object_save`). The other new methods are private.
- `preserve_order_taxes_on_recalculation()`, `restore_order_taxes_after_recalculation()` and `calculate_tax()` keep their signatures and hooks.
- We now call the core filters `woocommerce_apply_base_tax_for_local_pickup` and `woocommerce_local_pickup_methods` the way core does, to find the taxed address. Callbacks on them may run once more per recalculation.
- Known gap: the core filter `woocommerce_order_get_tax_location` is not applied to the address we send to TaxJar.
- No hooks of our own are added or removed. No options are added.
- Multisite: no options are read or written. The only data used is on the order and in the site's own tax table, so each site behaves on its own (not tested on multisite).
- WooCommerce Subscriptions (checked in 8.5.0): renewal orders are never recalculated with `calculate_totals()`, so they are not affected. Subscriptions whose address changes through its REST API or plan switching are looked up again (not tested).
- **Changed tests:**
  - Removed `test_rest_order_creation_is_untouched`. It checked that a new REST order was left to WooCommerce, which is the WOOTAX-346 bug. It is replaced by the create tests in the new test file.
  - The two WOOSHIP-2256 tests in `test-class-wc-connect-taxjar-integration.php` now make TaxJar unreachable, because an address change now asks it. What they check is unchanged: the tax is kept, never set to $0.

### Related issue(s)

Part B of [WOOTAX-341](https://linear.app/a8c/issue/WOOTAX-341) and [WOOTAX-346](https://linear.app/a8c/issue/WOOTAX-346).

Setup: **Automated taxes** on, a US store address in a state where you have nexus, and a REST API key with read/write access. Turn on logging in **WooCommerce > Settings > Tax** to see the TaxJar requests in **WooCommerce > Status > Logs** (`wc-services-taxes`).

1. **New order.** In **WooCommerce > Settings > Tax > Standard rates**, delete the rows for a town in your state (or pick a town no one has ordered from). Create an order there through the REST API:
   ```
   curl -X POST https://example.test/wp-json/wc/v3/orders \
     -u ck_xxx:cs_xxx -H 'Content-Type: application/json' \
     -d '{"shipping":{"country":"US","state":"CO","postcode":"80120","city":"Littleton","address_1":"8031 Southpark Circle"},
          "line_items":[{"product_id":<id>,"quantity":1}]}'
   ```
   - **Before:** $0 tax.
   - **With this PR:** the tax TaxJar gives for that address, and one `taxjar/v2/taxes` request in the log.
2. **Address change.** Move that order to another town in the same state:
   ```
   curl -X PUT https://example.test/wp-json/wc/v3/orders/<order_id> \
     -u ck_xxx:cs_xxx -H 'Content-Type: application/json' \
     -d '{"shipping":{"postcode":"80302","city":"Boulder","address_1":"1777 Broadway"}}'
   ```
   - **Before:** the tax stays as it was.
   - **With this PR:** the tax for the new address, plus the note `Tax recalculated with today's rates for …: $old → $new.`
3. **No change.** Send `{"billing":{"phone":"555-0100"}}`. No request appears in the log, and the tax does not change.
4. **Failure.** Block outgoing requests to the WooCommerce Tax server (or disconnect Jetpack), then repeat step 2 with another address. The tax stays as it was, and the note says it could not be updated.

Checked on a local store with real TaxJar answers: steps 1–3 made exactly 2 requests (create, address change) and none for the phone change.

Automated tests: new `tests/php/test-class-wc-connect-taxjar-order-address-lookup.php` (24 tests, real REST requests against a fake TaxJar proxy). The full suite is green (510). I broke the new code in 18 different ways (for example, ignoring the street, skipping the fallback to billing, taxing shipping at the wrong rate, or not keeping the tax when TaxJar fails), and each time at least one test failed.

### Checklist

- [x] unit tests


Abdalsalaam:
This makes REST, app, POS and code orders call TaxJar on every new order and address change, with no setting, and each lookup writes rate rows the way checkout does. The description says this was agreed on the P2 thread but there's no link. Can you add it here?

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -2714,7 +2747,16 @@ public function restore_order_taxes_after_recalculation( $and_taxes, $order ) {
...
 
-		if ( empty( $snapshot['base_changes']['has_changes'] ) ) {
+		$lookup = $snapshot['lookup'] ? $this->lookup_order_taxes( $order ) : null;
+
+		if ( is_array( $lookup ) ) {
+			$this->apply_looked_up_order_taxes( $order, $lookup, $snapshot );
+		} elseif ( empty( $snapshot['tax_lines'] ) ) {
+			// Nothing to keep: what WC calculated stands, as it did before the lookup existed.
+			if ( false === $lookup ) {
+				$this->add_order_tax_note( $order, __( 'Tax could not be calculated for this order, because the tax service did not answer. Check the tax on this order.', 'woocommerce-services' ) );
```
Abdalsalaam: This note also shows up when no request was sent. An order with no ZIP fails the local check in calculate_tax() and still gets "the tax service did not answer". Same for the note at 2769. Something like "Tax could not be looked up for this address" fits both.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -3154,32 +3239,401 @@ private function reapply_order_tax_rates( $order, $snapshot ) {
...
+		foreach ( $order->get_items( array( 'line_item', 'fee', 'shipping' ) ) as $item_key => $item ) {
+			if ( 'taxable' !== $item->get_tax_status() ) {
+				$item->set_taxes( false );
+				continue;
+			}
+
+			if ( 'shipping' === $item->get_type() ) {
+				$rates = $shipping_rates;
+			} elseif ( 'fee' === $item->get_type() ) {
+				$rates = $rates_by_class[ $item->get_tax_class() ] ?? array();
```
Abdalsalaam: A taxable fee only gets tax if a product in the same tax class is on the order. A new order with a reduced-rate product and a standard-class fee ended up with $0 tax on the fee and no note. It should at least get the "No tax was added for ..." note the reapply path writes.

classes/class-wc-connect-taxjar-integration.php
```diff
@@ -3154,32 +3239,401 @@ private function reapply_order_tax_rates( $order, $snapshot ) {
...
+	 * Built like the admin Recalculate request (same line items), with the address taken
+	 * from the order instead of the form. Items whose tax location is filtered to a
+	 * different address are asked for separately, as at checkout. Answers are cached
+	 * for an hour (see smartcalcs_cache_request()).
+	 *
+	 * @param WC_Order $order The recalculated order.
+	 * @return array|false|null Lookup result; false when any request failed; null when
+	 *                          the order has nothing to ask about.
+	 */
+	private function lookup_order_taxes( $order ) {
```
Abdalsalaam: This runs inside the REST request with the proxy's 60s timeout, and failed lookups aren't cached. If the Connect server is down, every POS address change or bulk sync waits up to 60s per call. Needs a shorter timeout here or a short-lived unreachable flag.

Abdalsalaam:
Found it, same P2 thread, George's reply on question 2 and Bartosz's follow-up "always, no setting": https://fulfillcodeable.wordpress.com/2026/09/07/in-person-and-mobile-pos-app-orders-are-taxed-incorrectly/ . Withdrawn. Merge order still matters: 3012 first, since on trunk the lookup rewrites a merchant's own rate row.

Abdalsalaam (approved):

